PR #213, opened 2022-05-09
:sparkles: #481 - Handle situation where the user has been made inact…"

alextreme:
Restoring PR for 211


annashamray (approved):
I've removed "disabled" parameter from form and form actions. 
@JostCrow if it should be present please revert my commit

PR #215, opened 2022-05-11
Feature/search feedback message

annashamray:
task - https://taiga.maykinmedia.nl/project/open-inwoner/task/532

### DIFF
--- a/src/open_inwoner/scss/components/Form/ButtonRadio.scss
+++ b/src/open_inwoner/scss/components/Form/ButtonRadio.scss
@@ -4,6 +4,7 @@
 
     &:checked ~ .button-radio__button {
       color: var(--color-primary);
+      cursor: default;
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -4,6 +4,7 @@
 
     &:checked ~ .button-radio__button {
       color: var(--color-primary);
+      cursor: default;
```
JostCrow, line 7: It seems weird that we do not want the pointer cursor. What is the reason for this?

JostCrow (approved):

src/open_inwoner/scss/components/Form/ButtonRadio.scss
annashamray: Before the button is pressed the cursor is a pointer. But after the button is pressed I decided to move it back to arrow, since the second pressing won't do anything. I can make it always a pointer

JostCrow: Thanks for the explanation. You do not have to change anything. This is indeed better.

PR #230, opened 2022-05-18
[#621] Improved logging

vaszig:
This should be merged after PR #239 is merged.


src/open_inwoner/utils/tests/test_timeline_logger.py
```diff
@@ -97,40 +97,6 @@ def test_user_does_not_have_add_permission(self):
...
-        add_form["created_by"] = self.user.id
-        add_form.submit()
-        log_entry = TimelineLog.objects.first()
-        log_url = reverse(
-            "admin:timeline_logger_timelinelog_delete",
-            kwargs={"object_id": log_entry.id},
-        )
-        response = self.app.post(log_url, user=self.user, expect_errors=True)
-        self.assertEquals(response.status_code, 403)
-
```
JostCrow, line 133: I would leave these tests in. I would even make the change that it is a superuser and not a normal user that is not allowed to change or delete.

JostCrow (changes requested):

alextreme (approved):

PR #231, opened 2022-05-18
Feature/invite with digid

annashamray:
task - https://taiga.maykinmedia.nl/project/open-inwoner/task/472


vaszig (commented):
Generally, it looks fine to me. The only thing that feels weird to me is when registering via digid. I have two contacts in the list view. Is it because of the mock version of digid?

contact_digid_register

annashamray:
@vaszig Thanks for catching it! Two contacts (yours and reverse) were created and they were not duplicated correctly 
I simplified the contact queryset making, it should work now

alextreme:
Works fine, thanks! Glad you're able to round this up properly.

Feedback from my testing:
1. Having users without name/email will cause problems, created task #636 as this is separate
2. I'd like a bit more attention to the user interface when confirming the invitation. The DigiD-only option shouldn't be screen-wide, and the DigiD+username option should make clear that you can register via one _or_ the other. My suggestion would be to put both options next to each other (left vs right). 
3. The BSN-in-emailadres situation is a bit odd, I created #637 to rectify this.

Point 2 can be completed within this PR, 1+3 can be done in separate ones.

alextreme (changes requested):
> I'd like a bit more attention to the user interface when confirming the invitation. The DigiD-only option shouldn't be screen-wide, and the DigiD+username option should make clear that you can register via one or the other. My suggestion would be to put both options next to each other (left vs right).

annashamray:
@alextreme I placed registration with Digid and with email into two columns next to each other

alextreme (approved):

PR #232, opened 2022-05-18
supported subfolder

JostCrow:


src/open_inwoner/conf/base.py
```diff
@@ -230,11 +230,32 @@
 ]
 
 MEDIA_ROOT = os.path.join(BASE_DIR, "media")
-MEDIA_SUBFOLDER = config("MEDIA_SUBFOLDER", "")
+MEDIA_SUBFOLDER = config("MEDIA_SUBFOLDER", "appels")
 
 if MEDIA_SUBFOLDER:
     MEDIA_ROOT = os.path.join(MEDIA_ROOT, MEDIA_SUBFOLDER)
 
+FILER_STORAGES = {
```
annashamray, line 238: This variable is reassigned below in this file, could you remove it?

annashamray (changes requested):

annashamray (commented):
I'm not sure if there can be problems with displaying existing filer images on staging environment and if we need to make a data migration to change path for these images? 
Staging env uses `MEDIA_SUBFOLDER` to separate media of different domains.

alextreme:
@annashamray I'll deal with this on the stagingenvironments manually, worst-case is that we have to have a copy for the time being. But a datamigration is more trouble than it's worth in this stage.

annashamray (approved):

PR #234, opened 2022-05-19
[#589] Fix/show map's title on a single line

vaszig:


annashamray (commented):
Not sure about this width, it looks funny
image

I'll add other reviewers to gather more opinions

alextreme (approved):
I agree that it would be best to revisit how the popup is designed and functions (this was also Roxanne's feedback during our weekly, as the map currently has limited user-value) but this has been placed on the backlog. For now this is sufficient in my eyes.

PR #236, opened 2022-05-19
[#619] Update/product import-export fields

vaszig:
This has to do with a small update for keeping the m2m fields consistent. They are all imported and exported based on the slug field.

### DIFF
--- a/src/open_inwoner/pdc/resources/export_resource.py
+++ b/src/open_inwoner/pdc/resources/export_resource.py
@@ -31,7 +31,7 @@ class ProductExportResource(resources.ModelResource):
     organizations = fields.Field(
         column_name="organizations",
         attribute="organizations",
-        widget=ManyToManyWidget(Organization, field="name"),
+        widget=ManyToManyWidget(Organization, field="slug"),
     )
     related_products = fields.Field(
         column_name="related_products",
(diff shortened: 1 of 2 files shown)


annashamray (commented):
I am blocked by the bug in task #635. I'll review this PR after that bug is fixed

src/open_inwoner/pdc/resources/export_resource.py
```diff
@@ -31,7 +31,7 @@ class ProductExportResource(resources.ModelResource):
     organizations = fields.Field(
         column_name="organizations",
         attribute="organizations",
-        widget=ManyToManyWidget(Organization, field="name"),
+        widget=ManyToManyWidget(Organization, field="slug"),
```
annashamray, line 34: I'm not sure about this one, because in the provided Excel file they use Organization name, not slug
Should the input for M2M fields be slugified on our end?

vaszig: The file that we have as a template on the import page has everywhere slug type names for m2m fields. I changed them as well because if we have name in the resource but slug type in the imported file we will have a problem. In my opinion it is better to have slugs because they are unique and helps with duplicates as well.

PR #241, opened 2022-05-23
Feature/style fixes

JostCrow:


vaszig (changes requested):
It seems that there is a difference between logo and footer logo appearance. The footer is a bit blurry.
- Logo
logo

- Footer (blurry)
footer_before


- Footer (proper)
footer_after

vaszig (approved):

PR #242, opened 2022-05-24
[#618] Feature/autofocus in mobile questionnaire

vaszig:


alextreme:
Leaving for Jorik to review, I'm not in a rush to get this in

src/open_inwoner/js/components/questionnaire/index.js
```diff
@@ -0,0 +1,10 @@
+const question = document.getElementById('question')
+const stepIndicators = document.querySelectorAll(
```
JostCrow, line 2: There should only be one active item. So I would say that we should not use querySelectorAll.
Also by selecting multiple (if there are) you will scroll to the first, than the second, than the third.

JostCrow (changes requested):

alextreme:
Merging and double-checking on the testenv

PR #247, opened 2022-05-25
[#581] Make location non clickable

vaszig:


annashamray (approved):

alextreme:
Merging and will double-check on testenv

PR #248, opened 2022-05-25
Feature/register necessary fields

annashamray:
tasks:
* https://taiga.maykinmedia.nl/project/open-inwoner/task/636
* https://taiga.maykinmedia.nl/project/open-inwoner/task/637

### DIFF
--- a/src/open_inwoner/accounts/forms.py
+++ b/src/open_inwoner/accounts/forms.py
@@ -59,6 +59,30 @@ class UserForm(forms.ModelForm):
         )
 
 
+class NecessaryUserForm(forms.ModelForm):
+    invite = forms.ModelChoiceField(
+        queryset=Invite.objects.all(),
+        to_field_name="key",
+        widget=forms.HiddenInput(),
+        required=False,
+    )
+
+    class Meta:
+        model = User
+        fields = (
+            "first_name",
+            "last_name",
+            "email",
+            "invite",
+        )
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["first_name"].required = True
+        self.fields["last_name"].required = True
+
+
 class ThemesForm(forms.ModelForm):
     class Meta:
         model = User
@@ -76,7 +100,7 @@ class ContactForm(forms.ModelForm):
     def __init__(self, user, create, *args, **kwargs):
         self.user = user
         self.create = create
-        return super().__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
 
     class Meta:
         model = Contact
--- a/src/open_inwoner/accounts/migrations/0043_change_digid_email.py
+++ b/src/open_inwoner/accounts/migrations/0043_change_digid_email.py
@@ -1,0 +1,30 @@
+from django.db import migrations
+from open_inwoner.utils.hash import generate_email_from_string
+from ..choices import LoginTypeChoices
+
+
+def change_bsn_email_to_hash(apps, _):
+    User = apps.get_model("accounts", "User")
+
+    for user in User.objects.filter(login_type=LoginTypeChoices.digid).all():
+        if user.email == f"user-{user.bsn}@bsn.com":
+            user.email = generate_email_from_string(user.bsn)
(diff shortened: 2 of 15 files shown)


vaszig (approved):
I think @JostCrow should have a look as well. have only some remarks-questions after reading and testing your code.

src/open_inwoner/accounts/migrations/0043_change_digid_email.py
```diff
@@ -0,0 +1,30 @@
...
+        if user.email == f"user-{user.bsn}@bsn.com":
+            user.email = generate_email_from_string(user.bsn)
+            user.save()
+
+
+def change_hash_email_to_bsn(apps, _):
+    User = apps.get_model("accounts", "User")
+
+    for user in User.objects.filter(login_type=LoginTypeChoices.digid).all():
+        if user.email == generate_email_from_string(user.bsn):
```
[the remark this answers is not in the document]
annashamray: In this PR I change how we generate an email for Digid users: before it was generated like `user-{user.bsn}@bsn.com`, but it's not a good thing to expose the user's BSN in the email, that's why I replace it with hash (this hash is not secure, but the email would be used temporarily before the user fills in the correct one, so it should do). 

This file is a data migration which replaces the old generated emails with the new ones and this function does the reverse thing: replace the hashed email with `user-{user.bsn}@bsn.com` email, since it is used when you are making a reverse migration.

src/open_inwoner/accounts/views/registration.py
```diff
@@ -1,24 +1,25 @@
...
+from ..forms import CustomRegistrationForm, NecessaryUserForm
+from ..models import Invite, User
 
 
-class CustomRegistrationView(LogMixin, RegistrationView):
-    form_class = CustomRegistrationForm
-
+class InviteMixin:
     def get_initial(self):
         initial = super().get_initial()
```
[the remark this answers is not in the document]
annashamray: This is the mixin class, which is used below for views inherited from `FormView`

src/open_inwoner/accounts/forms.py
```diff
@@ -59,6 +59,30 @@ class Meta:
         )
 
 
+class NecessaryUserForm(forms.ModelForm):
+    invite = forms.ModelChoiceField(
+        queryset=Invite.objects.all(),
```
alextreme, line 64: this will be evaluatied upon loading open-inwoner. Should be populated in the __init__

alextreme (changes requested):

src/open_inwoner/accounts/migrations/0043_change_digid_email.py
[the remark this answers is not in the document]
alextreme: I've explained this to Vasileios, so should be cleared up now.

I would like to replace 'bsn.com' with 'example.com'. This to ensure that we don't send email by accident to a domain which isn't under our control.

src/open_inwoner/utils/hash.py
```diff
@@ -0,0 +1,8 @@
+from hashlib import md5
+
+
+def generate_email_from_string(value: str) -> str:
+    """generate email address based on string"""
+    hashed_bsn = md5(value.encode(), usedforsecurity=False).hexdigest()
```
alextreme, line 6: ensure salt is used here, even though it's not used for security

src/open_inwoner/accounts/forms.py
alextreme: I think you indeed misunderstood. Doing this means afaik executing the queryset when starting/reloading runserver, the queryset won't be executed when instantiating the form. In my experience you need to make your queryset a callable, in that case it should work as expected.

However it could be that my knowledge is rusty, I noticed that they have an example in the Django documentation which does the same: https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/#a-full-example

alextreme: Lets leave it as is. I've tested adding a plantemplate on the testenvironment and this worked so I was incorrect, sorry!

alextreme (approved):

PR #249, opened 2022-05-27
[#640] Fix/only non-digid users access password reset-change

vaszig:


src/open_inwoner/accounts/views/profile.py
```diff
@@ -63,6 +63,10 @@ def get_context_data(self, **kwargs):
         else:
             context["contact_text"] = _("U heeft nog geen contacten.")
         context["questionnaire_exists"] = QuestionnaireStep.objects.exists()
+        context["backend_auth_is_email"] = (
```
JostCrow, line 66: The naming seems a bit weird to me. It is not clear what "backend_auth_is_email is doing. A better name would be "can_change_password"

Also all we want here is to see if the user is login_type != digid.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -1,16 +1,31 @@
...
+
+class LogPasswordChangeView(UserPassesTestMixin, LogMixin, PasswordChangeView):
+    def test_func(self):
+        if self.request.user.is_authenticated:
+            return self.request.user.login_type == LoginTypeChoices.default
+        return False
+
+    def handle_no_permission(self):
+        if self.request.user.is_authenticated:
+            return redirect(reverse("root"))
```
annashamray, line 25: Maybe include django message that this user can't change the password here?

annashamray (changes requested):

src/open_inwoner/accounts/forms.py
```diff
@@ -59,6 +60,15 @@ class Meta:
         )
 
 
+class CustomPasswordResetForm(PasswordResetForm):
+    def send_mail(self, *args, **kwargs):
+        email = self.cleaned_data.get("email")
+        user = User.objects.get(email=email)
+
+        if user.login_type == LoginTypeChoices.default:
+            return super().send_mail(*args, **kwargs)
```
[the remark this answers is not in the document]
alextreme: I've added a remark to the password reset template to state that it doesn't work when using DigiD.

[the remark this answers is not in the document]
alextreme: I've added a remark to the password reset template to state that it doesn't work when using DigiD.

src/open_inwoner/accounts/views/auth.py
vaszig: Is this needed? I am asking because this is only a secondary check if the digid user types explicitly the url for the password-change view. By default the user cannot see this button (it is not rendered). We have implemented the same  case in "Mijn aanvragen" page.

src/open_inwoner/accounts/forms.py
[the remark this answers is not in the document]
JostCrow: I like the way it is now. So you can not guess email addresses in the system. Also limiting ti for non-Digid users requires you to be logged in. But the view is for users that are not logged in.

src/open_inwoner/accounts/views/auth.py
JostCrow: Lets remove the function.

alextreme (approved):

annashamray (approved):

PR #253, opened 2022-06-01
show full name instead of email for action users

annashamray:
task - https://taiga.maykinmedia.nl/project/open-inwoner/task/502

I'm not sure should I change widget to 'autocomplete', I tried but it doesn't look like very beneficial, so I left the widget as it is


alextreme (approved):
Works for me, thanks!

PR #257, opened 2022-06-03
[#528] Feature/improve plan template popup

vaszig:


annashamray (changes requested):
Since it's a styling issue I'll be a little nitpicking 

* The margin between title (plan name) and the body below is bigger in the design
* The first column (preview__title) is too wide and it makes the data in the right column to look more crowded
* The file url ("bestand") should be clickable
* "X" button is missing in the top-right corner
* (nice-to-have) I think the popup in the design is a little wider, and with a lot of action templates the wider popup could look nicer, but I don't insist on this one
* (nice-to-have) I'd like to close modal by clicking outside of it. For now if you click outside you can even select radio button which is not nice. But it can be dealt with in another issue

I'll attach the screenshot with my modal so you can double check that it's displayed as expected 

image

vaszig:
Concerning the 4th and 6th comment, I didn't add the "x" button because I thought we should be consistent we the other popups as well (they don't have this button, only the "sluiten" one and you cannot click outside the popup). So, I believe that I should either change the other ones as well or keep it as it is.

annashamray:
1. Margin after the header is not something critical. We can leave it as it is
4. I personally think that 'x' button would be nice everywhere, but we can park this change for now

vaszig:
Yes, Jorik agrees with adding the "x" button as well, so I will change this.

alextreme (approved):
Agreeing with Anna here, the 'x' isn't necessary the work which is done in this PR is sufficient in my eyes too.

PR #259, opened 2022-06-08
Feature/actions history

annashamray:
task - https://taiga.maykinmedia.nl/project/open-inwoner/task/506

### DIFF
--- a/src/open_inwoner/utils/logentry.py
+++ b/src/open_inwoner/utils/logentry.py
@@ -28,7 +28,7 @@ def get_change_message(fields=None, form=None):
         form, form.changed_data
     )
 
-    return _("Changed {changed_fields}.").format(
+    return _("Changed: {changed_fields}.").format(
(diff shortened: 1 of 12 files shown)


src/open_inwoner/utils/logentry.py
```diff
@@ -28,7 +28,7 @@ def get_change_message(fields=None, form=None):
         form, form.changed_data
     )
 
-    return _("Changed {changed_fields}.").format(
+    return _("Changed: {changed_fields}.").format(
```
vaszig, line 31: I don't know if it's really necessary, but we could add here a different message when the user doesn't make an actual change to any field (like django admin's message).

actions

admin

annashamray: Thanks for catching it! I think there is no point in logging form submission without the actual change, so I excluded such cases from logging. Now it should be fixed.

JostCrow (approved):

PR #267, opened 2022-06-14
Feature/cases improvements

vaszig:
Tasks #657 and #658


JostCrow (approved):
It seems to be working correctly. The only downside is that the pages are really slow.

src/open_inwoner/accounts/views/cases.py
```diff
@@ -78,32 +116,42 @@ def get_context_data(self, **kwargs):
...
+                for status_type in status_types:
+                    if status.statustype == status_type.url:
+                        status.statustype = status_type.omschrijving
+
+            context["case"] = {
+                "start_date": case.startdatum,
+                "end_date": case.einddatum,
+                "description": case_type.omschrijving
+                if case_type
+                else _("No data available"),
```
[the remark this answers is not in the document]
vaszig: This condition exists only for the case of an exception during the request. We return None if something goes wrong with `fetch_single_case_type`, so we will have an attribute error (`case_type.omschrijving`)

src/open_inwoner/openzaak/statuses.py
```diff
@@ -33,7 +34,44 @@ def fetch_status_history(case_url: str) -> List[Status]:
...
+    except ClientError as e:
+        logger.exception("exception while making request", exc_info=e)
+        return []
+
+    status_types = factory(StatusType, response)
+
+    return status_types
+
+
+def fetch_specific_status_types(case_type: str) -> List[StatusType]:
```
annashamray, line 74: It looks like `fetch_specific_status_types` and `fetch_status_types` are basically the same function, the only difference is the input parameter. I think these functions can be merged into one  with `case_type` as an optional parameter. 
It will help to make code DRYer

annashamray (approved):
Except one last remark about merging status type functions, the rest looks nice

src/open_inwoner/openzaak/statuses.py
```diff
@@ -33,7 +34,24 @@ def fetch_status_history(case_url: str) -> List[Status]:
...
+            lambda url: client.retrieve("status", url=url),
+            status_urls,
+        )
+
+    statuses = factory(Status, list(_statuses))
+
+    return statuses
+
+
+def fetch_status_types(*case_type: str) -> List[StatusType]:
```
annashamray, line 54: Why do you user asterisk here?

vaszig: Because I am using this function with and without a variable (accounts/views/cases.py). Do you mean I should use a keyword argument?

annashamray: yes, please use a keyword argument with the default value. When * is used The tuple is initialized (https://docs.python.org/3/reference/compound_stmts.html#function-definitions), I don't think we need it here.

PR #272, opened 2022-06-15
Feature/action is for field

annashamray:
task - https://taiga.maykinmedia.nl/project/open-inwoner/task/698
and action list improvements:
* show `start_date` before `end_date` and in the same format
* add defaults for empty values to avoid showing 'None'


annashamray:
@alextreme I couldn't reproduct the empty dropdown, but anyway I changed the field.queryset for both cases - with and without a plan

Without a plan it should be a disabled dropdown

JostCrow (approved):
It works correct. The only small thing that I found is that when an input/select element is disabled, it does not change the cursor to not-allowed.

PR #277, opened 2022-06-17
[#684] Feature/add published boolean to theme and product 

vaszig:


src/open_inwoner/pdc/models/category.py
```diff
@@ -3,7 +3,19 @@
...
+from treebeard.mp_tree import MP_MoveHandler, MP_Node
+
+from open_inwoner.utils.managers import PublishedQueryset
+
+
+class PublishedMoveHandler(MP_MoveHandler):
+    def process(self):
+        if self.node.published and not self.target.published:
+            raise InvalidMoveToDescendant(
+                _("Published parent nodes cannot be children of an unpublished parent.")
```
JostCrow, line 16: This sentence is a bit weird.

JostCrow (commented):
Still need to go over the changes visually

JostCrow (approved):

src/open_inwoner/pdc/models/category.py
```diff
@@ -3,7 +3,19 @@
 from django.utils.translation import ugettext_lazy as _
 
 from filer.fields.image import FilerImageField
-from treebeard.mp_tree import MP_Node
+from treebeard.exceptions import InvalidMoveToDescendant
+from treebeard.mp_tree import MP_MoveHandler, MP_Node
+
+from open_inwoner.utils.managers import PublishedQueryset
```
annashamray, line 9: Usually we put model managers and querysets in the same app as related models. You can see an example in the `open_inwoner.accounts.managers` and `open_inwoner.accounts.query`
Could you move this queryset to the `accounts.pdc` folder?

src/open_inwoner/pdc/models/category.py
```diff
@@ -45,6 +62,7 @@ class Category(MP_Node):
     )
 
     node_order_by = ["slug"]
+    objects = PublishedQueryset.as_manager()
```
annashamray, line 65: Afaik you should use a subclass of `MP_NodeQuerySet` or `MP_NodeManager` as MP_Node model manager

annashamray (changes requested):

src/open_inwoner/pdc/models/category.py
vaszig: Thanks for this. This is indeed what is recommended in the documentation. I changed the class to MP_NodeQuerySet for the custom queryset.

PR #279, opened 2022-06-20
Feature/OIDC

JostCrow:


vaszig (changes requested):
As discussed with @JostCrow as well, the following should be added-fixed:

- Add settings-configuration details for the admin page to taiga ticket

- Check whether we receive the email from the external page or not

src/open_inwoner/configurations/models.py
```diff
@@ -265,6 +265,22 @@ class SiteConfiguration(SingletonModel):
...
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="oidc_logo",
+        help_text=_("Logo that can be used for the openid commect login method"),
+    )
+    openid_connect_login_text = models.CharField(
+        verbose_name=_("Openid Connect login text"),
+        max_length=250,
+        default="Login with Azure AD",
```
vaszig, line 279: The translation is missing.

vaszig (approved):

src/open_inwoner/configurations/models.py
JostCrow: This is not missing. It is just a default value. So they can change it later on.

JostCrow:
It indeed is a bit of a hack. But the problem is that an email address needs to be present. Otherwise only one user can login with OIDC if it stays empty.

src/open_inwoner/accounts/backends.py
```diff
@@ -24,3 +31,34 @@ class CustomAxesBackend(AxesBackend):
...
+
+        logger.debug("Creating OIDC user: %s", unique_id)
+
+        kwargs = {
+            "oidc_id": unique_id,
+            "email": email,
+            "login_type": LoginTypeChoices.oidc,
+        }
+
+        user = self.UserModel.objects.create_user(**kwargs)
```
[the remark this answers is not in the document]
JostCrow: Personally, I like to override as little as possible from external packages. That way, upgrading will be easier later on.

src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -19,9 +19,24 @@ <h1 class="h1">{% trans 'Welkom' %}</h1>
...
                         </a>
-                        {% link bold=True href='digid:login' text=_('Inloggen met DigiD') secondary=True icon='arrow_forward' %}
+                        {% link href='digid:login' text=_('Inloggen met DigiD') secondary=True icon='arrow_forward' extra_classes="link--digid" %}
                     {% endrender_card %}
 
+                {% get_solo 'mozilla_django_oidc_db.OpenIDConnectConfig' as oidc_config %}
+                {% get_solo 'configurations.SiteConfiguration' as site_config %}
+                {% if oidc_config.enabled %}
+                    {% render_card tinted=True direction='horizontal' %}
+                    {% if site_config.openid_connect_logo %}
```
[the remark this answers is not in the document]
JostCrow: Personally, for now it will be a no. Since this is what the package itself already provided.

src/open_inwoner/accounts/backends.py
```diff
@@ -24,3 +33,34 @@ class CustomAxesBackend(AxesBackend):
...
         if request:
             return super().authenticate(request, *args, **kwargs)
+
+
+class CustomOIDCBackend(OIDCAuthenticationBackend):
+    def create_user(self, claims):
+        """Return object for a newly created user account."""
+        unique_id = self.retrieve_identifier_claim(claims)
+
+        email = generate_email_from_string(bsn)
```
annashamray, line 43:
```suggestion
        email = generate_email_from_string(unique_id)
```

annashamray (approved):
Except one minor typo it looks very nice.

PR #282, opened 2022-06-29
Feature/digid logout

annashamray:
**Changes:**
1. use digid view to logout digid users (SLO support)
2. temporarily use github branch as a source code for `django-digid-eherkenning`. It would be pinned back when logout callback view is implemented


alextreme (approved):
Looks good!

PR #293, opened 2022-07-24
Fixed #745 -- published on False by default instead of True

alextreme:
Also updated the translations


alextreme:
FYI, admins wanted to have Published being False by default instead of True.

vaszig (approved):
In my opinion this makes more sense, it should be False by default to avoid publishing something that it is not "ready for publishing"

PR #294, opened 2022-07-24
Fixed #544 -- Ensuring highlighted categories are also shown when log…

alextreme:
…ged in


alextreme:
Investigated 544, and with that case I noticed that non-highlighted categories are shown before highlighted ones if you're logged in. As anonymous user it did work as expected.
I've restructured this a bit to make the 3 possible situations more clear, and ensured that if you as a user haven't selected categories that then you'll see the same categories as when you're not logged in.

PR #298, opened 2022-08-23
Did some work on accesibility

alextreme:


src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -3,7 +3,9 @@
 
 
 {% block header_image %}
+<aside aria-label="{% trans "Header afbeelding" %}">
```
svenvandescheur, line 6: Is there any reason these kinds of labels should be explicitly available to assistive technology. In other words: does it add value to for screen readers to read that this section shows only an image?

If this section is purely decorative, I think aria-hidden may be a better choice?

> Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. This can improve the experience for assistive technology users by hiding:

 > Purely decorative content, such as icons or images
 > Duplicated content, such as repeated text
 > Offscreen or collapsed content, such as menus

 > The presence of the aria-hidden attribute hides content from assistive technology but doesn't visually hide anything.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -3,7 +3,9 @@
...
     <img class="main-image" src="{{ hero_image_login }}" alt="Header image">
```
svenvandescheur, line 7: Can we add a proper alt label which describes the contents of the image?

src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n icon_tags button_tags %}
 
-<aside class="anchor-menu {% if desktop %}anchor-menu--desktop{% else %}anchor-menu--mobile{% endif %}">
+<aside class="anchor-menu {% if desktop %}anchor-menu--desktop{% else %}anchor-menu--mobile{% endif %}" aria-label="{% trans "Zijmenu mobiel" %}">
```
svenvandescheur, line 3: I think setting a nav tag for this implicitly sets a navigation role, If a label is used, the position is not relevant for the assistive technology so a label can better describe the menu's contents like "primary navigation"?

src/open_inwoner/components/templates/components/Button/Button.html
```diff
@@ -5,8 +5,9 @@
         class="{{ classes }}"
         href="{{href}}"
         title="{% firstof title text %}"
-        aria-describedby="{{ text }}"
+        {% if text %}aria-describedby="{{ text }}"{% endif %}
```
svenvandescheur, line 8: > The aria-describedby attribute lists the [id](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#attr-id)s of the elements that describe the object. It is used to establish a relationship between widgets or groups and the text that describes them.

 I think we should use aria-label here, and text should be mandatory to populate it. We can then choose not to render the text but only the aria-label as the icon is not useful for assistive technology.

src/open_inwoner/components/templates/components/Form/ChoiceRadio.html
```diff
@@ -11,4 +12,5 @@
         {% if choice.0 in data or checked %}checked="checked"{% endif %}
     />
     <label class="radio__label" for="id_{{ name }}_{{ index }}">{{ choice.1 }}</label>
```
svenvandescheur, line 14: I think nesting labels may confuse assistive technology, can we get rid of the inner label?

src/open_inwoner/components/templates/components/Form/Input.html
```diff
@@ -1,9 +1,11 @@
 {% load form_tags icon_tags %}
 
 <div class="form__control {{ extra_classes|default:"" }}">
-    <label class="label">
+    <label class="label {% if no_label %}label--gapless{% endif %}">
         {% if icon %}{% icon icon outlined=True %}{% endif %}
-        {% if not no_label %}{{ field.label }}{% endif %}
+        <label class="label__label {% if no_label %}label__label--hidden{% endif %}">
```
svenvandescheur, line 6: Same, I don't think lables should be nested.

src/open_inwoner/components/templates/components/Header/Header.html
```diff
@@ -3,16 +3,16 @@
...
-        {% logo src=config.logo.file|thumbnail_url:'logo' alt="" %}
+        {% logo src=config.logo.file|thumbnail_url:'logo' alt=config.logo.default_alt_text %}
 
         <div class="header__menu">
             <div class="header__menu-icon">
                 {% icon icon="menu" %}
             </div>
             <div class="header__submenu">
-                <nav class="header__actions">
+                <nav class="header__actions" aria-label="Zoek navigatie mobiel">
```
svenvandescheur, line 15: I don't think mobile specific functionality applies to assistive technology? Can we use only one navigation structure for assistive technology and hide the other using aria-hidden="true"?

src/open_inwoner/components/templates/components/Step/StepIndicator.html
```diff
@@ -1,7 +1,7 @@
-{% load link_tags %}
+{% load link_tags i18n %}
 
 {% spaceless %}
-    <aside class="step-indicator">
+    <aside class="step-indicator" aria-label="{% trans "Stappen indicator" %}">
```
svenvandescheur, line 4: I think a more useful label would be "step 1 out of 3" then hide the rest of the contents using aria-hidden?

src/open_inwoner/components/templates/components/Typography/Link.html
```diff
@@ -17,6 +17,7 @@
     {% endif %}
     {% if toggle_exclusive %}data-toggle-exclusive="{{ toggle_exclusive }}"{% endif %}
     aria-describedby="{{ text }}"
+    {% if hide_text %}aria-hidden="true"{% endif %}
```
svenvandescheur, line 20: This means that this link would not be available for screen readers if we want to only hide the text for visual browsers. I think the aria-label should always be populated and the context may be hidden (as long as aria-label is set) but aria-hidden does not represent the visual state of the element:

 > Use caution when using aria-hidden to hide visibly rendered content from assistive technologies. You should not be hiding visible content unless doing so improves the experience for users of assistive technologies by removing redundant or extraneous content. Only when identical or equivalent meaning and functionality is exposed to assistive technologies can removing visible content from the accessibility API be considered.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles

src/open_inwoner/js/components/accessibility/read.js
```diff
@@ -11,6 +11,17 @@ class ReadOut {
     } else {
       this.node.parentElement.remove()
     }
+
+    let voices = []
+    function populateVoiceList() {
+      voices = this.speechSynthesis.getVoices()
+      console.log(voices)
```
svenvandescheur, line 18: Console.log

src/open_inwoner/templates/pages/category/detail.html
```diff
@@ -3,7 +3,9 @@
 
 {% block header_image %}
 {% if object.image %}
-    <img class="main-image main-image--start-after" src="{{ object.image.url }}" alt="Header image">
+    <aside aria-label="{% trans "Header afbeelding" %}">
+        <img class="main-image main-image--start-after" src="{{ object.image.url }}" alt="Header image">
```
svenvandescheur, line 7: Can we use the imaga's alt text here?

src/open_inwoner/templates/pages/category/detail.html
```diff
@@ -35,7 +37,7 @@ <h1 class="h1">
     {% if questionnaire_roots %}
         <div class="grid">
             <div class="column column--start-1 column--span-6 ">
-                <aside class="questionnaire">
+                <aside class="questionnaire" aria-label="{% trans "zelfdiagnose" %}">
```
svenvandescheur, line 40: I'm not sure whether these labels add value since the assistive technology probably can read the textual content.

 > Note: aria-label is intended for use on interactive elements, or elements made to be interactive via other ARIA declarations, when there is no appropriate text visible in the DOM that could be referenced as a label

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/open_inwoner/templates/pages/product/detail.html
```diff
@@ -95,8 +97,8 @@ <h4 class="h4" id="links">{% trans "Links" %}</h4>
         {% render_column start=related_products_start span=4 compact=True %}
             {# Related products. #}
             {% if object.related_products.published.exists %}
-                <nav class="link-list">
-                    <h4 class="h4" id="see">{% trans "Zie ook" %}</h4>
+                <nav class="link-list" aria-label="{% trans "Zie ook links" %}">
```
svenvandescheur, line 100: I thiink "gerelateerd links" is a bit more descriptive.

src/open_inwoner/templates/pages/user-home.html
```diff
@@ -18,8 +18,12 @@ <h2 class="h2 h2--extra">
             {% render_card %}
                 <h3 class="h3">{{ plan.title }}</h3>
                 <p class="p">{{ plan.goal|truncatewords:20 }}</p>
-                <a class="button button--icon-before button--transparent button--secondary"
-                   href="{{ plan.get_absolute_url }}">
+                <a
+                    class="button button--icon-before button--transparent button--secondary"
+                    href="{{ plan.get_absolute_url }}"
+                    aria-describedby="{{ plan.title }}"
```
svenvandescheur, line 24: aria-describedby should probably be aria-label, and an interactive element should not be made hidden to assitive technologies using aria-hidden as the latter describes more the availability of the item and not it's visual state.

src/open_inwoner/accounts/templates/registration/login.html
svenvandescheur, line 6: I don't think the label adds much value if the section is purely decorative. I think it's better to hide the section from asssistive technology.

 > Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. This can improve the experience for assistive technology users by hiding:
 >
 > Purely decorative content, such as icons or images
Duplicated content, such as repeated text
Offscreen or collapsed content, such as menus
 > 
 > The presence of the aria-hidden attribute hides content from assistive technology but doesn't visually hide anything.

src/open_inwoner/accounts/templates/registration/login.html
svenvandescheur, line 7: Can add a proper alt text here describing the content of the image?


 > The alt attribute holds a text description of the image, which isn't mandatory but is incredibly useful for accessibility — screen readers read this description out to their users so they know what the image means. Alt text is also displayed on the page if the image can't be loaded for some reason: for example, network errors, content blocking, or linkrot.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img

src/open_inwoner/components/templates/components/Button/Button.html
```diff
@@ -5,8 +5,9 @@
...
         {% if href|startswith:"http" %}target="_blank"{% endif %}
+        {% if hide_text %}aria-hidden="true"{% endif %}
```
svenvandescheur, line 10: Aria-hidden doesn't describe the visual state but whether it's available to assistive technologies, I think this should be removed.

I think the aria-label (or title) should always be available and should describe the link. Whether we hide the text is stylistic but in all cases a textual description should be available to assistive technologies.

 > aria-hidden="true" should not be used on elements that can receive focus. Additionally, since this attribute is inherited by an element's children, it should not be added onto the parent or ancestor of a focusable element.

 > The presence of the aria-hidden attribute hides content from assistive technology but doesn't visually hide anything.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

svenvandescheur (changes requested):

src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
svenvandescheur: > User agents, such as screen readers targeting disabled users, can use this element to determine whether to omit the initial rendering of navigation-only content.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

src/open_inwoner/accounts/templates/registration/login.html
alextreme: applied aria-hidden

src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
alextreme: renamed to "Secundaire navigatie", kept as is with aria-label

src/open_inwoner/components/templates/components/Button/Button.html
alextreme: Applied both changes, using aria-label instead of aria-describedby and removed aria-hidden

src/open_inwoner/templates/pages/category/detail.html
alextreme: Attempted to do this but might have to be applied on other places too

src/open_inwoner/templates/pages/category/detail.html
alextreme: Removed the aria-label here

src/open_inwoner/templates/pages/product/detail.html
alextreme: Renamed aria-label

alextreme:
Will be bypassing the other comments and proper review, sorry! Wanted to get the changes on test and staging to ensure our auditor doesn't flag accessibility issues we've already resolved.

PR #311, opened 2022-09-27
Added very basic Open Forms integration

joeribekker:
Very basic integration of Open Forms building on top of a detail page to show the form directly how it would show in Open Forms. This works with CSP headers and you can nicely select a form on the product admin page, once you configure Open Forms client in the admin.

Make sure you configure the environment variable "OPEN_FORMS_DOMAIN" to activate proper CSP headers. I didn't see any place to document this environment variable

Improvements can definitely be:

* Tests that adding a form actually results the form page being rendered with the expected snippet.
* Better page layout to show the full form.
* Use of the future 0.1.1 library django-open-forms-client because now there is a default UUID in the migration that I don't want there.
* Optionally, you might want to set the CSP-domein via database settings. We already add the domain in the OpenForm Configuration model, so might hook into that.


vaszig (commented):
- As discussed with Joeri as well, this shoud not be merged until it is updated with the newer version of open-forms-client.
- The domain (OPEN_FORMS_DOMAIN) should be added via an environmental variable for now.

joeribekker:
Updated the lib and this PR. No default uuid in the migration needed anymore.

vaszig (commented):
I think something goes wrong with some forms concerning the max length of the `SlugFiled` of Django. Is this happening to you as well?
image

joeribekker:
Weird, I thought slugfields had a length of 100? I'll check it out. I specifically used the same field as Open Forms to prevent exactly this problem, sigh :)

joeribekker:
Increased length to 100 chars. Note, you need to rollback and migrate again to make this effective.

vaszig (approved):
Tested with the 0.2.2 newer version and it works fine!

alextreme (approved):

PR #326, opened 2022-11-10
[#879] Added configuration fields for questionnaire block's (sub-)titles

Bartvaderkin:


alextreme (approved):
Works as expected, I would have named the configurable_text key differently than 'widgets' ('questionnaire_page' would have been more logical imho) but as a quick-win for Enschede this is fine.

Bartvaderkin:
I updated the context key as suggested and undrafted the PR

PR #329, opened 2022-11-15
[#912] Harden access control on case-details/documents

Bartvaderkin:
Continues from PR #327 & taiga 890

### DIFF
--- a/src/open_inwoner/accounts/views/cases.py
+++ b/src/open_inwoner/accounts/views/cases.py
@@ -16,7 +16,9 @@ from view_breadcrumbs import BaseBreadcrumbMixin
 
 from open_inwoner.openzaak.cases import (
     fetch_case_information_objects,
+    fetch_case_information_objects_for_case_and_info,
     fetch_cases,
+    fetch_roles_for_case_and_bsn,
     fetch_single_case,
     fetch_specific_statuses,
     fetch_status_history,
@@ -153,6 +155,10 @@ class CaseDetailView(
         case = fetch_single_case(case_uuid)
 
         if case:
+            # check if we have a role in this case
+            if not fetch_roles_for_case_and_bsn(case.url, self.request.user.bsn):
+                raise PermissionDenied()
+
             documents = self.get_case_document_files(case)
 
             statuses = fetch_status_history(case.url)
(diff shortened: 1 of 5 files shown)


vaszig (commented):
This has been reviewed in #327  except for the addition for the user's roles. So, it seems ok to me and can be merged after the changes-update of the mentioned PR.

Bartvaderkin:
I'll rebase this once the other changes land and remark for review as Anna hasn't looked at this yet.

Bartvaderkin:
I've rebased this on the PR for #890 and marked this as ready for review but note the review only covers the commit marked with [#912] (the others marked [#890] are part of the other PR)

src/open_inwoner/accounts/views/cases.py
```diff
@@ -153,6 +155,10 @@ def get_context_data(self, **kwargs):
         case = fetch_single_case(case_uuid)
 
         if case:
+            # check if we have a role in this case
+            if not fetch_roles_for_case_and_bsn(case.url, self.request.user.bsn):
+                raise PermissionDenied()
```
annashamray, line 160: TBH checking permissions in `get_context_data` seems odd and scattering permission checks between `test_func` and  `get_context_data` makes it hart to understand the authorization logic. 
I'd suggest putting all permissions in one place (`test_func` or `get` or `dispatch`)

Since we already use mixins for permissions and this code is also used in `CaseDocumentDownloadView` it can also be put in the separate mixin

src/open_inwoner/accounts/views/cases.py
```diff
@@ -252,12 +259,27 @@ def handle_no_permission(self):
         return super().handle_no_permission()
 
     def get(self, *args, **kwargs):
-        info_object_uuid = kwargs["object_id"]
+        case_uuid = kwargs["object_id"]
```
annashamray, line 262: I'd also suggest to put `test_func` and permission checks in `get` in one place

Bartvaderkin:
@annashamray I made an attempt to your suggestion to centralize the access checks: I dumped the regular mixins and turned them into a dispatch()-override because of the order in which the mixins super().dispatch() methods would execute (and we want auth/bsn check before we grab data).

Feel free to merge/append/refactor if this holds up the case confidentiality tickets

annashamray (approved):
Looks nice!

alextreme (approved):

PR #333, opened 2022-11-17
[#896] Added table with additional fields to case status-page dashboard

Bartvaderkin:


Bartvaderkin:
This needs a little bit of styling from @svenvandescheur, it can be added to this PR or done after merge.

src/open_inwoner/components/templatetags/dashboard_tags.py
```diff
@@ -59,3 +68,71 @@ def case_dashboard(case: dict, **kwargs) -> dict:
...
+    Variables:
+        + case: dict | The case to be able to build the dashboard, fetching the documents and statusses of the case.
+
+    Extra context:
+        + table: TableConfig | The configuration of the table.
+    """
+
+    # build rows for data we actually have
+    rows = []
+    if case.get("identification"):
```
annashamray, line 90:
```suggestion
    if case.get("initiator"):
```

src/open_inwoner/openzaak/utils.py
```diff
@@ -52,6 +53,54 @@ def is_zaak_visible(zaak: Zaak) -> bool:
     return is_object_visible(zaak, config.zaak_max_confidentiality)
 
 
+def get_role_identification_display(rol: Rol) -> str:
```
annashamray, line 56: this comment is optional, but it looks like this function will be better as `Rol` method

src/open_inwoner/openzaak/utils.py
```diff
@@ -52,6 +53,54 @@ def is_zaak_visible(zaak: Zaak) -> bool:
...
+    elif rol.betrokkene_type == RolTypes.medewerker:
+        display = join(
+            value("voorletters"),
+            value("voorvoegsel_achternaam"),
+            value("achternaam"),
+        )
+
+    if not display:
+        # fallback to generic role description
+        return RolTypes.labels.get(rol.betrokkene_type, rol.betrokkene_type)
```
annashamray, line 99:
```suggestion
        return rol.get_betrokkene_type_display()
```

src/open_inwoner/openzaak/utils.py
annashamray, line 56:
```suggestion
def get_role_name_display(rol: Rol) -> str:
```
You don't show here any id fields, so I think that `name` or `description` is better here

src/open_inwoner/openzaak/tests/test_utils.py
```diff
@@ -93,3 +97,105 @@ def test_filter_info_object_visibility(self):
                 ),
             )
             self.assertFalse(is_info_object_visible(info_object, "non_existent_key"))
+
+    def test_get_role_identification_display(self):
```
annashamray, line 101: Nice test coverage!

annashamray (changes requested):
I personally think that the styling can be also done in this PR
But it's up to you :)

src/open_inwoner/components/templatetags/dashboard_tags.py
```diff
@@ -16,6 +16,15 @@ class DashboardConfig(TypedDict):
     metrics: list[Metric]
 
 
+class Row(TypedDict):
```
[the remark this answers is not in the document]
Bartvaderkin: This is the custom type hint, it isn't real class but a PEP 589 TypedDict. 

So it doesn't product an instance with methods but just a dictionary. I did it to follow the pattern from the existing code.

src/open_inwoner/components/templatetags/dashboard_tags.py
```diff
@@ -59,3 +68,71 @@ def case_dashboard(case: dict, **kwargs) -> dict:
...
+        )
+    if case.get("end_date_legal"):
+        rows.append(
+            {
+                "label": _("Wettelijke termijn"),
+                "value": case.get("end_date_legal"),
+            }
+        )
+
+    table: TableConfig = {"rows": rows}
```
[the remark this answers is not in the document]
Bartvaderkin: A TypedDict isn't a normal class and doesn't produce a real instance but a normal dict. But I'll make it dryer another way.

src/open_inwoner/openzaak/utils.py
Bartvaderkin: If this works how we want we could move it to the zgw consumers.

Bartvaderkin:
@annashamray I applied your suggestions and replied to your notes on the TypedDict.

src/open_inwoner/components/templatetags/dashboard_tags.py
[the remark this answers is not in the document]
annashamray: right, sorry, I didn't notice that it inherits from `TypedDict`

annashamray (approved):
:+1:

src/open_inwoner/components/templatetags/dashboard_tags.py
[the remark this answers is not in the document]
annashamray: Right, my bad, I haven't paid attention that it inherits from `TypedDict`

Bartvaderkin:
Ok, lets merge so sven can style in develop.

PR #334, opened 2022-11-17
Fixed MM internal #470 -- Ensuring correct mimetype is used for images, enabling DigiD Mock by default when using docker-compose

alextreme:
image


joeribekker (approved):
❤️

PR #336, opened 2022-11-18
[#889] Optimize cases listview

vaszig:
Further optimization can be done when we solve the problem with parallel requests in tests.

- Only required case types are fetched (we were retrieving all of them)
- Cache is used for case types, status types and current statuses


src/open_inwoner/accounts/views/cases.py
```diff
@@ -88,17 +87,31 @@ def get_context_data(self, **kwargs):
...
+        case_types = {}
+        status_types = {}
+        temp_status_types = []
+        current_statuses = {}
+
+        for case in cases:
+            # Fetch new case types
+            if case.zaaktype not in case_types.keys():
+                case_type = fetch_single_case_type(case.zaaktype)
+                case_types.update({case_type.url: case_type})
```
Bartvaderkin, line 100: We can use assignment instead if `.update()`, like `case_types[case_type.url] = case_type` (a fewe more below)

vaszig:
As discussed during our meeting I am going to update the code according to Bart's comments and I will add some explanation to the cache decorator as well. So, I am sorry Anna but I will need some more time before being able to merge this!

src/open_inwoner/accounts/views/cases.py
```diff
@@ -88,17 +87,31 @@ def get_context_data(self, **kwargs):
...
+
+            # Fetch list of case's status types and each one (unique) separately
+            temp_status_types += [st.url for st in fetch_status_types(case.zaaktype)]
+
+            for status_type_url in set(temp_status_types):
+                if status_type_url not in status_types:
+                    status_type = fetch_single_status_type(status_type_url)
+                    status_types.update({status_type.url: status_type})
```
annashamray, line 108: Let's discuss what's happening here:

1. For each zaak you fetch statustypes for its zaaktype. 
So it you have 100 cases with 1 zaaktype you will fetch the same statustypen 100 times 

2. For each fetched statustype that you have in `fetch_status_types`, you fetch it again but individually in ` fetch_single_status_type`. Why to do this? They are already fetched, you have all the data. 
So if you have 100 cases with 1 zaaktype, which has 3 statustypen, you will fetch statustypen 100 times and then you will fetch the same statustypen 300 times 

Let's not do this. 
What I suggest:
* don't use `fetch_single_status_type` here, it's excessive
* cache `fetch_status_types`  function for a day

annashamray (changes requested):

src/open_inwoner/accounts/views/cases.py
vaszig: This was exactly the opposite thing of optimization.. Apologies for that, I was confused when I was adding the cache decorator.

annashamray (approved):
Looks nice and neat!

Bartvaderkin (approved):

alextreme:
:see_no_evil:

PR #340, opened 2022-11-21
:lipstick: #894 - Check card component for MijnAanvragen listview.

svenvandescheur:


src/open_inwoner/templates/pages/cases/list.html
```diff
@@ -1,54 +1,48 @@
...
-                </div>
+                {% render_column start=forloop.counter_0|multiply:4 span=4 %}
+                    {% render_card compact=True title=case.identificatie %}
+                        {% render_list %}
+                            {% list_item case.current_status caption=_("Status") strong=False %}
+                            {% list_item case.start_date caption=_("Ontvangstdatum") strong=False %}
+                            {% list_item case.description caption=_("Omschrijving") strong=False %}
+                        {% endrender_list %}
+
+                        {% link href="accounts:case_status" object_id=case.uuid icon="arrow_forward" primary=True text=_("Bekijk zaak") %}
```
annashamray, line 23:
```suggestion
                        {% link href="accounts:case_status" object_id=case.uuid icon="arrow_forward" primary=True text=_("Bekijk aanvraag") %}
```

src/open_inwoner/components/templates/components/List/ListItem.html
```diff
@@ -1,7 +1,12 @@
 {% load link_tags %}
 <li class="list-item{% if active %} list-item--active{% endif %}">
     {% if href %}<a class="link" href="{{ href }}">{% endif %}
-    <h4 class="h4">{{ text }}</h4>
+    {% if caption %}<p class="p p--muted">{{ caption }}</p>{% endif %}
```
annashamray, line 4: I don't think it should be muted color:
* `  --color-mute: var(--color-gray);` and our `-color-gray` is very light and not very readable
* according to the design it should be the same color as headers in the action list, for example, which we implemented much darker

src/open_inwoner/components/templates/components/Card/RenderCard.html
```diff
@@ -1,7 +1,7 @@
 {% load link_tags %}
 
 {# Utilizes same template as Card. #}
-<div class="card{% if inline %} card--inline{% endif %}{% if tinted %} card--tinted{% endif %}{% if type %} card--type-{{ type }}{% endif %}">
+<div class="card{% if compact %} card--compact{% endif %}{% if inline %} card--inline{% endif %}{% if tinted %} card--tinted{% endif %}{% if type %} card--type-{{ type }}{% endif %}">
```
annashamray, line 4: I can't find `card--compact` CSS styles, where are they located?

src/open_inwoner/templates/pages/cases/list.html
```diff
@@ -1,54 +1,48 @@
...
```
annashamray, line 42:
```suggestion
                        {% link href="accounts:case_status" object_id=case.uuid icon="arrow_forward" primary=True text=_("Bekijk aanvraag") %}
```

annashamray (changes requested):
Please make CI green and happy

src/open_inwoner/components/templates/components/List/ListItem.html
svenvandescheur: I agree that the contrast is too low, but aren't the colors configurable and therefore; shouldn't the settings be changed in such a way that muted texts remain readable, even though they require less attention that the more relevant information?

Changing the implementation to suit the configuration would IMO make the configuration unreliable.

src/open_inwoner/components/templates/components/Card/RenderCard.html
svenvandescheur: I might missed something; will check.

src/open_inwoner/components/templates/components/List/ListItem.html
annashamray: Do you mean that we can configure colors in the admin? I think we can't configure mute color there

Anyway I think header should be styled the same way as headers in action list, it will solve all problems

svenvandescheur:
> Apart from @annashamray comments I see some differences comparing it with the design.
> 
> * There are no lines between the different fields (Personally I prefer @svenvandescheur approach)
> * When we have larger text in one of the three cards they are not consistent
> 
> image

Fixed

alextreme (approved):

src/open_inwoner/templates/pages/cases/list.html
annashamray: Now for open cases it's "Bekijk zaak" and for closed ones "Bekijk aanvraag"

annashamray (changes requested):
Only one minor remark about button text is left

I personally think that it would be prettier if the header, button and card text would be aligned on the left (now the text has offset), but if Alex is ok, I don't mind
image

alextreme:
Well, as long as the PR isn't merged I can change my mind :) 

Putting this in line with the design would be the way to go. So without the lines between the fields, and aligning the text to the left:
image

alextreme (changes requested):
Ensure this is in line with the design

annashamray (changes requested):
My comment with different button texts is still unresolved

alextreme:
I've applied Anna's suggestion

alextreme (approved):

PR #342, opened 2022-11-21
:sparkles: - Use htmx to async update messages.

svenvandescheur:


annashamray:
Is it possible to unit test? (if not then should we really use it?)

svenvandescheur:
> Is it possible to unit test? (if not then should we really use it?)

It uses the same logic as the previous implementation except async. The actual implementation is iirc 4 properties, testing those would just be retesting the htmx library.

annashamray:
I still think this should be tested. If unit testing is not a good idea, end-to-end tests could be introduced

svenvandescheur:
> I still think this should be tested. If unit testing is not a good idea, end-to-end tests could be introduced

Ok, let's see if I can add some wettest to. test the integration.

svenvandescheur:
@annashamray I've added selenium for tests for this.

alextreme (approved):
Does seem like a bit of HTML magic, but lets try this out.

@annashamray see if you can help Sven with the requirements FIXME for adding selenium

annashamray:
@alextreme dependencies didn't compile because there was a version conflict between our `certifi` and `selenium` `certifi`. I updated our version and it seems to work

src/open_inwoner/components/templates/components/Form/Form.html
```diff
@@ -9,6 +9,12 @@
     {% if data_confirm_title %}data-confirm-title="{{ data_confirm_title }}"{% endif %}
     {% if data_confirm_cancel %}data-confirm-cancel="{{ data_confirm_cancel }}"{% endif %}
     {% if data_confirm_default %}data-confirm-default="{{ data_confirm_default }}"{% endif %}
+    {% if async_selector %}
+        hx-{{ method|lower }}="{{ action|default:"./" }}"
```
annashamray, line 13: `action` variable is not documented in template tags

alextreme:
Great thanks @annashamray !

src/open_inwoner/components/templates/components/Form/Form.html
```diff
@@ -9,6 +9,12 @@
...
+        hx-select="{{ async_selector }}"
+        hx-target="{{ async_selector }}"
```
annashamray, line 15: Should we also add polling for messages? According to the docs it can be done with 1 attribute
@alextreme @svenvandescheur what do you think?

src/open_inwoner/components/templates/components/Form/Form.html
```diff
@@ -9,6 +9,12 @@
...
+        hx-swap="outerHTML"
```
annashamray, line 16: Why to use this tag? I commented it and nothing changed

src/open_inwoner/accounts/tests/test_inbox_page.py
```diff
@@ -117,3 +121,59 @@ def test_mark_messages_as_seen(self):
 
         self.assertFalse(message_sent.seen)
         self.assertTrue(message_received.seen)
+
+
+class MySeleniumTests(StaticLiveServerTestCase):
```
annashamray, line 126: Nice! Perhaps in the future we can also test it for Chrome

annashamray (changes requested):

src/open_inwoner/components/templates/components/Form/Form.html
annashamray: https://htmx.org/docs/#polling

.github/workflows/code-quality.yml
```diff
@@ -58,7 +58,7 @@ jobs:
         uses: actions/checkout@v2
       - name: Install dependencies
         run: |
-          npm install
+          npm install --legacy-peer-deps
```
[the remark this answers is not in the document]
svenvandescheur: I've added this line because sometimes the build seems to fail, maybe the runners are migrating to node. I preferably stick with a more recent (latest) node on the ci so we catch stuff early. Figuring out why/if we need legacy-peer-deps is IMO another task.

alextreme (approved):
Can be merged after Anna's approval

src/open_inwoner/components/templates/components/Messages/Messages.html
```diff
@@ -8,7 +8,12 @@ <h4 class="h4">{{ subject }}</h4>
 
 
     <div class="messages__body">
-        <ol class="messages__days">
+        <ol
+            class="messages__days"
+            hx-get="./"
+            hx-select=".messages__days"
+            hx-trigger="every 2s"
```
annashamray, line 15: Should we make it 5 seconds? @alextreme  ?

src/open_inwoner/components/templatetags/form_tags.py
```diff
@@ -44,6 +44,7 @@ def render_form(parser, token):
         {% endrender_form %}
 
     Variables:
+        + action: string | The action attribute of the form.
```
annashamray, line 47: I've just noticed that we already have `form_action` variable. Are they duplicates?

src/open_inwoner/components/templates/components/Messages/Messages.html
```diff
@@ -8,7 +8,12 @@ <h4 class="h4">{{ subject }}</h4>
 
 
     <div class="messages__body">
-        <ol class="messages__days">
+        <ol
+            class="messages__days"
+            hx-get="./"
```
annashamray, line 13: Please add test for polling

annashamray (changes requested):
Polling doesn't work on my local machine. Does it work for you?

svenvandescheur:
> Polling doesn't work on my local machine. Does it work for you?

yes

Bartvaderkin (approved):

src/open_inwoner/components/templates/components/Form/Form.html
annashamray: `action` is not documented again

annashamray (changes requested):
I have one remark about documentation which is not resolved 

Polling still doesn't work for me, but maybe the problem is in my config, so I'll approve after the template variable is documented

annashamray:
@Bartvaderkin @svenvandescheur 
Polling works on your local machine right? How do you test it?

My steps:
1. log in as use A, go to "Mijn berichten" page with user B
2. in other session hijack user B, go to "Mijn berichten" and send message to user A
3. Wait for this message to appear in the user A session
The message doesn't appear

svenvandescheur:
> @Bartvaderkin @svenvandescheur Polling works on your local machine right? How do you test it?
> 
> My steps:
> 
> 1. log in as use A, go to "Mijn berichten" page with user B
> 2. in other session hijack user B, go to "Mijn berichten" and send message to user A
> 3. Wait for this message to appear in the user A session
>    The message doesn't appear

I made screencapture:

oippolling

annashamray:
Thanks! I'll test it on our test env after deployment and will try to understand why it doesn't work locally for me

annashamray (approved):

PR #343, opened 2022-11-21
[#892] open cases view

annashamray:


annashamray:
I separated case list tests into 3 files and refactored them a little bit in order to test 2 views instead of 1
Sorry about that

alextreme (approved):

src/open_inwoner/accounts/views/cases.py
```diff
@@ -145,27 +131,67 @@ def get_cases(self):
...
+    def get_cases(self):
+        all_cases = super().get_cases()
+
+        cases = [case for case in all_cases if case.einddatum]
+        cases.sort(key=lambda case: case.einddatum, reverse=True)
+        return cases
+
+    def get_anchors(self) -> list:
+        return [
+            (reverse("accounts:my_closed_cases"), _("Lopende aanvragen")),
```
vaszig, line 189: This should redirect to `"accounts:my_open_cases"` , right?Now both anchors redirect to closed cases.

vaszig (commented):
Except for a small change in anchors it looks nice. Tests refactoring also looks nice.

src/open_inwoner/accounts/views/cases.py
annashamray: Nice catch!

Bartvaderkin (approved):
:+1:

PR #344, opened 2022-11-22
[#898] Added 'delete' operation to menu in Action list

Bartvaderkin:
Note it is a soft-delete.

Also note the Actions are a separate feature but also part of Plans/Samenwerken


src/open_inwoner/accounts/urls.py
```diff
@@ -57,7 +58,7 @@
     path("actions/<str:uuid>/edit/", ActionUpdateView.as_view(), name="action_edit"),
     path(
         "actions/<str:uuid>/delete/",
-        ActionUpdateView.as_view(),
+        ActionDeleteView.as_view(),
         name="action_delete",
     ),
```
Bartvaderkin, line 63: Weird: there already existed an `action_delete` that wasn't used/referred.

src/open_inwoner/components/templates/components/Action/Actions.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n form_tags icon_tags button_tags dropdown_tags action_tags link_tags %}
 
-{% render_form form=form method="GET" spaceless=True autosubmit=True form_action=form_action %}
 <div class="actions">
+    {% render_form form=form method="GET" spaceless=True autosubmit=True form_action=form_action %}
```
Bartvaderkin, line 4: @svenvandescheur I had to move the top search-filter form-tag so it doesn't wrap the whole table and I could nest mini-forms to do the POST to the delete view, with the confirm and everything.

src/open_inwoner/plans/views.py
```diff
@@ -320,6 +321,21 @@ def get_success_url(self) -> str:
         return self.object.get_absolute_url()
 
 
+class PlanActionDeleteView(ActionDeleteView):
+    # TODO do we need to do something depending on related users? (see the view above)
```
Bartvaderkin, line 325: :question: do we?

src/open_inwoner/accounts/tests/factories.py
```diff
@@ -58,6 +58,7 @@ class Meta:
 
     name = factory.Faker("first_name")
     created_by = factory.SubFactory(UserFactory)
+    is_deleted = False
```
vaszig, line 61: Is this required?Since it is False by default.

src/open_inwoner/plans/views.py
vaszig: I think we should inform the other users about deleting the action. Since we do that when a user edits a plan action. But, Alex and Anna can tell us their opinion.

src/open_inwoner/templates/export/plans/plan_export.html
```diff
@@ -43,6 +43,6 @@ <h2 class="h2 h2--extra">{% trans "Bestanden" %}</h2>
     {% endif %}
 
     <h2 class="h2 h2--extra">{% trans "Acties" %}</h2>
-    {% include "export/include/actions.html" with actions=object.actions.all %}
+    {% include "export/include/actions.html" with actions=object.actions.visible %}
```
vaszig, line 46: This needs update in the profile as well. We also show the number of open actions in the this page and export them there.

src/open_inwoner/accounts/tests/factories.py
Bartvaderkin: Hmm.. not really. I'll remove it.

src/open_inwoner/templates/export/plans/plan_export.html
Bartvaderkin: Good catch! I found the places in the profile and will update those as well.

src/open_inwoner/plans/views.py
Bartvaderkin: I've discussed this with Alex and we want the notification to be send, so I got to add that.

src/open_inwoner/accounts/models.py
```diff
@@ -503,6 +503,10 @@ class Action(models.Model):
         related_name="actions",
         help_text=_("The plan that the action belongs to. This can be left empty."),
     )
+    is_deleted = models.BooleanField(
+        verbose_name=_("Is deleted"),
+        default=False,
```
annashamray, line 508: Could you please add `help-text` here with info about soft-delete?

annashamray (commented):
It looks nice, I'll approve it after the notification is implemented.

In order not to change the queryset in all action-related things (which we have a lot here), I think we can have two model managers:
* in the default ('objects') we can rewrite `get_queryset` and `delete` methods 
* in the second model manager (`objects_full`) we can show all objects and use it in the admin 

In this case we won't worry that we forgot add `.visible()` somewhere in the code

You approach is also valid and I'll approve it after it notification part is finished

annashamray (approved):
:+1:

PR #346, opened 2022-11-23
:white_check_mark: - Add tests for compile_dependencies.

svenvandescheur:


alextreme:
I'm neutral about this one, @annashamray and @vaszig what do you think?

annashamray:
I don't mind but I think it should be run only if requirement of the `bin/compile_dependencies.sh` file is changed
We don't need it to run every time 
@svenvandescheur could you add `paths` to the gh action config?

vaszig:
I am not very familiar with this kind of checking or optimizations but I agree with Anna concerning when we should run this.

annashamray (changes requested):
Please limit the events which trigger the action

annashamray (changes requested):

annashamray (approved):

PR #347, opened 2022-11-23
[#901, #902, #903 ] Implemented Action-status toggle and added HTMX to backend

Bartvaderkin:
It is quite a lot for a simple button, but this also adds htmx and django-htmx to the project. 

Additionally I've added some infrastructure and related tests to be able to render a template-tag to a string. 

And then tied that into a base class for a View to handle hx-post requests and respond with a rendered template-tag (this gives convenient partial responses without extra boilerplate templates or hx-select's).

And then expanded on Sven's selenium integration testing.

And then the functionality exists in two flavours (plain Actions and Actions via Plans) so implementation and test cases almost double.

image


src/open_inwoner/templates/master.html
```diff
@@ -75,5 +75,6 @@ <h2 class="modal__title" id="modal__title"></h2>
         </div>
 
         <script nonce="{{ request.csp_nonce }}" src="{% static 'bundles/open_inwoner-js.js' %}" type="text/javascript"></script>
+        {% django_htmx_script %}
```
Bartvaderkin, line 78: This tag adds some development features when DEBUG=True

src/open_inwoner/components/utils.py
```diff
@@ -132,3 +135,118 @@ def render(self):
                 kwargs,
                 self.template_name,
             )
+
+
+class RenderableTag:
```
Bartvaderkin, line 140: This is a bit of a black box but the main part from InclusionTagWebTest has been used for testing for a while.

annashamray: Sorry, I don't understand why you use it. 
Can't we just use django `TemplateResponse` or `render` shortcut? It does the same things - resolves the template with it's name and then render the content. In the htmx view we return a piece of HTML, why not to do it the way django usually does it?

annashamray (commented):
`django-htmx` looks interesting,my main question is about the necessity of `RenderableTag`

src/open_inwoner/components/utils.py
Bartvaderkin: Because that would require to create boilerplate template files that would only have a {% load xyz %} and the tag we want to render with a repeat of all the arguments. This class basically creates that on the fly so we need less files and repeating of argument names/values.

src/open_inwoner/htmx/views.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    template_tag: RenderableTag = None
+
+    def get_template_tag_args(self, context: dict) -> dict:
+        """
+        override me and return a dict with the arguments of the template tag
+        """
+        return {
+            "request": self.request,
```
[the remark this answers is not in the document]
Bartvaderkin: So you can override tag arguments separately from context, just like a tag in the html doesn't always use the same named variables from the context.

Bartvaderkin:
@svenvandescheur Can you sign-off on this so we can get it merged? Afterwards we need a little style update on the dropdown and button style, see Anna's comment: https://github.com/maykinmedia/open-inwoner/pull/347#discussion_r1038261864

alextreme:
Happy to merge after Anna or Vasileios approves. Create a separate Taiga task for any styling still necessary

vaszig (approved):
The status toggle works as expected concerning the shared plans now. If Anna is also ok this can be merged.

annashamray (approved):
If other developers are ok with this approach, I approve. 
I'm curious where this approach would be developing, but it looks like worth experimenting

PR #348, opened 2022-11-24
[#929] Refactor contacts

vaszig:
- Removes Contact model
- Adds user_contacts and contacts_for_approval (field for invitations) to the User model
- Removes phonenumber from contact creation and the ability to modify a contact


src/open_inwoner/accounts/admin.py
```diff
@@ -60,6 +60,10 @@ class _UserAdmin(UserAdmin):
                 ),
             },
         ),
+        (
+            _("Contacts - invites"),
+            {"fields": ("user_contacts", "contacts_for_approval")},
```
alextreme, line 65: I would also add these to filter_horizontal (or perhaps raw_id_fields). The default M2M admin widget in Django is annoying and shouldn't be used imho.

src/open_inwoner/accounts/migrations/0048_auto_20221124_1515.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def populate_user_contacts(apps, schema_editor):
+    MyModel = apps.get_model("accounts", "Contact")
+    for contact in MyModel.objects.exclude(contact_user__isnull=True):
+        creator = contact.created_by
+        contact_user = contact.contact_user
+        creator.user_contacts.add(contact_user)
+
+        contact_user.first_name = contact.first_name
```
alextreme, line 22: populating user_contacts is appreciated, however I would avoid setting the first_name/last_name/phonenumber/function fields on the User in the migration.

alextreme:
Nice refactor! Went over it early to ensure we're on the same page, think this is going in the right direction :+1:

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -105,58 +65,66 @@ def crumbs(self):
...
+                    contact=cleaned_data["email"]
                 )
             ),
         )
-
-        self.log_addition(self.object, _("contact was created"))
         return HttpResponseRedirect(self.get_success_url())
 
 
 class ContactDeleteView(LogMixin, LoginRequiredMixin, DeleteView):
```
alextreme, line 108: I'd recommend to not use a DeleteView here. You're not deleting a User :)

src/open_inwoner/plans/admin.py
```diff
@@ -29,3 +29,5 @@ class PlanAdmin(UUIDAdminFirstInOrder, admin.ModelAdmin):
         "created_by",
     )
     inlines = [ActionInlineAdmin]
+    filter_horizontal = ("plan_contacts",)
+    exclude = ("contacts",)
```
alextreme, line 33: Is this exclude still necessary?

src/open_inwoner/plans/managers.py
```diff
@@ -3,13 +3,4 @@
...
-            Q(created_by=user)
-            | Q(contacts__contact_user=user)
-            | Q(contacts__created_by=user)
-        ).distinct()
-
-    def shared(self, user):
-        return self.filter(
-            Q(contacts__contact_user=user) | Q(contacts__created_by=user)
-        ).distinct()
+        return self.filter(Q(created_by=user) | Q(plan_contacts__id=user.id)).distinct()
```
alextreme, line 6: Maybe it makes sense to auto-add the creating user also to plan_contacts?

alextreme (changes requested):
Nice work! Big PR but this was imho as discussed, only a few minor comments

vaszig:
I have processed @alextreme 's reccomendations and I will push the changes as soon as the other guys have reviewed as well (I don't want to push new commits in case they have started their reviews).

@alextreme can I have a bit of explanation for the third comment?About adding the creating user to the plan_contacts

alextreme:
> @alextreme can I have a bit of explanation for the third comment?About adding the creating user to the plan_contacts

Well, if we want to simplify the Plans further we could make it so that the User creating the Plan is both registered in created_by and in the plan_contacts M2M. This way you no longer would need to check both the created_by FK and the plan_contacts M2M, you would only have to check plan_contacts.

Up to you, Bart and Anna to determine if this makes sense, but as we're refactoring the Plans significantly it would make sense to also do this if you three agree with me. Otherwise you can leave it as-is.

Bartvaderkin:
@vaszig I haven't started on this today so I don't mind new commits.

Bartvaderkin (changes requested):
I've added some code style feedback.

But the most important thing I'm missing is a test for the significant data migration. 

Here you can find a base class `TestMigrations` and some usage examples for a migration test: https://github.com/open-formulieren/open-forms/search?q=TestMigrations

Also I ran the migrations on my local database and I get some styling issues (but note the names were messy test data so the gaps might not mean anything)

Screenshot from 2022-11-30 11-04-46

Bartvaderkin:
Here you can find a base class `TestMigrations` and some usage examples for a migration test: https://github.com/open-formulieren/open-forms/search?q=TestMigrations

vaszig:
> 

- Changes according to comments have been made
- Tests for data migrations have been added but one is failing, I will investigate this with @Bartvaderkin 
- Concerning the styling, since we don't have a design for the creator's contactlist page,  I am only showing the contacts' data (first name, last name) if they have been approved.

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -22,35 +22,55 @@ <h1 class="h1" id="title">
...
-        {% endfor %}
+        {% else %}
+            {% for invite in pending_invitations %}
+                <tr>
+                    <th class="table__header">{{invite.invitee_first_name}} {{invite.invitee_last_name}}</th>
+                    <td class="table__item"></td>
+                    <td class="table__item">{{invite.invitee_email}}</td>
+                    <td class="table__item"></td>
+                    <td class="table__item">{% icon "check" extra_classes="icon icon--disabled" %}</td>
+                    <td class="table__item">{% button text=_('Stuur bericht') disabled=True %}</td>
```
annashamray, line 37: I agree with Bart - styling issue with disabled buttons should be addressed. It looks like the styling was messed up when `background-color` attribute was added into `.button--disabled` style

It can be done in the separate PR (perhaps, Sven can help). 
@vaszig  Please create a task for it

src/open_inwoner/utils/tests/test_migrations.py
```diff
@@ -0,0 +1,39 @@
+from django.db import connection
+from django.db.migrations.executor import MigrationExecutor
+from django.test import TestCase
+
+
+class TestMigrations(TestCase):
```
annashamray, line 6: This is cool!

src/open_inwoner/plans/managers.py
annashamray: Agree with Alex, it will simplify code in several places

src/open_inwoner/plans/forms.py
```diff
@@ -21,15 +21,15 @@ class PlanForm(forms.ModelForm):
...
-        extended_contacts = user.get_extended_active_contacts()
-        self.fields["contacts"].queryset = extended_contacts
-        self.fields["contacts"].choices = [
-            [c.id, f"{c.other_user_first_name} {c.other_user_last_name}"]
-            for c in extended_contacts
+
+        user_contacts = user.get_active_contacts()
+        self.fields["plan_contacts"].queryset = user_contacts
+        self.fields["plan_contacts"].choices = [
+            [c.id, c.get_full_name()] for c in user_contacts
```
annashamray, line 32: same comment as in contact form: I think we don't need to add choices manually here anymore. This line is unnecessary now

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -45,11 +41,16 @@ def test_contact_list_only_show_personal_contacts(self):
...
+        response = self.app.get(self.list_url, user=self.user)
+        self.assertContains(response, existing_user.first_name)
+
     def test_contact_filter(self):
-        begeleider = ContactFactory(
-            contact_user__contact_type=ContactTypeChoices.begeleider,
-            created_by=self.user,
-        )
+        begeleider = UserFactory(contact_type=ContactTypeChoices.begeleider)
+        self.user.user_contacts.add(begeleider)
```
annashamray, line 52: please check also that unwanted `contacts_for_approval` and `pending_invitations` are not shown

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -174,24 +127,59 @@ def test_contact_create_send_invite(self):
...
+            "__all__": [
+                _(
+                    "Het ingevoerde e-mailadres komt al voor in uw contactpersonen. Pas de gegevens aan en probeer het opnieuw."
+                )
+            ]
+        }
         self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context["form"].errors, expected_errors)
 
+    def test_email_required(self):
```
annashamray, line 163: nice test coverage!
Please add a test when a user tries to create contact with inactive user

annashamray (dismissed):
Good job! 
It's much neater and more readable when before! The simplified querysets have brought a joy to my heart :)

src/open_inwoner/accounts/forms.py
```diff
@@ -120,42 +121,27 @@ class ContactFilterForm(forms.Form):
...
         cleaned_data = super().clean()
         email = cleaned_data.get("email")
 
-        if self.create and email:
-            if (
-                self.user.contacts.filter(email=email).exists()
-                or self.user.assigned_contacts.filter(created_by__email=email).exists()
-            ):
+        if email:
+            if self.user.user_contacts.filter(email=email).exists():
```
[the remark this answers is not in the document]
vaszig: You are right, I will add it.

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -31,69 +30,30 @@ def crumbs(self):
...
+            base_qs = base_qs.filter(contact_type=type_filter)
+
+        return base_qs.order_by("-date_joined")
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+
+        user = self.request.user
+        context["pending_approvals"] = user.get_contacts_for_approval()
+        context["pending_invitations"] = user.get_pending_contacts()
```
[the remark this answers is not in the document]
vaszig: Same as above.

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -31,69 +30,30 @@ def crumbs(self):
...
-        return base_qs
+            base_qs = base_qs.filter(contact_type=type_filter)
+
+        return base_qs.order_by("-date_joined")
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+
+        user = self.request.user
+        context["pending_approvals"] = user.get_contacts_for_approval()
```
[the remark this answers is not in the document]
vaszig: I did not apply a filter because I thought since it's not a contact yet we shouldn't give extra information (except the email). I will change this if we should.

src/open_inwoner/accounts/models.py
```diff
@@ -20,21 +20,41 @@
...
     )
     email = models.EmailField(verbose_name=_("Email address"), unique=True)
+    phonenumber = models.CharField(
+        verbose_name=_("Phonenumber"),
+        blank=True,
+        default="",
+        max_length=15,
+        validators=[validate_phone_number],
+    )
+    function = models.CharField(
```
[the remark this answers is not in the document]
vaszig: As far as I am concerned it's not needed. @alextreme can confirm that and I can remove it.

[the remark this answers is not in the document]
alextreme: We don't use 'function' afaik, it can be removed. It also conflicts a bit with the 'contact_type' field so would be better to not have it in.

alextreme (approved):

src/open_inwoner/accounts/views/contacts.py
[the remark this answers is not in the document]
vaszig: Discussed during weekly meeting, we will not apply a filter here.

src/open_inwoner/accounts/forms.py
```diff
@@ -305,12 +291,10 @@ def __init__(self, user, **kwargs):
 
         super().__init__(**kwargs)
 
-        extended_contact_users = User.objects.get_extended_contact_users(self.user)
-        choices = [
-            [u.email, f"{u.first_name} {u.last_name}"] for u in extended_contact_users
-        ]
+        contact_users = self.user.get_active_contacts()
+        choices = [[u.email, f"{u.first_name} {u.last_name}"] for u in contact_users]
```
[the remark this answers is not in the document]
vaszig: Discussed with Anna, it complicates things rather than simplifying them.

PR #349, opened 2022-11-28
[#899] Reselect template and contact on Plan-create validation error

Bartvaderkin:
This was caused by custom widget template code doing an `in` operator, but the primary key of an object is an integer while the querydict has strings so it would never find the value.


annashamray:
https://github.com/maykinmedia/open-inwoner/pull/348 should be merged before this PR

Bartvaderkin:
I updated this on #348

annashamray (approved):

PR #350, opened 2022-11-28
[#909] Added FAQ to Product page

Bartvaderkin:


vaszig (commented):
It looks nice to me. The only thing, which I don't know if we want to add, is to protect the addition of both category and product generally and not only via a form (as you have already done in the clean method).

Bartvaderkin:
@vaszig Good point. I'll try to add a `CheckConstraint` so it can't save even if something bypasses the `clean()`.

annashamray (approved):
Looks very nice! :+1:

PR #352, opened 2022-11-29
:lipstick: #941 - Stylen tabel aanvraag details.

svenvandescheur:


Bartvaderkin (approved):
Splitting the table into tags for every element seems like a lot but it does follow the pattern.

PR #354, opened 2022-12-01
[#948] Harden `fetch_roles_for_case_and_bsn()` to work with e-Suite

Bartvaderkin:
e-Suite doesn't support query on both zaak and BSN

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_case_detail.py
+++ b/src/open_inwoner/openzaak/tests/test_case_detail.py
@@ -151,15 +151,29 @@ class TestCaseDetailView(WebTest):
             volgnummer=2,
             is_eindstatus=False,
         )
-        self.role = generate_oas_component(
+        self.user_role = generate_oas_component(
             "zrc",
             "schemas/Rol",
-            url=f"{ZAKEN_ROOT}rollen/f33153aa-adnatuurlijk_persoon2c-4a07-ae75-15add5891",
+            url=f"{ZAKEN_ROOT}rollen/f33153aa-ad2c-4a07-ae75-15add5891",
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -253,12 +267,12 @@ def _setUpMocks(self, m):
...
-            f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}&omschrijvingGeneriek={RolOmschrijving.initiator}",
-            json=paginated_response([self.role]),
+            f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}",
+            json=paginated_response([self.user_role, self.not_our_user_role]),
         )
         m.get(
-            f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}&betrokkeneIdentificatie__natuurlijkPersoon__inpBsn={self.user.bsn}",
-            json=paginated_response([self.role]),
+            f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}&omschrijvingGeneriek={RolOmschrijving.initiator}",
+            json=paginated_response([self.user_role]),
```
Bartvaderkin, line 275: Learning moment: requests.mocker matches on partial URLs in reverse order: 

I had to move the mock with both `zaak` and `omschrijving` below the one with only a `zaak` or it would always return the one with only `zaak` even if my request had the `omschrijving` parameter.

alextreme (approved):
Lets try this out, I've let Andy know about hitting this limitation in e-Suite so this might be a workaround we can remove in the future.

PR #358, opened 2022-12-02
Feature/895 pagination

annashamray:
Since we have to filter cases on client side (separate open cases from closed) we can't use `page` query param to Zaken API for pagination and we have to implement it also on the client side. 

There were **3 options** considered to implement pagination:
1. Each time request all cases, filter them and then paginate
  \+ We show client all the data
  \- Performance-heavy, especially on staging environment, where we have thousands cases for 1 person
2. Use `ordering` query param when requesting cases from Zaken API. 
  \+ It would be an optimal solution: we could order cases on the end date and limit the request to only open ones or closed ones.
  \- Too bas this query param doesn't work in eSuite
3. Set the maximum amount of requests we do to cases resource in Zaken API
  \+ Improve performance
  \- We don't show all cases
  \- We limit requests, not the amount of cases, and then we separate open and closed ones, it means it is possible to show the user 3 closed cases and hide 30 older ones. But for regular user in production it is very unusual situation

@alextreme and I decided to stick to the 3rd option for now, because it won't affect users in production much, since it's rare for a user to have hundreds of cases, but it will improve our performance on staging env

* only 100 cases can be retrieved and processed, e-Suite returns 20 cases per page, so no more than 5 requests to cases could be done
* these 100 cases are then separated on the open and closed ones, ~~each of them we limit to 50 cases. Not sure if this step is necessary, I'd like to hear your opinion~~ **Upd. removed thanks to** @alextreme 
* these open or closed cases are paginated. We request related resources (statuses, documents. etc) only for the cases shown on the page

### DIFF
--- a/src/open_inwoner/accounts/views/cases.py
+++ b/src/open_inwoner/accounts/views/cases.py
@@ -89,7 +89,6 @@ class CaseAccessMixin(AccessMixin):
 
 class CaseListMixin(PaginationMixin):
     paginate_by = 9
-    max_cases = 50
     template_name = "pages/cases/list.html"
 
     def get_cases(self):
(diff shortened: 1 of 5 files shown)


vaszig (approved):
I agree that the best approach is the 3rd one. It looks nice (CI fails though) but I cannot test it completely in my local machine as I fetch only a few cases using the bsn we usually use. I assume we can see that in the right way when we deploy.

Bartvaderkin (approved):
My only observation is the paginator shows when there aren't any cases but that might be common for all list views.

PR #359, opened 2022-12-05
[#952] Use client and server certificates to download documents from ZGW backend

Bartvaderkin:


alextreme (approved):
Seems in order, currently can't be tested on any other environment than acc but implementation seems okay. Can be merged after CI passes

PR #362, opened 2022-12-06
Added Github workflow "Check for model changes not present in the migrations"

Bartvaderkin:
This runs `manage.py makemigrations --check --dry-run` as part of the build, which will fail if there are changes in the model code that aren't part of the migrations up to this point.

So when you run makemigrations we won't have mystery migrations show up for code you didn't touch and have to decide if it was something important or just a label or choices change.

This also means a label change requires a migration.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -49,6 +49,47 @@ jobs:
       - name: Run black
         run: |
           black --check --diff src
+
+  migrations:
+    name: Check for model changes not present in the migrations
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:12
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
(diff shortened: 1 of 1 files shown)


Bartvaderkin:
This is expected (but needs to be solved):

```
Migrations for 'mail_editor':
  /opt/hostedtoolcache/Python/3.9.15/x64/lib/python3.9/site-packages/mail_editor/migrations/0012_auto_20221206_1434.py
    - Alter field id on mailtemplate
    - Alter field language on mailtemplate
Error: Process completed with exit code 1.
```

annashamray:
Good idea :+1:

alextreme:
Discussed, can only be included after fixing the missing migration in maileditor

Bartvaderkin:
@vaszig I'm merging this, so the Code Quality workflow will fail if there are changes in the models that don't have a migration.

PR #363, opened 2022-12-06
[#897] Feature/contacts approval flow

vaszig:


src/open_inwoner/accounts/views/contacts.py
```diff
@@ -125,3 +130,34 @@ def post(self, request, *args, **kwargs):
...
+            self.update_contact(sender, receiver, (approved or rejected))
+            return HttpResponseRedirect(self.success_url)
+
+    def update_contact(self, sender, receiver, type_of_approval):
+        if type_of_approval == "approve":
+            sender.contacts_for_approval.remove(receiver)
+            sender.user_contacts.add(receiver)
+            self.log_change(sender, _("contact was approved"))
+
+        elif type_of_approval == "reject":
```
Bartvaderkin, line 161: I think it is convention to not use raw POST but always try to go for a Form, and use Choices instead of magic strings so validation is a bit harder and more regular. 

The two buttons can have the same name in the HTML but different values and then the Form is just a CharField with the choices for the incoming data.

Also, looking at the post method: what happens if there is no approved or rejected? It could `raise BadRequest` since there is no HTML for this view.

Bartvaderkin (changes requested):
I put my notes in the code. 

I also noticed this that might existed before: maybe these buttons could be closer? The Accept button is really far from the Reject, and the Reject button label says Terug (Back)

Screenshot from 2022-12-07 14-15-50

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -22,11 +61,13 @@ <h1 class="h1" id="title">
         </tr>
     </thead>
     <tbody>
-        {% if not pending_invitations and not pending_approvals and not object_list %}
+        {% if not pending_invitations and not contacts_for_approval and not pending_approval and not pending_approval_list and not object_list %}
```
Bartvaderkin, line 64: (this is a lot.. but then I don't know how to do it differently without making it worse so meh :wink: )

Bartvaderkin:
I forgot to mention this: there is a small CSS bug in the first header of the contacts table (probably from earlier):

image

vaszig:
> I put my notes in the code.
> 
> I also noticed this that might existed before: maybe these buttons could be closer? The Accept button is really far from the Reject, and the Reject button label says Terug (Back)
> 
> Screenshot from 2022-12-07 14-15-50

Yes, this was there before I touch the frontend!It seems weird but if @alextreme wants I can create a task for it.

vaszig:
> I forgot to mention this: there is a small CSS bug in the first header of the contacts table (probably from earlier):
> 
> image

Yes, there is a task in taiga about this. A change in the cases tables has broken all tables.

src/open_inwoner/accounts/views/contacts.py
vaszig: Discussed this with Anna as well and I decided to go for the same approach as the one in deleting a contact. That's why you saw the ids in the for loop.

src/open_inwoner/accounts/models.py
```diff
@@ -235,13 +235,15 @@ def get_active_contacts(self):
     def get_contacts_for_approval(self):
         return self.contacts_for_approval.filter(is_active=True)
 
-    def get_all_contacts(self):
```
vaszig, line 238: I removed this method as it complicated things when we want to refer to different kind of contacts.

src/open_inwoner/accounts/views/contacts.py
Bartvaderkin: Ok, if it's same pattern as the other and Anna is ok with it then it's fine.

vaszig: Let's wait for @annashamray 's approval as well because she hasn't seen this implemented.

alextreme:
> > I put my notes in the code.
> > I also noticed this that might existed before: maybe these buttons could be closer? The Accept button is really far from the Reject, and the Reject button label says Terug (Back)
> > Screenshot from 2022-12-07 14-15-50
> 
> Yes, this was there before I touch the frontend!It seems weird but if @alextreme wants I can create a task for it.

Task #960 created, noticed it before so lets make this prettier.

src/open_inwoner/scss/components/Contacts/Contacts.scss
```diff
@@ -0,0 +1,57 @@
...
+    form {
+      flex: 1;
+    }
+  }
+
+  p {
+    margin: 0;
+  }
+
+  .btn-reject {
```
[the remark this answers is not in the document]
vaszig: Since it's a button class, I followed the existing pattern in the project (button--danger, button--disabled etc.) and removed this rule from contacts scss file. I added it to the button.scss.

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -41,7 +41,12 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
         user = self.request.user
-        context["pending_approvals"] = user.get_contacts_for_approval()
+        pending_approvals = User.objects.get_pending_approvals(user)
+        if pending_approvals.count() == 1:
+            context["pending_approval"] = pending_approvals.get()
+        else:
+            context["pending_approval_list"] = pending_approvals
```
annashamray, line 48: it looks excessive. Why not to have just `context["pending_approval_list"] = pending_approvals` and check the count in the template?

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -8,6 +8,47 @@ <h1 class="h1" id="title">
     {% button href="accounts:contact_create" text=_("Nieuw contact toevoegen") icon="person" primary=True %}
 </h1>
 
+<!--List of contact requests: receiver's view-->
+{% if pending_approval or pending_approval_list %}
+    <div class="approval{% if pending_approval %} approval--solo{% endif %}">
+        <div class="approval__header">
+            <h2 class="h2">{% trans "U bent toegevoegd als contactpersoon" %}</h2>
+            <p class="p">{% trans "By accepting you agree on sharing your personal data (first name, last name and phonenumber)." %}</p>
```
annashamray, line 16: I don't see this sentence it in the design (https://projects.invisionapp.com/share/Y211V6JRHPC4#/screens/470424529). It the design updated somewhere else?

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -8,6 +8,47 @@ <h1 class="h1" id="title">
...
+        {% if pending_approval %}
+            <div class="approval__single">
+                <p class="p">
+                    {% blocktrans with full_name=pending_approval.get_full_name %}{{ full_name }} heeft u toegevoegd aan diens contactpersonen.{% endblocktrans %}
+                </p>
+                {% url 'accounts:contact_approval' uuid=pending_approval.uuid as approval_url %}
+                {% render_form form=None method="POST" form_action=approval_url id="approval_form" spaceless=True show_notifications=True %}
+                    {% csrf_token %}
+                    {% button_row %}
+                        {% button text=_("Weigeren") name="contact_reject" value="reject" secondary=True bordered=True %}
```
annashamray, line 27: I don't think the value is needed here. I think the button name is enough (and in the view you can just check something like `"contact_reject" in request.POST`

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -8,6 +8,47 @@ <h1 class="h1" id="title">
...
+            <h2 class="h2">{% trans "U bent toegevoegd als contactpersoon" %}</h2>
+            <p class="p">{% trans "By accepting you agree on sharing your personal data (first name, last name and phonenumber)." %}</p>
+        </div>
+        {% if pending_approval %}
+            <div class="approval__single">
+                <p class="p">
+                    {% blocktrans with full_name=pending_approval.get_full_name %}{{ full_name }} heeft u toegevoegd aan diens contactpersonen.{% endblocktrans %}
+                </p>
+                {% url 'accounts:contact_approval' uuid=pending_approval.uuid as approval_url %}
+                {% render_form form=None method="POST" form_action=approval_url id="approval_form" spaceless=True show_notifications=True %}
```
annashamray, line 24: I could be dryer (with moving this form into a separate template tag and reusing it in the loop below), but not necessary though

src/open_inwoner/scss/components/Contacts/Contacts.scss
```diff
@@ -0,0 +1,53 @@
+.approval {
+  display: flex;
```
annashamray, line 2: Hmm I don't see why it should be flex

src/open_inwoner/scss/components/Contacts/Contacts.scss
```diff
@@ -0,0 +1,53 @@
...
+  flex-direction: column;
+  background-color: var(--color-accent-lighter);
+  margin-top: var(--gutter-width);
+  border-radius: var(--border-radius);
+
+  &__header {
+    padding: var(--spacing-medium);
```
annashamray, line 9: You include this padding into all child tags of `.approval`. Whi not to change the padding of the parent class itself?

src/open_inwoner/accounts/views/contacts.py
annashamray: I think there is a little misunderstanding. 
I said that there is no need in adding additional forms into `ContactListView` context - we can just generate forms in the template since it's only csrf token, two buttons and 1 action. It's how the delete buttons for contacts are done. 
Security-wise it should be ok, since we send csrf token. 

But it's a different story with the view which process POST request (`ContactApprovalView`). 
Here we can use forms to escape and validate user input, especially if we use values in buttons. We can use only button names like django admin does, and they do it without extra validation. I don't know is it a best practice or not.

annashamray (commented):
Nice! I'll approve after CI is happy

src/open_inwoner/scss/components/Contacts/Contacts.scss
alextreme: Discussed, as long as this is consistent with the design this is fine

alextreme:
Removing @annashamray , for @Bartvaderkin to double-check after Vasileios's changes

Bartvaderkin (approved):

Bartvaderkin:
:heavy_check_mark: :+1:

PR #364, opened 2022-12-07
#950 bump zgw-consumers to 0.23.2

annashamray:
`zgw_consumers` 0.23.2 includes bugfix of `get_paginated_results` function, which should help with duplicated results (PR - https://github.com/maykinmedia/zgw-consumers/pull/58)

**NOTE!** 
`zgw_consumers` 0.21.0 has a breaking change - moving `Certificate` model into the separate library (https://github.com/maykinmedia/zgw-consumers/blob/main/CHANGELOG.rst#0210-2022-08-31)
After this PR is merged and deployed into staging the Open Zaak certificate configuration should be checked:
* requesting cases should work
* download documents should also work

### DIFF
--- a/src/open_inwoner/openzaak/tests/factories.py
+++ b/src/open_inwoner/openzaak/tests/factories.py
@@ -1,5 +1,5 @@
 import factory
-from zgw_consumers.constants import CertificateTypes
+from simple_certmanager.constants import CertificateTypes
 from zgw_consumers.models import Certificate, Service
 
 
--- a/src/open_inwoner/openzaak/tests/factories.py
+++ b/src/open_inwoner/openzaak/tests/factories.py
@@ -1,6 +1,7 @@
 import factory
 from simple_certmanager.constants import CertificateTypes
-from zgw_consumers.models import Certificate, Service
+from simple_certmanager.models import Certificate
+from zgw_consumers.models import Service
 
 
 class ServiceFactory(factory.django.DjangoModelFactory):
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,6 +20,7 @@ brotli==1.0.9
     # via fonttools
 certifi==2022.9.24
     # via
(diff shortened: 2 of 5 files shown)


src/open_inwoner/openzaak/tests/factories.py
```diff
@@ -1,5 +1,5 @@
 import factory
-from zgw_consumers.constants import CertificateTypes
+from simple_certmanager.constants import CertificateTypes
 from zgw_consumers.models import Certificate, Service
```
Bartvaderkin, line 3: We should update the factory as well.

Although it is interesting this test apparently still passes on this PR even though we save another model. I guess because the model is a copy? https://github.com/maykinmedia/open-inwoner/blob/2ae82764306fb19ab748ceb99db055679d52b87f/src/open_inwoner/openzaak/tests/test_documents.py#L308-L318

Bartvaderkin (changes requested):
I wonder if migrating the certificates would be a deployment hassle we should cover but @alextreme can decide on that.

src/open_inwoner/openzaak/tests/factories.py
annashamray: yeah, afaik `zgw_consumers.Service` model still exists but it's a proxy for `simple_certmanager.Service` model

annashamray: I've changed import to use "real" model :)

alextreme:
I'm aware of the breaking change in zgw-consumers, we'll have to deal with this on the test/acceptance environments. Better to do this earlier than later...

src/open_inwoner/openzaak/tests/factories.py
Bartvaderkin: Cool, let's resolve!

Bartvaderkin (approved):

alextreme:
:pray:

PR #365, opened 2022-12-07
 [#938] Added logging when viewing Zaak information

Bartvaderkin:


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -336,13 +336,7 @@ def test_status_is_retrieved_when_user_logged_in_via_digid(self, m):
...
         self._setUpMocks(m)
 
-        response = self.app.get(
-            reverse(
-                "accounts:case_status",
-                kwargs={"object_id": "d8bbdeb7-770f-4ca9-b1ea-77b4730bf67d"},
-            ),
-            user=self.user,
-        )
+        response = self.app.get(self.case_detail_url, user=self.user)
```
Bartvaderkin, line 339: Not task related but I replaced some noisy boilerplate reverse()'s below.

src/open_inwoner/openzaak/tests/test_documents.py
```diff
@@ -179,6 +179,17 @@ def test_document_content_is_retrieved_when_user_logged_in_via_digid(self, m):
             response.headers["Content-Length"], str(len(self.informatie_object_content))
         )
 
+    def test_document_retrieval_logs_case_identification_and_file(self, m):
```
Bartvaderkin, line 182: Earlier I appended the logging test to the general test, but it is more readable to have them separate and not test multiple things at same time.

src/open_inwoner/accounts/views/cases.py
```diff
@@ -51,6 +59,7 @@ class CaseAccessMixin(AccessMixin):
 
     When retrieving a case :
     - users BSN has a role for this case
+    - case confidentiality is lower then globally configured
```
annashamray, line 62:
```suggestion
    - case confidentiality is not higher than globally configured
```
Nice doc string

annashamray (changes requested):

annashamray (approved):

PR #371, opened 2022-12-08
[#911] Added configuration options and logic to hide unused functionality

Bartvaderkin:
We had to do some interpretation on the ticket, because Actions are a thing with or without Plan/Samenwerken. 

So I made it so there is an option to disable Samenwerken and remove it from menu but keep general Action functionality, and another option to also disable the general Actions (which requires disabled Plan).

Additionally there was an ask to hide QuestionaireStep (Zelfcheck) if there are no published QuestionaireStep's, which was more of a bug-fix then a feature request.

It might make sense to overview the code per-commit as they are reasonably clean.


src/open_inwoner/configurations/models.py
```diff
@@ -354,6 +369,13 @@ class Meta:
     def __str__(self):
         return str(_("Site Configuration"))
 
+    def clean(self):
+        super().clean()
+
+        if self.show_plans and not self.show_actions:
```
alextreme, line 375: Netjes opgelost :)

alextreme (approved):

alextreme:
@vaszig thanks, glad you double-checked :+1:

Bartvaderkin:
@vaszig I don't see it here but I'm still interested about the migration feedback, but anyway, there are two because I was trying to make clean commits instead of piling everything into one.

vaszig:
> @vaszig I don't see it here but I'm still interested about the migration feedback, but anyway, there are two because I was trying to make clean commits instead of piling everything into one.

Yes, it's clear enough. It makes sense to have them separately according to the nice and clear functionality-based commits.

PR #374, opened 2022-12-09
[#904] Feature/expand action log messages

vaszig:


src/open_inwoner/accounts/tests/test_logging.py
```diff
@@ -528,12 +528,63 @@ def test_action_update_is_logged(self):
         self.assertEqual(
             log_entry.extra_data,
             {
-                "message": "Changed: Naam.",
+                "message": [_("Naam changed to Updated name.")],
```
Bartvaderkin, line 531: Variable text in translations for tests should still use formatting like the original, because this will generate a new translatable with hardcoded 'Updated'.

And using the same formatting placeholders as the source, like `_("{label} changed to {updated_data}.").format(<...etc...>`

Bartvaderkin (changes requested):

src/open_inwoner/accounts/tests/test_logging.py
Bartvaderkin: This looks good though, like it reproduces exactly what the tested code does.

I'll try and ask Alex what works for the way this project does translations.

Bartvaderkin (approved):

alextreme:
If the translations become a problem I'll let you guys know, this is fine imho

PR #376, opened 2022-12-12
Added setting to silence house-keeping signals that clutter tests

Bartvaderkin:
The haalcentraal signal isn't very useful on CI, neither is the User 'last_login' type of logging functionality. But they create a log of noise that makes the CI/dev logs less usable.


Bartvaderkin:
As I was afraid we actually test the signals, so clobbering them like this will fail.

Bartvaderkin:
Closing this as we'll cover this with a setting for haalcentraal and mocks/mutes in the tests.

PR #379, opened 2022-12-12
:lipstick: #946 - Only use word-break: break-all; in mobile.

svenvandescheur:


Bartvaderkin (approved):

Bartvaderkin:
@svenvandescheur the prettier formatter is blocking

PR #380, opened 2022-12-12
[#927] Added cache decorator to some case-views to improve navigation

Bartvaderkin:


alextreme (commented):
Maybe it would be good to make these durations into Django settings?

I'm doubting the zaak/informatieobject caching, it might be viable for ~30 seconds but 3 minutes feels a bit too long. And with 30 seconds it might just not be worthwhile anymore.

Feel free to discuss this with Anna if you want. Could also be something that we enable after the demo.

Bartvaderkin:
Yea, I expected those to be settings but they weren't. I'll make the change on the PR.

Bartvaderkin:
@alextreme I appended the PR with settings for the timeouts and removed caching on two calls that currently don't need it because they are only used in the download view.

src/open_inwoner/openzaak/cases.py
```diff
@@ -16,7 +18,7 @@
 
 
 # cache for 3 minutes to quickly switch between open and closed cases
```
vaszig, line 20: Maybe confusing since this is modified by the settings now. We can add that the default value is 3 minutes. If you think it's excessive we can leave it as is.

vaszig (commented):
For `fetch_specific_status`  function (openzaak/cases) we don't use a variable because it's only used once right?

vaszig (approved):

src/open_inwoner/openzaak/cases.py
Bartvaderkin: Yes you are right, the comment doesn't have to be there. I appended a commit to remove it.

PR #381, opened 2022-12-13
[#964] Feature/first-last name validation

vaszig:


src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -319,6 +363,58 @@ def test_submit_not_unique_email_different_case(self):
...
+        user = UserFactory.create(
+            first_name="",
+            last_name="",
+            login_type=LoginTypeChoices.digid,
+        )
+        invalid_characters = "/\"\\,.:;'"
+
+        for char in invalid_characters:
+            with self.subTest(char=char):
+                get_response = self.app.get(self.url, user=user)
```
Bartvaderkin, line 377: The name of your response variable changes in every test :D

I'd stick with just `response` unless something specific is going on and/or you really need to access multiple responses at the same time (additionally `get_response` particular reads like a callable)

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -193,6 +193,58 @@ def test_save_filled_form(self):
         self.assertEquals(self.user.postcode, "1013 RM")
         self.assertEquals(self.user.city, "Amsterdam")
 
+    def test_save_with_invalid_first_name_chars_fails(self):
+        invalid_characters = "/\"\\,.:;'"
+
+        for char in invalid_characters:
+            with self.subTest(char=char):
+                response = self.app.get(self.url, user=self.user)
+                self.assertEquals(response.status_code, 200)
```
Bartvaderkin, line 202: This is pattern that happens a lot in this codebase, but you don't need the separate `self.assertEquals(response.status_code, 200)` because the WebTest client already tests for the status code. Although the default accepts both some 2xx and 3xx, if if it must be just 200 you need to specify it:

`response = self.app.get(self.url, user=self.user, status=200)`

Bartvaderkin (changes requested):
Only two minor style/consistency comments. 

I do note the particular set of invalid characters seems a bit limited but I see in the ticket this was specifically asked for.

Bartvaderkin (approved):

alextreme:
Thanks for this, better safe than sorry!

PR #384, opened 2022-12-13
[#908] Feature/ability to modify email address

vaszig:
I have also modified some queries concerning email in some places which I think they should exist for avoiding duplications in db. Let me know if you agree with that.

### DIFF
--- a/src/open_inwoner/accounts/forms.py
+++ b/src/open_inwoner/accounts/forms.py
@@ -54,6 +54,7 @@ class UserForm(forms.ModelForm):
         fields = (
             "first_name",
             "last_name",
+            "email",
             "phonenumber",
             "birthday",
             "street",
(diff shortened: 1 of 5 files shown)


alextreme:
@vaszig how have to dealt with the situation where user A changes his emailadres to that of user B?

vaszig:
> @vaszig how have to dealt with the situation where user A changes his emailadres to that of user B?

In this case we have the same validation error, which indicates that a user with this email already exists.

Bartvaderkin (changes requested):
I initially wanted to complain about the dense blocks like `test_updating_a_field_without_modifying_email_succeeds` but the existing tests there are like that too so it's probably fine to leave it as it is now and keep to the surrounding pattern.

Bartvaderkin (approved):

PR #388, opened 2022-12-14
[#972] Changed status ordering to reverse() to handle eSuite timeless-datetimes

Bartvaderkin:


vaszig (approved):
Seems logical approach to me since we cannot use datumStatusGezet.  I don't see though somewhere in the documentation any details about the order of the results. Maybe my dutch doesn't help

PR #390, opened 2022-12-14
[#982] Issue/plan edit view with previous user

vaszig:
This is a fix based on the way we have implemented the init method of PlanForm. If you think we want to improve this by refactoring the related template tags I can create a new task-enhancement.


Bartvaderkin (approved):
I'll approve it to get the ticket processed but this is technically kludge-on-kludge-on-kludge (one of mine).

We need the enhancement ticket with @svenvandescheur in the loop to make the custom form widget template code/tags work with regular Django form functionality, like initial/selected and error messages. This must come up in other projects so I'm interested how this is usually solved.

Particular the radio-buttons, checkboxes en selects.

alextreme:
Ugh, agree with Bart normally this isn't necessary

PR #394, opened 2022-12-15
[#911] Added 'indicatieInternOfExtern' access check, also refactored cases-views tests

Bartvaderkin:
Note: the important change was to `is_zaak_visible()`

Then I found this wasn't applied to the list-views at all, and that required a refactor because we needed the filtering before the pagination which required to move the zaaktype retrieval code around.

But then I found the issue with the property casing in the mocks, and some other test noise like the cache-reset mixin and test data setup.

As a bonus: less noisy UserFactory, and improved ZGW api_model dataclass annotations that support swapping link URI's to objects because that is how it was used.

### DIFF
--- a/src/open_inwoner/accounts/views/cases.py
+++ b/src/open_inwoner/accounts/views/cases.py
@@ -78,7 +78,12 @@ class CaseAccessMixin(AccessMixin):
             if not fetch_roles_for_case_and_bsn(self.case.url, request.user.bsn):
                 return self.handle_no_permission()
 
-            # check confidentiality level
+            # resolve case-type
+            self.case.zaaktype = fetch_single_case_type(self.case.zaaktype)
+            if not self.case.zaaktype:
+                return self.handle_no_permission()
+
+            # check if case + case-type are visible
             if not is_zaak_visible(self.case):
                 return self.handle_no_permission()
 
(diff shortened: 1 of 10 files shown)


src/open_inwoner/accounts/views/cases.py
```diff
@@ -78,7 +78,12 @@ def dispatch(self, request, *args, **kwargs):
             if not fetch_roles_for_case_and_bsn(self.case.url, request.user.bsn):
                 return self.handle_no_permission()
 
-            # check confidentiality level
+            # resolve case-type
+            self.case.zaaktype = fetch_single_case_type(self.case.zaaktype)
+            if not self.case.zaaktype:
+                return self.handle_no_permission()
```
vaszig, line 84: I don't know if this is done for readability purposes but we have five times the same return statement.

vaszig (approved):
As far as I am concerned, I don't see something weird. Only a small comment-question.

src/open_inwoner/accounts/views/cases.py
Bartvaderkin: Yep, to keep the decision procedure more readable (without adding extra noisy workarounds) it just returns after every failed condition.

PR #395, opened 2022-12-16
[#925] Fix/validate different-case email in admin

vaszig:


alextreme (changes requested):
image

I double-checked just in case as it felt obvious looking at the code: This ensures you can't change a User anymore via the admin without changing their emailadres to something new :)

alextreme (approved):

PR #396, opened 2022-12-16
[#996] Fix/layout from admin to frontend

vaszig:
Screenshot from 2022-12-16 10-30-46


vaszig:
Double checked all the configurable text fields and the login text was the only one which had no linebreaks.

alextreme (approved):
Thanks for checking!

PR #397, opened 2022-12-19
[#967] Refactored Inbox views to use user UUID, cleaned code

Bartvaderkin:
With "cleaned code" I mean I renamed all the `self.me` / `me`'s I ran into.

### DIFF
--- a/src/open_inwoner/accounts/views/inbox.py
+++ b/src/open_inwoner/accounts/views/inbox.py
@@ -5,7 +5,7 @@ from urllib.parse import unquote
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
-from django.urls import reverse_lazy
+from django.urls import reverse, reverse_lazy
 from django.utils import formats
 from django.utils.translation import gettext as _
 from django.views.generic import FormView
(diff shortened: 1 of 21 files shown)


Bartvaderkin:
I added a commit that disables the `async_selector` on the form-submit, because it replaces then whole `.messages` element including the form and emoji/file widget without re-hydrating them. 

I tried using the same technique we used for the Dropdown component (in `src/open_inwoner/js/components/index.js`) but the Emoji and MessageFile javascript code isn't as clean: it keeps global references and event listeners so the htmx swap can't remove everything.

src/open_inwoner/accounts/views/inbox.py
```diff
@@ -126,24 +142,31 @@ def get_form_kwargs(self):
...
         return HttpResponseRedirect(f"{url}#messages-last")
 
     def get(self, request, *args, **kwargs):
-        """Mark all messages as seen for the receiver"""
         context = self.get_context_data()
 
-        # Redirect to the end of page.
-        # Redirecting to a hash doesn't work, so we need to change the url.
-        # Alter URL with redirected query param in order to go to the last message (#messages-last).
+        """
```
vaszig, line 156: By removing this part here it seems that the `message.seen`  is never updated to True. That's why Mijn berichten (navigation bar) is not updated with the number of seen=False messages.
image

vaszig (changes requested):

src/open_inwoner/accounts/views/inbox.py
Bartvaderkin: Good catch. I'll need to fix it and see about the tests.

vaszig: Tested again the code and works as expected. Maybe something was wrong with the browsers.

vaszig (approved):

alextreme (approved):
Approving in order to get v1.1 ready, thanks guys

PR #398, opened 2022-12-20
[#993,#994,#1001] Fix/notifications layout and icons

vaszig:


Bartvaderkin:
I note the font-size for the form-errors is different:

image

vaszig:
> I note the font-size for the form-errors is different:
> 
> image

Thanks. I will double check this since they have the same class and it's weird.

Bartvaderkin (approved):

src/open_inwoner/scss/components/Notification/_Notification.scss
```diff
@@ -25,6 +26,14 @@
     --notification-color-text: var(--color-success);
   }
 
+  &--compact {
+    margin: var(--spacing-medium) 0;
+
+    p {
```
Bartvaderkin, line 32: IIRC there was something about our CSS code-style where we don't use tag names in selectors (hence the `<p class="p">` everywhere, but I'm not sure. I think it might have to do with specificity of the selector? Maybe @svenvandescheur has ideas about this.

PR #401, opened 2022-12-21
[#933] Added automatic search submit when clicking on search filter

Bartvaderkin:


vaszig (commented):
The code looks good to me and makes sense, but when I apply a filter it takes for ever. Is this happening only in my local machine?

Bartvaderkin:
@vaszig It is fine here, I think it is probably your elasticsearch (it is a bit memory hungry). You could try a reboot or `manage.py search_index --rebuild`.

vaszig:
Yes, I tried to rebuild but I had the same issue. If it's working for you then it's fine.

vaszig (approved):

PR #402, opened 2022-12-21
[#914] Feature/make subtitles-intro fields optional

vaszig:


Bartvaderkin (approved):
:+1: So nice and tidy.

PR #403, opened 2022-12-22
Configured Github workflows to use CI settings and requirements

Bartvaderkin:


alextreme (approved):
Altijd handiger om dev en ci settings te splitsen

PR #404, opened 2022-12-22
[#976] Implemented first Playwright tests for action status toggle

Bartvaderkin:
WIP to see if this runs on CI


alextreme:
Playwright seems trivial to install and the tests are imho much more readable than Selenium. I wouldn't mind merging this in to see how it goes over the next few weeks, it's better than the disabled Seleniumtests.

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -571,22 +566,15 @@ def setUp(self) -> None:
...
 
 
-@skip("skipped for now because of random CI failures, ref Taiga #963")
-class PlanActionStatusFirefoxSeleniumTests(
-    FirefoxSeleniumMixin,
-    _PlanActionStatusSeleniumMixin,
-    ActionStatusSeleniumBaseTests,
-    StaticLiveServerTestCase,
+class ChromePlanActionStatusPlaywrightTests(
+    _PlanActionStatusSeleniumMixin, ActionsPlaywrightTests
```
vaszig, line 570: We could rename the mixin above, since we won't have selenium in the project but we can do this when we add the test we are skipping for now.

vaszig (approved):
I am not familiar with playwright but went through this and seems nice. Since my local machine and the CI are happy let's give it a try!
And by the way, I don't know if this is still draft for some reason or by mistake.

alextreme:
If there was a reason why it's still a draft we'll see soon enough :)

src/open_inwoner/plans/tests/test_views.py
Bartvaderkin: Oops, I was sloppy.

PR #406, opened 2022-12-23
[#869] Fix/redirect to registration page when digid via invite fails

vaszig:


src/open_inwoner/accounts/views/invite.py
```diff
@@ -40,4 +42,11 @@ def get_object(self, queryset=None):
             self.log_system_action(_("invitation used"), invite)
             raise Http404(_("The invitation was already used"))
 
+        if user.is_authenticated:
+            if user.email != invite.invitee_email:
```
alextreme, line 46: Comments:
- I would suggest to remove the check on the emailadres, this to avoid the situation where the user registers under a different emailadres than that the invite was sent. You can also do a case-insensitive comparison if you think this is too lax.
- After accepting and saving the invite, you can add request.user to the contacts of the user sending the invite. You can then return the accepting user to Mijn Contacten, with a message that the user has accepted the invitation.

src/open_inwoner/accounts/views/invite.py
```diff
@@ -41,3 +43,30 @@ def get_object(self, queryset=None):
             raise Http404(_("The invitation was already used"))
 
         return invite
+
+    def dispatch(self, request, *args, **kwargs):
+        invite = self.get_object()
+        user = request.user
+
+        # mark invitation as accepted and add inviter to the user's contacts as well
+        if user.is_authenticated:
```
alextreme, line 52: Please only do this if the invite has not yet been accepted earlier.

alextreme:
One comment but besides that this is good to be merged imho

src/open_inwoner/accounts/views/invite.py
vaszig: Done!
I hadn't added this here since we have an early check in `get_object()` where we show a 404 if the invite is already accepted.

alextreme (approved):

PR #408, opened 2023-01-02
Added @multi_browser() decorator to playwright utils and applied to cases

Bartvaderkin:


alextreme (approved):
Fine imho as long as it doesn't lead to flakey tests or a slow CI

PR #410, opened 2023-01-04
[#969] Feature/add django yubin

vaszig:


vaszig:
The cronjobs for sending the emails are in another PR concerning deployment.

Bartvaderkin (changes requested):
We should also add it to the admin-index fixture so we can view the messages in the admin.

Bartvaderkin:
Ugh, I assumed we had a signal to reload admin-index fixture after a migrate, but we don't. 

I can't run `python src/manage.py loaddata django-admin-index` and can't see it used anywhere, maybe I misunderstood.

vaszig:
Yes indeed..We have to run `python src/manage.py loaddata src/open_inwoner/conf/fixtures/django-admin-index.json` (realized that needs the complete path after some tries..)

alextreme (approved):

Bartvaderkin (approved):

PR #412, opened 2023-01-04
[#997] Issue/OpenIdConnect improvements

vaszig:


vaszig:
Discussed with Bart as well and decided to use `email__iexact` for oidc. This is used in the registration and the necessary form as well.

alextreme (approved):
Seems like the proper way to deal with this.

PR #414, opened 2023-01-05
[#884] Added dynamic html-titles to all pages

Bartvaderkin:


src/open_inwoner/pdc/views.py
```diff
@@ -202,12 +213,17 @@ def get_context_data(self, **kwargs):
...
+class ProductFormView(
+    CommonPageMixin, BaseBreadcrumbMixin, CategoryBreadcrumbMixin, DetailView
+):
     template_name = "pages/product/form.html"
     model = Product
     breadcrumb_use_pk = False
     no_list = True
 
+    def page_title(self):
+        return " ".join(self.object.name, _("Formulier"))
```
vaszig, line 225: I assume you want an f-string here or something like this. We will get an error in join.

vaszig (changes requested):
Nice mixin!Tested locally the frontend  and works fine. I have only one comment..

src/open_inwoner/pdc/views.py
Bartvaderkin: Yes, you are right. I added a commit to fix this.

PR #417, opened 2023-01-05
[#801] Fix/close menus with escape button

vaszig:


Bartvaderkin:
@vaszig I think this has been covered by Sven's solution for some other issues like the focus handling (the menu is now a <dialog>)

vaszig:
@Bartvaderkin thanks for letting me know. I will double check this as soon as Svenv's PRs are merged.

alextreme:
@vaszig can be checked, Svens PR was merged afaik

alextreme:
Discussed: This is for being able to close (for instance) the dropdown menu in Mijn Profiel with an escape-button

image

jiromaykin:
@vaszig This seems to work well (and it doesn't even have the Header click-trap problem) but when I check out this branch the dropdown seems to be styled differently now (extra borders in different places)?
For some reason the border is visible here, where it doesn't in different versions:

`
element.style {
}
.dropdown .dropdown__item .button, .dropdown .dropdown__item .flatpickr-clear {
    border-color: var(--font-color-body); [.....]`

accountmenu

vaszig:
It's an old branch so the borders are not updated here. 
It works because it only implements what we wanted..the escape key. Svenv's approach is more generic and affects a lot of code, that's why we have the problem.

jiromaykin:
> It's an old branch so the borders are not updated here.

Sorry @vaszig yes, I see now - I would think this PR might no longer be needed, once https://github.com/maykinmedia/open-inwoner/pull/466 (escape-button combo with focus-trap on header/search) is reviewed and merged?

vaszig:
> Sorry @vaszig yes, I see now - I would think this PR might no longer be needed, once #466 (escape-button combo with focus-trap on header/search) is reviewed and merged?

Yes indeed, we can close this PR as soon as we merge 466.

src/open_inwoner/scss/components/Header/PrimaryNavigation.scss
```diff
@@ -98,4 +98,9 @@
   &:not(#{&}--dismissed) > &__list > &__list-item:hover &__list {
     transform: scaleY(1);
   }
+
+  &__list--hide {
```
jiromaykin, line 102: is this style used anywhere? (not if we do not use the primary-navigation JavaScript from this PR?).

src/open_inwoner/js/components/primary-navigation/index.js
```diff
@@ -0,0 +1,24 @@
+const primaryNavItems = document.querySelectorAll(
+  '.primary-navigation .primary-navigation__list'
```
jiromaykin, line 2: I'm assuming all of this could be deleted due to the new "_src/open_inwoner/js/components/header/primary-navigation.jssrc/open_inwoner/js/components/header/primary-navigation.js_" file?

jiromaykin (changes requested):
The Dropdown JS we still need; the PrimaryNavigation might not be necessary anymore after all of the updates from PR https://github.com/maykinmedia/open-inwoner/pull/466

src/open_inwoner/scss/components/Header/PrimaryNavigation.scss
vaszig: No, it is not needed now. I will remove it.

src/open_inwoner/js/components/primary-navigation/index.js
vaszig: Yes, indeed this will be removed as well.

jiromaykin (approved):

PR #426, opened 2023-01-12
Fixed fixture discovery and added post-migrate admin-index fixture reloading

Bartvaderkin:
The first part fixes the discovery of fixtures, so you no longer need full path or extension.

The second part is a script `bin/generate_admin_index_fixture.sh` to manually run after we changed the admin-index, like after adding new models or links.

The third part adds a post-migrate signal handler so whenever `manage.py migrate` is run it will synchronize the admin-index so we're all looking at the same menu.

There is another level after this I didn't add for now, where we'd add a task to the CI workflow that fails if there are new models that aren't explicitly added or excluded from admin-index, but it can wait.

### DIFF
--- a/bin/generate_admin_index_fixture.sh
+++ b/bin/generate_admin_index_fixture.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+#
+# Dump the current (local database) admin layout to a JSON fixture. This
+# overwrites the existing one.
+#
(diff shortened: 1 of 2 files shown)


Bartvaderkin:
@Bartvaderkin Update this PR without the signal and auto-loading.

alextreme (approved):

PR #430, opened 2023-01-17
[#1052-4]Feature/ haalcentraal contact data

vaszig:


alextreme:
General idea is good, however thinking about the edit-profile view there isn't much point in showing all the disabled fields if the User is prepopulated, and the opt-out structure feels a bit brittle (easy to add new fields to the form that might be made editable by accident).

How about we go for 2 forms: one where the User is prepopulated and one where this isn't the case. The former would only contain the fields that are actually editable.

vaszig:
@alextreme I agree, the only reason I did this in one form was because the task is about making specific fields readonly. Now that we are going for two separate forms this does not seem to be needed.

alextreme (approved):
This is fine. Was there a separate task for updating the NecessaryUserForm?

vaszig:
> This is fine. Was there a separate task for updating the NecessaryUserForm?

Yes, this is the task (#1055) that I am working now.

PR #431, opened 2023-01-19
[#1060] Feature/prepopulate address fields from haalcentraal on profile page

vaszig:


alextreme (approved):
Thanks, good to expand this as requested.

PR #433, opened 2023-01-20
[#1056] Feature/add second logo field

vaszig:
Since there is no specific design for this, I kept it on the left side and I didn't change the grid concerning the image (it is expanded to the whole grid horizontally). Perhaps this should be reviewed by Svenv as well


alextreme (approved):
"hmm" about CSS Grid approach in the footer, but this issue is unrelated to that and solves the problem imho.

Bartvaderkin (approved):

PR #435, opened 2023-01-24
[#811] Case notification base code

Bartvaderkin:
Let's review this part before I gets out of hand:

This is everything to accept notifications from Openzaak, including filtering the case/status/role/users, up to actually sending the message.

So these are basically two parts: the generic ZGW notification webhook and the first of the OIP specific part, but the PR touches a bunch of unrelated things because I updated the TimelineLog system-action wrapper.

I have a branch on top of this that adds the ZaakTypeConfig (to also handle eSuite's' informeren) and sending the actual email.

(not a lot of this would run locally so if the tests look good it's good)


src/open_inwoner/openzaak/api_models.py
```diff
@@ -137,10 +137,52 @@ class Resultaat(ZGWModel):
...
 @dataclass
 class Status(ZGWModel):
     url: str
     zaak: Union[str, Zaak]
     statustype: Union[str, StatusType]
     datum_status_gezet: Optional[datetime] = None
     statustoelichting: Optional[str] = ""
+
+
+@dataclasses.dataclass
```
vaszig, line 162: Is it different from the one used above?It seems to be the same import.

src/open_inwoner/openzaak/auth.py
```diff
@@ -0,0 +1,67 @@
...
+                    "verify_iat": False,
+                    "require": ["client_id"],
+                },
+            )
+        except InvalidTokenError:
+            continue
+        else:
+            return subscription
+
+    raise InvalidAuthForClientID(client_id)
```
vaszig, line 67: I don't get when this line should be hit.

src/open_inwoner/openzaak/models.py
```diff
@@ -1,4 +1,6 @@
 from django.db import models
+from django.db.models import UniqueConstraint
+from django.utils import timezone
```
vaszig, line 3: We don't use these somewhere as I can see.

src/open_inwoner/openzaak/api/urls.py
```diff
@@ -0,0 +1,13 @@
...
+from open_inwoner.openzaak.api.views import ZakenNotificationsWebhookView
+
+app_name = "openzaak"
+
+urlpatterns = [
+    path(
+        "notifications/webhook/zaken",
+        ZakenNotificationsWebhookView.as_view(),
+        name="notifications-webhook-zaken",
+    ),
```
vaszig, line 12: A minor detail..if we want be consistent with the naming conventions of the project we use `notifications_webhook_zaken`

src/open_inwoner/urls.py
```diff
@@ -76,6 +76,10 @@
     path("accounts/", include("django_registration.backends.one_step.urls")),
     path("accounts/", include("django.contrib.auth.urls")),
     path("api/", include("open_inwoner.api.urls", namespace="api")),
+    path(
+        "api/openzaak/",
+        include("open_inwoner.openzaak.api.urls", namespace="openzaak-api"),
```
vaszig, line 81: Same as above, naming with underscores.

src/open_inwoner/openzaak/api_models.py
Bartvaderkin: Not really, just sloppy use of auto-imports. Fixed it.

src/open_inwoner/openzaak/auth.py
Bartvaderkin: This happens when we do have a Subscription(s) for the `client_id` we got from the unverified JWT, but the secret(s) is wrong (eg: it didn't hit the return in the loop above).

src/open_inwoner/openzaak/api/urls.py
Bartvaderkin: Yes we want.

vaszig (approved):

alextreme:
Okay, lets give this a try on the testenvironment due to the dependancy on OpenZaak/OpenNotificaties

PR #436, opened 2023-01-25
[#1065] Fixed skiplink focus visibility

Bartvaderkin:


vaszig (approved):
I will approve this as it seems to work as expected. Not the best person for reviewing the actual css modifications.

PR #439, opened 2023-01-25
[#1047,#1063] Feature/add upload documents to the case detail page

vaszig:


src/open_inwoner/accounts/views/cases.py
```diff
@@ -371,6 +380,51 @@ def get_case_document_files(self, case) -> List[SimpleFile]:
             )
         return documents
 
+    def get_information_object_type_choices(self) -> List[str]:
```
Bartvaderkin, line 383: I think in django 'choices' are generally the combination of values & labels (eg: the list of tuples). Then you also don't have to do additional noisy work where you'd want to use them (eg: in get_form_kwargs/get_context etc).

For now its fine as this will be refactored anyway.

Bartvaderkin (changes requested):
Cool. I think the CSS needs some margin, and maybe we should alphabetically sort the extensions?

Screenshot from 2023-01-31 10-15-27

I don't like the thin field labels but it's the standard look from the components so fine.

vaszig:
Thanks for looking at the css as well!I noticed that and has been updated on the refactor branch.

Bartvaderkin (approved):

Bartvaderkin:
@vaszig  I approved but it looks like there are some (new) merge conflicts.

PR #442, opened 2023-01-31
[#1089] Feature/BRP retrieval upon every login

vaszig:


alextreme (approved):
Thanks, seems okay to me and the additional tests look solid. Will require further testing in test and on acceptance.

PR #443, opened 2023-01-31
[#1023] Added ZaakTypeInformatieObjectTypeConfig model and import

Bartvaderkin:
I'm not 100% done with the tests for edge-cases but the command itself runs, and I also want to add some helpers to retrieve the right things from the cases-detail view.

So I put it in draft so we can look at it if we want.


src/open_inwoner/openzaak/admin.py
```diff
@@ -57,8 +59,33 @@ def queryset(self, request, queryset):
...
+        "omschrijving",
+        "informatieobjecttype_url",
+        "zaaktype_uuids",
+    ]
+    ordering = ("omschrijving",)
+
+    def has_add_permission(self, request, obj):
+        return False
+
+    def has_delete_permission(self, request, obj=None):
```
vaszig, line 80: I am not sure maybe I didn't understand something but do we want to allow delete? I mean since we just show data that we retrieve and the next day it will be again there. This has to do generally with the delete operation (ZaakTypeConfig as well).

src/open_inwoner/openzaak/catalog.py
```diff
@@ -202,6 +193,34 @@ def fetch_single_case_type(case_type_url: str) -> Optional[ZaakType]:
     return case_type
 
 
+@cache_result(
```
vaszig, line 196: This function exists twice in the file, maybe after rebasing.

src/open_inwoner/openzaak/models.py
```diff
@@ -141,8 +141,20 @@ class ZaakTypeConfig(models.Model):
...
+
+    # notifications
     notify_status_changes = models.BooleanField(default=False)
 
+    # documents
+    external_document_upload_url = models.URLField(
+        verbose_name=_("Document upload URL"),
+        blank=True,
+    )
+    document_upload_enabled = models.BooleanField(
```
vaszig, line 153: Does this have to do with external uploading?We have the same below for `ZaakTypeInformatieObjectTypeConfig` and its name is a bit confusing, at least for me.

vaszig (changes requested):

src/open_inwoner/openzaak/admin.py
Bartvaderkin: Allowing delete like this is mostly for development and testing purposes, which is why I put it as superuser only.

src/open_inwoner/openzaak/catalog.py
Bartvaderkin: Well spotted!

We really need to clean these at some point because what started as a few handy helper is growing out of control. I made a ticket 1096 on Taiga for this.

src/open_inwoner/openzaak/models.py
Bartvaderkin: I think the Taiga ticket 1045 explains best how it is supposed to work (it was my ticket so you might not have seen this.).

Maybe the name is not ideal; we can rename to fit the actual implementation (easy to migrate).

Bartvaderkin:
@vaszig I processed your feedback. 

Admittedly I also force-pushed again, because I used the wrong issue number in the commit (should be 1045 instead of 1023) :confused:

vaszig (approved):

PR #445, opened 2023-02-02
[#1076] Fix/refactor case document upload form

vaszig:
This has to do only with the upload form functionality, the refactoring after adding the ZaakType configs. 
The submission part is covered by the next task (taiga #1062).


Bartvaderkin (approved):
:+1: Neat!

alextreme (approved):

PR #448, opened 2023-02-03
[#1062] Feature/upload documents to openzaak

vaszig:


src/open_inwoner/openzaak/documents.py
```diff
@@ -87,3 +94,37 @@ def download_document(url: str) -> Optional[Response]:
...
+
+    client = build_client("document")
+    if client is None:
+        return
+
+    document_body = {
+        "bronorganisatie": source_organization,
+        "creatiedatum": date.today().strftime("%Y-%m-%d"),
+        "titel": title,
+        "auteur": "Open Inwoner Platform",
```
alextreme, line 111: I would use the name of the logged-in user in OIP for this. OIP is just a vessel, we didn't create the document :)

src/open_inwoner/openzaak/documents.py
```diff
@@ -87,3 +94,37 @@ def download_document(url: str) -> Optional[Response]:
...
+        "inhoud": base64.b64encode(file.read()).decode("utf-8"),
+        "bestandsomvang": file.size,
+        "bestandsnaam": file.name,
+        "taal": "nld",
```
alextreme, line 115: I think we're missing status="definitief" here.

alextreme:
Good first attempt, interested in how this holds up with large files

src/open_inwoner/openzaak/documents.py
vaszig: I thought that the status is defined by another source (definitief - (Definitief) Informatieobject door bevoegd iets of iemand vastgesteld dan wel ontvangen). I will update this with the status field as well.

alextreme: Looking at this again I think you can leave it out, as long as it's set on the InformatieObject in OpenZaak (check the admin) then it's okay. Setting the status would then only be required if it's a document which is not yet definitive, which isn't applicable for us.

alextreme (approved):

PR #450, opened 2023-02-03
[#1046] Feature/add external url link to case detail

vaszig:
I haven't found a specific design for this so I added just another section. If there is one somewhere else please let me know.

Screenshot from 2023-02-03 10-56-01

### DIFF
--- a/src/open_inwoner/accounts/views/cases.py
+++ b/src/open_inwoner/accounts/views/cases.py
@@ -35,6 +35,7 @@ from open_inwoner.openzaak.documents import (
 )
 from open_inwoner.openzaak.models import (
     OpenZaakConfig,
+    ZaakTypeConfig,
     ZaakTypeInformatieObjectTypeConfig,
 )
 from open_inwoner.openzaak.utils import (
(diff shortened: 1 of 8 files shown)


alextreme:
@vaszig please adhere to this design: 

https://projects.invisionapp.com/share/Y211V6JRHPC4#/screens/464009119

image

eg. Only show the "Document toevoegen" button in the external-URL case.

Rationale:
- The user shouldn't care if it is an internal or external system
- A button makes it clear that an action can be done
- A Title for separation between "Documenten" and adding a document to this list isn't necessary
- We should allow only one or the other document upload method per zaaktype (this can be validated upon when configuring the zaaktype in OIP)

I do agree that showing the "external site" icon within the button may be necessary for accessibility purposes, but for now I'd like to follow the design where possible.

vaszig:
After the changes:

Internal upload enabled

Screenshot from 2023-02-03 15-45-39

External upload enabled

Screenshot from 2023-02-03 15-45-15

alextreme (approved):
This is a good first step. May require a bit of tweaking and I'll have to ensure translations are entered.

PR #452, opened 2023-02-03
:lipstick: [#1082] Word/letter wrapping

jiromaykin:
Issue: on mobile view for locations, names/words break on all letters, instead of words.
(Both inside the headers of card-containers, as well in headers of leaflet map titles)
Note: "_While word-break: break-word is deprecated, it has the same effect, when specified, as word-break: normal and overflow-wrap: anywhere — regardless of the actual value of the [overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) property._"

wordbreakall

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -87,7 +87,8 @@ development machine.
    .. code-block:: bash
 
        $ npm install
-       $ npm ci --legacy-peer-deps
+       or as an alternative: npm ci --legacy-peer-deps
+       $ npm run build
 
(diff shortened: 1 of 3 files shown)


alextreme (approved):

alextreme:
Werking mag op de testomgeving gecontroleerd worden. INSTALL aanpassingen waren niet helemaal juist volgens mij (het is of `npm install` of `npm ci`, maar `npm run build` blijft nodig)

PR #462, opened 2023-02-07
[#1109] Feature/update plan list view

vaszig:
- Status column shows 'Open' (open plan) or 'Afgerond' (expired plan)
- Openstaande acties shows the number of 'open' actions (if status is 'open' or 'approval') with the proper message ('Actie vereist')

### DIFF
--- a/src/open_inwoner/components/templatetags/helpers.py
+++ b/src/open_inwoner/components/templatetags/helpers.py
@@ -64,3 +64,22 @@ def as_attributes(attribute_dict):
     if not attribute_dict:
         return ""
     return format_html_join(" ", '{}="{}"', attribute_dict.items())
+
+
+@register.simple_tag
+def get_plan_participants(plan, user):
+    """
+    Returns the participants of the plan as a string (without the creator).
+
+    Usage:
+        {% get_plan_participants plan=plan user=request.user %}
+
+    Variables:
+        + plan: Plan | The plan we should generate the contacts-participants for.
+        + user: User | The logged in user.
+    """
+    if not plan:
+        return ""
+
+    plan_contacts = plan.get_other_users(user=user)
+    return ", ".join([contact.get_full_name() for contact in plan_contacts])
--- a/src/open_inwoner/components/templatetags/helpers.py
+++ b/src/open_inwoner/components/templatetags/helpers.py
@@ -64,22 +64,3 @@ def as_attributes(attribute_dict):
     if not attribute_dict:
         return ""
     return format_html_join(" ", '{}="{}"', attribute_dict.items())
-
-
-@register.simple_tag
-def get_plan_participants(plan, user):
-    """
-    Returns the participants of the plan as a string (without the creator).
-
-    Usage:
-        {% get_plan_participants plan=plan user=request.user %}
-
-    Variables:
-        + plan: Plan | The plan we should generate the contacts-participants for.
-        + user: User | The logged in user.
-    """
-    if not plan:
-        return ""
-
-    plan_contacts = plan.get_other_users(user=user)
-    return ", ".join([contact.get_full_name() for contact in plan_contacts])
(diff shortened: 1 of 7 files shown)


src/open_inwoner/components/templatetags/helpers.py
```diff
@@ -64,3 +64,22 @@ def as_attributes(attribute_dict):
     if not attribute_dict:
         return ""
     return format_html_join(" ", '{}="{}"', attribute_dict.items())
+
+
+@register.simple_tag
+def get_plan_participants(plan, user):
```
alextreme, line 70: Feel free to check with Bart, but I wouldn't use a templatetag for this, feels more like something you would set in your view(s)

src/open_inwoner/templates/pages/plans/list.html
```diff
@@ -1,11 +1,48 @@
...
+{% load i18n button_tags string_tags pagination_tags helpers %}
 
 {% block content %}
     <h1 class="h1">
         {% trans "Samenwerken" %}
         {% button href="plans:plan_create" text=_("Start nieuwe samenwerking") primary=True icon="group" icon_outlined=True %}
     </h1>
     {% optional_paragraph configurable_text.plans_page.plans_intro %}
-    {% card_container plans=object_list columns=2 %}
+
```
alextreme, line 10: Note that the userstory is about having listview as an expanded table when the User.contact_type is Begeleider. For regular users the card-listview is fine.

vaszig: Yes indeed, but in my notes from the weekly I had that we would go for one list view. I will change this according to the user.

vaszig: @alextreme discussed this with the PO and we will go for two options, one for the regular users (old list view) and one for the 'begeleider' (the updated one).

src/open_inwoner/plans/models.py
```diff
@@ -158,3 +155,12 @@ def get_other_users(self, user=None):
         from open_inwoner.accounts.models import User
 
         return User.objects.filter(id__in=user_ids)
+
+    def get_status(self):
```
Bartvaderkin, line 159: I think the multi-line style is more readable as you don't need to read the whole thing, scales better when modified and has better diffs.

```
if self.end_date > date.today():
    return _("Open")
else:
    return _("Afgerond")  
```

src/open_inwoner/components/templatetags/helpers.py
Bartvaderkin: I kinda agree, I would put this in in get_context_data().

I'm not a big fan of template tags for many reasons, but also because they reduce visibility of what is going on. (especially database access and other IO). Sven & Jorik got a bit wild with them in this project.

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -561,6 +564,113 @@ def test_plan_action_status_not_your_action(self):
...
+
+    @freeze_time("2022-01-01")
+    def test_plan_list_renders_expected_data_for_expired_plan_when_begeleider(self):
+        self.user.contact_type = ContactTypeChoices.begeleider
+        self.user.save()
+        self.plan.end_date = date(2022, 1, 1)
+        self.plan.save()
+
+        response = self.app.get(self.list_url, user=self.user)
+        rendered_end_date = response.pyquery("tbody .table__item")[1].text
```
Bartvaderkin, line 599: I see a lot of indexes into the pyquery result list (plain integer indexes are almost magic numbers so not great).

It is fine for now but not so great to edit or debug later.

Bartvaderkin (approved):

vaszig:
Discussed with Alex and added only one small fix concerning pagination to this PR because it's not yet merged in.

PR #463, opened 2023-02-07
:wheelchair: #1084 Check borders after accessibility changes

jiromaykin:
Some borders need to be reverted to original design, because: The 'Toegankelijkheidsaudit' does not explicitly ask for borders around these areas + This Level AA requirement reads: "...Incidental: Text[...]that are part of an inactive user interface component, that are pure decoration,[...], have no contrast requirement..."
"pure decoration" in the definition of WCAG = "serving only an aesthetic purpose, providing no information, and having no functionality"

https://taiga.maykinmedia.nl/project/open-inwoner/task/1084

Answer to my question on Stack Overflow: https://stackoverflow.com/questions/75379770/do-borders-that-visually-group-related-content-have-a-contrast-requirement

“... WCAG does not require you to have a border around the grouped elements. […] there aren't any WCAG requirements for the contrast. [1.4.11 Non-text Contrast](https://www.w3.org/TR/WCAG21/#non-text-contrast) comes close but that requirement is for "Interface Components" and "Graphical Objects". Your tile might contain "Interface Components" (such as a link) but the tile itself is not an "Interface Component" so 1.4.11 does not apply. That doesn't mean you shouldn't try to have better contrast, but it's not required by WCAG.”

### DIFF
--- a/src/open_inwoner/scss/components/Form/ButtonRadio.scss
+++ b/src/open_inwoner/scss/components/Form/ButtonRadio.scss
@@ -27,7 +27,7 @@
   }
 
   &__button {
-    border: 1px solid var(--color-gray-light);
+    border: var(--border-width-thin) solid var(--color-gray-light);
     border-left: 0;
     background-color: var(--color-gray-lightest);
     color: var(--color-gray-lighter);
--- a/src/open_inwoner/scss/components/Form/ButtonRadio.scss
+++ b/src/open_inwoner/scss/components/Form/ButtonRadio.scss
@@ -12,16 +12,38 @@
       color: var(--color-primary);
       cursor: default;
     }
+    &#negative {
+      &:checked ~ .button-radio__button {
+        color: #cf1c08;
+        cursor: default;
+      }
+    }
+    &#positive {
+      &:checked ~ .button-radio__button {
+        color: var(--color-green-dark);
+        cursor: default;
+      }
+    }
   }
 
   &__button {
-    border: 1px solid #d2d2d2;
+    border: 1px solid var(--color-gray-light);
     border-left: 0;
     background-color: var(--color-gray-lightest);
     color: var(--color-gray-lighter);
     display: inline-block;
     padding: 12px;
     cursor: pointer;
+
+    &[for='negative'] {
+      border-top-right-radius: var(--border-radius);
+      border-bottom-right-radius: var(--border-radius);
+    }
+
+    &[for='positive'] {
+      border-top-left-radius: var(--border-radius);
+      border-bottom-left-radius: var(--border-radius);
+    }
   }
 
   &:first-child {
--- a/src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
+++ b/src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
@@ -34,7 +34,7 @@
     }
 
     &__list-item:nth-child(2) {
-      border-top: var(--border-width) solid var(--color-mute-deco);
+      border-top: var(--border-width) solid var(--color-gray);
     }
 
     &__list-item {
--- a/src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
+++ b/src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
@@ -13,12 +13,12 @@
   }
 
   &__list-item {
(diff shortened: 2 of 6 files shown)


jiromaykin:
I needed to add a new color class `--color-mute-deco` because some elements still do need a bigger contrast with a different color `--color-mute` (like "input" interface components and other form inputs) but others look better with the original design.

src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
```diff
@@ -13,12 +13,12 @@
   }
 
   &__list-item {
-    border-bottom: var(--border-width) solid var(--color-mute);
+    border-bottom: var(--border-width) solid var(--color-gray);
```
vaszig, line 16: `--border-width` is 2px, maybe we should adhere to the card and dropdown buttons size -> 1px?
If so, let's create a variable for border size and set it to 1px, just to be reusable.

image

vaszig (commented):
I assume that dropdowns have to remain the same for accessibility reasons, right? Maybe I am losing something in the task..

- dropdowns

image

vaszig:
Feedback in search page? It doesn't seem to have a border-radius set.

image

src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
jiromaykin: Yes, you are right, I think this should become 1px as well; i will add to it.

jiromaykin:
@vaszig Do you mean you think that there should be an extra border there? 
I don't think so. When I look at the design in Invision I do not see a set border, so this should stay as is (also because that feedback text-area border would only be 'decorative' and not 'functional' for accessibility)

feedback

vaszig:
> @vaszig Do you mean you think that there should be an extra border there? I don't think so. When I look at the design in Invision I do not see a set border, so this should stay as is (also because that feedback text-area border would only be 'decorative' and not 'functional' for accessibility)

image

No, I mean the `border-radius` here. Not adding extra borders.

jiromaykin:
> No, I mean the `border-radius` here. Not adding extra borders. I see that the design has but we don't.

Ah, I see, you mean the voting-labels; well spotted!
Yes, I will add a radius around those.

alextreme:
Thanks @vaszig for the detailed check! :+1:

src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
```diff
@@ -34,7 +34,7 @@
     }
 
     &__list-item:nth-child(2) {
-      border-top: var(--border-width) solid var(--color-mute);
+      border-top: var(--border-width-thin) solid var(--color-gray);
```
vaszig, line 37: This changes only the second child each time. Let's set the border to your new variable for all the children.

image

src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -12,16 +12,38 @@
       color: var(--color-primary);
       cursor: default;
     }
+    &#negative {
```
vaszig, line 15: We have nice buttons now!If you think the border on focus is a problem feel free to to fix it. I am not sure about it.

image

vaszig (changes requested):
Only two small remarks ant I think it's ready to be merged. @alextreme can tell us his opinion as well.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: Yes that is not the way it should be, eventually, but I want to collect that in a new ticket for all accessibility issues that are not quite on par yet.

src/open_inwoner/scss/components/Questionnaire/Questionnaire.scss
jiromaykin: Oh wow, I was not very smart with this - i will correct it.

alextreme:
> Only two small remarks ant I think it's ready to be merged. @alextreme can tell us his opinion as well.

@vaszig My opinion is that we'll likely have to iterate on this, we won't be able to get this exactly according to the design in one go but I'm happy with these and other steps being taken.

vaszig (approved):

PR #468, opened 2023-02-13
:wheelchair: [#1078] frontend a11y issues

jiromaykin:
multiple issues:

- [x] background of success messages (fix errors where a comment breaks the HSL color codes)
- [x] area of link too large around home-banner image


vaszig (commented):
- Success message works fine!
- I still see problems with the logo while trying different size images. 




- Maybe we should target another class (ex. the image class) and not the parent one (logo).

vaszig (approved):

PR #473, opened 2023-02-16
:art: [#1131] Dropdown on mobile partly rendered off-screen

jiromaykin:
issue in Taiga: [#1131 Dropdown on mobile partly rendered off-screen](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1131)
In the original design it already showed the drop down should have been aligned to the right:
(https://projects.invisionapp.com/share/Y211V6JRHPC4#/screens/470665006)

Screenshot from 2023-02-16 14-24-27

### DIFF
--- a/src/open_inwoner/scss/components/Dropdown/Dropdown.scss
+++ b/src/open_inwoner/scss/components/Dropdown/Dropdown.scss
@@ -11,7 +11,7 @@
     flex-direction: column;
     padding: var(--spacing-medium) 0;
     position: absolute;
-    left: 0;
+    right: 0;
     border-radius: var(--border-radius);
     background-color: var(--color-white);
     border: 1px solid var(--color-mute);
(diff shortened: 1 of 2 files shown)


src/open_inwoner/scss/components/Dropdown/Dropdown.scss
```diff
@@ -11,7 +11,7 @@
     flex-direction: column;
     padding: var(--spacing-medium) 0;
     position: absolute;
-    left: 0;
+    right: 0;
```
vaszig, line 14: I don't see a reason to revert this and remove the code below imo. Testing the mobile and desktop versions in the test envs I have the expected left and right results.

acties

files

vaszig: If this is what we want for the different size screens, the code looks good to me. @alextreme can you please give us your opinion?

vaszig (approved):

PR #476, opened 2023-02-16
[#1144] Fix/update birthday date format in mydata

vaszig:


Bartvaderkin (approved):
Technically it'd be better to parse the API data to Python data (eg: a Date), and then format that using Django's standard date localization.

PR #479, opened 2023-02-17
:art: [#1139] login-screen validation-text

jiromaykin:
issue with overflowing UL's in registration view:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/1139

### DIFF
--- a/src/open_inwoner/scss/components/Form/Form.scss
+++ b/src/open_inwoner/scss/components/Form/Form.scss
@@ -80,6 +80,19 @@
     width: 100%;
   }
 
+  /// Registration -start.
+
+  &__control > .label ul {
+    padding-left: var(--card-spacing);
+  }
+
+  &__control > .label ul *:not(.button) {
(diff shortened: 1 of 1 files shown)


src/open_inwoner/scss/components/Form/Form.scss
```diff
@@ -169,3 +169,11 @@
     }
   }
 }
+
+/// Registration.
+
+#registration-form .form__control label.label ul {
```
Bartvaderkin, line 175: IIRC there was some sort of rule/method to not use element names in the selectors, something with specificity (hence why everything has class names matching the tags).

jiromaykin: @Bartvaderkin I've changed it now - it's a unique situation that i think only occurs on the Registration page. + it's not valid HTML to have a UL tag (=block) inside a label (=inline) - but this is quick fix for now.

Bartvaderkin (approved):

PR #480, opened 2023-02-20
[#1143] Fix/replace f string with format in translations

vaszig:


alextreme (approved):
Sure, I'll add the translation for this after the sprint no worries.

PR #481, opened 2023-02-20
:iphone: [#1140] mobile: Samenwerken-listview

jiromaykin:
Responsive design: fix issue where header text and button should be on top of each other, not next to each other;
see https://taiga.maykinmedia.nl/project/open-inwoner/task/1140 
⚠ Note: the current structure of templates, (where the entire H1 heading is surrounding the button), makes word-breaking behave strange again sometimes (causes letter-breaking in very long words on _very_ tiny screens).

### DIFF
--- a/src/open_inwoner/components/templates/components/Header/AccessibilityHeader.html
+++ b/src/open_inwoner/components/templates/components/Header/AccessibilityHeader.html
@@ -1,7 +1,7 @@
 {% load i18n link_tags %}
 
 <div class="accessibility-header">
-    <ul class="accessibility-header__list" role="complementary" aria-label="{% trans "Toegangkelijkheids help" %}">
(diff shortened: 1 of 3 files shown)


jiromaykin:
Apart from the word-break-row/column fix, I've fixed a typo in translations (_without_ generating 'compilemessages' etcetera).

vaszig (approved):
It looks better!
I am not sure about the change in translations. Alex can have a look as well.

Bartvaderkin (commented):
Also not sure about the change in the .po: we usually don't generate or commit those

alextreme:
I don't mind too much about also updating the translations, I tend to do this at the end of each sprint in a separate branch/PR. Having multiple branches with each their own translation changes can lead to a mess, so I propose having a "Translation Ducky" at the office to avoid this.

PR #484, opened 2023-02-21
[#1123] Feature/search functionality in plan list view

vaszig:
FYI, we use auto-submit forms (via js) in some components. We have this for `input` and `select` html tags but afaik we only use it for select now, that's why I removed input selector. Another solution would be to add some time delay in order to be easier for the user to type something in the search input (before the form is auto-submitted).

### DIFF
--- a/src/open_inwoner/plans/views.py
+++ b/src/open_inwoner/plans/views.py
@@ -44,11 +44,13 @@ class BasePlanFilter:
     """
 
     def get_filtered_plans(self, plans, available_contacts):
-        args = self.request.GET
-        plan_contacts = args.get("plan_contacts")
-        status = args.get("status")
-        query = args.get("query")
+        plan_contacts = self.request.GET.get("plan_contacts")
+        status = self.request.GET.get("status")
+        query = self.request.GET.get("query")
         today = date.today()
+
+        if not (plan_contacts or status or query):
+            return plans
 
         if plan_contacts:
             plans = plans.filter(plan_contacts__uuid=plan_contacts)
@@ -111,7 +113,6 @@ class PlanListView(
         context = super().get_context_data(**kwargs)
         user = self.request.user
         initial_qs = self.get_queryset()
-        args = self.request.GET
         plans = {"extended_plans": False}
 
         # render the extended plan list view when a begeleider is logged in
@@ -119,10 +120,7 @@ class PlanListView(
             plans["extended_plans"] = True
             available_contacts = self.get_available_contacts_for_filtering(initial_qs)
 
-            if "plan_contacts" in args or "status" in args or "query" in args:
-                filtered_plans = self.get_filtered_plans(initial_qs, available_contacts)
-            else:
-                filtered_plans = initial_qs
+            filtered_plans = self.get_filtered_plans(initial_qs, available_contacts)
 
             # sort the filtered plans based on if they are open or closed
             open_plans = filtered_plans.filter(end_date__gt=date.today())
@@ -138,7 +136,7 @@ class PlanListView(
 
             # instantiate filter form
             context["plan_filter_form"] = PlanListFilterForm(
-                data=args, available_contacts=available_contacts
+                data=self.request.GET, available_contacts=available_contacts
             )
 
             # prepare plans for frontend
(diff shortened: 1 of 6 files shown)


src/open_inwoner/plans/views.py
```diff
@@ -39,12 +40,14 @@ def dispatch(self, request, *args, **kwargs):
...
     """
-    This will filter the plans according to the user's selection.
+    This will filter the plans according to the user's selection or query.
     """
 
-    def get_filtered_plans(self, plans):
-        plan_contacts = self.request.GET.get("plan_contacts")
-        status = self.request.GET.get("status")
+    def get_filtered_plans(self, plans, available_contacts):
+        args = self.request.GET
```
Bartvaderkin, line 47: `args` is not a good general use variable name as it is strongly associated with variable function arguments (eg `args` & `kwargs`). For query parameters `params` is a good choice. 

(also I'm not 100% convinced you'd need to alias it here and below but heh)

src/open_inwoner/plans/views.py
```diff
@@ -101,15 +111,16 @@ def get_context_data(self, **kwargs):
...
         # render the extended plan list view when a begeleider is logged in
         if user.contact_type == ContactTypeChoices.begeleider:
             plans["extended_plans"] = True
+            available_contacts = self.get_available_contacts_for_filtering(initial_qs)
 
-            # filter plans if necessary
-            if "plan_contacts" in self.request.GET or "status" in self.request.GET:
-                filtered_plans = self.get_filtered_plans(initial_qs)
+            if "plan_contacts" in args or "status" in args or "query" in args:
+                filtered_plans = self.get_filtered_plans(initial_qs, available_contacts)
```
Bartvaderkin, line 123: This construction is not great: now both get_context_data() and get_filtered_plans() need to know about these GET arguments. 

So I'd move this decision into get_filtered_plans().

Bartvaderkin (changes requested):
Works pretty good. One extra feedback on existing code: I'd expect the bold title in the 'Samenwerking' column in the list also clickable, and to have something in the aria-label of the arrow for accessibility.

Also I noticed the filter widgets don't have a regular label so I'd expect they need something as well.

vaszig:
@Bartvaderkin I tried to fix everything you mentioned except for the labels in filters. We have the same approach in actions filters (and maybe somewhere else too), so if we do need them I will create a separate task for all of them.

src/open_inwoner/templates/pages/plans/list.html
```diff
@@ -51,7 +51,7 @@ <h1 class="h1">
                                 <div class="table__item--notification-danger">{% trans "Actie vereist" %}</div>
                             {% endif %}
                         </td>
-                        <td class="table__item">{% button text="" href=plan.get_absolute_url icon="arrow_forward" icon_outlined=True transparent=True %}</td>
+                        <td class="table__item">{% button text="plan_detail" hide_text=True href=plan.get_absolute_url icon="arrow_forward" icon_outlined=True transparent=True %}</td>
```
Bartvaderkin, line 54: I think this should be `text=plan_detail` (without quotes)

Bartvaderkin (changes requested):

Bartvaderkin:
It is possible the action filters weren't there when this was initially reviewed for accessibility.

Anyway, we need to have something for accessibility so if it is not part of this then a new ticket would be good.

src/open_inwoner/templates/pages/plans/list.html
vaszig: For me it doesn't work without quotes, because it then means it's a variable but I need a string here.
When I have the quotes I see the following:

image

vaszig:
> Anyway, we need to have something for accessibility so if it is not part of this then a new ticket would be good.

Yes, you are right. Task #1183 [https://taiga.maykinmedia.nl/project/open-inwoner/task/1183](url) wad created.

src/open_inwoner/templates/pages/plans/list.html
Bartvaderkin: My example was bad, I meant `text=plan.title`, or something else useful that makes sense to an accessibility user.

Having literally aria-label="plan_detail" doesn't seem helpful.

Bartvaderkin (approved):

PR #487, opened 2023-02-21
[#1154] Product location view

vaszig:


src/open_inwoner/pdc/tests/test_product_location.py
```diff
@@ -227,3 +227,48 @@ def test_exception_is_handled_when_city_and_postcode_are_not_provided(
...
+
+        response = self.app.get(
+            reverse("pdc:location_detail", kwargs={"uuid": product_location.uuid})
+        )
+
+        self.assertEqual(response.context["products"].get(), product)
+
+    def test_locations_details_are_rendered(self):
+        product = ProductFactory()
+        product_location = ProductLocationFactory(
```
Bartvaderkin, line 246: We should add email and phonenumber to the factory?

(for a Dutch phonenumber I used `phonenumber = factory.Faker("numerify", text="0#########")`, because the regular phonenumber provider has extra characters)

src/open_inwoner/pdc/tests/test_product_location.py
```diff
@@ -227,3 +227,48 @@ def test_exception_is_handled_when_city_and_postcode_are_not_provided(
...
+        product_location = ProductLocationFactory()
+        published_product.locations.add(product_location)
+        draft_product.locations.add(product_location)
+
+        response = self.app.get(
+            reverse("pdc:location_detail", kwargs={"uuid": product_location.uuid})
+        )
+        products = response.context["products"]
+
+        self.assertEqual(products.count(), 1)
```
Bartvaderkin, line 273: The .get() on the next line already checks if there is a single record.

src/open_inwoner/scss/components/Product/product-location.scss
```diff
@@ -0,0 +1,14 @@
...
+  padding: 0 0 var(--spacing-giant) 0;
+
+  &__details {
+    display: flex;
+    gap: var(--spacing-extra-large);
+    margin-top: var(--spacing-small);
+  }
+
+  &__address p:nth-child(2),
+  &__contact p:nth-child(2) {
```
Bartvaderkin, line 11: Why `nth-child(2)`  here? Feels like a magic number

src/open_inwoner/templates/pages/product/location_detail.html
```diff
@@ -0,0 +1,26 @@
...
+        <h1 class="h1">{{ object.name }}</h1>
+
+        <div class="location__details">
+            <div class="location__address">
+                <p class="p">{{ object.address_line_1 }}</p>
+                <p class="p">{{ object.address_line_2 }}</p>
+            </div>
+            <div class="location__contact">
+                <p class="p">{{ object.phonenumber }}</p>
+                <p class="p">{{ object.email }}</p>
```
Bartvaderkin, line 16: Some of these fields aren't required so this could print empty `<p>`'s

src/open_inwoner/templates/pages/product/location_detail.html
```diff
@@ -0,0 +1,26 @@
...
+            </div>
+        </div>
+    </div>
+        
+    {% if products %}
+        {% card_container products=products small=True %}
+    {% else %}
+        {% trans "There are no products available at this location." %}
+    {% endif %}
+{% endblock content %}
```
Bartvaderkin, line 26: There is a no-newline warning here I don't see very often.

My editor has a setting like "Ensure file ends in newline" that is handy to catch this.

src/open_inwoner/pdc/tests/test_product_location.py
```diff
@@ -227,3 +227,48 @@ def test_exception_is_handled_when_city_and_postcode_are_not_provided(
             ],
         )
         mock_geocoding.assert_called_once_with(" ,  ")
+
+
+class TestLocationDetailView(WebTest):
```
Bartvaderkin, line 232: I would collapse these three test functions into one, like they are so simple this just generates more code.

Then you can group statements with some comments like `# adding one published and one draft product`, and later `# only the published product shows` and `# check location details`

Bartvaderkin (changes requested):

src/open_inwoner/pdc/tests/test_product_location.py
vaszig: Here I follow the wiki's best practices which says not to define non-required fields. But if you think I should, I will of course add it.

Bartvaderkin: The wiki is more of a starting point or guideline you can ignore if you can make a defendable case for the transgression. 

I think factories should be useful and help get usable test data.

Bartvaderkin (approved):

src/open_inwoner/scss/components/Product/product-location.scss
vaszig: I think I made this more generic for purposes like you mentioned (multiple <p>'s)

PR #489, opened 2023-02-22
[#824] Added explicit through-model with django-ordered-model for Product/Category

Bartvaderkin:


src/open_inwoner/pdc/migrations/0049_auto_20230222_0901.py
```diff
@@ -8,7 +8,7 @@ def truncate_product_summary(apps, schema_editor):
     for product in MyModel.objects.all():
         if len(product.summary) > 300:
             product.summary = product.summary[:300]
-            product.save()
+            product.save(update_fields=["summary"])
```
vaszig, line 11: I did a lot of changes concerning migrations..thanks for that!

src/open_inwoner/pdc/tests/test_product.py
```diff
@@ -94,6 +95,25 @@ def test_only_published_products_exist_on_product_finder_page_when_logged_in(sel
...
+        product2 = ProductFactory(categories=(category,))
+
+        response = self.app.get(
+            reverse("pdc:category_detail", kwargs={"slug": category.slug})
+        )
+        self.assertEqual(list(response.context["products"]), [product1, product2])
+
+        # grab the ordered many-2-many and move ordering
+        prodcat = CategoryProduct.objects.get(category=category, product=product1)
+        prodcat.down()
```
vaszig, line 110: I would test the real order as well by `prodcat.order` . But a bit excessive?

vaszig (approved):
It looks really nice, tested admin locally as well.

src/open_inwoner/pdc/migrations/0049_auto_20230222_0901.py
Bartvaderkin: I'd probably been fine with the optimisations but its a good pattern to not do unnecessary work, especially in bulk operations.

PR #492, opened 2023-02-22
Removed HTMX debug template-tag as it is not in active use and generates too much CSP noise

Bartvaderkin:
If we need it we can put it back, possibly add a fix that uses `request.csp_nonce` like the other script-tags.

### DIFF
--- a/src/open_inwoner/templates/master.html
+++ b/src/open_inwoner/templates/master.html
@@ -82,6 +82,5 @@
         </dialog>
 
         <script nonce="{{ request.csp_nonce }}" src="{% static 'bundles/open_inwoner-js.js' %}" type="text/javascript"></script>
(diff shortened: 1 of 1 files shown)


vaszig (approved):

alextreme:
Could have been made conditional based on settings.DEBUG, but I don't know if this is useful during development or not.

Bartvaderkin:
@alextreme It already does that: https://github.com/adamchainz/django-htmx/blob/7b5a796fc6a336c728163837a87bc09705d69c24/src/django_htmx/jinja.py#L8

But it doesn't add the nonce so you still get tons of CSP log spam when DEBUG=True

alextreme:
@Bartvaderkin makes sense, thanks!

PR #493, opened 2023-02-22
Improved codestyle in case views and related code

Bartvaderkin:
Some stuff got merged when I wasn't around to review so I'd fix it myself :wink:

### DIFF
--- a/src/open_inwoner/accounts/views/cases.py
+++ b/src/open_inwoner/accounts/views/cases.py
@@ -170,9 +170,9 @@ class CaseListMixin(CaseLogMixin, PaginationMixin):
                     "end_date": getattr(case, "einddatum", None),
                     "description": case.omschrijving,
                     "zaaktype_description": case.zaaktype.omschrijving,
-                    "current_status": case.status.statustype.omschrijving
-                    if case.status
-                    else "",
+                    "current_status": (
+                        case.status.statustype.omschrijving if case.status else ""
+                    ),
                 }
             )
         return updated_cases
@@ -306,25 +306,6 @@ class CaseDetailView(
                 status_type = status_types_mapping[status.statustype]
                 status.statustype = status_type
 
-            # documents
-            internal_upload_enabled = ZaakTypeInformatieObjectTypeConfig.objects.get_visible_ztiot_configs_for_case(
-                self.case
-            ).exists()
-            external_upload_enabled = ZaakTypeConfig.objects.get_visible_zt_configs_for_case_type_identification(
-                self.case.zaaktype.identificatie
-            ).exists()
-
-            if external_upload_enabled:
-                external_upload_url = (
-                    ZaakTypeConfig.objects.get_visible_zt_configs_for_case_type_identification(
-                        self.case.zaaktype.identificatie
-                    )
-                    .get()
-                    .external_document_upload_url
-                )
-            else:
-                external_upload_url = ""
-
             context["case"] = {
                 "identification": format_zaak_identificatie(
                     self.case.identificatie, config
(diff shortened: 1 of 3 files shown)


src/open_inwoner/accounts/views/cases.py
```diff
@@ -306,25 +306,6 @@ def get_context_data(self, **kwargs):
...
-                external_upload_url = (
-                    ZaakTypeConfig.objects.get_visible_zt_configs_for_case_type_identification(
-                        self.case.zaaktype.identificatie
-                    )
-                    .get()
-                    .external_document_upload_url
-                )
-            else:
-                external_upload_url = ""
-
```
Bartvaderkin, line 327: I moved this to a function because this gnarly block clutters the get_context_data. It doesn't need anything from there, so it can get it's own space.

src/open_inwoner/accounts/views/cases.py
```diff
@@ -170,9 +170,9 @@ def process_cases(self, cases: List[Zaak]) -> List[Zaak]:
                     "end_date": getattr(case, "einddatum", None),
                     "description": case.omschrijving,
                     "zaaktype_description": case.zaaktype.omschrijving,
-                    "current_status": case.status.statustype.omschrijving
-                    if case.status
-                    else "",
+                    "current_status": (
+                        case.status.statustype.omschrijving if case.status else ""
+                    ),
```
Bartvaderkin, line 175: Brace conditionals for better grouping after blacking.

src/open_inwoner/accounts/views/cases.py
```diff
@@ -348,16 +329,42 @@ def get_context_data(self, **kwargs):
                 ),
                 "statuses": statuses,
                 "documents": documents,
-                "internal_upload_enabled": internal_upload_enabled,
-                "external_upload_enabled": external_upload_enabled,
-                "external_upload_url": external_upload_url,
                 "allowed_file_extensions": sorted(config.allowed_file_extensions),
+                "anchors": self.get_anchors(statuses, documents),
             }
-            context["anchors"] = self.get_anchors(statuses, documents)
```
Bartvaderkin, line 356: We're building a dictionary just above so this doesn't have to be it's own assignment.

src/open_inwoner/accounts/views/cases.py
```diff
@@ -423,30 +429,28 @@ def handle_document_upload(self, request, form):
 
         file = cleaned_data["file"]
         title = cleaned_data["title"]
-        user_choice = cleaned_data["type"].id
+        document_type = cleaned_data["type"]
```
Bartvaderkin, line 432: The ModelChoiceField returns an instance of the object we want, so no need to take its ID and then later do a `.get()` to retrieve a copy of that same object.

src/open_inwoner/openzaak/documents.py
```diff
@@ -119,9 +120,7 @@ def upload_document(
         "status": "definitief",
         "indicatieGebruiksrecht": False,
         "taal": "dut",
-        "informatieobjecttype": ZaakTypeInformatieObjectTypeConfig.objects.get(
-            id=user_choice
-        ).informatieobjecttype_url,
```
Bartvaderkin, line 124: We already had that object from the ModelChoiceField :cry: 

(see note above)

src/open_inwoner/openzaak/documents.py
```diff
@@ -100,9 +101,9 @@ def upload_document(
     user: SimpleLazyObject,
     file: InMemoryUploadedFile,
     title: str,
-    user_choice: int,
+    informatieobjecttype_url: str,
```
Bartvaderkin, line 104: This function doesn't need to know about `ZaakTypeInformatieObjectTypeConfig`, it just wants the URL for the informatieobjecttype.

src/open_inwoner/openzaak/cases.py
```diff
@@ -303,8 +303,9 @@ def fetch_single_result(result_url: str) -> Optional[Resultaat]:
     return result
 
 
-def connect_case_with_document(case_url: str, document_url: str) -> dict:
+def connect_case_with_document(case_url: str, document_url: str) -> Optional[dict]:
```
Bartvaderkin, line 306: If there is a normal return value but also a bare `return` (eg: None) then the return type should be `Optional`

vaszig (approved):
Thanks for the nice refactoring!

PR #495, opened 2023-02-23
[#1156] Add email and phonenumber to map and location cards

vaszig:


src/open_inwoner/scss/components/Map/_Map.scss
```diff
@@ -20,3 +20,15 @@
...
+.leaflet-content-name {
+  & > h4 {
+    margin: 0 !important;
+  }
+}
+
+.leaflet-content-details {
+  display: flex;
+  row-gap: var(--spacing-small);
+  flex-wrap: wrap;
```
jiromaykin, line 33: You would need to add this to it: 
`flex-direction: column;`
Everything else looks good! 
Explanation: The leaflet-map address etc. will look weird if locations have long names + while still breaking neatly on a hyphen (the location details would be **next** to each other in stead of on top of each other, making it difficult to read without a margin.):
image

src/open_inwoner/js/components/map/index.js
```diff
@@ -75,15 +75,21 @@ class Map {
...
-            .join(' ')}
-          </p>
+        <div class="leaflet-content-name">
+          <h4 class="h4">
+            ${displayName}
+          </h4>
+        </div>
+        <div class="leaflet-content-details p--no-margin">
+          ${finalHTML}
+        </div>
```
Bartvaderkin, line 92: These values aren't html escaped.

Bartvaderkin (changes requested):
In the map bubble, I think we'd want each property on it's own line like on the detail page? It is unusual to concatenate them together.

vaszig:
I had to specify the version of `babel-loader` package since it was updated automatically to the latest version (not compatible with the CI node version).


image

Bartvaderkin (approved):

jiromaykin:
> I had to specify the version of `babel-loader` package since it was updated automatically to the latest version (not compatible with the CI node version).

Which CI node-version are we using? Should we keep NodeJS to the latest version or not? I don't know much about this yet :-)  so I wonder how we will keep track of CI node and change Babel-parser to latest again?

jiromaykin (approved):
Looks all good to me now ,when viewed in browser.

vaszig:
> > I had to specify the version of `babel-loader` package since it was updated automatically to the latest version (not compatible with the CI node version).
> 
> Which CI node-version are we using? Should we keep NodeJS to the latest version or not? I don't know much about this yet :-) so I wonder how we will keep track of CI node and change Babel-parser to latest again?

In CI we are using node 12.22.12, which is not compatible with the latest version of babel-loader. I am not an expert concerning all these, but I know that it's better to pin a problematic package instead of upgrading node (which may cause problems to several packages then).

PR #496, opened 2023-02-23
:lipstick: [#1171] make phone-number in footer clickable

jiromaykin:
This will turn the phone-number into a 'call-able' href.
https://taiga.maykinmedia.nl/project/open-inwoner/task/1171
There is no design for this situation, but I am styling the phonenumber as a default linkstyle.


jiromaykin:
I am not sure if I've added the new translations correctly 
(and if I should actually add the comments into the `.po` file myself with those linenumbers in it)

src/open_inwoner/configurations/models.py
```diff
@@ -209,6 +209,13 @@ class SiteConfiguration(SingletonModel):
         blank=True,
         help_text=_("Visiting intro text on the footer section."),
     )
+    footer_visiting_phonenumber = models.CharField(
```
vaszig, line 212: - You can add a phone validator here. We use it in this kind of fields (`utils/validators/validate_phone_number`) to prevent people from entering unwanted values.
- `max_length` should have the necessary value (you can put 15 here)

vaszig (changes requested):
Except for a small detail, it looks good to me. I am not sure about the translations again. I thought that we were doing the translations once and  not for each branch separately.

vaszig (approved):

Bartvaderkin (approved):

PR #499, opened 2023-02-24
[#1158] Add image field for begeleider user

vaszig:


src/open_inwoner/accounts/admin.py
```diff
@@ -23,7 +24,14 @@ class _UserChangeForm(UserChangeForm):
     def clean(self, *args, **kwargs):
         cleaned_data = super().clean(*args, **kwargs)
 
+        if (
+            cleaned_data.get("photo")
+            and cleaned_data.get("contact_type") != ContactTypeChoices.begeleider
+        ):
+            raise ValidationError(_("Only a 'begeleider' user can add a photo."))
```
alextreme, line 31: I wouldn't limit this to 'begeleider' only, even though it might be requested from a viewpoint of a begeleider if another type of user wants to upload a foto I'd suggest to allow this.

Of course this does open up possible 'uncomfortable' situations, so I'll add a task to double-check this.

vaszig: Yes, I went only for begeleiders because it's the phrase that the user story starts with. I have based task 1191 on this branch as well, should I wait?

alextreme: No need to wait, lets continue with the assumption that this is begeleider-only. 1192 is open to check this assumption.

src/open_inwoner/accounts/models.py
```diff
@@ -66,6 +69,13 @@ class User(AbstractBaseUser, PermissionsMixin):
         max_length=15,
         validators=[validate_phone_number],
     )
+    photo = models.ImageField(
+        verbose_name=_("Begeleider photo"),
```
alextreme, line 73: Lets just use "Foto" (or "Photo")

I don't mind Dutch and English, but combining this in one label hurts :)

src/open_inwoner/accounts/models.py
```diff
@@ -23,7 +22,11 @@
 
 from .choices import ContactTypeChoices, LoginTypeChoices, StatusChoices, TypeChoices
 from .managers import ActionQueryset, DigidManager, UserManager, eHerkenningManager
-from .query import InviteQuerySet, MessageQuerySet, UserQuerySet
+from .query import InviteQuerySet, MessageQuerySet
+
+
+def generate_uuid_photo_name(instance, filename):
+    return "photo_{uuid}/{filename}".format(uuid=uuid4(), filename=filename)
```
Bartvaderkin, line 29: Let's use a sub-directory for the profile pictures, and I think you should only use the extension from the filename (because the name if the file could be personal data). 

So `profile/ea78fd4c-6708-4248-a6e3-91ff9e907a20.jpg`

Bartvaderkin (changes requested):
Are you sure you want to name the field 'photo'? It feels odd, usually it is a 'profile picture' or 'image'?

src/open_inwoner/accounts/models.py
```diff
@@ -23,7 +23,14 @@
 
 from .choices import ContactTypeChoices, LoginTypeChoices, StatusChoices, TypeChoices
 from .managers import ActionQueryset, DigidManager, UserManager, eHerkenningManager
-from .query import InviteQuerySet, MessageQuerySet, UserQuerySet
+from .query import InviteQuerySet, MessageQuerySet
+
+
+def generate_uuid_image_name(instance, filename):
+    filename, file_extension = os.path.splitext(filename)
+    return "profile/{uuid}/{file_extension}".format(
```
Bartvaderkin, line 31: The last separator should be a dot (instead of a slash) so it is still a valid openable file: "profile/{uuid}.{file_extension}"

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #501, opened 2023-02-24
[#1191] Add user image to the profile settings

vaszig:
This is based on #499 which is not yet totally reviewed and merged.

### DIFF
--- a/src/open_inwoner/accounts/forms.py
+++ b/src/open_inwoner/accounts/forms.py
@@ -18,7 +18,12 @@ from open_inwoner.pdc.models.category import Category
 from open_inwoner.utils.forms import LimitedUploadFileField, PrivateFileWidget
 from open_inwoner.utils.validators import validate_charfield_entry
 
-from .choices import EmptyContactTypeChoices, EmptyStatusChoices, LoginTypeChoices
+from .choices import (
+    ContactTypeChoices,
+    EmptyContactTypeChoices,
+    EmptyStatusChoices,
+    LoginTypeChoices,
+)
 from .models import Action, Document, Invite, Message, User
 
 
@@ -56,7 +61,25 @@ class CustomRegistrationForm(RegistrationForm):
         raise ValidationError(_("This user has been deactivated"))
 
 
-class UserForm(forms.ModelForm):
+class BaseUserForm(forms.ModelForm):
+    class Meta:
+        model = User
+        fields = (
+            "display_name",
+            "email",
+            "phonenumber",
+            "photo",
+        )
+
+    def __init__(self, user, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["photo"].help_text = None
+        if user.contact_type != ContactTypeChoices.begeleider:
+            self.fields["photo"].widget = forms.HiddenInput()
+
+
+class UserForm(BaseUserForm):
     class Meta:
         model = User
         fields = (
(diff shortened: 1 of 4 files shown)


src/open_inwoner/accounts/forms.py
```diff
@@ -56,33 +61,46 @@ def clean_email(self):
...
-            "city",
+            "image",
         )
 
+    def __init__(self, user, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields["image"].help_text = None
+        if user.contact_type != ContactTypeChoices.begeleider:
+            self.fields["image"].widget = forms.HiddenInput()
```
Bartvaderkin, line 79: This feels unusual, why kill the help_text, and why the hidden field? Shouldn't we just `del self.fields["image"]`?

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -346,6 +350,86 @@ def test_expected_form_is_rendered(self):
...
         self.assertEqual(type(form), BrpUserForm)
 
+    def test_image_is_saved_when_begeleider_and_default_login(self):
+        self.user.contact_type = ContactTypeChoices.begeleider
+        self.user.save()
+
+        image = Image.new("RGB", (10, 10))
+        byteIO = io.BytesIO()
+        image.save(byteIO, format="png")
+        img_bytes = byteIO.getvalue()
```
Bartvaderkin, line 360: If you have repeating blocks of setup or utility code in the tests it is helpful to move it to a method or function so we have less uninteresting/noisy/distracting stuff in the tests.

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -346,6 +350,86 @@ def test_expected_form_is_rendered(self):
...
+
+        response = self.app.get(self.url, user=self.user, status=200)
+        form = response.forms["profile-edit"]
+        form["image"] = Upload("test_image.png", img_bytes, "image/png")
+        form_response = form.submit()
```
Bartvaderkin, line 365: This block is also showing up often while not very interesting to what we're testing.

Might as well put these together and make a method like `upload_test_image_to_profile_edit_page()` or something.

Bartvaderkin (changes requested):

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -346,6 +363,63 @@ def test_expected_form_is_rendered(self):
...
+            self.user.image.file
+
+        self.user.refresh_from_db()
+
+        self.assertIsNotNone(self.user.image.file)
+
+    def test_image_is_not_saved_when_no_begeleider_and_default_login(self):
+        img_bytes = self.create_test_image_bytes()
+        with self.assertRaises(AssertionError) as e:
+            self.upload_test_image_to_profile_edit_page(img_bytes)
```
Bartvaderkin, line 400: Why would this raise an AssertionError, and why is this assertion new in the test? I was looking at the diff of the commit but I don't see anything like that in the code this replaces. I also don't see anything in the new helpers that could cause this.

Also (related), in general: if you use `assertRaises` it can be very readable to use `assertRaisesMessage` or `assertRaisesRegex` so it is more clear what kind of error happens. Sometimes its obvious (like the ValueError's) but like here it is hard to determine.

Bartvaderkin (changes requested):

src/open_inwoner/accounts/tests/test_profile_views.py
vaszig: I added the `AssertionError` because I changed the form init method (now if the user is not a begeleider I delete the field from the form, as you suggested in your previous comment, so it cannot find the field in the form).

Bartvaderkin (approved):
Very nice update.

PR #502, opened 2023-02-27
[#1149] Implemented e-Suite form's open submission API and added to cases views

Bartvaderkin:
Note: the zgw-consumers update was needed because e-Suite's returns bad pagination fields and I had to modify the `get_paginated_results` utility. (and then it also had some changes itself in the way it ships the enum-choices).


src/open_inwoner/openzaak/tests/test_utils.py
```diff
@@ -190,7 +190,7 @@ def test_get_role_name_display(self):
                     "voornamen": "",
                 },
             )
-            expected = RolTypes.labels[RolTypes.natuurlijk_persoon]
+            expected = RolTypes.natuurlijk_persoon.label
```
Bartvaderkin, line 193: This is because the zgw-consumers dependency update

src/open_inwoner/openzaak/api_models.py
```diff
@@ -140,10 +140,10 @@ class Rol(ZGWModel):
     betrokkene_identificatie: Optional[dict] = None
 
     def get_betrokkene_type_display(self):
-        return RolTypes.values[self.betrokkene_type]
+        return RolTypes[self.betrokkene_type].label
 
     def get_omschrijving_generiek_display(self):
-        return RolOmschrijving.values[self.omschrijving_generiek]
+        return RolOmschrijving[self.omschrijving_generiek].label
```
Bartvaderkin, line 147: This is because the zgw-consumers dependency update

src/open_inwoner/templates/pages/cases/submissions.html
```diff
@@ -0,0 +1,40 @@
...
+                    <td class="table__item">{% trans "Formulier" %}</td>
+                    <td class="table__item">{% trans "Laatste wijziging" %}</td>
+                    <td class="table__item">{% trans "Einde geldigheid" %}</td>
+                    <td class="table__item">&nbsp;</td>
+                </tr>
+            </thead>
+            <tbody>
+                {% for submission in submissions %}
+                    <tr>
+                        <th class="table__header">{% link href=submission.vervolg_link text=submission.naam target="_blank"  %}</th>
```
vaszig, line 26: `target` is not a variable of the template tag. Link component uses by default `target='_blank'` if href starts with 'http'

src/open_inwoner/templates/pages/cases/submissions.html
```diff
@@ -0,0 +1,40 @@
...
+                        <td class="table__item table__item--no-lb">{{submission.datum_laatste_wijziging|date:"d-m-Y"}}</td>
+                        <td class="table__item table__item--no-lb">{{submission.eind_datum_geldigheid|date:"d-m-Y"}}</td>
+                        <td class="table__item">{% button text=submission.naam hide_text=True href=submission.vervolg_link target="_blank" icon="arrow_forward" icon_outlined=True transparent=True %}</td>
```
vaszig, line 29: Same as above with target.

vaszig (changes requested):
I would change the order of the fields in admin as well. Now the FormsAPI is all the way at the bottom and does not help too much. 
Tested locally and works fine!

src/open_inwoner/templates/pages/cases/submissions.html
Bartvaderkin: > Link component uses by default target='_blank' if href starts with 'http'

Ugh.. that seems like a bad assumption or generalisation, but let's not think about it for now and stash that on the template-tag issue pile. I'll remove the variable for now.

Bartvaderkin:
@vaszig I removed the template variable and defined a fieldset for the OpenZaakConfigAdmin (also put the less interesting/huge things in a collapse group).

vaszig (approved):

PR #477, opened 2023-02-16
:art: [#1132] Select-element for contacts-page design

jiromaykin:
select tag needs arrow
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1132

After discussing with Roxanne (designer):

- [x] translation/textually "soort contact" and "type contact" should be the same; so pick one and use consistently (translations), either use "soort contact" everywhere, or use "type contact" everywhere.
- [x] align the Select element to the right
- [x] add icon with expand_more arrow/caret
- [x] discuss with designer if all filter-pages should look consistent
- [ ] delete this PR/branch due to mess with translations and wrong name, so make new branch/PR

PR #503, opened 2023-02-27
:lipstick: [#1132] Select-element caret for contacts-page

jiromaykin:
select tag needs arrow/caret
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1132

After discussing with Roxanne (designer):

- [x] translation/textually "soort contact" and "type contact" should be the same; so pick one and use consistently (TODO later in translations), either use "soort contact" everywhere, or use "type contact" everywhere.
- [x] align the Select element to the right
- [x] add icon with expand_more arrow/caret
- [x] discuss with designer if all filter-pages should look consistent
- [x] delete [the old PR/branch](https://github.com/maykinmedia/open-inwoner/pull/477) due to mess with translations and wrong name there
- [x] refactor Filters section to look like all the others (like in Plans and Actions)


src/open_inwoner/scss/views/_view.scss
```diff
@@ -2,6 +2,30 @@
...
+        grid-template-columns: 1fr 1fr;
+        justify-content: space-between;
+        width: 100%;
+      }
+      @media (max-width: 767px) {
+        > .form__control {
+          display: block;
+          max-width: 90vw;
+
+          select {
```
Bartvaderkin, line 18: here was a theeory about element names in CSS selectors.

Bartvaderkin (changes requested):

vaszig (commented):
It looks nice on desktop version but what about the mobile one? (there is no space between the two elements and is the input correctly aligned?)

image

jiromaykin:
> It looks nice on desktop version but what about the mobile one? (there is no space between the two elements and is the select correctly aligned?)

Thank you - for now I have changed this one to look more like the other pages that have a filter - but this means we are repeating the same pattern for Plans/Actions/Contacts - so in future I probably should refactor them into 1 single JS file (not sure yet...).

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -3,127 +3,143 @@
...
-{% form id="contact-filter" form_object=form method="GET" columns=2 no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action %}
+    <div class="contacts">
+        {% render_form id="contact-filter" form=form method="GET" no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action class="form form--columns-2 form--spaceless form--autosubmit form--horizontal" %}
+            <div class="form__control contacts__filters">
+                <div class="contacts__filter-button">
+                    {% button text=_("Filter") type="button" bordered=True %}
+                </div>
+                <div class="contacts__filter-container">
+                    <p class="p">{% trans "Soort contactpersoon" %}</p>
+                    {% input form.type no_label=True no_help=True icon="expand_more" class="label input" id="id_type" %}
```
vaszig, line 64: - Same as above. Input template tag does not have a `class` variable, so nothing gets applied here.
- If we define an id let's give a more representative name (something like `id_contact_type)`

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -3,127 +3,143 @@
...
-    {% endwith %}
-{% endif %}
+        {% endwith %}
+    {% endif %}
 
-{% form id="contact-filter" form_object=form method="GET" columns=2 no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action %}
+    <div class="contacts">
+        {% render_form id="contact-filter" form=form method="GET" no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action class="form form--columns-2 form--spaceless form--autosubmit form--horizontal" %}
+            <div class="form__control contacts__filters">
+                <div class="contacts__filter-button">
```
vaszig, line 59: It would be better to keep the names consistent. The parent is different from the child (`contacts__filters` and `contacts__filter`

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -3,127 +3,143 @@
...
-    </div>
-    {% endwith %}
-{% endif %}
+        {% endwith %}
+    {% endif %}
 
-{% form id="contact-filter" form_object=form method="GET" columns=2 no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action %}
+    <div class="contacts">
+        {% render_form id="contact-filter" form=form method="GET" no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action class="form form--columns-2 form--spaceless form--autosubmit form--horizontal" %}
+            <div class="form__control contacts__filters">
```
vaszig, line 58: Do we need the `form__control` class here?I am asking because the form generates this class by default when it's needed.

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -3,127 +3,143 @@
...
-        {% endif %}
-    </div>
-    {% endwith %}
-{% endif %}
+        {% endwith %}
+    {% endif %}
 
-{% form id="contact-filter" form_object=form method="GET" columns=2 no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action %}
+    <div class="contacts">
+        {% render_form id="contact-filter" form=form method="GET" no_actions=True horizontal=True spaceless=True autosubmit=True form_action=form_action class="form form--columns-2 form--spaceless form--autosubmit form--horizontal" %}
```
vaszig, line 57: The `class` variable is not part of the template tag definition. The classes you define inside this variable are automatically generated by the template tag since you have passed `spaceless=True`, `autosubmit=True`, `horizontal=True`.

vaszig (changes requested):

vaszig:
Some of my comments should be present in my previous review but I was lost in the html changes (formatting) and didn't mention them.

src/open_inwoner/templates/pages/profile/contacts/list.html
jiromaykin: Thanks! I have no idea why I did it like this...

src/open_inwoner/templates/pages/profile/contacts/list.html
jiromaykin: :-D I was using the naming convention from the 'Actions' files `open_inwoner/js/components/actions/index.js` - where it seems a collection of filters is called as the multiple 'filters',  but I can change this, no prob. 
I will then also have to change this in the Actions files, or else we will have different conventions.

vaszig: No problem then, you can keep it as is.

jiromaykin:
I was trying to solve a merge conflict by rebasing but somehow got everyone else's commits into this branch as well :-(
The only files needed for this PR are:

- src/open_inwoner/js/components/contacts/index.js
- src/open_inwoner/scss/components/Contacts/Contacts.scss
- src/open_inwoner/templates/pages/profile/edit.html

jiromaykin:
@vaszig @Bartvaderkin  I was trying to solve a merge conflict by _rebasing_ so hopefully this all went correct!
This looks a bit crazy now, with all the rebased commits, but my own code should be correct...

Bartvaderkin:
@jiromaykin This rebase didn't really go as it should, your PR should list all other commits. Also merging from develop back into a feature branch is not what we'd like to see (this really messes with the git history). 

If you can't fix this branch it is also an option to make a new branch and do things like Cherry-Pick from this, or if all else fails just manually copy your changes (using Browse At Revision).

PR #504, opened 2023-02-28
[#1186] Hiding Geboorteland from Mijn Gegevens (BRP)

alextreme:


vaszig (approved):
If it's not temporary removal, we should maybe update the view as well (MyDataView ), by removing the field from there as well.

alextreme:
> If it's not temporary removal, we should maybe update the view as well (MyDataView ), by removing the field from there as well.

Good idea, fixed

PR #505, opened 2023-02-28
[#1159] Add profile image when accepting an invitation

vaszig:
This is based on #499.

### DIFF
--- a/src/open_inwoner/templates/pages/profile/contacts/list.html
+++ b/src/open_inwoner/templates/pages/profile/contacts/list.html
@@ -51,9 +51,9 @@
                 </ul>
             {% endif %}
         </div>
-        {% if approvals_count == 1 %}
+        {% if approvals_count == 1 and request.user.image %}
             <div class="avatar">
-                <img class="" src="{{ request.user.image|thumbnail_url:'avatar' }}" alt="{{ image.default_alt_text }}" />
+                <img src="{{ request.user.image|thumbnail_url:'avatar' }}" alt="{{ image.default_alt_text }}" />
             </div>
         {% endif %}
     {% endwith %}
--- a/src/open_inwoner/templates/pages/profile/contacts/list.html
+++ b/src/open_inwoner/templates/pages/profile/contacts/list.html
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n button_tags link_tags icon_tags pagination_tags form_tags dropdown_tags messages_tags %}
+{% load i18n button_tags link_tags icon_tags pagination_tags form_tags dropdown_tags messages_tags thumbnail %}
 
 
 {% block content %}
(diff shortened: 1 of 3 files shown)


alextreme:
> @alextreme do we want this for multiple invitations as well? Below you can see the result for a single pending invitation:
> 
> contact_approval_with_img
> 
> It's a bit complicated - needs a design imo, if we want this for the multiple invitations as well
> 
> multiple_contacts

Doing this for the single-invitation UI is sufficient in my eyes.

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -11,46 +11,53 @@ <h1 class="h1" id="title">
...
-                {% url 'accounts:contact_approval' uuid=pending_approval.uuid as approval_url %}
-                {% render_form form=None method="POST" form_action=approval_url id="approval_form" spaceless=True show_notifications=True %}
-                    {% csrf_token %}
-                    {% button_row %}
-                        {% button text=_("Weigeren") name="contact_reject" secondary=True bordered=True %}
-                        {% button text=_("Accepteren") name="contact_approve" primary=True %}
-                    {% endbutton_row %}
-                {% endrender_form %}
+            <div class="avatar">
+                <img class="" src="{{ request.user.image|thumbnail_url:'avatar' }}" alt="{{ image.default_alt_text }}" />
```
Bartvaderkin, line 56: Empty `class=""` here.

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -11,46 +11,53 @@ <h1 class="h1" id="title">
 <!--List of contact requests: receiver's view-->
 {% if pending_approvals %}
     {% with pending_approvals.count as approvals_count %}
-    <div class="approval{% if approvals_count == 1 %} approval--solo{% endif %}">
-        <div class="approval__header">
-            <h2 class="h2">{% trans "U bent toegevoegd als contactpersoon" %}</h2>
-            <p class="p">{% trans "By accepting you agree on sharing your personal data (first name, last name and phonenumber)." %}</p>
+    <div class="approval-container{% if approvals_count == 1 %}--solo{% endif %}">
+        <div class="approval{% if approvals_count == 1 %} approval--solo{% endif %}">
```
Bartvaderkin, line 15: I noticed a double `approval--solo` if/else here but I don't know if it is a thing.

Bartvaderkin (changes requested):
I noticed the PR had the wrong ticket number (1559), so I edited it, but I also noticed the commit had 1559 (branch is 1159).

src/open_inwoner/templates/pages/profile/contacts/list.html
vaszig: The parent is `container-approval--solo` and the child `approval--solo`,  but I couldn't think of another way for having this condition in 2 different class names.

Bartvaderkin: I'd expect you could select on the parent? 

```
.approval-container .approval { /* normal rules for .approval */ }
.approval-container--solo .approval { /* rules for .approval when solo */  }
```

Bartvaderkin (approved):

alextreme:
@jiromaykin feel free to double-check

jiromaykin (commented):
Where is the image for Begeleiders set? If I login as a Begeleider, i cannot reach the admin dashboard + I don't see a way to upload an avatar in the Contacts section?

vaszig:
> Where is the image for Begeleiders set? If I login as a Begeleider, i cannot reach the admin dashboard + I don't see a way to upload an avatar in the Contacts section?

The field is visible for all users in the admin (user). You can set it only if the user is a begeleider type of user.
If you cannot reach the admin page means that you are trying to log in as a regular user and not as an administrator.

image

jiromaykin:
> ...You can set it only if the user is a begeleider type of user. If you cannot reach the admin page means that you are trying to log in as a regular user and not as an administrator.

I can see the image can be set only for Begeleider users; and I guess it only becomes visible in the invitation message? So now this means that Begeleider users cannot set their image themselves + it is hardly ever visible, not even in the contacts-list table?

vaszig:
> I can see the image can be set only for Begeleider users; and I guess it only becomes visible in the invitation message? So now this means that Begeleider users cannot set their image themselves + it is hardly ever visible, not even in the contacts-list table?

Yes, it's only for begeleiders because this is the request (at least for now). This means that a begeleider is able to add an image (via the admin page) and yes, it's not visible anywhere else apart from the invitation that you can see above. The point of the task is to be able to see an image of the begeleider who wants to add a user as a contact.

jiromaykin (approved):
All seems good: correct image + correct names are shown in the invitation.

PR #506, opened 2023-02-28
:lipstick: [#1067] Mobile menu-structure with submenu toggle

jiromaykin:
This PR can only be reviewed properly when checked-out and run in the browser.
See issue https://taiga.maykinmedia.nl/project/open-inwoner/task/1067

- [x] remove borders around hamburger when open
- [x] move Search bar to top of menu (but under hamburger)
- [x] remove UL dots (`list-style: none;`)
- [x] show 'close' cross when menu is opened; show 'hamburger' when menu is closed
- [x] switch order of text-actions to bottom on mobile (so 'logged in' state is shown on bottom of menu)
- [x] add caret for pages with subpages
- [x] change color of Opened subpage-item to Primary color
- [x] make top pages NOT clickable IF they have subpages
- [x] make subpages still clickable when shown, but not reachable by Tab when closed
- [x] show/hide subpages on toggle
- [x] align icon for user and arrow when logged in etc [as in mobile design here](https://projects.invisionapp.com/share/3E135L29X8BP#/screens/471881065).
- [x] improve margins and lineheights, also between header logo and search field, check for borders
- [x] check z-indexes and overflow
- [x] check all in Browserstack for Android etc.


jiromaykin:
@vaszig and @bart-maykin : don't be afraid to review this - the best way to do this, is to checkout the branch and actually view it in the browser in moblie views; (just looking at all the code might be a bit too much for humans :-)

src/open_inwoner/templates/master.html
```diff
@@ -33,7 +33,8 @@
     </head>
 
     <body hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'>
-        {% header categories=menu_categories request=request breadcrumbs=breadcrumbs search_form=search_form has_general_faq_questions=has_general_faq_questions %}
+        {% get_solo 'configurations.SiteConfiguration' as config %}
```
vaszig, line 36: This is not necessary, we already load the utils so we have access to the context_processors (like we do above with the colors and below with the `has_general_faq_questions`). So, you can add `show_plans=show_plans` to the header template tag.

src/open_inwoner/templates/master.html
```diff
@@ -1,4 +1,4 @@
-{% load static i18n header_tags card_tags footer_tags button_tags notification_tags anchor_menu_tags view_breadcrumbs utils session_tags django_htmx %}<!DOCTYPE html>
+{% load static i18n header_tags card_tags footer_tags button_tags notification_tags anchor_menu_tags view_breadcrumbs utils session_tags django_htmx solo_tags %}<!DOCTYPE html>
```
vaszig, line 1: `solo_tags` not needed here, we have utils.

vaszig (changes requested):
When testing it on the browser looks good. Generally I think it would be easier and better to have smaller PR's (here we combine 14 tasks) when we have to do with changes that may affect a lot of css and html files (It has to do with one task, indeed, but I think we should split this up in the future).

src/open_inwoner/components/templates/components/Header/Header.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n logo_tags header_tags form_tags link_tags solo_tags icon_tags thumbnail %}
+{% load i18n logo_tags header_tags form_tags link_tags solo_tags icon_tags thumbnail string_tags solo_tags %}
```
vaszig, line 1: `solo_tags` was already here, we can remove the added one.

src/open_inwoner/templates/master.html
Bartvaderkin: Heh, @vaszig is correct. Note this get_solo was both my suggestion (because we looked at the other code that did something like that), and I even added the line to the context processor earlier: https://github.com/maykinmedia/open-inwoner/blob/95cbe9dfa17ccc250e0766879ceb0a3f98b84d1d/src/open_inwoner/utils/context_processors.py#L79 :confused:

Bartvaderkin (changes requested):
Very cool. One note, when I'm logged in my long username the layout pushes sideways:

image

src/open_inwoner/components/templates/components/Header/Header.html
```diff
@@ -11,43 +11,103 @@
...
+                        </li>
+
+                        {% if request.user.is_authenticated %}
+                            <li class="primary-navigation__list-item">
+                                {% link text=_("Mijn profiel") href='accounts:my_profile' icon="inventory_2" icon_position="before" icon_outlined=True %}
+                            </li>
+                            <li class="primary-navigation__list-item">
+                                {% link text=_('Mijn berichten') href='accounts:inbox' icon="inbox" icon_position="before" %}
+                                {% if request.user.get_new_messages_total %}
+                                    {% with "("|addstr:request.user.get_new_messages_total|addstr:")" as message_total %}
```
Bartvaderkin, line 60: Pretty impressive filter use :smile: 

The `request.user.get_new_messages_total` is doing queries though, so it is better to put its result into a variable and re-use it, instead of querying the same thing twice.

jiromaykin: @Bartvaderkin  I copied all of the already existing code from the PrimaryNavigation template to the Header template, so I did not come up with these myself; this will be a learning experience for me!

jiromaykin:
> ...One note, when I'm logged in my long username the layout pushes sideways

Oh dear... @Bartvaderkin, I will think of a way to solve this for someone named Wolfeschlegelsteinhausenbergerdorff-Senior and will probably use "get_short_name" for mobile view + I'll probably talk to the designer.

src/open_inwoner/components/templates/components/Header/Header.html
jiromaykin: With the help of Vasileios I added `{% with request.user.get_new_messages_total as total_messages %}` in both header and Primary navigation.

alextreme:
> > ...One note, when I'm logged in my long username the layout pushes sideways
> 
> Oh dear... @Bartvaderkin, I will think of a way to solve this for someone named Wolfeschlegelsteinhausenbergerdorff-Senior and will probably use "get_short_name" for mobile view + I'll probably talk to the designer.

Lets not let perfect be the enemy of good, cutting off very long names on mobile would be acceptable in my eyes

src/open_inwoner/templates/master.html
jiromaykin: I think I did it correctly now, but keeping this comment for future reference.

jiromaykin:
> [...]cutting off very long names on mobile would be acceptable in my eyes

OK, I will add a cut off -  I only just found out something like `text-overflow: ellipsis` exists in CSS so I will use that.

src/open_inwoner/utils/context_processors.py
```diff
@@ -76,6 +76,7 @@ def settings(request):
         "siteimprove_id": config.siteimprove_id,
         "siteimprove_enabled": config.siteimprove_enabled,
         "show_plans": config.show_plans,
+        "show_cases": config.show_plans,
```
Bartvaderkin, line 79: Typo in `"show_cases": config.show_plans,`

PR #507, opened 2023-03-01
[#1163] Add optional banner to the plan list view

vaszig:


src/open_inwoner/utils/context_processors.py
```diff
@@ -65,6 +65,7 @@ def settings(request):
             config.hero_image_login.file.url if config.hero_image_login else ""
         ),
         "favicon": (config.favicon.file.url if config.favicon else ""),
+        "plans_banner": (config.plans_banner.file.url if config.plans_banner else ""),
```
Bartvaderkin, line 68: Does this have to be in global context for every page? I'd expect this only in the context of the plan view(s) where this is displayed.

(I'm aware all dynamic text for every page is also in here (and some specific things like the general_faq_questions) but that's a legacy decision we got to live with for now)

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #508, opened 2023-03-01
[#1182] Updated ZGW OAS copies for testing to ~1.2.0 from VNG-Realisatie

Bartvaderkin:


Bartvaderkin:
Sadly this is a no-go for now, due to support limitations in the testing infrastructure:

https://github.com/maykinmedia/zgw-consumers/issues/62

PR #516, opened 2023-03-03
[#1172] Removal of opening Productview links in new tab

vaszig:

PR #517, opened 2023-03-03
[#1128] Add icon to cases document upload button

vaszig:
Based on #516 .

### DIFF
--- a/src/open_inwoner/templates/pages/cases/status.html
+++ b/src/open_inwoner/templates/pages/cases/status.html
@@ -42,7 +42,7 @@
                 {% elif case.external_upload_enabled %}
                     <p class="p">{% trans "By clicking the button below you can upload a document. This is an external link and you will be redirected to a different system." %}</p>
                     {% button_row %}
-                        {% button href=case.external_upload_url text=_("Document toevoegen") primary=True %}
+                        {% button href=case.external_upload_url text=_("Document toevoegen") primary=True icon="open_in_new" icon_position="after" %}
                     {% endbutton_row %}
                 {% endif %}
 
--- a/src/open_inwoner/templates/pages/cases/status.html
+++ b/src/open_inwoner/templates/pages/cases/status.html
@@ -42,7 +42,7 @@
                 {% elif case.external_upload_enabled %}
                     <p class="p">{% trans "By clicking the button below you can upload a document. This is an external link and you will be redirected to a different system." %}</p>
                     {% button_row %}
-                        {% button href=case.external_upload_url text=_("Document toevoegen") primary=True icon="open_in_new" icon_position="after" %}
+                        {% button href=case.external_upload_url text=_("Document toevoegen") title=_("Opens new window") primary=True icon="open_in_new" icon_position="after" %}
                     {% endbutton_row %}
                 {% endif %}
 


vaszig:
Only the last commit is for review.

Bartvaderkin (approved):

src/open_inwoner/templates/pages/cases/status.html
```diff
@@ -42,7 +42,7 @@ <h2 class="h2" id=>{% trans "Document toevoegen" %}</h2>
                 {% elif case.external_upload_enabled %}
                     <p class="p">{% trans "By clicking the button below you can upload a document. This is an external link and you will be redirected to a different system." %}</p>
                     {% button_row %}
-                        {% button href=case.external_upload_url text=_("Document toevoegen") primary=True %}
+                        {% button href=case.external_upload_url text=_("Document toevoegen") primary=True icon="open_in_new" icon_position="after" %}
```
jiromaykin, line 45: For accessibility purposes the icon here needs an aria-label with the translatable text "Opens new window" - but it's also OK to put the aria-label in the Anchor link if that's easier.

jiromaykin (changes requested):
So here is an official request :-) with the same accessibility issue as stated: 
"For accessibility purposes the icon here needs an aria-label with the translatable text-string "Opens new window" - but it's also OK to put the aria-label in the Anchor link if that's easier."

vaszig:
> So here is an official request :-) with the same accessibility issue as stated: "For accessibility purposes the icon here needs an aria-label with the translatable text-string "Opens new window" - but it's also OK to put the aria-label in the Anchor link if that's easier."

@jiromaykin done. Added the aria-label and rebased in order to have a cleaner branch as well.

jiromaykin:
A 'title' attribute actually is not accessible for screen-readers (but yes, it is still necessary for other users, so keep the title like you have now);
but it still needs an actual _**aria-label** attribute_; 
in HTML, it looks like this:
`aria-label="Opens new window"`
(but with translation)

Maybe I am saying too much here, but just a little more info:
- Do not use the attribute with the similar name "aria-labelledby" here because that's only for ID's.
- If you cannot add the aria-label, another trick to make this accessible is by adding an invisible Span element that contains the text "Opens new window" and **[is made invisible with CSS](https://webaim.org/techniques/css/invisiblecontent/)** by adding a new css class to the project, like 'visuallyhidden' or 'sr-only' and set its properties something like this:

`.sr-only {
position:absolute;
left:-10000px;
top:auto;
width:1px;
height:1px;
overflow:hidden;
}`

I hope this is clear and helpful.

vaszig:
> A 'title' attribute actually is not accessible for screen-readers (but yes, it is still necessary for other users, so keep the title like you have now); but it still needs an actual _**aria-label** attribute_; in HTML, it looks like this: `aria-label="Opens new window"` (but with translation)
> 
> Maybe I am saying too much here, but just a little more info:
> 
>     * Do not use the attribute with the similar name "aria-labelledby" here because that's only for ID's.
> 
>     * If you cannot add the aria-label, another trick to make this accessible is by adding an invisible Span element that contains the text "Opens new window" and **[is made invisible with CSS](https://webaim.org/techniques/css/invisiblecontent/)** by adding a new css class to the project, like 'visuallyhidden' or 'sr-only' and set its properties something like this:
> 
> 
> `.sr-only { position:absolute; left:-10000px; top:auto; width:1px; height:1px; overflow:hidden; }`
> 
> I hope this is clear and helpful.

By defining the variable `title` of the template tag (which, if defined, adds an aria-label attribute according to the template) I wanted to add an aria-label attribute. Maybe I don't get something here but with my changes there is an aria-label attribute:

image

jiromaykin:
> By defining the variable `title` of the template tag (which, if defined, adds an aria-label attribute according to the template)

Oh wow, I did not know that - sorry - i will approve this :-)

jiromaykin (approved):

vaszig:
> Oh wow, I did not know that - sorry - i will approve this :-)

No problem..the template tags are a bit complicated. Thanks for the helpful comments though!

PR #519, opened 2023-03-06
:art: [#1168] Make Theme cards entirely clickable

jiromaykin:
See issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1168 
Theme/Category- cards need to be entirely clickable, unless they show Subthemes/Products.


jiromaykin:
The biggest challenge here was the fact that it is impossible to have a-hrefs inside a-hrefs in HTML, so this is the same card but with an if/else state - in future i will need to simplify these, but this is for the Theme cards only for now:

Screenshot from 2023-03-06 16-34-47

Bartvaderkin (approved):

vaszig (approved):

PR #520, opened 2023-03-07
:art: [#1132] Add select-caret and filters for contacts

jiromaykin:
select tag needs arrow/caret
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1132

After discussing with Roxanne (designer):

- [x] translation/textually "soort contact" and "type contact" should be the same; so pick one and use consistently (TODO later in translations), either use "soort contact" everywhere, or use "type contact" everywhere.
- [x] align the Select element to the right
- [x] add icon with expand_more arrow/caret
- [x] discuss with designer if all filter-pages should look consistent
- [x] delete [the old PR/branch](https://github.com/maykinmedia/open-inwoner/pull/503) due to mess with merge-conficts, translations and wrong name.
- [x] refactor Filters section to look more like all the others (like in Plans and Actions)
+ Today I just found out that these filters should only be visible in the ‘Begeleider’ view, but for now I'll just make it look like all the other filter-views that we have. + Adding a Search field would be a task for later future.


jiromaykin:
Because I'm surrounding the Contacts section with a `Div`, I had to indent the code inside; hence why it looks like there is a lot of 'new' code.

src/open_inwoner/js/components/contacts/index.js
```diff
@@ -0,0 +1,18 @@
+class Contacts {
+  constructor(node) {
+    this.node = node
+    this.node.addEventListener('click', this.toggleOpen.bind(this))
+  }
+
+  toggleOpen(event) {
+    event.preventDefault()
+    this.node.parentElement.parentElement.classList.toggle(
```
Bartvaderkin, line 9: For the future: selecting up with a .parent.parent type of lookup is generally not a great pattern. It's better to select the outside first and then grab sub-parts, or use a real selector.

Bartvaderkin (approved):

PR #523, opened 2023-03-08
[#1160] Add cropping for profile image

vaszig:


jiromaykin (changes requested):
The problem here is really in the file:
 **src/open_inwoner/scss/components/Form/Form.scss**

where this selector is causing multiple problems inside the project, it is selecting _everything_ that is not a button (but...... almost _everything_ is NOT a button anyway.... 😳 so it's too generic):

`&__control > .label *:not(.button) {
    width: 100%;
  }`

So my proposal is to replace this with multiple slightly more specific selectors there, like this: 

```
&__control > .label .input,
  &__control > .label .input[type='file'] {
    width: 100%;
    max-width: 100%;

    @media (min-width: 768px) {
      max-width: 50%;
    }
  }
```
I think i have selected all types of form elements here, that should still work in all other pages, 
if not, new classes/attributes/wildcards can be added to it: I have also added a mobile-first/media-query
We need to make the forms look smaller on Desktop anyway, like in the design:
https://projects.invisionapp.com/share/UF134RWMWCK8#/screens/471596624
They can be 100% on mobile.
Not using '100vw' here because that may cause horizontal scrolling in some places.

jiromaykin:
> @jiromaykin ...with the change above (width for desktop at 50%) we had some weird results (see screenshots below). [...] create another task for updating generally the forms according to the design.

Thanks for checking even more - I will add this task to my mobile-sprint list, so what you did here is fine for now.

jiromaykin (approved):

Bartvaderkin (approved):
It'd be neat if the form inputs would render conditionally, so we'd only have the form render logic in once place (the `if user.contact_type == "begeleider"` in the form class) but that's for another time.

PR #524, opened 2023-03-08
[#1185] Outgoing requests tracking

vaszig:
This is an approach according to the fact that we already use requests library (same for the external libraries we use for the APIs), by adding a hook to each request (based on a session). Also this approach was discussed with Sergei and Bart at some point.

@Bartvaderkin @alextreme feel free to have a first look and if you agree I can continue with making this more stable.


alextreme:
I think this would be a neat way to do this. Couple of thoughts:
- a Django setting to toggle this on/off (I think the default should be on if installed, but this allows it to easily be disabled if not wanted/necessary)
- Tests :P  
- Split this off into a separate/external Django app to allow easy reuse (can be before or after we've tested it out in OIP)

Looking further this only works for requests when making use of requests-sessions, correct? So this does mean that plain `requests.get()` calls would need to be rewritten. I'd prefer to also track requests when sessions aren't used (so we have a drop-in component for this), but I don't know if that's possible and for an initial version that we can try out as long as we see the BRP and ZGW requests being sent this would be sufficient.

vaszig:
> Looking further this only works for requests when making use of requests-sessions, correct? So this does mean that plain `requests.get()` calls would need to be rewritten. I'd prefer to also track requests when sessions aren't used (so we have a drop-in component for this), but I don't know if that's possible and for an initial version that we can try out as long as we see the BRP and ZGW requests being sent this would be sufficient.

Requests library sends under the hood all requests through `requests.Session`, so a plain request with `requests.get("url")` is logged as well ([https://github.com/psf/requests/blob/main/requests/api.py#L58](url)) I hope this is what you mean.

alextreme:
Compromise is to put this in src/external_api_logging/ to allow easier migration to a third-party Django app

src/open_inwoner/conf/base.py
```diff
@@ -370,9 +383,22 @@
...
+            "propagate": True,
+        },
     },
 }
 
+
+#
+# LOG OUTGOING REQUESTS
+#
+LOG_OUTGOING_REQUESTS_ENABLED = config("LOG_OUTGOING_REQUESTS_ENABLED", default=False)
```
Bartvaderkin, line 398: I think we might not need `LOG_OUTGOING_REQUESTS_ENABLED`? It now just works like `LOG_OUTGOING_REQUESTS_DB_SAVE` (and the stream logging is now controlled by LOGGING)

src/log_outgoing_requests/tests/test_logging.py
```diff
@@ -0,0 +1,109 @@
...
+
+        methods = [
+            ("POST", requests.post, m.post),
+            ("PUT", requests.put, m.put),
+            ("PATCH", requests.patch, m.patch),
+            ("DELETE", requests.delete, m.delete),
+            ("HEAD", requests.head, m.head),
+        ]
+
+        for method, func, mocked in methods:
```
Bartvaderkin, line 50: :+1: nice

src/log_outgoing_requests/log_requests.py
```diff
@@ -0,0 +1,36 @@
...
+
+logger = logging.getLogger("requests")
+
+
+def hook_requests_logging(response, *args, **kwargs):
+    """
+    A hook for requests library in order to add extra data to the logs
+    """
+    extra = {"requested_at": timezone.now(), "req": response.request, "res": response}
+    logger.debug("External request", extra=extra)
```
Bartvaderkin, line 16: For consistency this message string should also be 'Outgoing' instead of 'External' (this applies everywhere from the old name)

src/log_outgoing_requests/models.py
```diff
@@ -0,0 +1,63 @@
...
+        verbose_name=_("Timestamp"),
+        help_text=_("This is the date and time the API call was made."),
+    )
+    trace = models.TextField(
+        verbose_name=_("Trace"),
+        blank=True,
+        null=True,
+        help_text=_("Text providing information in case of request failure."),
+    )
+
```
Bartvaderkin, line 55: I noticed we don't store headers here, but we do print then in the stream logger.

src/log_outgoing_requests/admin.py
```diff
@@ -0,0 +1,23 @@
...
+        "method",
+        "response_ms",
+        "timestamp",
+        "trace",
+    )
+    readonly_fields = fields
+    list_display = fields
+    list_filter = ("method", "status_code", "hostname")
+    search_fields = ("path",)
+    date_hierarchy = "timestamp"
```
Bartvaderkin, line 23: Couple of cosmetics/UIX points:

- Add full URL readonly field to admin detail (for copy-paste convenience when debugging issues)
- Remove `trace` column form admin list (too large)
- Override admin `has_change_permission`/`has_add_permission` (not relevant here, but delete could be in case of sensitive data issue)
- Add more fields to search_fields (`params, query_params, hostname`)

src/log_outgoing_requests/formatters.py
```diff
@@ -0,0 +1,31 @@
...
+                ---------------- response ----------------
+                {res.status_code} {res.reason} {res.url}
+                {reshdrs}
+                
+            """
+            ).format(
+                req=record.req,
+                res=record.res,
+                reqhdrs=self._formatHeaders(record.req.headers),
+                reshdrs=self._formatHeaders(record.res.headers),
```
Bartvaderkin, line 28: I like the header information for debugging, but I'm not sure if we can store that in logs. @alextreme can pitch in on that.

src/log_outgoing_requests/admin.py
```diff
@@ -0,0 +1,23 @@
+from django.contrib import admin
+
+from .models import OutgoingRequestsLog
+
+
+@admin.register(OutgoingRequestsLog)
```
Bartvaderkin, line 6: Good name; 'outgoing' is a much better term then 'external'

Bartvaderkin (changes requested):

src/log_outgoing_requests/handlers.py
```diff
@@ -0,0 +1,37 @@
...
+            trace = None
+
+            # save only the requests coming from the library requests
+            if record and record.getMessage() == "External request":
+                if record.exc_info:
+                    trace = traceback.format_exc()
+
+                parsed_url = urlparse(record.req.url)
+
+                kwargs = {
```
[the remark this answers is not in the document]
Bartvaderkin: I think we also want to store the protocol (eg: HTTPS or not).

src/log_outgoing_requests/models.py
```diff
@@ -1,41 +1,56 @@
...
 
 
 class OutgoingRequestsLog(models.Model):
+    url = models.URLField(
+        verbose_name=_("URL"),
+        blank=True,
+        default="",
+        help_text=_("The url of the outgoing request."),
+    )
     hostname = models.CharField(
```
Bartvaderkin, line 15: I'd put a comment here to note that `.hostname` is obviously part of `.url` but we also have it as a field so we can filter on it in the admin.

src/log_outgoing_requests/models.py
```diff
@@ -61,3 +76,8 @@ def __str__(self):
         return ("{hostname} at {date}").format(
             hostname=self.hostname, date=self.timestamp
         )
+
+    @cached_property
+    def query_params(self):
+        parsed_url = urlparse(self.url)
+        return parsed_url.query
```
Bartvaderkin, line 83: This is ok for now since we only take the query, but if we'd want more url components you'd cache the result and access that like so:

```python
@cached_property
def url_parsed(self):
    return urlparse(self.url)

@property
def query_params(self):
    return self.url_parsed.query

@property
def scheme(self):
    return self.url_parsed.scheme
```

Bartvaderkin (approved):

PR #526, opened 2023-03-10
:lipstick: [#1169] Make images/icons have static sizes in cards

jiromaykin:
issue here: https://taiga.maykinmedia.nl/project/open-inwoner/task/1169 
This would solve the problem where 'icons' no longer look like this (with potentially cut-off heads): when the screensize gets smaller than the container - the only way to do this would be to set static sizes for the cards on all screenwidths, and inform the user which sizes need to be used for icons.

Note: this is not for images in Product cards.

ToDo:

- [x] set statics max-sizes for Grid columns and Cards: max. of 230 px/columnsize wide, image max-height:  100px. Except for the cards on Registration page (and elsewhere?)

other todo's to make the card align with design will be done later when decision is made about horizontal scroll and about the number of carsds in 1 column on mobile

cut-off-heads

### DIFF
--- a/src/open_inwoner/pdc/views.py
+++ b/src/open_inwoner/pdc/views.py
@@ -60,7 +60,7 @@ class HomeView(CommonPageMixin, TemplateView):
     def get_context_data(self, **kwargs):
         config = SiteConfiguration.get_solo()
 
-        limit = 3 if self.request.user.is_authenticated else 4
+        limit = 4
         kwargs.update(categories=Category.objects.published().order_by("name")[:limit])
         kwargs.update(product_locations=ProductLocation.objects.all()[:1000])
         kwargs.update(
@@ -86,7 +86,7 @@ class HomeView(CommonPageMixin, TemplateView):
             and self.request.user.selected_themes.exists()
         ):
             kwargs.update(
-                categories=self.request.user.selected_themes.order_by("name")[:3]
+                categories=self.request.user.selected_themes.order_by("name")[:limit]
             )
         elif highlighted_categories:
             kwargs.update(categories=highlighted_categories)
--- a/src/open_inwoner/scss/views/Home.scss
+++ b/src/open_inwoner/scss/views/Home.scss
@@ -1,0 +1,11 @@
+/// Cards on Home Page and Theme page
+
+.home {
+  .card-container {
+    grid-template-columns: repeat(auto-fit, 228px);
+
+    .card {
+      max-width: 360px;
+    }
+  }
+}
--- a/src/open_inwoner/accounts/templates/accounts/registration_necessary.html
+++ b/src/open_inwoner/accounts/templates/accounts/registration_necessary.html
@@ -3,6 +3,7 @@
 
 {% block content %}
 
+<div class="registration-grid">
 {% render_grid %}
     {% render_column span=9 %}
         {% render_card tinted=True  %}
@@ -11,5 +12,6 @@
         {% endrender_card %}
     {% endrender_column %}
 {% endrender_grid %}
+</div>
 
 {% endblock content %}
--- a/src/open_inwoner/accounts/templates/django_registration/registration_form.html
+++ b/src/open_inwoner/accounts/templates/django_registration/registration_form.html
@@ -3,6 +3,7 @@
 
 {% block content %}
 
+<div class="registration-grid">
 {% render_grid %}
     {% render_column start=5 span=5 %}
         {% render_card direction='horizontal' tinted=True %}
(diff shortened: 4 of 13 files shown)


alextreme:
> @alextreme We could decide to add this version for now, as a form on '_intermediate but improved_' state - this version doesn't have the 'fallback' option removed yet though - should we make it required to always choose an icon?

No, as discussed on monday we can keep the fallback option as-is, as long as we don't cause weird aspect ratio's anymore this should be fine.

Happy to merge if you and Vasileios are happy with this

jiromaykin:
This intermediate stage still needs to be reviewed.

src/open_inwoner/pdc/views.py
```diff
@@ -60,7 +60,7 @@ def page_title(self):
     def get_context_data(self, **kwargs):
         config = SiteConfiguration.get_solo()
 
-        limit = 3 if self.request.user.is_authenticated else 4
+        limit = 4
```
vaszig, line 63: We can remove this variable now as the limit is not dynamic any more. Let's use the number directly in each queryset.

vaszig (changes requested):
I am a bit confused with this PR. I understood that we want the changes only for the theme-cards and not for the product ones (which will be fixed in another PR). If I am wrong let me know please.

jiromaykin:
> I am a bit confused with this PR. I understood that we want the changes only for the theme-cards and not for the product ones 

This is a PR that will change overall responsiveness of all of the columns so everything looks better already before continuing to fine-tuning later; so this one is more about the weird resizing of ALL of the card-columns -this means the Theme card images will remain stable - the Product card images don't need responsiveness for now so they don't need to change here;
Example: if you change screenwidths gradually in the old version it's all very 'elastic' but in this PR all cards have become 'block like' and will drop to the next line if the screenwidth becomes smaller - hope this helps a bit.

### old columns that need to become 'static' on mobile example:

Screenshot from 2023-03-17 16-44-57

src/open_inwoner/pdc/views.py
jiromaykin: In future I may have to make a difference between the mobile limit and the desktop limit, which hopefully can be done in the templates somewhere (not sure yet), so I guess this can be cleaned up.

jiromaykin: @vaszig  I now see the 'limit' is used 5 times in this file; is it really a good idea to remove the 'limit' variable? I would think it is cleaner to keep it and reuse it instead of the value? (in javascript I would keep the variable, not sure if this is also Python convention).

vaszig: I would use a variable if I wanted first to evaluate something and then reuse it (Not in Python, in every language). Here we don;t need to evaluate something, it's just an integer. But this is my knowledge, @Bartvaderkin can you tell us your opinion please?

Bartvaderkin: @vaszig @jiromaykin I would keep the variable, so we can see all the usages are linked to the same value. If we'd put the plain number everywhere you get a magic number situation where it isn't clear which numbers are related or just happen to have the same value.

jiromaykin: @Bartvaderkin @vaszig Hoorrah! in Javascript I would use a '**const**' for numbers like these - so I declare this PR as finished and ready for review again :-)

jiromaykin:
@vaszig This may not look very impressive, but hopefully this works well as an in-between stage with some new CSS classes for future views; hopefully I haven't made any columns/views act all weird here.

src/open_inwoner/scss/views/Home.scss
```diff
@@ -0,0 +1,11 @@
+/// Cards on Home Page and Theme page
+
+.home {
+  .card-container {
+    grid-template-columns: repeat(auto-fit, 228px);
```
vaszig, line 5: Maybe we could add a specific rule for the mobile version as it seems a bit weird to my eyes, but you know better.

image

vaszig (approved):

src/open_inwoner/scss/views/Home.scss
jiromaykin: @vaszig For mobile: horizontal scrolling will be added later in a different issue so it will look better, but the cards will need to remain this same size.

jiromaykin: What do you think @alextreme ?

alextreme: This is fine for now, the scrolling is something Jiro will continue on but these changes are a necessary first step for that.

PR #527, opened 2023-03-13
[#1177] Show new plan contact count in menu

Bartvaderkin:


src/open_inwoner/plans/migrations/0013_auto_20230223_1115.py
```diff
@@ -6,6 +6,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
+        ("accounts", "0055_user_image"),
```
Bartvaderkin, line 9: I had to edit this existing migration to get the migration test to work.

vaszig (commented):
One remark concerning the creator of the plan. Do we want to show the count to the creator as well?I would expect this only for the contact, since the creator is already aware of the plan at this point.

src/open_inwoner/accounts/models.py
```diff
@@ -282,6 +283,12 @@ def is_email_of_contact(self, email):
             or self.contacts_for_approval.filter(email=email).exists()
         )
 
+    def get_plan_contact_new_count(self):
```
vaszig, line 286: Would it be better if we put these in the `query.py`?

Bartvaderkin: query.py is usually for things about collections of the model, while this is instance specific so can be an instance method

vaszig (changes requested):
As mentioned in the previous comment, the number is shown for the creator as well.

Bartvaderkin:
@vaszig Review is mostly about last two commits, although I had to rebase and force push.

vaszig (approved):

alextreme:
Can be merged

PR #528, opened 2023-03-13
:lipstick: [#1180] Make scroll-to-top anchor appear on every page

jiromaykin:
-make scroll-to-top appear everywhere on all mobile pages
-make it disappear when user is scrolled to top completely= make it appear only when scrolling down
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1180

### DIFF
--- a/src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
+++ b/src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
@@ -1,0 +1,15 @@
+// Get the button
+const mybutton = document.getElementById('scroll-anchor')
+
+// When the user scrolls down 80px from the top of the document, show the button
+window.onscroll = function () {
+  scrollFunction()
+}
+
+function scrollFunction() {
+  if (document.body.scrollTop > 80 || document.documentElement.scrollTop > 80) {
+    mybutton.style.display = 'block'
+  } else {
+    mybutton.style.display = 'none'
+  }
+}
--- a/src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
+++ b/src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
@@ -2,14 +2,18 @@
 const mybutton = document.getElementById('scroll-anchor')
 
 // When the user scrolls down 80px from the top of the document, show the button
-window.onscroll = function () {
-  scrollFunction()
-}
-
-function scrollFunction() {
-  if (document.body.scrollTop > 80 || document.documentElement.scrollTop > 80) {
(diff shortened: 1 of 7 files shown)


jiromaykin:
This is such an extremely tiny new 'component', I didn't think it would be necessary to create new JS folders and a new component library (it replaces a list-item that was part of the anchor-menu) - but feel free to suggest otherwise.

src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
```diff
@@ -0,0 +1,15 @@
+// Get the button
+const mybutton = document.getElementById('scroll-anchor')
+
+// When the user scrolls down 80px from the top of the document, show the button
+window.onscroll = function () {
+  scrollFunction()
+}
```
Bartvaderkin, line 7: In general try to use `addEventListener` instead of assigning to (the global) `.on<xyz>` handler, eg: use `window.addEventListener("scroll", .. ` here.

Also I recall there was something about scroll events and performance. I'm not 100% sure it applies here but if I google about it I see stuff about jankiness and using `{ passive: true }` when adding the event listener (and a whole bunch of hacky solutions with intervals).

Bartvaderkin (changes requested):

src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
jiromaykin: The intervals are good for old browsers, but we support the modern ones with Passive event listener now :)

Bartvaderkin: Very cool.

Bartvaderkin (approved):

src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
```diff
@@ -0,0 +1,19 @@
+// Get the button
+const mybutton = document.getElementById('scroll-anchor')
```
vaszig, line 2: Just a general remark, we use camel-case names in js `(myButton).` No need of changes here.

vaszig (approved):

src/open_inwoner/js/components/anchor-menu/scroll_to_top.js
jiromaykin: @vaszig Thanks, it seemed weird to use camelCase here because CSS is case-insensitive, but after some investigation, it seems it's 'convention' to use camelCase with ID's.

PR #532, opened 2023-03-14
[#1059] Added support for SVG logos

Bartvaderkin:
Note I had to update some dependencies to get SVG support in django-filer

### DIFF
--- a/src/open_inwoner/scss/components/Logo/Logo.scss
+++ b/src/open_inwoner/scss/components/Logo/Logo.scss
@@ -1,13 +1,9 @@
 .logo {
   &__image {
     max-width: 100%;
-    height: auto;
 
     @media (min-width: 768px) {
       display: block;
-    }
-    @media (max-width: 455px) {
-      height: auto;
     }
   }
 }
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -58,7 +58,9 @@ cryptography==38.0.4
     #   mozilla-django-oidc
     #   pyopenssl
 cssselect2==0.4.1
-    # via weasyprint
+    # via
+    #   svglib
+    #   weasyprint
 defusedxml==0.7.1
     # via
     #   django-digid-eherkenning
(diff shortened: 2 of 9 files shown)


Bartvaderkin:
@jiromaykin I'm not 100% sure if the CSS changes are OK, as you explicitly added those `height:auto;` 's, but it doesn't work if I keep them. 

Note I also needed to set a height-attribute on the SVG's image element for it to show up, but then that would break the non-blurry-images for regular pixel images so I made it adaptive.

I got some SVG's from here: https://www.cityofenschede.com/en/brand (particularly the wide logo)

jiromaykin:
> [...]'m not 100% sure if the CSS changes are OK, as you explicitly added those `height:auto;` 's, but it doesn't work if I keep them.

OK @Bartvaderkin  - i needed those heights for mobile, in order to not prevent the images stretch/shrink really weirdly  - but i will check to see what happens with SVG's and other images. And... maybe different thing happens if the user uploads SVG's that have an inline style?

src/open_inwoner/scss/components/Logo/Logo.scss
```diff
@@ -1,13 +1,9 @@
 .logo {
   &__image {
     max-width: 100%;
```
jiromaykin, line 3: This seems completely fine actually - I've done some really weird testing with strange SVG's and the only SVG's that do not behave well are the ones with inline-styling and **without** an explicit `viewbox` - so: if users upload correct and valid SVG's  then the styling behaves well (something to note for ourselves in case future users report problems with SVG).

jiromaykin (approved):

src/open_inwoner/scss/components/Logo/Logo.scss
Bartvaderkin: Ugh yes invalid or weird SVG's are a complication. The SVG used should at least look normal in a browser/viewer without margins (viewbox and transforms etc)

PR #534, opened 2023-03-15
[#1208] Refactored cases views to not use stale cached catalog collections

Bartvaderkin:
Also renamed and commented a few fetch-functions


vaszig (approved):
Looks nice, I assume the task number though it's wrong. Is it 1206?

PR #535, opened 2023-03-15
[#1238] Markdown CTABUTTON option in product detail

vaszig:


vaszig:
This is a first approach which implements the button for all ckeditor "Editors" (not only for the content of product). If @alextreme @Bartvaderkin agree I can continue with implementing the js because it requires creating a new custom editor.

I created another filter for ckeditor content for products since the initial got complicated enough and  was used by other pages as well.

src/open_inwoner/pdc/admin/product.py
```diff
@@ -93,6 +93,14 @@ def display_categories(self, obj):
...
     display_categories.short_description = "categories"
 
+    def get_form(self, request, obj=None, **kwargs):
+        form = super().get_form(request, obj, **kwargs)
+
+        if obj:
+            form.base_fields["link"].widget.attrs.update({"data-product-id": obj.id})
+            form.base_fields["form"].widget.attrs.update({"data-product-id": obj.id})
+        return form
+
```
Bartvaderkin, line 103: Is this still necessary now the button doesn't need to carry the ID?

src/open_inwoner/utils/ckeditor.py
```diff
@@ -28,15 +30,77 @@ def get_rendered_content(content):
...
-                icon = soup.new_tag("span")
-                icon.attrs.update(
-                    {
-                        "aria-hidden": "true",
-                        "aria-label": "Opens in new window",
-                        "class": "material-icons",
-                    }
-                )
-                icon.append("open_in_new")
-                element.append(icon)
```
Bartvaderkin, line 40: I notice the original of this function has more code then what is left after splitting it for the product version. Eg the new `get_rendered_content()` doesn't add the icon but adds a _blank target.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -28,15 +30,77 @@ def get_rendered_content(content):
...
+        ("h5", "h5"),
+        ("h6", "h6"),
+        ("img", "image"),
+        ("li", "li"),
+        ("p", "p"),
+        ("a", "link link--secondary"),
+        ("table", "table table--content"),
+        ("th", "table__header"),
+        ("td", "table__item"),
+    ]
```
Bartvaderkin, line 59: If this is the same data for each function then it is better to move it out of the function to a shared variable so we don't forget to update both if we need changes in the future.

src/open_inwoner/js/admin/ckeditor/editor.js
```diff
@@ -19,8 +19,42 @@ import Table from '@ckeditor/ckeditor5-table/src/table'
...
+class CTARequestButtonPlugin extends Plugin {
+  init() {
+    const editor = this.editor
+
+    editor.ui.componentFactory.add('addRequest', () => {
+      // The button will be an instance of ButtonView.
+      const button = new ButtonView()
+
+      button.set({
+        label: 'Add request button',
```
Bartvaderkin, line 36: I would change the label to something like "Add call-to-action button"

src/open_inwoner/utils/ckeditor.py
```diff
@@ -28,15 +30,77 @@ def get_rendered_content(content):
...
+                if "[CTAREQUESTBUTTON]" in element.text:
+                    # decompose the element when product doesn't have either a link or a form
+                    if not (product.link or product.form):
+                        element.decompose()
+                        continue
+
+                    # icon
+                    icon = soup.new_tag("span")
+                    icon.attrs.update(
+                        {"aria-label": "Aanvraag starten", "class": "material-icons"}
```
Bartvaderkin, line 73: I think all the UI text (labels etc) should be marked as translatable.

Bartvaderkin (changes requested):

src/open_inwoner/utils/ckeditor.py
vaszig: Yes you are right, I did this in a generic way as we had it before [https://taiga.maykinmedia.nl/project/open-inwoner/task/1172](url). Because the icon was added only for the product view.

src/open_inwoner/pdc/admin/product.py
vaszig: Correct, I will fix it.

vaszig:
After discussing this with Alex, we decided that there is no reason to add a new ckeditor widget only for this purpose. We just search for the specific text and render the button.

alextreme (approved):

alextreme:
@vaszig feel free to merge if you're happy with this.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -28,15 +30,77 @@ def get_rendered_content(content):
...
-                icon.attrs.update(
-                    {
-                        "aria-hidden": "true",
-                        "aria-label": "Opens in new window",
-                        "class": "material-icons",
-                    }
-                )
-                icon.append("open_in_new")
-                element.append(icon)
+                element.attrs["target"] = "_blank"
```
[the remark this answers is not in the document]
vaszig: This is not for the product page, it has to do with all the other <a> tags. Do you mean that we don't want anywhere?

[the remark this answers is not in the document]
alextreme: Ah okay. I'll wait for Bart to do a double-check.

Bartvaderkin (changes requested):

vaszig:
I made a small update because the button was not properly rendered (had the arrow after the text).

Bartvaderkin (approved):

PR #538, opened 2023-03-16
:lipstick: [#1179] Add breakpoint for inputs, fix tags and other minor fixes

jiromaykin:
Minor fixes, see Taiga issue 1179

- [x] make inputs less wide (involves adding classes and fix width for Search input)
- [x] mobile menu needs black close icon instead of ‘forced system-color’
- [x] hover on button of 'confirm' notification needs light text color
- [x] on mobile and desktop: rounder border for “actie vereist” in begeleider view
- [x] background color of sent messages needs to be the theme/city color but with light opacity
- [x] add flex-wrap to tags for mobile:

tags-wrap

### DIFF
--- a/src/open_inwoner/accounts/templates/django_registration/registration_form.html
+++ b/src/open_inwoner/accounts/templates/django_registration/registration_form.html
@@ -4,7 +4,7 @@
 {% block content %}
 
 {% render_grid %}
-    {% render_column start=1 span=6 %}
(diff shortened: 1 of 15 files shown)


vaszig (changes requested):
Looks really nice!I have found only one problem about the labels in the forms (especially for the messages where it is obvious to the user):


profile
messages

vaszig (approved):

PR #539, opened 2023-03-16
[#1249] Add headers to outgoing requests logging

vaszig:


src/log_outgoing_requests/log_requests.py
```diff
@@ -30,7 +31,7 @@ def install_outgoing_requests_logging():
     Session._original_request = Session.request
 
     def new_request(self, *args, **kwargs):
-        kwargs.setdefault("hooks", {"response": hook_requests_logging})
+        self.hooks["response"].append(hook_requests_logging)
```
vaszig, line 34: This was changed in case we add more hooks to the session

src/log_outgoing_requests/handlers.py
```diff
@@ -29,6 +31,8 @@ def emit(self, record):
                     "res_content_type": record.res.headers.get("Content-Type", ""),
                     "timestamp": record.requested_at,
                     "response_ms": int(record.res.elapsed.total_seconds() * 1000),
+                    "req_headers": record.req.headers,
+                    "res_headers": record.res.headers,
```
Bartvaderkin, line 35: What kind of data format is this?

Bartvaderkin (changes requested):
I'm missing a test (modification) for the header saving

And looking at the code I miss test for some other properties? Like the two content-type's and params.

src/log_outgoing_requests/handlers.py
```diff
@@ -14,11 +14,17 @@ def emit(self, record):
 
             # save only the requests coming from the library requests
             if record and record.getMessage() == "Outgoing request":
+                exclude_headers = ["Authorization"]
+                safe_req_headers = {
+                    k: v
+                    for k, v in record.req.headers.items()
+                    if k not in exclude_headers
+                }
+
```
Bartvaderkin, line 23: Maybe it'd be better to keep the exclude_headers's name but replace the value with something like `***hidden***`? Now we can't see in the logs if the request had Authorization at all.

src/log_outgoing_requests/tests/test_logging.py
```diff
@@ -63,12 +61,33 @@ def test_expected_data_is_saved_when_saving_enabled(self, m):
                 self.assertEqual(request_log.req_content_type, "")
                 self.assertEqual(request_log.res_content_type, "")
                 self.assertEqual(request_log.response_ms, 0)
+                self.assertEqual(
+                    request_log.req_headers,
+                    (
+                        "{'User-Agent': 'python-requests/2.26.0', 'Accept-Encoding': 'gzip, deflate, br', 'Accept': '*/*', 'Connection': 'keep-alive'}"
+                        if method == "HEAD"
+                        else "{'User-Agent': 'python-requests/2.26.0', 'Accept-Encoding': 'gzip, deflate, br', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Length': '0'}"
```
Bartvaderkin, line 69: This is a bit much for an if/else expression, and blacks pretty ugly.

Also I'm not convinced this is a good format to store the headers in: if you want to keep a structure then it'd be JSON (and a JSONfield), not this Python repr() format (and without linebreaks it would not be very nice for debugging).

Would it be more useful to store them the way they are in HTTP? Just name/value pairs each on a line? Do we need all these braces and quotes?

src/log_outgoing_requests/tests/test_logging.py
```diff
@@ -63,12 +61,33 @@ def test_expected_data_is_saved_when_saving_enabled(self, m):
...
+                    ),
+                )
+                self.assertEqual(request_log.res_headers, "{}")
```
Bartvaderkin, line 72: The test should include some data and check for every field we store (I miss `res_headers`, also `req_content_type`, `res_content_type` and `params`)

Bartvaderkin (changes requested):

src/log_outgoing_requests/tests/test_logging.py
```diff
@@ -58,26 +71,40 @@ def test_expected_data_is_saved_when_saving_enabled(self, m):
...
+                        "Authorization: ***hidden***\n"
+                        "Content-Type: text/html"
+                        if method in ["HEAD", "GET"]
+                        else "User-Agent: python-requests/2.26.0\n"
+                        "Accept-Encoding: gzip, deflate, br\n"
+                        "Accept: */*\n"
+                        "Connection: keep-alive\n"
+                        "Authorization: ***hidden***\n"
+                        "Content-Type: text/html\n"
+                        "Content-Length: 0"
```
Bartvaderkin, line 93: There is way too much in this if/else expression, and it is very hard to see which parts are the same and what is the difference between each case.

As a general rule if/else expressions are not so great;  for anything non trivial it is almost always more readable to use normal if/else statements.

src/log_outgoing_requests/handlers.py
```diff
@@ -35,9 +34,13 @@ def emit(self, record):
...
-                    "res_headers": record.res.headers,
+                    "req_headers": self.format_headers(safe_req_headers),
+                    "res_headers": self.format_headers(record.res.headers),
                     "trace": trace,
                 }
 
                 OutgoingRequestsLog.objects.create(**kwargs)
+
+    def format_headers(self, headers):
+        result = textwrap.dedent("\n".join(f"{k}: {v}" for k, v in headers.items()))
```
Bartvaderkin, line 45: Is `.dedent()` needed here? We just built that string without indents?

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #540, opened 2023-03-16
:art: [#1240] Add 'Aanvraag starten' button in mobile anchor menu

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1240 
for Product pages only

### DIFF
--- a/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
+++ b/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
@@ -1,4 +1,4 @@
-{% load i18n icon_tags button_tags %}
+{% load i18n icon_tags link_tags button_tags %}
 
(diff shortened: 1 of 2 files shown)


src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n icon_tags button_tags %}
+{% load i18n icon_tags link_tags button_tags %}
```
vaszig, line 1: "link_tags" is not used in this template, it can be removed.

vaszig (changes requested):
Do we also need the button at the bottom of the text?We render this via the product detail html. Maybe we can remove this since they will have the ability to also add a cta-button inside the product's content.

image

jiromaykin:
> Do we also need the button at the bottom of the text? [.....] Maybe we can remove this since they will have the ability to also add a cta-button inside the product's content.

But that is a different issue - this PR is _only_ about the mobile menu.

vaszig (approved):

PR #544, opened 2023-03-17
[#1239] Remove grey lines from product detail page

vaszig:


src/open_inwoner/templates/pages/product/detail.html
```diff
@@ -24,6 +24,8 @@
 {% endblock sidebar_content %}
 
 {% block content %}
+    {% with question_set=object.question_set.exists files=object.files.exists conditions=object.conditions.exists locations=object.locations.exists product_links_exist=product_links.exists published_related_products=object.related_products.published.exists contacts=object.contacts.exists %}
```
Bartvaderkin, line 27: I'm not convinced this is a good idea. For one the names are not great (storing booleans in things that are named like collections).

Also the amount of queries is a bit much maybe? 

If we're aliasing anyway then just alias the query so Django can cache the result and do both the the if/else and iteration on the same result (instead of doing .exists() everywhere or doing multiple queries for the same things).

src/open_inwoner/templates/pages/product/detail.html
```diff
@@ -44,130 +50,125 @@ <h1 class="h1" id="title">
...
+                                {% render_column span=6 %}
+                                    <span class="p">{{ contact.role|default:_('Telefoonnummer') }}</span>
+                                {% endrender_column %}
+
+                                {% render_column start=7 span=6 %}
+                                    {% link href='tel:'|add:contact.phonenumber text=contact.phonenumber primary=True %}
+                                {% endrender_column %}
+                    {% endif %}
+
+                    {% if contact.email %}
```
Bartvaderkin, line 157: Indentation seems weird here.

Bartvaderkin (changes requested):

src/open_inwoner/templates/pages/product/detail.html
```diff
@@ -34,6 +36,10 @@ <h1 class="h1" id="title">
     <p class="p">{{ object.summary }}</p>
     {{ object.content|ckeditor_content|safe }}
 
+    {% if question_set or files or conditions or locations or product_links or published_related_products or contacts %}
+        <hr class="divider">
```
jiromaykin, line 40: If would probably have done this differently, by not adding a real HR divider, but making it a CSS only **border** that would appear if the class "has" something inside it or not - I was hoping Django would NOT show Divs if they do not contain anything. If the templates just simply do not render Divs that do not contain anything, you can do something in CSS like this:

`.product-info:has(.faq),
.product-info:has(.map),
.product-info:has(.link-list),
.product-info:has(.notification) {
  padding-top: var(--spacing-giant);
  border-top: 2px solid var(--color-gray-light);
}`

But this does not work in Firefox...!

jiromaykin (changes requested):

jiromaykin:
> [...] right now the design doesn't look good anymore (because of different spaces between sections), so I think we need to pass this by the designer.

A little addition to this: The inconsistent spacing depends on which elements the user will add to a product page; so on a page like this: http://127.0.0.1:8000/themas/zorg-en-ondersteuning/products/logeeropvang/ I feel like the visual design becomes ugly because:
- there is less space between the map-location card and the links, 
- but then there is always a lot of space around the green notifcation on product pages. 
- And then there is not that much space between the FAQ and the condition-list; yet a bigger space between the condition-list and the map.

This does not comply to design laws that say 'keep spacing consistent' - but i think we just need to leave this for now and ask the designer or ask for user feedback.

alextreme:
Postponed for this release, can be revisited after the current changes.

PR #546, opened 2023-03-20
:lipstick: [#1233] Add required fields indicators

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1233
the 'required' attribute exists but is not shown in any styling or indicator, so the following is needed:

- [x] - show asterisks after/before each required field (CSS) 
do _for the Upload component AND (if it's easy) ALSO add everywhere else_
- [x] - show one caption in the top of forms that contain required fields “Velden met een ster zijn verplicht.”/"Fields marked with * are mandatory"; don't show this line if there are no required fields present (Django);
do this _for the Upload component AND (if it's easy) ALSO add everywhere else_
- [x] -  make this line not break up the grid (will look ugly)
- [x] Add translatable text, with a red asterisk "Velden gemarkeerd met een * zijn verplicht"

Needs to look somewhat like this,
not sure yet if the asterisk needsto be before or after the Labels:
 
Screenshot from 2023-03-28 12-42-12


jiromaykin:
@Bartvaderkin Have you got a suggestion on how to do this 'the smart way'? 
Vasilios and I looked at this together and think we may need to "... have to update every form template tag that we have in the project (with **has_required**)."
I guess I want to make a generic solution for something that still needs to be able to deal with exceptions.
Currently we only need this for the Upload component, but if there is a simple way to add this caption to all other forms, we may as well do this.

Bartvaderkin:
Should this '* verplicht veld' line always be visible? If so, how does it interact with the field and non-field errors (the ones that show up if backend validation fails?

src/open_inwoner/components/templates/components/Form/Error.html
```diff
@@ -1,2 +1,10 @@
-{% load notification_tags %}
-{% notifications messages=messages compact=True %}
+{% load i18n button_tags icon_tags button_tags icon_tags notification_tags %}
+
+<div class="notifications notifications__error">
+   <div class="notification notification__error notification--compact">
+    {% for message in messages %}
+        {% icon icon="warning_amber" icon_position="before" outlined=True %}
+        {% notification icon=message.icon icon_position="before" type=message.level_tag message=message.message closable=False compact=compact %}
```
Bartvaderkin, line 7: Good to swap the `notifications`(meervoud)-tag for plain HTML, but I note it then still uses the same multi-purpose `notification`-tag and styles for the for validation error messages, so I assume only the wrapping div was an issue and not the inner notification component.

vaszig, line 7: Why do we add `closable=False` here? Don't we want the user to be able to close the notification?

src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -49,7 +49,7 @@ <h1 class="h1">{% trans 'Welkom' %}</h1>
 
                 {% if login_allow_registration %}
                     {% render_card tinted=True %}
-                        {% render_form id="login-form" method="POST" form=form %}
+                        {% render_form id="login-form" method="POST" form=form show_required=True %}
```
vaszig, line 52: I think we need something `align-items: center;` here?


error-login

vaszig (changes requested):
Also when uploading a file via the user's profile page we don't have the "verplicht veld" text.

profile-file

jiromaykin:
> Also when uploading a file via the user's profile page we don't have the "verplicht veld" text.

This 'cloud' functionality may be removed in future I think? I'm not sure, need to check.

jiromaykin:
> Also when uploading a file via the user's profile page we don't have the "verplicht veld" text.

This part of the site will be removed, as in PR #572

src/open_inwoner/components/templates/components/Form/Error.html
jiromaykin: @vaszig I need a bit of help with differentiation (maybe add a new 'warning' template instead of an error template?); 
because most of the time we actually don't need the close-button at all. Like: if a user forgets to fill a required field; then the user just needs to do that and re-submit; clicking away errors would have no function in that case. 
I am not sure which forms/errors actually need a close function and how much validation we have. 
Perhaps we only need a close-button in the Inbox Messages field?

jiromaykin: After consultation with @Bartvaderkin the decision is to make the Error template more static, and to not have a close-button.

vaszig (approved):

alextreme:
resolved conflict, can be merged after all is green

PR #550, opened 2023-03-21
[#1248] Redirect user to next after login

vaszig:
Next is not used somewhere in the project (for instance if we click from an email the link to approve a contact we are not redirected to the contacts) so I added it for all the possible login methods.

### DIFF
--- a/src/open_inwoner/components/templates/components/Typography/Link.html
+++ b/src/open_inwoner/components/templates/components/Typography/Link.html
@@ -1,7 +1,7 @@
 {% load i18n helpers icon_tags %}
 
 <a
-    href="{{href}}"
+    href="{{href}}{% if next %}?next={{next}}{% endif %}"
     {% if id %}{{ id }}{% endif %}
     {% if download %}download{% endif %}
     {% if blank is True %}target="_blank"{% endif %}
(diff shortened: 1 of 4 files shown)


src/open_inwoner/components/templates/components/Typography/Link.html
```diff
@@ -1,7 +1,7 @@
 {% load i18n helpers icon_tags %}
 
 <a
-    href="{{href}}"
+    href="{{href}}{% if next %}?next={{next}}{% endif %}"
```
Bartvaderkin, line 4: Simply appending `?next=` breaks if the `href` already has a query string. It'd be better if a filter or template tag could do this.

Bartvaderkin (changes requested):

src/open_inwoner/components/templatetags/link_tags.py
```diff
@@ -112,3 +112,15 @@ def get_classes():
...
+@register.filter
+def addnexturl(href, next_url):
+    """
+    Concatenates href & next_url.
+    """
+    if "?" in href:
+        href += f"&next={next_url}"
+        return href
+
+    return href + f"?next={next_url}"
```
Bartvaderkin, line 126: Unnecessary concatenation if we're already doing a template in the other half of the statement:

````python
return f"{href}?next={next_url}"
````

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #551, opened 2023-03-22
[#1270] Show proper image in contact approval

vaszig:


src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n button_tags link_tags icon_tags pagination_tags form_tags dropdown_tags messages_tags thumbnail %}
+{% load i18n button_tags link_tags icon_tags pagination_tags form_tags dropdown_tags messages_tags thumbnail cropping %}
```
vaszig, line 2: FYI this template tag is part of the external library.

Bartvaderkin (approved):

jiromaykin (approved):
The padding in the avatar could just be "var(--spacing-large)" instead of 15/10px I think?
Anyway: Seems to function well - even though I think the white background behind the avatar is not pleasant when it is white and has such a thin border, but let's see if in future there will be feedback from users.

PR #553, opened 2023-03-23
[#1268] Disable case document button when form is being submitted

vaszig:


Bartvaderkin (commented):
Seems reasonable, maybe it should change style to show disabled state?

jiromaykin:
How can I test this in the front-end? Do I need to add a new Aanvraag through Open Forms first?

jiromaykin:
How can I test this in the front-end? Do I need to add a new Aanvraag through Open Forms first?

src/open_inwoner/js/components/cases/index.js
```diff
@@ -0,0 +1,16 @@
+class DisableSubmitButton {
+  constructor(form) {
+    this.form = form
+    this.form.addEventListener('submit', this.disableButton.bind(this))
+  }
+
+  disableButton() {
+    const submitButton = this.form.querySelector('button[type="submit"]')
+    submitButton.setAttribute('disabled', 'true')
```
jiromaykin, line 9: in the Cases styles you could add something similar to this:

`.button[type="submit"]:disabled {
    background-color: var(--color-primary-darker);
    color: var(--color-gray)
  }`

jiromaykin (changes requested):
I suggested something that needs to be checked if it works, maybe you'll need a more specific selector for the disabled button inside a form.

src/open_inwoner/js/components/cases/index.js
vaszig: Thanks for the hint. I added something similar using the same pattern we use for other disabled buttons in the project.

jiromaykin (approved):

PR #556, opened 2023-03-24
[#1258] Fix order of phonenumber and email in product locations

vaszig:


src/open_inwoner/js/components/map/index.js
```diff
@@ -74,31 +74,50 @@ class Map {
...
+      email,
+      ...properties
+    } = feature.properties
+
     const displayName = name ? escapeHTML(name) : ''
     const locationDetailView = location_url ? escapeHTML(location_url) : ''
+    const displayAddress1 = address_line_1 ? escapeHTML(address_line_1) : ''
+    const displayAddress2 = address_line_2 ? escapeHTML(address_line_2) : ''
+    const displayPhonenumber = phonenumber ? escapeHTML(phonenumber) : ''
+    const displayEmail = email ? escapeHTML(email) : ''
```
Bartvaderkin, line 92: I see a repeating pattern that isn't very DRY.

Lets create a function that wraps the imported `escapeHTML()` and does the check for empty/null/undefined and either return empty string or call the imported function instead of putting ternary if/else expressions everywhere.

Also I'm not convinced you need a whole set of copies of the same variables, like it'd mostly just increases risk of accidentally using the wrong ones.

Bartvaderkin (changes requested):

src/open_inwoner/js/components/map/index.js
vaszig: You are right I' ll put this in a function. Concerning the variables, I thought it would be better since I use them multiple times below.

Bartvaderkin (approved):

src/open_inwoner/js/components/map/index.js
```diff
@@ -74,31 +87,50 @@ class Map {
...
+      address_line_2,
+      phonenumber,
+      email,
+      ...properties
+    } = feature.properties
+
+    const displayName = escapeVariableText(name)
+    const locationDetailView = escapeVariableText(location_url)
+    const displayAddress1 = escapeVariableText(address_line_1)
+    const displayAddress2 = escapeVariableText(address_line_1)
```
Bartvaderkin, line 103: address_line_1 is twice here

Bartvaderkin (changes requested):

src/open_inwoner/js/components/map/index.js
vaszig: Thanks for catching this

Bartvaderkin (approved):

PR #559, opened 2023-03-27
[#1274] Downgraded ZGW role-retrieval's error-level in notification handler for less Sentry spam

Bartvaderkin:


alextreme:
Can be merged

PR #562, opened 2023-03-30
Simplified error handling codestyle in ZGW fetch-functions

Bartvaderkin:


vaszig (commented):
Maybe we should fix this in `open_inwoner/haalcentraal/utils.py` as well, since you did nicer now!

Bartvaderkin:
Good idea, I appended a commit for this.

vaszig (approved):

PR #563, opened 2023-03-31
[#1296] Make openzaak,catalogi,documenten,formulieren api fields in OpenZaakConfig optional

vaszig:


Bartvaderkin (approved):
Interesting these were already nullable but not blank but heh.

PR #565, opened 2023-04-03
[#847] First pass at implementing Klanten & Contacten API

Bartvaderkin:


Bartvaderkin:
I rebased on django-cms and applied necessary changes and implemented the new fields.

alextreme (approved):

PR #569, opened 2023-04-04
[#1306-7] Update case and case type texts in case list and status views

vaszig:
I have also removed the case type description from the status page as it is not in the design.


src/open_inwoner/accounts/views/cases.py
```diff
@@ -347,8 +350,14 @@ def get_context_data(self, **kwargs):
...
                 ),
-                "description": self.case.omschrijving,
-                "type_description": self.case.zaaktype.omschrijving,
+                "description": (
+                    self.case.omschrijving
+                    if self.case.omschrijving
+                    else self.case.zaaktype.omschrijving
+                ),
+                "type_description": (
+                    self.case.zaaktype.omschrijving if self.case.omschrijving else ""
```
Bartvaderkin, line 359: I note the condition looks at a different attribute then the left side but suppose it is intentional and related to the regular `description` just above.

Bartvaderkin (approved):

alextreme (approved):

alextreme:
@Bartvaderkin ready for review now

Bartvaderkin (approved):

PR #570, opened 2023-04-04
[#1308] Hide upload type from case document when only one type exists

vaszig:


Bartvaderkin (approved):
I approve it because it'll work and gets validated, but technically it is weird to send values through a form if we already know the value at the backend (because it is the one possible choice).

alextreme:
I agree with Bart, might need a second look later on (using 2 separate Django forms would have been neater). But okay for now.

PR #573, opened 2023-04-05
[#1310] Add  description field to ZaakTypeConfig

vaszig:


src/open_inwoner/accounts/views/cases.py
```diff
@@ -386,7 +386,18 @@ def get_upload_info_context(self, case: Zaak):
         else:
             external_upload_url = ""
 
+        try:
+            case_type_config = ZaakTypeConfig.objects.get(
+                identificatie=case.zaaktype.identificatie,
+                catalogus__url=case.zaaktype.catalogus,
+            )
```
Bartvaderkin, line 393: This should have used the ZaakTypeConfig queryset methods. (`filter_catalog()`, possibly also `filter_case_type()`)

But this function seems to do a lot of repeated queries that could be simplified (I see `get_visible_zt_configs_for_case_type_identification()` twice, and this new addition overlaps with it as well.

Also I see the old comments in the queryset how these should be refactored, and I'm not sure if you can safely do a `.get()` (and there isn't a case that returns multiple objects) maybe I should pick this up and do the refactoring and amend the PR.

vaszig: Yes, it would be nice. Thank you Bart

Bartvaderkin:
@vaszig I amended the PR with the refactor. 

One notable result was that some code didn't query for the catalog, but that is necessary because in the spec (eg: Open Zaak) the `ZaakType.identificaties` are namespaced per Catalog (in eSuite catalog-resource doesn't exist and the namespace is global). 

But the queryset methods check for that so I had to modify the tests to set a catalog.

Bartvaderkin:
@vaszig as discussed I rebased to fix the CI issue and it is green now. 

But I can't request review from you because you were the creator of the PR :confused: 

(I created an issue on GH itself because this doesn't feel very 'collaborative coding': https://github.com/orgs/community/discussions/52253)

src/open_inwoner/accounts/views/cases.py
```diff
@@ -371,28 +371,30 @@ def get_context_data(self, **kwargs):
...
+                case.zaaktype
+            ).get()
+        except ObjectDoesNotExist:
+            external_upload_enabled = False
             external_upload_url = ""
+            case_type_config_description = ""
+        else:
+            external_upload_enabled = True
+            external_upload_url = ztc.external_document_upload_url
+            case_type_config_description = ztc.description
```
vaszig, line 394: I see the description is set for external upload but where is the internal one?

Bartvaderkin:
@alextreme We collaboratively fixed the issue in @vaszig's PR but somehow I cannot approve the PR anymore even though I was marked as reviewer, so maybe you can.

src/open_inwoner/templates/pages/cases/status.html
```diff
@@ -32,14 +32,23 @@ <h2 class="h2" id="documents">{% trans 'Documenten' %}</h2>
...
 
                 {% if case.internal_upload_enabled %}
-                    <h2 class="h2" id=>{% trans "Document toevoegen" %}</h2>
+                    <h2 class="h2">{% trans "Document toevoegen" %}</h2>
+                    {% if case.case_type_config_description %}
+                        <p class="p">{{case.case_type_config_description}}</p>
+                    {% endif %}
+
                     <p class="p">
                         {% blocktranslate with max_filesize=openzaak_config.max_upload_size allowed_extensions=case.allowed_file_extensions|join:', ' %}
```
alextreme, line 41: As discussed with Groningen, if specified the case_type_config_description will overrule this default description.

So the idea is that we either show the default (with max filesize and the formats allowed) or show the case type description, but not both

alextreme:
@Bartvaderkin sure, one request did come in from Groningen about this that I've added as a remark to the PR

alextreme (approved):

PR #574, opened 2023-04-05
[#1326] Fix phonenumber, email and url in product location cards

vaszig:
This fixes [https://taiga.maykinmedia.nl/project/open-inwoner/issue/1252](url) as well.

### DIFF
--- a/src/open_inwoner/scss/components/Typography/Link.scss
+++ b/src/open_inwoner/scss/components/Typography/Link.scss
@@ -75,6 +75,8 @@
   }
 }
 
-.link + .link {
+.link
+  + .link
+  :not(.card__body--flex > a > span, .leaflet-content-details > a > span) {
   margin-left: var(--spacing-small);
 }
--- a/src/open_inwoner/scss/components/Typography/Link.scss
+++ b/src/open_inwoner/scss/components/Typography/Link.scss
@@ -74,9 +74,3 @@
     transform: scale(0.75);
   }
 }
-
-.link
-  + .link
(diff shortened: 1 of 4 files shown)


jiromaykin (changes requested):
Good to know it needs to be a string - I see one strange thing: there is a little bit of space before the last link, so (probably) the link-list inside leaflets need to have a margin-left set to zero:

Screenshot 2023-04-06 at 09 49 42

Bartvaderkin (approved):

vaszig:
@jiromaykin I hope I did this in the right way! I added this rule for the product location cards as well.


image

src/open_inwoner/scss/components/Typography/Link.scss
```diff
@@ -75,6 +75,8 @@
   }
 }
 
-.link + .link {
+.link
+  + .link
+  :not(.card__body--flex > a > span, .leaflet-content-details > a > span) {
```
jiromaykin, line 80: it seems like the linter broke off the ":not" selector so it doesn't work anymore; also this ":not" selector is a bit dangerous to use if it's just 1 or 2 exceptions; so better to correct the original style, which was faulty and not specific enough, because it dealt with the wrong exception (not your blame :-D ).
The style for the numbers in the main Navigation here should onl be:
`.primary-navigation {
  &__list-item .link.link--secondary {
    margin-left: var(--spacing-small);
  }
}`
declared outside of the ".link" curly braces.
This way all the other links everywhere will not be affected.

jiromaykin (changes requested):

jiromaykin (approved):

PR #580, opened 2023-04-13
:lipstick: [#1309] Make upload widget correspond with design

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1309 

multiple tasks:

- [x] - needs custom template for input with type=file
- [x] - form needs to be overridden with a render_form
- [x] - toggle open/close interface to show/hide file-upload input field
- [x] - show filename info
- [x] - show filesize info
- [x] Disable submit button when no file is chosen and on cancel

other task:

- [x] - fix failing test "`@requests_mock.Mocker() /class TestCaseDetailView(ClearCachesMixin, WebTest):`'
- [x] - errors need to be shown and scrolled into view
- [x] - client-side errorhandling (browser defaults) needs to be overriden somehow, so the error notifications actually gets shown (with "novalidate"?)
- [x] - add reset 'close' button to clear the field and the information

### DIFF
--- a/src/open_inwoner/components/templates/components/Form/FileInput.html
+++ b/src/open_inwoner/components/templates/components/Form/FileInput.html
@@ -1,0 +1,21 @@
+{% load i18n form_tags button_tags icon_tags %}
+
+<div class="form__control {{ extra_classes|default:"upload" }}">
+    <div id="inputfile-group" class="inputfile-group">
+        <label class="label">
+            <span class="input-file">
+            {{ field|addclass:"inputfile" }}
+
+            <label class="label__label">
+            {% icon icon="cloud_upload" icon_position="before" outlined=True %}
+                {{ text }}
+                {% if field.field.required %}<span class="label__label--required"> * </span>{% endif %}
+            </label>
+            </span>
+
+            {% if field.errors %}
+                {% errors errors=field.errors %}
+            {% endif %}
+        </label>
+    </div>
+</div>
--- a/src/open_inwoner/components/templates/components/Form/FileInput.html
+++ b/src/open_inwoner/components/templates/components/Form/FileInput.html
@@ -1,7 +1,7 @@
-{% load i18n form_tags button_tags icon_tags %}
+{% load i18n form_tags icon_tags %}
 
 <div class="form__control {{ extra_classes|default:"upload" }}">
-    <div id="inputfile-group" class="inputfile-group">
+    <div class="inputfile-group">
         <label class="label">
             <span class="input-file">
             {{ field|addclass:"inputfile" }}
@@ -9,8 +9,8 @@
             <label class="label__label">
             {% icon icon="cloud_upload" icon_position="before" outlined=True %}
                 {{ text }}
-                {% if field.field.required %}<span class="label__label--required"> * </span>{% endif %}
-            </label>
+
+            </label>{% if field.field.required %}<span class="label__label--required"> * </span>{% endif %}
             </span>
 
             {% if field.errors %}
--- a/src/open_inwoner/js/components/upload-document/file-input-errors.js
+++ b/src/open_inwoner/js/components/upload-document/file-input-errors.js
@@ -1,0 +1,17 @@
+const documentUpload = document.getElementById('formOpenUpload')
+const uploadError = document.querySelectorAll('.notifications')
+// get info
+const getFormInfo = document.querySelectorAll('.form__control--info')
+
+// if errors are present, scroll and trigger opened state
+if (uploadError.length > 1) {
+  console.log(uploadError)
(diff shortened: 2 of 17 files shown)


src/open_inwoner/templates/pages/cases/status.html
```diff
@@ -11,55 +11,93 @@
...
+                            </div>
+                        </section>
+
+                        <div class="form__control--info">{% input form.title no_label=False no_help=True class="label input" id="id_title" %}</div>
+
+                        <div class="form__control--info">{% input form.type no_label=False no_help=True icon="expand_more" icon_position="after" class="label input" id="id_title" extra_classes="file-type__select" %}</div>
+
+                        <div class="form__control upload">
+                            <div class="upload__container">
+                                {% file_input form.file text="Bestanden uploaden" no_label=False no_help=True extra_classes="file-upload" %}
```
vaszig, line 72: Let's make the text translatable as well. The required field indicator, in the button, looks a bit weird to me, but you know better.

required

src/open_inwoner/js/components/upload-document/file-input-errors.js
```diff
@@ -0,0 +1,17 @@
+const documentUpload = document.getElementById('formOpenUpload')
+const uploadError = document.querySelectorAll('.notifications')
+// get info
+const getFormInfo = document.querySelectorAll('.form__control--info')
+
+// if errors are present, scroll and trigger opened state
+if (uploadError.length > 1) {
+  console.log(uploadError)
```
vaszig, line 8: I assume this can be removed.

src/open_inwoner/js/components/upload-document/file-input-errors.js
```diff
@@ -0,0 +1,17 @@
...
+  documentUpload.scrollIntoView({
+    block: 'center',
+    behavior: 'smooth',
+  })
+  documentUpload.classList.add('upload--open')
+  for (let i = 0, max = getFormInfo.length; i < max; i++) {
```
vaszig, line 14: Can't we use only variable "i" here?Why do we need "max" as well?

src/open_inwoner/js/components/upload-document/show-file-info.js
```diff
@@ -0,0 +1,91 @@
...
+
+if (fileUploadInput) {
+  // Convert the file size to a readable format
+  const formatFileSize = function (bytes) {
+    const suffixes = ['B', 'kB', 'MB', 'GB', 'TB']
+    const i = Math.floor(Math.log(bytes) / Math.log(1024))
+    return `${(bytes / Math.pow(1024, i)).toFixed(2)} ${suffixes[i]}`
+  }
+  // initial values
+  submitUpload.style.cssText = `
```
vaszig, line 17: Why do we use inline css?

vaszig (changes requested):

src/open_inwoner/js/components/upload-document/file-input-errors.js
Bartvaderkin: My guess is to save the limit condition (because `.length` is a calculated property)

But `for-i` loops are terrible, it is much better preferred to use `.forEach()` on the queryselector result, something like:

``
getFormInfo.forEach(elem => elem.style.display = 'none')
``

(a `for-of` loop would be nice but I don't know if that works with every weird JS itearble thing)

src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -0,0 +1,21 @@
+{% load i18n form_tags button_tags icon_tags %}
+
+<div class="form__control {{ extra_classes|default:"upload" }}">
+    <div id="inputfile-group" class="inputfile-group">
```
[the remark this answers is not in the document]
jiromaykin: @vaszig I've done a bit of research on naming conventions, since we don't have one standard yet in this project.
I found it's not _technically_ necessary, but 'good practice' to have different names for ID's and classes in HTML.
Plus: I will set ID-names with an underscore from now on, and classes with a dash (because for classes the dash can be used as a modifier). I was using camelCase for some ID names, because of javascript, but that's not useful for CSS.

jiromaykin:
@vaszig and @Bartvaderkin : I have rebased this one to develop so it works with the CMS.

alextreme:
Discussed, can be reviewed again by @vaszig . We might need to revisit setting the style display=none in show-file-info.js, but this can be double-checked on the testenvironment after merging

vaszig (approved):
It seems really nice!I don't see any problems concerning cps headers in my local environment, so we have to make sure we don't have any on the other environments as well.

PR #582, opened 2023-04-13
[#1364] Add django log outgoing requests library

vaszig:


Bartvaderkin (changes requested):
Looks good, great ratio of added-vs-removed code :smile: 

One cosmetic note, it'd be neater if we start to do cleaner PR/merges: like the "Trying to debug" and "Fixed tests" commits don't need to be in the git log forever. 

Personally once I'm done with a thing and I'm the only one working on it I soft-reset the branch, make a clean commit and force-push it (if I wasn't doing Amend-Commit in the first place)

PR #592, opened 2023-04-20
[#1394] Add banner plugin

vaszig:


src/open_inwoner/cms/banner/cms_plugins.py
```diff
@@ -0,0 +1,17 @@
...
+
+@plugin_pool.register_plugin
+class BannerPlugin(CMSPluginBase):
+    model = Banner
+    name = _("Banner Plugin")
+    render_template = "cms/banner/banner_plugin.html"
+
+    def render(self, context, instance, placeholder):
+        context = super().render(context, instance, placeholder)
+        return context
```
Bartvaderkin, line 17: If this method just calls super() with same parameters it can be removed.

src/open_inwoner/cms/banner/models.py
```diff
@@ -0,0 +1,19 @@
...
+from filer.fields.image import FilerImageField
+
+
+class Banner(CMSPlugin):
+    title = models.CharField(_("title"), max_length=250, default="")
+    image = FilerImageField(
+        verbose_name=_("Banner image"),
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
```
Bartvaderkin, line 14: Is `SET_NULL` what we want here? I'd expect `PROTECT` so you can't accidentally remove the homepage image.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #593, opened 2023-04-24
Added common page extension and integrated with (old) menu

Bartvaderkin:
This supports filtering menu on requires auth/BSN, setting a number-indicator lookup and setting a menu icon.


src/open_inwoner/cms/extensions/cms_menus.py
```diff
@@ -0,0 +1,75 @@
...
+from menus.menu_pool import menu_pool
+
+from open_inwoner.accounts.choices import LoginTypeChoices
+from open_inwoner.cms.extensions.constants import IndicatorChoices
+from open_inwoner.cms.extensions.models import CommonExtension
+
+
+def lookup_plan_contacts(request, namespace) -> int:
+    if request.user.is_authenticated:
+        return request.user.get_plan_contact_new_count() or 123
```
Bartvaderkin, line 17: `or 123` ? :wink:

src/open_inwoner/cms/extensions/cms_menus.py
```diff
@@ -0,0 +1,75 @@
+import random
```
vaszig, line 1: These two lines can be removed as well, I don't see them used somewhere

vaszig (approved):
Besides the small remark about imports and the numbers you already mentioned, it looks nice!

vaszig (approved):
It looks nice!My question is don't we need an icon for the profile as well?

Bartvaderkin:
Probably, but these are temporary icons, @jiromaykin might replace them.

PR #597, opened 2023-04-26
[#1397] Add apphook configuration for profile page

vaszig:
I added the link to apphook configurations in the admin (generally) to the toolbar as it is shown below:


src/open_inwoner/cms/profile/cms_appconfig.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    notifications = models.BooleanField(
+        verbose_name=_("Communicatievoorkeuren"),
+        default=True,
+        help_text=_("Designates whether 'notifications' section is rendered or not."),
+    )
+
+
+class ProfileConfigForm(AppDataForm):
+    title = forms.CharField()
```
Bartvaderkin, line 53: We don't use this field so let's remove it for now.

Bartvaderkin (changes requested):
As discussed in the call it would be convenient if the link to the config model's admin edit/create page could be added to the CMS toolbar if you're at the page of the apphook.

alextreme:
Discussed, will need rebasing due to changes on develop. Bart will double-check to determine if the configuration is in a logical place via the cms toolbar (Pagina's menu)

vaszig:
@Bartvaderkin I updated the branch and added a simple item to the toolbar. It's only enabled when we are on the profile page. Let me know if you agree with this approach.

image

alextreme:
Reviewed with @vaszig and approved, but double-check from @Bartvaderkin is requested due to the remarks made yesterday

alextreme (approved):

Bartvaderkin (approved):

PR #598, opened 2023-04-28
[#1404] Implemented possible/partial solutions to support CSP in CMS pages

Bartvaderkin:
Instead of blanketing everything in (conditional) unsafe-inlines for staff users I found that we can detect (most) of the django-cms script tags based on a data attribute.

So I made a middleware that adds the CSP nonce to all the script-tags that do that, and this works for most functionality except some vendored jquery library that doesn't seem essential. Except this approach is flawed, as an attacker could just inject evil tags with that data attribute and its basically `script-src 'unsafe-line'` with extra steps.

An alternate way that uses a django-sekizai block postprocessor to add the nonce, this is safer as it only processes the scripts added to sekizai (eg: significant smaller attack surface). A problem here is that it needs some work to cover other cms/admin template tag output (like the cms_toolbar-tag). So this also doesn't process the plugin placeholders but I think that is fine.

We should discus this as each solution has some flaws.

### DIFF
--- a/src/open_inwoner/cms/csp/middleware.py
+++ b/src/open_inwoner/cms/csp/middleware.py
@@ -1,0 +1,59 @@
+import logging
+
+from django.http import HttpResponse
+
+from lxml.etree import ParseError, ParserError
+from lxml.html import document_fromstring, tostring
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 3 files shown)


Bartvaderkin:
I'm aware Bandit is being difficult about lxml, I'll fix it if we want this.

PR #600, opened 2023-05-01
[#1404] Disable CSP when authenticated as staff user

Bartvaderkin:
Closes #598


alextreme (approved):
Approved, CSP is only necessary for DigiD users. Staff/admin users are except

PR #604, opened 2023-05-02
[#1407] Harden plugins to not crash site if cms-app is not active

vaszig:


src/open_inwoner/cms/utils/plugin_mixins.py
```diff
@@ -0,0 +1,14 @@
+from cms.models import Page
+from cms.plugin_base import CMSPluginBase
+
+
+class CMSActiveAppMixin:
+    app_hook = None
+
+    @property
+    def render_plugin(self):
+        return Page.objects.published().filter(application_urls=self.app_hook).exists()
```
Bartvaderkin, line 10: I think this should raise an error if `app_hook` is `None`, just to be safe.

Not sure what type, maybe ValueError? It doesn't really matter as this is mostly to catch development issues.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

alextreme:
@vaszig please resolve the conflicts ( was due to my merging of #606 )

PR #605, opened 2023-05-03
[#1312] Make banner height adjustable

vaszig:


jiromaykin:
@vaszig This doesn't have the option to select a preview or 'location aware cropping' - is that something we will add later?

vaszig:
> @vaszig This doesn't have the option to select a preview or 'location aware cropping' - is that something we will add later?

Discussed with @alextreme and decided to go for a simple height adjustment (via the form upload) which will be done only for the frontend.

jiromaykin (approved):
Works nice! (although we need to add to the manual the banner will not be visible until the user is logged out, so the admin won't see it either when logged in).

PR #610, opened 2023-05-09
[#1316] Add title-description to banner plugin

vaszig:
Discussed with Alex as well, this does not make sense to be combined in one plugin. We need the title and description to be available in another section (in the main-grid and not in the banner), so a second plugin was created for this purpose.


src/open_inwoner/cms/banner/models.py
```diff
@@ -5,8 +5,12 @@
...
 
 
-class Banner(CMSPlugin):
-    title = models.CharField(_("title"), max_length=250, default="")
+class BannerImage(CMSPlugin):
+    name = models.CharField(
+        _("Name"),
+        max_length=250,
+        help_text=_("The name of the image (this will not be shown on the page)"),
+    )
```
Bartvaderkin, line 13: Why do we need a name-field if we're not showing it?

Bartvaderkin (changes requested):

src/open_inwoner/cms/banner/models.py
vaszig: I was thinking that we do not have any required fields, so maybe I should remove name and make image a required field?

jiromaykin (commented):
So this text area is really only visible when logged in? Becasue when you're logged out, there is a different 'welkom' area visible that is still there and cannot be removed by the user and is configarable from the Admin.

So when  am logged out i still see the 'Koptekst homepage' area on Home because that is a required field - i am not sure if that one needs to be required and visible still - what do you think @alextreme ?
Screenshot 2023-05-11 at 10 21 59

src/open_inwoner/cms/banner/models.py
vaszig: Discussed with Bart, name fields are not useful. Image and title will be the required fields for the two models (`BannerImage` and `BannerText`).

alextreme: Reasoning why I didn't request to remove 'name' is that I often see plugins with an (internal) name which is only used for plugin.__str__ , this to make the name visible in the CMS when editting the page.

This is mostly used for CMS pages with long lists of plugins, so I agree that the benefit for only the bannerplugin is quite limited.

alextreme:
> So this text area is really only visible when logged in? Becasue when you're logged out, there is a different 'welkom' area visible that is still there and cannot be removed by the user and is configarable from the Admin.
> 
> So when am logged out i still see the 'Koptekst homepage' area on Home because that is a required field - i am not sure if that one needs to be required and visible still - what do you think @alextreme ?

This was discussed with @vaszig as a workable intermediate solution (populating the anonymous page via the Algemene Configuratie, and the logged-in homepage via the CMS). But I agree it isn't ideal, I would like to have the anonymous homepage configurable via the CMS too but that would be a next step.

Bartvaderkin (approved):

Bartvaderkin (changes requested):
Last week I started adding tests for the plugins, for example (some others as well): 

https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/cms/products/tests/test_plugin_product_location.py

We should do that for all plugins.

Bartvaderkin (approved):

PR #613, opened 2023-05-11
[#1378] Updated new menus to use dynamic pages

Bartvaderkin:
This assumes you have the authenticated pages as children of the homepage in the tree, and use 'home' as ID in the page's advanced settings.

The products/onderwerpen are still rendered as they were, until we move the content to actual pages.

Note I also added a context manager that adds some data to check if an app is active.

Relates to #591


jiromaykin (commented):
Seems to work great!
I assume the icons before the pages will be added later on as an option?
Or is that already active somewhere?

Screenshot 2023-05-12 at 11 23 00

Screenshot 2023-05-12 at 11 23 45

jiromaykin:
@alextreme and @Bartvaderkin I rebased my own branch from PR591, which maybe resulted in conflicts in this branch? 
I hope this branch can be approved so everything can be merged and I can clean up the menu and remove all unnecessary comments in [Pull Request 591](https://github.com/maykinmedia/open-inwoner/pull/591)

alextreme (approved):
Could use some caching to avoid determining the active apphooks on every request, but considering it's a single query this should be okay for now

PR #615, opened 2023-05-11
[#1442] Fix django-admin-index fixture

vaszig:


Bartvaderkin (changes requested):
The change in formatting makes it hard to see actual changes in the diff.
 
Do you generate this with the `generate_admin_index_fixture.sh` script from the bin folder? 

If I look at the GIT history I see some other changes that look a bit suspicious.

vaszig:
> The change in formatting makes it hard to see actual changes in the diff.
> 
> Did you generate this with the `generate_admin_index_fixture.sh` script from the bin folder?
> 
> If I look at the GIT history I see some other changes that look a bit suspicious.

@Bartvaderkin  I did this in this way but black formatted the json file, so I reverted it and it works as expected

Bartvaderkin:
@vaszig I don't think you're supposed to black json :D

Check what your using as githook or script so it filters on python files.

Bartvaderkin (approved):

PR #618, opened 2023-05-15
[#1452] Added middleware to hide django-cms toolbar when not 2FA verified

Bartvaderkin:
I tried a bunch of approaches, like putfing an if/else around the template tag, extending the regular ToolbarMiddleware, extending the template tag and some others, but this is the simplest that doesn't break anything.

It seems to work in browser, if it is acceptable I'll figure out how to create a test for this tomorrow.

### DIFF
--- a/src/open_inwoner/cms/utils/middleware.py
+++ b/src/open_inwoner/cms/utils/middleware.py
@@ -1,5 +1,7 @@
 from django.http import HttpResponseRedirect
 
+from cms.toolbar.utils import get_toolbar_from_request
+
 from open_inwoner.configurations.models import SiteConfiguration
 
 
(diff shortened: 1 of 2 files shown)


src/open_inwoner/cms/utils/middleware.py
```diff
@@ -24,3 +26,25 @@ def __call__(self, request):
...
+
+class DropToolbarMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if not self.use_toolbar(request):
+            request.session["cms_edit"] = False
+            request.session["cms_preview"] = False
+            request.session["cms_toolbar_disabled"] = True
```
Bartvaderkin, line 39: We might not even need these but the CMS code is really messy in when/where is looks into these so I'd just clobber them before setting the most important toolbar.show_toolbar = False

alextreme (approved):

PR #620, opened 2023-05-16
[#1452] Improved DropToolbarMiddleware to honour settings.TWO_FACTOR_FORCE_OTP_ADMIN

Bartvaderkin:
Also added tests and a mixin to improve 2FA testing

### DIFF
--- a/src/open_inwoner/cms/tests/test_middleware.py
+++ b/src/open_inwoner/cms/tests/test_middleware.py
@@ -1,0 +1,70 @@
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django.urls import reverse
+
+from pyquery import PyQuery as pq
+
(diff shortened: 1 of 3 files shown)


src/open_inwoner/cms/tests/test_middleware.py
```diff
@@ -0,0 +1,70 @@
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django.urls import reverse
+
+from pyquery import PyQuery as pq
```
Bartvaderkin, line 5: I originally used Webtest and it's pyquery extension, but to get the 2factor test setup working I had to use mixin code which depends on Django's regular TestCase/client, so I had to do my own pyqueerying.

alextreme:
:partying_face:

PR #621, opened 2023-05-16
[#1439] Add htmx to cases

vaszig:


vaszig:
@jiromaykin FYI if you want to freeze the spinner and review it properly you can add a breakpoint to the cases (cms/cases/views/cases.py -> `InnerOpenCaseListView.get_cases()`) . Let me know if you need any help.

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -0,0 +1,184 @@
...
+                logger.debug(
+                    "CaseAccessMixin - permission denied: case {self.case.url} is not visible"
+                )
+                return self.handle_no_permission()
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def handle_no_permission(self):
+        if self.request.user.is_authenticated:
+            return TemplateResponse(self.request, "pages/cases/403.html")
```
Bartvaderkin, line 92: neat!

Bartvaderkin (changes requested):
Htmx stuff looks pretty good.

One thing that bothers me is the javascript components don't chain from the node that is found by the selector, but use global selectors to grab the various sub-elements and attach event handlers. 

This is not accurate and could give memory leaks because if the selector matches the new html fragment the component could still find and (duplicated) attach to elements outside of the fragment. Eg: multiple instantiations of the component code could connect to the same elements.

jiromaykin:
> ...FYI if you want to freeze the spinner and review it properly you can add a breakpoint to the cases[...].

Thanks! I am excited to test this!

jiromaykin (changes requested):
Maybe it's just me, but when the list view has loaded (any list open/closed/Lopende) - I still see the spinner at the bottom after data has loaded (spinner should disappear here after loading is done),
When this happens it seems like javascript is becoming unresponsive because I cannot open the desktopmenu navigation anymore from this page.
 
listview-aanvraag

vaszig:
> > Maybe it's just me, but when the list view has loaded (any list open/closed/Lopende) - I still see the spinner at the bottom after data has loaded
> 
> Additionally: the main-navigation still works if I log in as a DigiD user - but it still has the spinner still visible after data has loaded.

@jiromaykin have you reloaded-rebuilt js in your local machine?It seems weird because I don't seem to have a problem.

src/open_inwoner/scss/components/Spinner/Spinner.scss
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+.loader-container {
+  &--hide {
+    display: none;
+  }
+
+  .rotate {
+    animation: rotate 2.5s linear infinite;
```
jiromaykin, line 16: I feel this is just a tiny bit too slow, even if you have to wait long - 1.5 seconds would feel better.
All else looking good!

jiromaykin (changes requested):

jiromaykin:
> 
> [...] have you reloaded-rebuilt js in your local machine?It seems weird because I don't seem to have a problem.

Sorry @vaszig - I thought the Watch function did a clean rebuild. It's working correctly after one build.

src/open_inwoner/js/components/upload-document/show-file-info.js
```diff
@@ -1,95 +1,111 @@
...
+    // get the closest parent element (the form section element)
+    const documentUpload = this.fileUploadInput.closest('#form_upload')
 
-  fileUploadInput.addEventListener('change', function (e) {
-    const files = e.target.files
+    const sizeInfo = documentUpload.querySelector('#upload_size')
+    const nameInfo = documentUpload.querySelector('#upload_name')
+    const validationInfo = documentUpload.querySelector('#upload_error')
+    const closeButton = documentUpload.querySelector('#close_upload')
+    const submit_upload = documentUpload.querySelector('#submit_upload')
```
Bartvaderkin, line 17: This still uses ID's, eg: global selectors. I'm not sure what browsers do with duplicate ID's even if you'd `querySelector from a one of their ancestor elements, or how this works with closest. 

The point of this is making sure that when hydrating one fragment we're not accidentally select elements elsewhere in the page. 

In this case I'll not block on this as there won't be multiple of this element.

Bartvaderkin (approved):
I still see a lot of global selectors (eg: ID's, `getElementById(x)` or `querySelector("#x")` in the components. 

As everything works I'll approve to land this larger PR and added Taiga issue 1544 (and earlier 1511) to clean this up.

PR #622, opened 2023-05-16
[#1461] Only show 'Bestanden' if there are files to show

alextreme:


vaszig (approved):
Maybe we want to modify the if statement to the export html as well:
 https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/templates/export/profile/profile_export.html#L52

PR #624, opened 2023-05-17
[#1463] Add phonenumber field to registration form

vaszig:


src/open_inwoner/accounts/forms.py
```diff
@@ -43,11 +44,22 @@ class Meta:
...
             "password2",
             "invite",
         )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # make phonenumber required when 2fa-sms login is enabled
+        config = SiteConfiguration.get_solo()
+        if config.login_2fa_sms is False:
```
Bartvaderkin, line 58: Let's keep it pythonic and not compare to booleans:

`if not config.login_2fa_sms:`

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #626, opened 2023-05-19
[#1464] Add 2fa sms verification for regular login

vaszig:


vaszig:
The potential problem that Github found has to do with an anchor tag which uses a POST request to resend a token. @alextreme @Bartvaderkin you can see after the review if it's really a security issue.

Bartvaderkin:
It seems safe as it only handles POST so with CSRF protection the referer has to come from one of our pages (eg: this referer header cannot be spoofed by a third party).

src/open_inwoner/accounts/gateways.py
```diff
@@ -0,0 +1,93 @@
...
+        else:
+            logging.debug('Sent SMS to %s: "%s"', to, self.get_message(token))
+            return True
+
+
+def load_gateway(path):
+    module, attr = path.rsplit(".", 1)
+    mod = import_module(module)
+    cls = getattr(mod, attr)
+    return cls()
```
Bartvaderkin, line 84: This is basically django's `import_string`

src/open_inwoner/js/components/twofactor-sms/index.js
```diff
@@ -0,0 +1,26 @@
...
+      xhr.open('POST', this.getAttribute('href'))
+      xhr.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded')
+      xhr.onload = function () {
+        if (xhr.status === 200) {
+          window.location.reload()
+        } else {
+          window.location.reload()
+        }
+      }
+      xhr.send('csrfmiddlewaretoken=' + encodeURIComponent(csrfToken))
```
Bartvaderkin, line 23: We have a `getCsrfTokenFromDom()` in `js/utils.js`

Also instead of using ancient XHR we should in general use the new Fetch API.

src/open_inwoner/conf/base.py
```diff
@@ -826,6 +827,21 @@
 MOZILLA_DJANGO_OIDC_DB_CACHE = "oidc"
 MOZILLA_DJANGO_OIDC_DB_CACHE_TIMEOUT = 1
 
+#
+# 2FA SMS Verification
+#
+
+MESSAGE_BIRD_ACCESS_KEY = os.getenv("MESSAGE_BIRD_ACCESS_KEY")
+ACCOUNTS_USER_TOKEN_EXPIRE_TIME = 300
+ACCOUNTS_SMS_MESSAGE = "Open Inwoner Platform Bevestigingscode: {token} (deze code is geldig voor 5 minuten)"
```
Bartvaderkin, line 836: Display string should be marked as (lazy) translatable

src/open_inwoner/utils/validators.py
```diff
@@ -36,6 +36,14 @@ def validate_phone_number(value):
...
 
 
+def format_phone_number(value):
+    if value[0:2] == "00":
+        value = "+{val}".format(val=value[2:])
+    elif value[0] == "0":
+        value = "+31{val}".format(val=value[1:])
+    return value.strip().replace("-", "").replace(" ", "")
+
+
```
Bartvaderkin, line 46: Lets add some quick & simple tests to verify this works for anything that passes the validator. 

Also, in general prefer to use f-strings instead of .format()

src/open_inwoner/accounts/views/login.py
```diff
@@ -0,0 +1,369 @@
...
+        params = {
+            self.redirect_field_name: redirect_to,
+            "user": signer.sign(user.pk),
+        }
+
+        if not user.phonenumber:
+            # Redirect to add mobile phone number view
+            return HttpResponseRedirect(
+                reverse("add_phone_number") + "?" + urlencode(params)
+            )
```
Bartvaderkin, line 56: In Django there is also the `redirect()` shortcut that is slightly nicer then HttpResponseRedirect, and usually we like `furl()` but in this case it's probably fine as is.

src/open_inwoner/accounts/views/login.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        auth_login(
+            self.request,
+            self.user_cache,
+            backend="open_inwoner.accounts.backends.UserModelEmailBackend",
+        )
+
+    @method_decorator(never_cache)
+    def dispatch(self, request, *args, **kwargs):
+        UserModel = get_user_model()
```
Bartvaderkin, line 123: In this project we don't support swappable User models so you don't need `get_user_model()` an can just hard-code to the real model.

src/open_inwoner/accounts/views/login.py
```diff
@@ -0,0 +1,369 @@
...
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_verification_location(self, request, user):
+        url_part = reverse(self.url_name)
+        location = request.META.get("HTTP_REFERER", url_part)
+
+        if "user=" not in location:
+            params = {"user": signer.sign(user.pk)}
+            location = location + "?" + urlencode(params)
```
Bartvaderkin, line 235: Let's use `furl()` for URL parsing and manipulation.

src/open_inwoner/conf/base.py
```diff
@@ -826,6 +827,21 @@
...
+ACCOUNTS_SMS_GATEWAY = {
+    "BACKEND": config(
+        "ACCOUNTS_SMS_GATEWAY_BACKEND", "open_inwoner.accounts.gateways.Dummy"
+    ),
+    "API_KEY": config("ACCOUNTS_SMS_GATEWAY_API_KEY", "openinwoner"),
+    "ORIGINATOR": config("ACCOUNTS_SMS_GATEWAY_ORIGINATOR", "openinwoner"),
+}
+
```
Bartvaderkin, line 844: I see both MESSAGE_BIRD_ACCESS_KEY and ACCOUNTS_SMS_GATEWAY_API_KEY which feels suspicious, as I don't see the MESSAGE_BIRD_ACCESS_KEY being used.

src/open_inwoner/accounts/views/login.py
```diff
@@ -0,0 +1,369 @@
...
+        else:
+            self.user_cache = request.user
+        if not self.user_cache:
+            return HttpResponseRedirect(reverse("login"))
+
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_verification_location(self, request, user):
+        url_part = reverse(self.url_name)
+        location = request.META.get("HTTP_REFERER", url_part)
```
Bartvaderkin, line 231: Thinking about the Gitbub scan result: do we actually need to look at HTTP_REFERER? Or can we just use the fallback it is doing here (with self.url_name)?

Bartvaderkin (changes requested):
I added some notes. but it is a mix of actual changes and general feedback and questions.

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,542 @@
...
+
+        self.assertEqual(response.url, reverse("pages-root"))
+
+        response = self.client.get(response.url, follow=True)
+
+        self.assertEqual(
+            [str(m) for m in response.context["messages"]],
+            [
+                "Het is vanwege een storing tijdelijk niet mogelijk om in te loggen, probeer het over 15 minuten nogmaals. "
+                "Mocht het inloggen na meerdere pogingen niet werken, neem dan contact op met de Open Inwoner Platform."
```
Bartvaderkin, line 153: There are not marked as translatable.

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,542 @@
...
+            self.assertTrue(mock_gateway_send.called)
+            self.assertEqual(mock_gateway_send.call_count, 1)
+            self.assertEqual(
+                mock_gateway_send.call_args[1]["to"], self.user.phonenumber
+            )
+
+            location = reverse("verify_token") + "?" + urlencode(params)
+
+            # we have a limit of 25 visit times per 5 minutes
+            for login_attempt in range(1, 26):
```
Bartvaderkin, line 376: Maybe just lower the limit for the test instead of just hammering through 25 tries :laughing:

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,542 @@
...
+                "username": self.user.email,
+                "password": "secret",
+            },
+            follow=True,
+        )
+
+        self.assertContains(response, "Account verificatie")
+        self.assertTrue(mock_gateway_send.called)
+        self.assertEqual(mock_gateway_send.call_count, 1)
+        self.assertEqual(mock_gateway_send.call_args[1]["to"], self.user.phonenumber)
```
Bartvaderkin, line 78: You can assert on the mock itself: `mock.assert_called_once()` or `mock.assert_called_once_with(...)` (etc, see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock)

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,542 @@
...
+                "username": self.user.email,
+                "password": "secret",
+            },
+            follow=True,
+        )
+        self.assertContains(response, "Account verificatie (stap 1 van 2)")
+        self.assertFalse(mock_gateway_send.called)
+
+        response = self.client.post(
+            response._request.get_full_path(),
```
Bartvaderkin, line 481: This feels weird:

````
    response = self.client.post(
            response._request.get_full_path(),
````

Maybe just explicitly use the real url? Now there is indirection and we have to search and think to read it.

File: src/open_inwoner/accounts/tests/test_gateways.py [hunk 14]
```diff
@@ -0,0 +1,34 @@
...
+
+        phone_number = "0123456789"
+        token = "123456"
+        result = gateway.send(to=phone_number, token=token)
+
+        self.assertTrue(result)
+        self.assertTrue(mock_message_create.called)
+        self.assertEqual(mock_message_create.call_count, 1)
+        self.assertEqual(mock_message_create.call_args[0][0], "OIP")
+        self.assertEqual(mock_message_create.call_args[0][1], "0123456789")
```
Bartvaderkin, line 30: More `mock.assert_called_once_with()`

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,542 @@
+import datetime
+import time
+from unittest.mock import patch
+
+from django.conf import settings
+from django.core.signing import TimestampSigner
+from django.test import TestCase, override_settings
+from django.test.utils import freeze_time as dj_freeze_time
```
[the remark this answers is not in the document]
vaszig: Discussed with @Bartvaderkin, this is used because freezegun causes a problem with time.

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
vaszig: The first response redirects the user to the phonenumber view so we have to use this in order to be aware of the url with the query params.

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -0,0 +1,535 @@
...
+            data={
+                "next": "",
+                "username": self.user.email,
+                "password": "secret",
+            },
+            follow=True,
+        )
+
+        self.assertNotContains(response, "Account verificatie")
+        self.assertRedirects(response, reverse("pages-root"))
```
Bartvaderkin, line 56: Is this assertRedirect correct if we use `follow=True` above?

Also this test relates to my other comment about the authentication backend and 2fa?

Bartvaderkin (changes requested):
Sorry, but I added another round of feedback.

Also I think we need to re-(re-)consider using webtest, because the forms and redirects are not really tested, but we'll call about it.

src/open_inwoner/accounts/backends.py
```diff
@@ -19,14 +22,31 @@ class UserModelEmailBackend(ModelBackend):
...
+        config = SiteConfiguration.get_solo()
+
+        if username and password and not config.login_2fa_sms:
+            try:
+                user = get_user_model().objects.get(email__iexact=username)
+                if check_password(password, user.password):
+                    return user
+            except get_user_model().DoesNotExist:
+                # No user was found, return None - triggers default login failed
+                return None
```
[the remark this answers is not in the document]
vaszig: This was tested in localhost and the user can log in as expected. I have added a test for this situation as well.

alextreme:
@vaszig see if you can round this up tomorrow, as I don't want to release v1.7 without it

src/open_inwoner/accounts/tests/test_auth_2fa_sms.py
```diff
@@ -255,23 +211,19 @@ def test_login_token_with_max_attempts(self, mock_gateway_send):
...
-            other_user.refresh_from_db()
+            verify_token_form = response.forms["verify-token-form"]
+            verify_token_form["token"] = sent_token
+            response = verify_token_form.submit()
+
             self.assertRedirects(response, reverse("pages-root"))
 
         # make sure the first user can login after a minute has passed
         with freeze_time("2023-05-22 12:01:01"):
             response = self.client.post(
```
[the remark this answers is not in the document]
vaszig: I am using `self.client` because I cannot inject the CSRF token in the post request when I am using webtest. 
That's why we have used in the past and in another PR now, the class attribute `csrf_checks = False`

https://github.com/django-webtest/django-webtest/blob/master/README.rst

alextreme (approved):
Approved for now to allow for v1.7 release

PR #629, opened 2023-05-22
[#1416] Add mijn vragen to profile apphook config

vaszig:


Bartvaderkin (changes requested):
Can we call this "questions" instead of "my_questions"?

Bartvaderkin (approved):

PR #631, opened 2023-05-23
[#1476] Production fixes for install without digid namespace

alextreme:


alextreme:
A few hotfixes were required for Deventer this morning, digid wasn't a reversable namespace. Will likely need improvements but this seems to have done the trick.

vaszig (approved):
We have the try-except problem (pass) but I assume it's ok for now and we can create a task for improving these checks.

PR #633, opened 2023-05-23
[#1482] Added 'infix' (tussenvoegsel) to User model

Bartvaderkin:


src/open_inwoner/haalcentraal/tests/test_signal.py
```diff
@@ -31,15 +30,16 @@ def test_signal_updates_users_data_when_logged_in_via_digid_v_2(self, m):
         user.bsn = "999993847"
         user.save()
 
-        updated_user = User.objects.filter(email=user.email)
+        user.refresh_from_db()
```
vaszig, line 33: Thanks for the updated part!

vaszig (changes requested):
Looks good! Let's add the infix field to the `profile/edit.html` as well, because we render the fields manually.

Bartvaderkin:
> Looks good! Let's add the infix field to the `profile/edit.html` as well, because we render the fields manually.

Thanks, weird how I missed that. I appended a commit for it.

vaszig (approved):

PR #634, opened 2023-05-23
[#1475] Harden apimock against directory traversal

Bartvaderkin:
relates to comment on #632


alextreme (approved):
Dank, blij dat we opletten :)

PR #640, opened 2023-06-01
[#1448] Add banner plugins tests

vaszig:


src/open_inwoner/cms/banner/tests/test_plugin_banner.py
```diff
@@ -0,0 +1,76 @@
...
+class TestBannerImage(TestCase):
+    def test_banner_image_is_rendered_in_plugin(self):
+        pil_image = Image.new("RGB", (10, 10))
+        tmp_file_path = "{}/{}".format(settings.MEDIA_ROOT, "tmp.jpg")
+        pil_image.save(tmp_file_path, "JPEG")
+
+        tmp_image_file = open(tmp_file_path, "rb")
+        file_obj = File(tmp_image_file)
+        instance = FilerImage()
+        instance.file.save("image.jpg", file_obj)
```
Bartvaderkin, line 26: This feels weird. 

For one it is not very DRY: let's create a function instead of repeating complicated code blocks (if you search for `Image.new` in de code you'll find similar snippets). 

Maybe we don't even need that and could create a `FilerImageFactory` (maybe with a ImageFileFactory or some function)

If not a factory but our own function then it seems unnecessary to create a tmp file? I think you can save the `pil_image` to a `BytesIO` (instead of a path), and then pass the BytesIO to the `instance.file.save` (instead of the `open()`-ed `file_obj`).

Bartvaderkin (changes requested):

Bartvaderkin (approved):

Bartvaderkin:
@vaszig I approved this but there is a merge conflict.

PR #641, opened 2023-06-01
[#1439] Fixed htmx re-hydration issue

Bartvaderkin:


src/open_inwoner/js/components/index.js
```diff
@@ -42,18 +42,29 @@ const elementWrappers = [
...
   }
+  // clean-up removed elements
+  activeElements = activeElements.filter((elt) => elt.isConnected)
 }
 
-htmx.onLoad(() => {
-  document.body.addEventListener('htmx:afterSwap', (event) => {
-    wrapComponentsOf(event.target)
-  })
-})
```
Bartvaderkin, line 59: I hooked into the wrong event before: apparently htmx onload is not a general ("htmx is ready") but already the event we wanted to use, eg: when a new element is added. In effect this meant that this kept adding event listeners and make a huge invisible mess.

src/open_inwoner/js/components/anchor-menu/anchor-menu.js
```diff
@@ -1,21 +1,15 @@
...
 }
 
-htmx.onLoad(() => {
-  document.body.addEventListener('htmx:afterSwap', (event) => {
-    const anchors = document.querySelectorAll(CreateGumshoe.selector)
-    ;[...anchors].forEach((anchor) => new CreateGumshoe(anchor))
-  })
-})
+const anchors = document.querySelectorAll(CreateGumshoe.selector)
+;[...anchors].forEach((anchor) => new CreateGumshoe(anchor))
```
Bartvaderkin, line 15: I created Taiga ticket #1511 to do something about these self-hydrating components.

vaszig (approved):
Thanks Bart, I will try to finish with the other issues as well.

PR #642, opened 2023-06-01
[#1453] Add-fix missing icon-options to Page extension

vaszig:


Bartvaderkin (changes requested):
Looks good, but since we changed the available values I think we need a data migration to update existing data.

alextreme (approved):
Leaving re-selecting the icons for the pages to the municipalities

PR #647, opened 2023-06-05
[#1509] fix template 'form.html'

pi-sigma:
Removed sidebar and form overview from the form template so that they are no longer displayed once a form has been started.

Taiga: #1509

### DIFF
--- a/src/open_inwoner/pdc/views.py
+++ b/src/open_inwoner/pdc/views.py
@@ -228,17 +228,6 @@ class ProductFormView(
             (_("Formulier"), self.request.path),
         ]
 
-    def get_context_data(self, **kwargs):
-        product = self.get_object()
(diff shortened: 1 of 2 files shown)


vaszig (changes requested):
Let's remove the anchors from the view (`ProductFormView`) since we don't use the sidebar anymore. It shows errors when Gumshoe tries to access them.

image

pi-sigma:
I set up integration with openforms locally and the solution works:

oi-of-local

On the other hand, in the test environment:

oi-of-test

@alextreme Does the test environment already include this patch?

alextreme:
@pi-sigma no, this PR is not merged yet so it won't be on the testenvironment.

After you make your changes hit the 're-review' icon next to your reviewers name

vaszig (approved):

PR #650, opened 2023-06-05
:bug: [#1467] Fix mobile navigation for old iOS devices

jiromaykin:
See if navigation can be fixed for old iPhones that cannot be updated anymore to use any iOS since 15.4
(in order to make dialog tag work)
solution: replace dialog tag  with div and change flexbox styles.

I have simplified javascript too - perhaps too simple this time? Needs review.

### DIFF
--- a/src/open_inwoner/components/templates/components/Header/Header.html
+++ b/src/open_inwoner/components/templates/components/Header/Header.html
@@ -25,7 +25,7 @@
             </span>
         {% endif %}
 
-            <dialog class="header__submenu">
+            <div class="header__submenu">
 
             {% if cms_apps.products %}
             <nav class="header__actions" aria-label="Zoek navigatie mobiel">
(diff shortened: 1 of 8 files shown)


jiromaykin:
@Bartvaderkin and @vaszig No one has reviewed this yet 🙂 I know it is annoying but at least I need to know if my javascript is in anyway okay or not (it should all still work in modern browsers, so that is easier to test - if you really wish to test its behavior on iOS devices, you can use Browserstack, for which I can give you credentials).

vaszig (approved):
Tested on IPhone 8 and 13 in browserstack and works fine.

jiromaykin:
> * I do note this particular component uses a `Component` base class that does some fake-React setState() stuff that seems undesirable (like json transcoding to deep copy).

Yes @Bartvaderkin the 'fake-state' stuff is what I struggle to understand repeatedly since that was built by a different developer who has a much deeper grasp of javascript than me; so this might return as an issue.

Bartvaderkin (approved):

PR #652, opened 2023-06-05
[#1478] Showing message.success after a profile(-related) save

alextreme:


src/open_inwoner/accounts/views/profile.py
```diff
@@ -129,6 +129,7 @@ def get_object(self):
     def form_valid(self, form):
         form.save()
 
+        messages.success(request, _("Uw wijzigingen zijn opgeslagen"))
```
vaszig, line 132: `self.request`

src/open_inwoner/accounts/views/profile.py
```diff
@@ -164,7 +165,7 @@ def get_object(self):
 
     def form_valid(self, form):
         form.save()
-
+        messages.success(request, _("Uw wijzigingen zijn opgeslagen"))
```
vaszig, line 168: `self.request`

src/open_inwoner/accounts/views/profile.py
```diff
@@ -259,7 +260,7 @@ def get_form_kwargs(self):
...
```
vaszig, line 263: `self.request`

vaszig (changes requested):
A small change and it's ready to go.

alextreme:
More like: don't break stuff Alex :thisisfine:

PR #654, opened 2023-06-07
[#1508] My profile messages

pi-sigma:
Links for sending messages to contacts are displayed if and only if a message page exists and is set to be published.

Taiga: #1508

### DIFF
--- a/src/open_inwoner/accounts/tests/test_contact_views.py
+++ b/src/open_inwoner/accounts/tests/test_contact_views.py
@@ -6,8 +6,8 @@ from django.test import override_settings
 from django.urls import reverse
 from django.utils.translation import ugettext_lazy as _
 
+from cms import api
 from django_webtest import WebTest
-from PIL import Image
 
 from open_inwoner.accounts.models import User
 from open_inwoner.utils.tests.helpers import create_image_bytes
@@ -85,10 +85,40 @@ class ContactViewTests(WebTest):
             ),
         )
 
-    def test_contact_list_show_link_to_messages(self):
-        message_link = reverse("inbox:index", kwargs={"uuid": self.contact.uuid})
+    def test_messages_enabled_disabled(self):
+        """Assert that `Stuur Bericht` is displayed if and only if the message page is published"""
+
+        # case 1: no message page
         response = self.app.get(self.list_url, user=self.user)
-        self.assertContains(response, message_link)
+
+        # self.assertNotIn(_("Stuur bericht"), response)
+        self.assertNotContains(response, _("Stuur bericht"))
+
+        # case 2: unpublished message page
+        page = api.create_page(
+            "Mijn Berichten",
+            "cms/fullwidth.html",
+            "nl",
+            slug="berichten",
+        )
+        page.application_namespace = "inbox"
+        page.save()
+
+        response = self.app.get(self.list_url, user=self.user)
+
+        self.assertNotContains(response, _("Stuur bericht"))
+
+        # case 3: published message page
+        page.publish("nl")
+        page.save()
+
+        response = self.app.get(self.list_url, user=self.user)
+
+        icons = response.pyquery(".material-icons-outlined")
+        message_icon = next((icon for icon in icons if icon.text == "message"), None)
+        message_button_text = message_icon.tail.strip()
+
+        self.assertEqual(_("Stuur bericht"), message_button_text)
 
     def test_contact_list_show_reversed(self):
         other_contact = UserFactory(first_name="reverse_contact_user_should_be_found")
@@ -366,19 +396,12 @@ class ContactViewTests(WebTest):
 
         self.assertContains(response, self.user.first_name)
         self.assertContains(response, self.user.last_name)
-        self.assertContains(
-            response, reverse("inbox:index", kwargs={"uuid": self.user.uuid})
-        )
 
         # Sender contact list page
         response = self.app.get(self.list_url, user=self.user)
 
         self.assertContains(response, existing_user.first_name)
         self.assertContains(response, existing_user.last_name)
-        self.assertContains(
-            response,
-            reverse("inbox:index", kwargs={"uuid": existing_user.uuid}),
-        )
 
     def test_post_with_no_params_in_contact_approval_returns_bad_request(self):
         existing_user = UserFactory(email="ex@example.com")
--- a/src/open_inwoner/accounts/views/contacts.py
+++ b/src/open_inwoner/accounts/views/contacts.py
@@ -11,6 +11,7 @@ from django.views.generic.edit import FormView
 from mail_editor.helpers import find_template
 from view_breadcrumbs import BaseBreadcrumbMixin
 
+from open_inwoner.cms.utils.page_display import inbox_page_is_published
 from open_inwoner.utils.views import CommonPageMixin, LogMixin
 
 from ..forms import ContactCreateForm, ContactFilterForm
@@ -48,6 +49,7 @@ class ContactListView(
         context["contacts_for_approval"] = user.get_contacts_for_approval()
         context["pending_invitations"] = user.get_pending_invitations()
         context["form"] = ContactFilterForm(data=self.request.GET)
+        context["inbox_page_is_published"] = inbox_page_is_published()
         return context
 
 
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -20,6 +20,7 @@ from open_inwoner.accounts.choices import (
     LoginTypeChoices,
     StatusChoices,
 )
+from open_inwoner.cms.utils.page_display import inbox_page_is_published
 from open_inwoner.haalcentraal.utils import fetch_brp_data
 from open_inwoner.questionnaire.models import QuestionnaireStep
 from open_inwoner.utils.mixins import ExportMixin
@@ -97,6 +98,7 @@ class MyProfileView(
             published=True
         ).exists()
         context["can_change_password"] = user.login_type != LoginTypeChoices.digid
+        context["inbox_page_is_published"] = inbox_page_is_published()
 
         return context
 
--- a/src/open_inwoner/cms/utils/page_display.py
+++ b/src/open_inwoner/cms/utils/page_display.py
@@ -1,0 +1,45 @@
+"""Utilities for determining whether CMS pages are published"""
+
+
+from cms.models import Page
+
+from open_inwoner.cms.cases.cms_apps import CasesApphook
+from open_inwoner.cms.collaborate.cms_apps import CollaborateApphook
+from open_inwoner.cms.inbox.cms_apps import InboxApphook
+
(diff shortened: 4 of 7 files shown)


pi-sigma:
- UX: after hiding the link to send messages for active contacts in case the messaging page is unpublished, should we also get rid of the link for contacts that still need approval? Or should the link be displayed as disabled so as to signal to the user: messaging is available in principle, but for this particular contact it is temporarily disabled (current situation)?

- Implementation: the solution filters `Page` objects by `application_namespace` and checks for the first page if it is published. At first I naively assumed that the result of filtering by `application_namespace` would be unique, but the resulting queryset has *two* elements (with `is_draft=True` and  `is_draft=False` respectively). Both however are `is_published=True`. I double-checked and noticed noticed that this is the case for *every* `application_namespace`. I'll be looking into the CMS documentation to see if I can figure out why this is (and whether the implementation is wrong), but if someone with more experience with CMS could shed some light on this, that would be great.

src/open_inwoner/accounts/views/profile.py
```diff
@@ -86,6 +86,7 @@ def get_context_data(self, **kwargs):
             context[
                 "contact_text"
             ] = f"{', '.join(contact_names)}{'...' if contacts.count() > 3 else ''}"
+            context["has_contacts"] = True
```
vaszig, line 89: Where is this used?

src/open_inwoner/cms/profile/cms_appconfig.py
```diff
@@ -52,3 +53,11 @@ class ProfileConfig(AppHookConfig):
         default=True,
         help_text=_("Designates whether 'questions' section is rendered or not."),
     )
+
```
vaszig, line 56: I am not sure if this should be here or in the Profile context view. It's not exactly part of the ProfileConfig model.

src/open_inwoner/cms/profile/cms_appconfig.py
```diff
@@ -52,3 +53,11 @@ class ProfileConfig(AppHookConfig):
...
+    @property
+    def messages_enabled(self) -> bool:
+        """
+        :returns: True if the message page is set to be published, False otherwise
+        """
+        message_page = Page.objects.filter(application_namespace="inbox").first()
```
vaszig, line 62: The same comment as in contacts above.

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -48,8 +49,17 @@ def get_context_data(self, **kwargs):
...
         context["form"] = ContactFilterForm(data=self.request.GET)
+        context["messages_enabled"] = self.messages_enabled
         return context
 
+    @property
+    def messages_enabled(self) -> bool:
+        """
+        :returns: True if a message page exists and is set to be published, False otherwise
+        """
+        message_page = Page.objects.filter(application_namespace="inbox").first()
```
vaszig, line 60: Let's use `application_namespace=InboxApphook.app_name` to be sure. 
I would add `publisher_is_draft=False` to the filter in order to get only the non-draft here.

https://docs.django-cms.org/en/latest/topics/publishing.html

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -116,7 +116,9 @@ <h2 class="h2">{% trans "U bent toegevoegd als contactpersoon" %}</h2>
                         <td class="table__item">{{contact.email}}</td>
                         <td class="table__item"></td>
                         <td class="table__item">{% icon "close" extra_classes="icon icon--danger" %}</td>
-                        <td class="table__item">{% button text=_('Stuur bericht') icon="message" icon_position="before" icon_outlined=True transparent=True disabled=True %}</td>
+			{% if not messages_enabled %}
```
vaszig, line 119: I think it doesn't make sense to show a disabled button if the functionality is also disabled (for both the accepted and not users). It's not something that the user can have even if his invitation is accepted. @alextreme can tell us his opinion as well.

If we do show the disabled button, we have to fix the styling.

image

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -85,10 +86,40 @@ def test_contact_filter_without_any_contacts(self):
...
+        self.assertNotIn("Stuur bericht", response)
+
+        # case 3: published message page
+        page.publish("nl")
+        page.save()
+
+        response = self.app.get(self.list_url, user=self.user)
+
+        html = response.content.decode("utf-8")
+        doc = PyQuery(html)
```
vaszig, line 118: FYI you can find-access the anchor tag by `response.pyquery("a[title='Stuur bericht']")`.

vaszig (changes requested):

src/open_inwoner/templates/pages/profile/contacts/list.html
alextreme: Agreed, I think I mentioned earlier to not show this button if mijn berichten is disabled

src/open_inwoner/accounts/tests/test_contact_views.py
pi-sigma: I forgot that you can directly use pyquery on the response object. Nice!

src/open_inwoner/cms/profile/cms_appconfig.py
pi-sigma: Makes sense. I moved the function to a separate utils module as it is the same for the profile and contacts view.

vaszig (changes requested):

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -85,10 +85,38 @@ def test_contact_filter_without_any_contacts(self):
...
-    def test_contact_list_show_link_to_messages(self):
-        message_link = reverse("inbox:index", kwargs={"uuid": self.contact.uuid})
+    def test_messages_enabled_disabled(self):
+        """Assert that `Stuur Bericht` is displayed if and only if the message page is published"""
+
+        # case 1: no message page
         response = self.app.get(self.list_url, user=self.user)
-        self.assertContains(response, message_link)
+
+        self.assertNotIn("Stuur bericht", response)
```
vaszig, line 94: Should be translatable.

vaszig (changes requested):
I am sorry but I now noticed that the asserts you are doing concerning the texts in the response are not translatable. There are several across the PR (I mentioned only the first one), let's fix these as well.

src/open_inwoner/cms/utils/page_display.py
```diff
@@ -0,0 +1,45 @@
+"""Utilities for determining whether CMS pages are published"""
+
+
+from cms.models import Page
+
+from open_inwoner.cms.cases.cms_apps import CasesApphook
+from open_inwoner.cms.collaborate.cms_apps import CollaborateApphook
+from open_inwoner.cms.inbox.cms_apps import InboxApphook
+
+cms_apps = {
```
vaszig, line 10: Since you did this for some other apphooks apart from inbox (which has to do with the specific PR), maybe you should add the remaining apphooks as well (products, profile etc.).

And I am not sure if we do need separate-additional functions for each apphook. Maybe a bit different name of the main function `_is_published` could be a solution in order not to confuse this with cms's Page model method.

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -85,10 +85,40 @@ def test_contact_filter_without_any_contacts(self):
...
-        message_link = reverse("inbox:index", kwargs={"uuid": self.contact.uuid})
+    def test_messages_enabled_disabled(self):
+        """Assert that `Stuur Bericht` is displayed if and only if the message page is published"""
+
+        # case 1: no message page
         response = self.app.get(self.list_url, user=self.user)
-        self.assertContains(response, message_link)
+
+        # self.assertNotIn(_("Stuur bericht"), response)
+        self.assertNotContains(response, _("Stuur bericht"))
```
vaszig, line 95: Thanks for the translations!

vaszig (changes requested):

src/open_inwoner/templates/pages/profile/contacts/list.html
```diff
@@ -103,8 +103,10 @@ <h2 class="h2">{% trans "U bent toegevoegd als contactpersoon" %}</h2>
                         <td class="table__item">{{invite.invitee_email}}</td>
                         <td class="table__item"></td>
                         <td class="table__item">{% icon "close" extra_classes="icon icon--danger" %}</td>
-                        <td class="table__item">{% button text=_('Stuur bericht') icon="message" icon_position="before" icon_outlined=True transparent=True disabled=True %}</td>
-                        <td class="table__item">{% button icon="settings" icon_position="before" text="" hide_text=True transparent=True disabled=True %}<td>
+			{% if inbox_page_is_published %}
+			    <td class="table__item">{% button text=_('Stuur bericht') icon="message" icon_position="before" icon_outlined=True transparent=True disabled=True %}</td>
```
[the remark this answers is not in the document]
pi-sigma: Thanks for pointing this out. It seems ugly to warp the entire `td` element in an `if/else` because of `colspan` (though it works). Is there a more elegant solution?

src/open_inwoner/cms/utils/page_display.py
pi-sigma: I'll update the module for the other apphooks.

The idea was to use a single private function for the implementation and several public functions for the interface. `_is_published` should never be used outside these public functions. I deliberately used a name that's close to the naming used in the CMS Page model. Do you still see the danger of confusion here? In that case I can rename it.

src/open_inwoner/templates/pages/profile/contacts/list.html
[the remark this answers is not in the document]
vaszig: We already use a lot the if statements in the templates and I don't think it's a problem to do it here as well. We can do something like this instead of wrapping the entire element:
`<td class="table__item"  {% if inbox_page_is_published %}colspan="2" {% else %}colspan="1"{% endif %}`

Another solution would be to add a variable to the `get_context_data` method of the view. It's up to you!

src/open_inwoner/cms/utils/page_display.py
vaszig: - Yes you are right, no need for adding the rest of the apphooks for now.

- There is no need to rename the function, this would be good to do it if we used only one function. Basically, I just wanted to understand why we created a function for each apphook separately instead of just passing the str to a single function. But you have explained this to me so you can keep it the way you have it.

src/open_inwoner/templates/pages/profile/contacts/list.html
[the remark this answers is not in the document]
pi-sigma: Thanks for the suggestion!

vaszig (approved):

alextreme (approved):

PR #656, opened 2023-06-07
[#1503] Refactored name validation

pi-sigma:
- refactored name validation in the user model, contact form, and case upload form, replacing our custom `validate_charfield_entry` with a `NameValidator` that inherits from Django's inbuilt `RegexValidator`. 
- added HTML escape in the contact drop down


pi-sigma:
Another point: while the code prevents execution of HTML, it doesn't address the question why the HTML isn't automatically escaped by Django, as it normally is.

vaszig:
> @vaszig I thought about removing `validate_charfield_entry` entirely and replacing it with `NameValidator`. On the other hand, the function is used in various places (see the failing tests which I overlooked). Secondly, `validate_charfield_entry` does reject / so prevents the execution of JavaScript. I still prefer the approach with a class-based RegexValidator since it is more idiomatic (?), but I'd like to hear what you think before I make any further changes.

I think I would go for the class validator since we need to avoid some additional characters. But each solution needs the tests to be modified, so we don't avoid this. @Bartvaderkin can you tell us your opinion please?

src/open_inwoner/js/components/autocomplete/index.js
```diff
@@ -2,8 +2,14 @@ import autoComplete from '@tarekraafat/autocomplete.js/dist/autoComplete'
...
 const autocompleteField = (node) => {
   const choices = JSON.parse(
-    node.dataset.autocompleteChoices.replace(/'/g, '"')
+    node.dataset.autocompleteChoices
+      .replace(/'/g, '"')
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;')
   )
+  console.log('node', node)
```
vaszig, line 11: Do we need them?

vaszig (commented):
A small remark concerning the name of the validator..

I see that we are now using the `NameValidator` class for several fields in the User model and the name of the validator is not really representative. Maybe something that declares it's for the charfields?

Concerning escaping characters with js, @Bartvaderkin should review this as well.

pi-sigma:
@Bartvaderkin HTML escape in the dropdown is fixed. However, I'm not quite sure whether this was simply an oversight on our part or whether escaping is the responsibility of the library we're using for autocompletion (JavaScript type-ahead widgets are a known source of XSS-vulnarability in Django but I'm not sure about the present case).

Bartvaderkin (changes requested):

src/open_inwoner/js/components/autocomplete/index.js
```diff
@@ -1,15 +1,14 @@
...
 
 const autocompleteField = (node) => {
-  const choices = JSON.parse(
-    node.dataset.autocompleteChoices
-      .replace(/'/g, '"')
-      .replace(/&/g, '&amp;')
-      .replace(/</g, '&lt;')
-      .replace(/>/g, '&gt;')
+  let choices = JSON.parse(
+    document.getElementById('autocomplete-choices').textContent
```
Bartvaderkin, line 6: See the note in the template.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

vaszig (approved):

PR #658, opened 2023-06-07
[#1519] remove files from profile sidebar if files are empty

pi-sigma:
Related: #622 

Taiga: #1519


jiromaykin (approved):
Looks and behaves good - Note: we hardly use the 'issues' branch - usually something is a 'fix' or part of our 'feature' branche.

PR #659, opened 2023-06-08
:lipstick: [#1498] Add missing icons to plan-edit actions view

jiromaykin:
Make the edit-plan/add-Action look like: 
https://projects.invisionapp.com/share/XP134RWRK3MG#/screens/471522253 
Also there is another page that has missing icons (Start nieuwe samenwerking):
https://projects.invisionapp.com/share/XP134RWRK3MG#/screens/471522250 

Note: due to adding extra class I had to change indentation of the templates.

### DIFF
--- a/src/open_inwoner/scss/components/Form/DocumentUpload.scss
+++ b/src/open_inwoner/scss/components/Form/DocumentUpload.scss
@@ -43,13 +43,18 @@ input[type='file'] {
   color: var(--color-primary-lighter);
   border: 1px solid var(--color-mute);
   border-radius: var(--border-radius);
-  display: flex;
-  flex-direction: column;
-  min-width: var(--mobile-xs-width);
+  display: inline-block;
+  min-width: 100%;
   max-width: 500px;
   padding: 0;
   overflow: hidden;
   cursor: pointer;
+
+  @media (min-width: 768px) {
+    display: flex;
+    flex-direction: column;
+    min-width: 500px;
+  }
 }
 
 /// File-input button site wide
@@ -69,7 +74,11 @@ input[type='file']::file-selector-button {
   white-space: nowrap;
   padding: 0 12px 0 12px;
   margin: 0 12px 0 0;
-  width: 200px;
+  width: 150px;
+
+  @media (min-width: 768px) {
+    width: 200px;
+  }
 }
 
 .button-row .upload-button--disabled {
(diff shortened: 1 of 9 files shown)


src/open_inwoner/scss/components/Form/DocumentUpload.scss
```diff
@@ -43,13 +43,18 @@ input[type='file'] {
...
-  flex-direction: column;
-  min-width: var(--mobile-xs-width);
+  display: inline-block;
+  min-width: 100%;
   max-width: 500px;
   padding: 0;
   overflow: hidden;
   cursor: pointer;
+
+  @media (min-width: 768px) {
```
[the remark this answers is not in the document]
jiromaykin: @pi-sigma Thanks :-) yes, i will add a new variable - we will need it more for certain designs.

src/open_inwoner/templates/pages/profile/actions/edit.html
```diff
@@ -3,27 +3,29 @@
...
 
 {% block content %}
-<h1 class="h1">
-    {% if object.pk %}
-        {% trans "Wijzig actie" %}
-    {% else %}
-        {% trans "Creeer actie" %}
-    {% endif %}
-</h1>
+    <div class="plan-edit">
```
pi-sigma, line 6: Do we need the indent on the first element? It seems that we don't have a consistent policy on this. Sometimes the first element after a `block` tag is indented, sometimes it isn't.

vaszig: The indentation in the whole file looks good to me (for the block content and the div)

vaszig (commented):
I think the `expand-more` icon should follow the `onderwerpen` styling and should change direction when we click it, but I am not sure what's the right approach here.

jiromaykin:
> I think the `expand-more` icon should follow the `onderwerpen` styling and should change direction when we click it, but I am not sure what's the right approach here.

@vaszig I disagree slightly, the Onderwerpen is a different type of interface element than a select, same goes for the FAQ accordion;
if we wish to change all Select elements (which isn't in the design) that should be a new issue.

vaszig:
> @vaszig I disagree slightly, the Onderwerpen is a different type of interface element than a select, same goes for the FAQ accordion; if we wish to change all Select elements (which isn't in the design) that should be a new issue.

You know better!It was just a thought and mostly for my better understanding.

pi-sigma (approved):
Looking good! I've left a few remarks but they are tangential to the PR since they concern issues that exist in the original version as well.

PR #660, opened 2023-06-08
:zap:  [#1485] Hide extra expiration warning before DigiD session ends

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1485

**_Note: can only be reproduced when logging in as a DigiD user!
and only works when clicking the buttons - not by clicking outside modal_** so clicking outside the buttons needs to be removed.

Explanation: clicking outside the modal, closes it, but also bypasses any necessary Confirm/Submit.

1. There is one warning that needs to stay (14 minutes in - but it has to disappear after 15 minutes, which it doesn't right now):
Screenshot 2023-06-08 at 11 50 26

2. This one needs to stay - it appears after 15 minutes of inactivity - if this one is clicked, it should redirect to the LOGIN page, not to the Home page:
Screenshot 2023-06-08 at 11 47 02

But if you keep on having inactivity after more than 15 minutes the first modal needs to be invisible (the first one pops up after clicking the second one again, should not happen).

### DIFF
--- a/src/open_inwoner/js/components/session/index.js
+++ b/src/open_inwoner/js/components/session/index.js
@@ -33,21 +33,17 @@ class SessionTimeout {
       this.warnTime * 1000
     )
     this.expiredTimeout = setTimeout(
-      this.showExpiredModal,
+      this.showExpiredModal.bind(this),
       (this.expiryAge + 1) * 1000
     )
   }
 
   setDataset() {
-    console.log('setDataset')
     this.expiryAge = parseInt(this.element.dataset.expiryAge)
     this.warnTime = parseInt(this.element.dataset.warnTime)
-    console.log('this.expiryAge', this.expiryAge)
-    console.log('this.warnTime', this.warnTime)
   }
 
   showWarningModal() {
-    console.log('showWarningModal')
     if (this.userActive) {
       this.restartSession()
       return
@@ -113,6 +109,7 @@ class SessionTimeout {
       this.restartNoActivity.bind(this),
       30 * 1000
     )
+    // SESSION_COOKIE_AGE in seconds - (minus) 30 = warnTime
   }
 
   restartNoActivity() {
--- a/src/open_inwoner/accounts/tests/test_auth.py
+++ b/src/open_inwoner/accounts/tests/test_auth.py
@@ -1265,7 +1265,7 @@ class TestLoginLogoutFunctionality(AssertRedirectsMixin, WebTest):
         """Test that a user is able to log out and page redirects to root endpoint."""
         # Log out user and redirection
         logout_response = self.app.get(reverse("logout"), user=self.user)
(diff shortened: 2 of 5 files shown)


jiromaykin:
@alextreme @Bartvaderkin Met deze heb ik erg lang geworsteld en om onbegrijpelijke redenen lijkt dit nu wel te werken met gewenste effect: dus 1) de redirect voert naar de inlog-pagina i.p.v. home, en 2) de Warning modal lijkt niet meer te blijven hangen wanneer de Expiration modal verschijnt - dubbelchecken nodig.

src/open_inwoner/js/components/session/index.js
```diff
@@ -28,22 +28,21 @@ class SessionTimeout {
...
-      this.showExpiredModal,
-      (this.expiryAge + 1) * 1000
-    )
+    this.warningTimeout = setTimeout(() => {
+      this.showWarningModal.bind(this)()
+    }, this.warnTime * 1000)
+
+    this.expiredTimeout = setTimeout(() => {
+      this.showExpiredModal()
+    }, (this.expiryAge + 1) * 1000)
```
Bartvaderkin, line 37: (we discussed use of bind() and arrow functions in person)

Bartvaderkin (commented):
@jiromaykin Note both of the changed parts do the same thing in two ways:

- the first one uses an arrow function and no bind() (and it works because the arrow function preserves `this` so we can call the method on it).
- the second one just bind()'s without arrow function  (and it works because we explicitly bind `this` to the function reference so it has an explicit `this`).

jiromaykin:
> ...... the second one just bind()'s without arrow function  (and it works because we explicitly bind `this` to the function reference so it has an explicit `this`).

Yes @Bartvaderkin so this should be correct now? There are multiple things not quite right with the Modal from which this is constructed - the time reset doesn't work for the Session modal when a user is typing for example, and the second modal that pops up after clicking, doesn't behave correctly when clicking outside of the button - i will need to rebuild this completely and put it back into draft.

jiromaykin:
@alextreme @pi-sigma This one could use a thorough testing (not part of the sprint).

pi-sigma (changes requested):
I no longer see the `Uncaught DOMExceptio` I reported earlier. On the other hand, I'm getting `Uncaught TypeError: this.configureModal is not a function at showExpiredModal` as soon as the session expires. Needs another look

jiromaykin:
> I no longer see the `Uncaught DOMExceptio` I reported earlier. On the other hand, I'm getting `Uncaught TypeError: this.configureModal is not a function at showExpiredModal` as soon as the session expires. Needs another look

@pi-sigma I think you need to do a proper `Git Pull` and `NPM run watch` - because I do not see these errors anymore.

pi-sigma (approved):

PR #645, opened 2023-06-02
[#1484] Changing title and adding explaination to the profile-deactiv…

alextreme:
…ation modal popup

PR #663, opened 2023-06-08
:bug: [#1523] Set empty confirmation text for file-deletion modal

jiromaykin:
If the text is not set or if it is an empty string - do not make it render as "undefined" but rather as an empty string in the modal.
Screenshot 2023-06-08 at 09 52 32

### DIFF
--- a/src/open_inwoner/js/components/confirmation/index.js
+++ b/src/open_inwoner/js/components/confirmation/index.js
@@ -13,7 +13,8 @@ class Confirmation {
       const modalId = document.getElementById('modal')
       const modal = new Modal(modalId)
       modal.setTitle(this.form.dataset.confirmTitle)
-      modal.setText(this.form.dataset.confirmText)
+      // Only show confirmation if text is set
+      modal.setText(this.form.dataset.confirmText || '')
       modal.setClose(this.form.dataset.confirmCancel)
       modal.setConfirm(
         this.form.dataset.confirmDefault,


Bartvaderkin (changes requested):
Is the right place to fix this? I'd assume this would be something for the Javascript? Like why is it even trying to print an undefined value?

jiromaykin:
> Is the right place to fix this? I'd assume this would be something for the Javascript? Like why is it even trying to print an undefined value?

@Bartvaderkin In some cases there is a description but in other cases there is not - it's.a new attribute coming from here: https://github.com/maykinmedia/open-inwoner/pull/645 
But of course I can see if I can make a JS conditional then.

Bartvaderkin:
@jiromaykin Sure, but if there is no description (because there is no data attribute) then it shouldn't just print something non-sensible. (yes I know javascript is annoying with undefined values)

src/open_inwoner/js/components/confirmation/index.js
```diff
@@ -13,7 +13,10 @@ class Confirmation {
       const modalId = document.getElementById('modal')
       const modal = new Modal(modalId)
       modal.setTitle(this.form.dataset.confirmTitle)
-      modal.setText(this.form.dataset.confirmText)
+
+      if (modal.setText === true) {
+        modal.setText(this.form.dataset.confirmText)
+      }
```
Bartvaderkin, line 19: Hmm :thinking: I think you meant:

````js
if (this.form.dataset.confirmText) {
    modal.setText(this.form.dataset.confirmText)
}
// or maybe just
modal.setText((this.form.dataset.confirmText || ""))
````

Bartvaderkin (changes requested):

src/open_inwoner/js/components/confirmation/index.js
jiromaykin: Oooooh @Bartvaderkin - the shorthand version looks neat!

Bartvaderkin (approved):

src/open_inwoner/js/components/confirmation/index.js
Bartvaderkin: @jiromaykin note as javascript shorthands goes it has edge cases, like you could never print the number 0.

jiromaykin: @Bartvaderkin Thanks, I am assuming in case of strings we will never have zero, but I will remember this for future oddities.

PR #666, opened 2023-06-09
[#1439] Implemented playwright tests for cases htmx view

Bartvaderkin:


Bartvaderkin:
@vaszig I've marked it as ready for review, but only if it's green.

vaszig (approved):

PR #667, opened 2023-06-12
:lipstick: [#1529] Improve contact-form styles

jiromaykin:
Contact form needs icon in select + different placeholder text, and styles for textarea.
https://projects.invisionapp.com/share/UF134RWMWCK8#/screens/471596624
In order to just change the Select styles, I needed to break open the form and turn it into a render_form.

Side Note: all Selects in the entire projects should 'actually' have a muted style, but it is impossible to do this with CSS if all fields have a value AND are not 'required' (or else all of the Options inside the Select will look muted) - **only required and invalid values can be targeted with CSS across most common browsers**.
"...The problem is that these elements have very different default looks across browsers, and while you can style them in some ways, some parts of their internals are literally impossible to style..." (new issue nr.#1556)

### DIFF
--- a/src/open_inwoner/components/templates/components/Contact/ContactForm.html
+++ b/src/open_inwoner/components/templates/components/Contact/ContactForm.html
@@ -1,9 +1,35 @@
-{% load i18n form_tags %}
+{% load i18n form_tags grid_tags %}
 
-<div class="contactform__content">
-    {% if has_form_configuration %}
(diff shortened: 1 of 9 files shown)


jiromaykin:
@Bartvaderkin it seems the "ContactFormTestCase" test is faling - wondering which field causes this problem?

vaszig (approved):

PR #668, opened 2023-06-12
:lipstick: [#1528] Make cards responsive for very large desktop screens

jiromaykin:
for screens wider than 1600 pixels set grid margins in such a way it will always show 4 cards next to each other.
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1528

Note: **unfortunately it is not possible to use css variables INSIDE a media-query screen-width declaration:
"...The var() function can not be used as property names, selectors, or anything else besides property values. ..."**

Some standard screensizes to test this with:

1600 x 900 High Definition Plus (HD+)
1920 x 1080 Full High Definition (FHD)
1920 x 1200 Wide Ultra Extended Graphics Array (WUXGA)
2560 x 1440 Quad High Definition (QHD)
3440 x 1440 Wide Quad High Definition (WQHD)
3840 x 2160 4K or Ultra High Definition (UHD)

### DIFF
--- a/src/open_inwoner/scss/components/CardContainer/CardContainer.scss
+++ b/src/open_inwoner/scss/components/CardContainer/CardContainer.scss
@@ -18,7 +18,17 @@
     }
 
     &--columns-4 {
-      grid-template-columns: repeat(auto-fit, 228px);
+      grid-template-columns: repeat(
+        auto-fit,
+        calc(230px - var(--spacing-small))
+      );
+
+      @media (min-width: 1600px) {
+        grid-template-columns: repeat(
+          auto-fit,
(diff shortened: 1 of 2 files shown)


src/open_inwoner/scss/views/App.scss
```diff
@@ -3,6 +3,7 @@
 
   --container-width: 1024px;
   --mobile-s-width: 768px;
+  --desktop-hd-width: 1600px;
```
vaszig, line 6: This can be removed right? Or is it there in order to have an order with all kinds of resolutions?

vaszig (changes requested):

vaszig (approved):

PR #669, opened 2023-06-13
[#1458] Update klant from profile form data

Bartvaderkin:


vaszig (commented):
What about the "roepnaam"?Is it missing because of e-suite?

Bartvaderkin:
@vaszig the ticket was spurious, as 'roepnaam' doesn't exist in the Klanten API :laughing:

vaszig (approved):

PR #670, opened 2023-06-13
[#1483] Update profile email/phone from klant on login

Bartvaderkin:
Also improved the logging for the earlier contact-api system logs. And cleaned up some f-strings from earlier.


src/open_inwoner/utils/tests/helpers.py
```diff
@@ -77,6 +82,9 @@ def getTimelineLogDump(self) -> str:
     def dumpTimelineLog(self):
         print(self.getTimelineLogDump())
 
+    def resetTimelineLogs(self):
```
vaszig, line 85: Useful!

vaszig (approved):

PR #671, opened 2023-06-13
[#1455] Add contact form to case detail page

vaszig:


alextreme:
Merging this in without review yet. @Bartvaderkin please double-check this tomorrow, but this gives Vasileios the option to double-check on the testenvironment.

If there are blocking requests I'll revert the commit and skip this part in the demo, if there are requests for improvement please create a separate task for these so Vasileios can clean them up.

jiromaykin:
@alextreme I did not have the chance to review this on my part-time day off :-) but I'd like to review it when needed.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -290,74 +388,112 @@ def handle_document_upload(self, request, form):
...
+            "medewerkerIdentificatie": {
+                "identificatie": config.register_employee_id,
+            },
+        }
 
-        # check if this info_object should be visible
-        config = OpenZaakConfig.get_solo()
-        if not is_info_object_visible(info_object, config.document_max_confidentiality):
-            raise PermissionDenied()
+        contactmoment = create_contactmoment(data, klant=klant)
```
Bartvaderkin, line 478: I still wonder about the situation when we don't get a klant and cannot create one either, in which case we'd loose the connection between the person and the contactmoment. We probably discussed this?

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -280,7 +373,12 @@ def handle_document_upload(self, request, form):
                         filename=file.name
                     ),
                 )
-                return HttpResponseRedirect(self.get_success_url())
+
+                return HttpResponseClientRedirect(
+                    reverse(
+                        "cases:case_detail", kwargs={"object_id": str(self.case.uuid)}
+                    )
+                )
```
Bartvaderkin, line 381: I think all the `HttpResponseClientRedirect`'s should be a `redirect()` shortcut.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -290,74 +388,112 @@ def handle_document_upload(self, request, form):
...
-
-    def handle_no_permission(self):
-        # plain error and no redirect
-        raise PermissionDenied()
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["contact_form"] = self.get_form()
+        context["hxpost_contact_form"] = reverse(
+            "cases:case_detail_contact_form", kwargs=self.kwargs
+        )
```
Bartvaderkin, line 498: Not 100% sure the naming is great: 

If `contact_form` has a `_form` postfix because it holds a form, then `hxpost_contact_form` suggests it is also a form? 

Same for `hxpost_document_form` earlier. Maybe it should be `_action` because it sort-of does what the regular form action attributes does?

Bartvaderkin (commented):
@vaszig I added reviews notes, but it'll obviously have to be a new PR.

src/open_inwoner/cms/cases/views/status.py
Bartvaderkin: I was mistaken, this was a HTMX redirect that doesn't do reverse()

src/open_inwoner/cms/cases/tests/test_contactform.py
```diff
@@ -0,0 +1,287 @@
...
+    def test_form_is_shown_if_open_klant_configured(self, m):
+        self._setUpMocks(m)
+
+        self.assertTrue(self.ok_config.has_api_configuration())
+
+        response = self.app.get(self.case_detail_url, user=self.user)
+        contact_form = response.pyquery("#contact-form")
+
+        self.assertTrue(response.context["case"]["contact_moments_enabled"])
+        self.assertTrue(contact_form)
```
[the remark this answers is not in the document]
vaszig: This will be added when task [https://taiga.maykinmedia.nl/project/open-inwoner/task/1566](url) will be picked up

src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -35,10 +42,7 @@
 @override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
 class CasesPlaywrightTests(ClearCachesMixin, PlaywrightSyncLiveServerTestCase):
```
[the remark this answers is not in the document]
vaszig: Issue https://taiga.maykinmedia.nl/project/open-inwoner/issue/1565 was created by Bart for this.

PR #673, opened 2023-06-14
[#1399] Ensuring Onderwerpen are sorted based on the configured order…

alextreme:
… in the admin


alextreme:
You solved the tree-ordering issue, however the order of Categories shown via the CMS plugin was still alphabetical instead of by path :see_no_evil:

vaszig (approved):

PR #674, opened 2023-06-15
[#1460] refactor Single Sign On

pi-sigma:
Added option to restrict login via OpenId Connect to admin users.

Taiga: #1460


src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -3,8 +3,12 @@
 from django.contrib.auth import get_user_model
 from django.test import TestCase
 from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
```
Bartvaderkin, line 6: If we don't have module level translatables we can use the normal, non-lazy version.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -207,3 +212,44 @@ def test_error_first_cleared_after_succesful_login(
...
+    "mozilla_django_oidc_db.mixins.OpenIDConnectConfig.get_solo",
+    return_value=OpenIDConnectConfig(enabled=True),
+)
+class OIDCConfigTest(WebTest):
+    def test_admin_only_enabled(self, mock_oidc_config):
+        """Assert that the OIDC login option is only displayed for login via admin"""
+
+        config = SiteConfiguration.get_solo()
+        config.openid_admin_only = True
+        config.save()
```
Bartvaderkin, line 227: Let's pick one of these two patterns to modify the solo objects for testing. So either @patch both or put both in regular code (like SiteConfiguration is now).

If we do the latter then the test case also needs the ClearCachesMixin, or we'll get unstable results because the solo objects are cached in the cache framework (and Django for some reasons doesn't isolate the caches the way it does the database, hence the custom mixin).

src/open_inwoner/configurations/models.py
```diff
@@ -396,6 +396,11 @@ class SiteConfiguration(SingletonModel):
             "The text that should display when OpenId connect is set as a login method"
         ),
     )
+    openid_admin_only = models.BooleanField(
```
Bartvaderkin, line 399: I'm not 100% sure if the is the most accurate name and field for what this does (conceptually if an 'only' boolean is false then you'd expect it to mean 'both' (and not 'the other' as in this case).

It could be a choicefield with choices for 'admins' and 'users' (with default 'admins', and in the future we could add another choice for both 'admins and users'.

Then for convenience usage in the template add two `@property's on the model, something like `openid_enabled_for_admin` and `openid_enabled_for_users` and use those in the templates.

Bartvaderkin (changes requested):

Bartvaderkin:
@pi-sigma Oops, apparently Github still submitted the Outdated commentsI copied to not be Outdated (I deleted them now)

src/open_inwoner/configurations/models.py
```diff
@@ -452,6 +453,14 @@ def matomo_enabled(self):
...
         return bool(self.siteimprove_id)
 
+    @property
+    def openid_enabled_for_admin(self):
+        return self.openid_display == "admin"
+
+    @property
+    def openid_enabled_for_regular_users(self):
+        return self.openid_display == "regular"
+
```
Bartvaderkin, line 463: These should compare to the choices on the class (eg: `OpenIdDisplayChoices.admin` instead of their literal string value).

Bartvaderkin (changes requested):

src/open_inwoner/templates/admin/login.html
```diff
@@ -0,0 +1,18 @@
+{% extends "admin/login.html" %}
+{% comment %} SPDX-License-Identifier: EUPL-1.2 {% endcomment %}
+{% comment %} Copyright (C) 2020 Dimpact {% endcomment %}
```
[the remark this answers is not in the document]
Bartvaderkin: @alextreme Do we need these copyright references?

Bartvaderkin (commented):
Looks good, only question is if we need those copyright references.

pi-sigma:
@alextreme Can you take a look at the copyright notice Bart mentioned (see comment above) and let me know should it stay or should it go?

src/open_inwoner/templates/admin/login.html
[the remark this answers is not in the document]
Bartvaderkin: @pi-sigma I asked Alex about this and we can remove these license comments.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #678, opened 2023-06-16
[#1546] Fix anchors (mobile) in cases detail page

vaszig:


src/open_inwoner/js/components/anchor-menu/anchor-menu-mobile.js
```diff
@@ -24,3 +24,13 @@ export class AnchorMobile {
...
+
+/**
+ * Controls the toggling of anchor list specifically for the cases-detail page
+ */
+export class AnchorMobileOOB extends AnchorMobile {}
+
+const anchorTogglesOob = document.querySelectorAll(AnchorMobileOOB.selector)
+;[...anchorTogglesOob].forEach(
+  (anchorToggleOob) => new AnchorMobileOOB(anchorToggleOob)
+)
```
Bartvaderkin, line 36: I'm not sure what is going on here: the AnchorMobileOOB inherits the selector from AnchorMobile so in this file you'd hit the same nodes twice. 

And in the hydration part this would do the same thing as before this PR (because it is the same code running from the same selector).

Bartvaderkin (changes requested):

Bartvaderkin:
Closing this as the issue was fixed in #679 

@vaszig The reason this PR seems to solve it was because of the multiple initialization added listeners multiple times and those interacted in a way they overrode each other.

PR #679, opened 2023-06-16
[#1546] Hotfixed broken anchor menu

Bartvaderkin:


vaszig (approved):
Thanks for the detailed explanation!

PR #681, opened 2023-06-19
:sparkles: [#1537] Adding cookie-consent banner

jiromaykin:
Cookie-banner that respects the GDPR privacy laws, and gives the user the chance to dismiss the non-essential cookies only.
https://taiga.maykinmedia.nl/project/open-inwoner/task/1537

IF the Admin does NOT want to use the cookiebanner, THEN analytics cookies need to be set to true.
IF Admin wants to use the banner, THEN analytics need to be set to false when cookies are rejected by the local user, and to true when approved. + Analytics must be on by default, unless there is a cookiebanner set.

- [x] style modal pop-up to be open onLoad
- [x] make cookiebanner optional in config (not visible)
- [x] add default Admin configuration texts (Scroll down to **Cookie consent** in http://127.0.0.1:8000/admin/configurations/siteconfiguration/)
- [x] reject all or accept all approve/dismiss
- [x] remember user choice across context/all pages for 90days(?) - this means the cookiebanner won't show for 90 days when it is clicked, but it will come back again due to the expirationTime in JS
- [x] make reject script actually reject ALL from Google/Matomo/Siteimprove
- [x] make the cookiebanner work for **failing tests** (error "intercepts pointer events... retrying click action" => ERROR: test_action_status (failing tests: 1. file: open_inwoner/accounts/tests/test_action_views.py, 2. test: open_inwoner/accounts/tests/test_action_views.py 'test_action_status' 3. file: open_inwoner/cms/cases/tests/test_htmx.py 4. htmx: test_cases, 5. file open_inwoner/plans/tests/test_views.py, 6. test "test_action_status")
- [x] add migrations

Challenge: (not in this issue) not making the cookie-banner appear on the Privacy policy page? Right now you first have to click it it in order to be able to read it; (if the cookie-banner is shown, it has to block all other interactions).


jiromaykin:
@Bartvaderkin and @pi-sigma I don't know too much writing tests yet (so a review of what I did in `accounts/tests/test_action_views.py` and in `cms/cases/tests/test_htmx.py` is much appreciated) + and I am not sure if my way of adding context-processors (in order to configure the banner from the admin) is the correct way to do it.... so feel free to leave feedback. for improvement.

src/open_inwoner/accounts/tests/test_action_views.py
```diff
@@ -327,10 +327,19 @@ def setUp(self) -> None:
...
 
+        # find the decline button inside the cookie-banner
+        cookiebanner_banner = page.locator(".cookie-banner")
+        cookiebanner_rejectbutton = page.locator(".reject-button")
+        # close cookie-banner
+        cookiebanner_rejectbutton.click()
+        # check cookie-banner state
+        expect(cookiebanner_banner).to_be_hidden()
+
         # find action element and the dropdown widget
```
pi-sigma, line 343: 1. The test is for the action view, which has nothing specifically to do with cookies. This would need to be a separate test.
2. If we're going to test the hiding of the cookie banner, we should also test that it's hidden when the 'accept' button is clicked.

However, I'm not sure this is even needed. We can confirm manually that the banner is hidden on click. What we do need to check automatically is that the analytics scripts are not rendered when the cookie is not set. You should modify the test in `src/open_inwoner/configurations/tests/test_analytics.py` (I should have done this myself when I changed the test, but didn't think of it back then).

src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -348,10 +348,19 @@ def test_cases(self, m):
...
 
+        # find the decline button inside the cookie-banner
+        cookiebanner_banner = page.locator(".cookie-banner")
+        cookiebanner_rejectbutton = page.locator(".reject-button")
+        # close cookie-banner
+        cookiebanner_rejectbutton.click()
+        # check cookie-banner state
+        expect(cookiebanner_banner).to_be_hidden()
+
         # expected anchors
```
pi-sigma, line 364: See previous comment

alextreme:
Discussed with Jiro and Paul:
- test needs to be added to check that the analytics aren't used if the cookies aren't accepted
- test_action_views needs to be modified, expect-changes about the cookiebanner should be removed (it should work without or be mocked)

alextreme:
3 failing tests:

2023-08-01T09:02:21.1027134Z ERROR: test_action_status (open_inwoner.accounts.tests.test_action_views.ActionsPlaywrightTests)
2023-08-01T09:02:21.1236333Z ERROR: test_cases (open_inwoner.cms.cases.tests.test_htmx.CasesPlaywrightTests)
2023-08-01T09:02:21.1397757Z ERROR: test_action_status (open_inwoner.plans.tests.test_views.PlanActionStatusPlaywrightTests)

src/open_inwoner/accounts/tests/test_action_views.py
```diff
@@ -325,7 +336,9 @@ def setUp(self) -> None:
         )
         self.action_list_url = reverse("profile:action_list")
 
-    def test_action_status(self):
+    def test_action_status(self, mock_solo):
```
jiromaykin, line 339: Nice! @pi-sigma I had no idea... now this is ready for review again.

alextreme (approved):

PR #684, opened 2023-06-20
[#1524] Make onderwerpen in contact form orderable

vaszig:


src/open_inwoner/openklant/migrations/0005_auto_20230621_1143.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.15 on 2023-06-21 09:43
+
+from django.db import migrations
+
+
+def populate_order_of_subjects(apps, schema_editor):
+    Subject = apps.get_model("openklant", "ContactFormSubject")
+    for subject in Subject.objects.all():
+        subject.order = subject.pk
+        subject.save()
```
Bartvaderkin, line 10: Technically this could have been an `Subject.objects.update(order=F('id'))` so it doesn't need to load all the records from the database to python and back, but it doesn't matter too much in this case.

Bartvaderkin (approved):
Cool. I've put a small note for the future.

PR #686, opened 2023-06-21
[#1560] Add linebreaks to contactmoment detail view

vaszig:


alextreme (approved):
To be double-checked on the testenv regarding other tables other than the contactmoment-details

PR #687, opened 2023-06-21
[#1553] Fix duplicate messages in contact form

vaszig:
When both email and api are activated the success/error message is shown twice.

### DIFF
--- a/src/open_inwoner/openklant/views/contactform.py
+++ b/src/open_inwoner/openklant/views/contactform.py
@@ -61,39 +61,23 @@ class ContactFormView(CommonPageMixin, LogMixin, BaseBreadcrumbMixin, FormView):
             return messages.add_message(
                 self.request, messages.SUCCESS, _("Vraag verstuurd!")
             )
-        return messages.add_message(
-            self.request,
-            messages.ERROR,
-            _("Probleem bij versturen van de vraag."),
-        )
+        else:
+            return messages.add_message(
+                self.request,
+                messages.ERROR,
+                _("Probleem bij versturen van de vraag."),
+            )
 
     def form_valid(self, form):
         config = OpenKlantConfig.get_solo()
+
+        success = True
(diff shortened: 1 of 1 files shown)


src/open_inwoner/openklant/views/contactform.py
```diff
@@ -60,12 +56,45 @@ def get_context_data(self, **kwargs):
...
+                self.get_result_message(success=True)
+            else:
+                self.get_result_message(success=False)
+
+        # both
+        elif config.register_email and config.register_contact_moment:
+            if registered_by_email and registered_by_api:
+                self.get_result_message(success=True)
+            else:
+                self.get_result_message(success=False)
```
Bartvaderkin, line 96: Can we simplify this? Something like:

````python
success = True
if config.register_email:
    success = self.register_by_email(form, config.register_email) and success
if config.register_contact_moment:
    success = self.register_by_api(form, config) and success

self.get_result_message(success=success)
````

**Review by Bartvaderkin — CHANGES_REQUESTED** [review]

File: src/open_inwoner/openklant/views/contactform.py [hunk 2]
```diff
@@ -60,12 +56,45 @@ def get_context_data(self, **kwargs):
         context["has_configuration"] = config.has_form_configuration()
         return context
 
+    def get_result_message(self, success=False):
+        if success:
+            return messages.add_message(
+                self.request, messages.SUCCESS, _("Vraag verstuurd!")
+            )
+        return messages.add_message(
```
Bartvaderkin, line 64: Can we put the second message in an else-block? So both parts have same indentation level.

src/open_inwoner/openklant/views/contactform.py
vaszig: Thanks for the nice and simple solution..

Bartvaderkin (approved):

PR #688, opened 2023-06-21
[#1547] Fix digid logout

pi-sigma:
Taiga: [#1547](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1547)

### DIFF
--- a/src/open_inwoner/accounts/middleware.py
+++ b/src/open_inwoner/accounts/middleware.py
@@ -29,15 +29,21 @@ class NecessaryFieldsMiddleware:
             # If the user is currently not editing their information, but it is required
             # redirect to that view.
 
-            # DigiD can be disabled, in which case the digid app isn't available
-            digid_logout = "/digid/logout/"
             try:
                 digid_logout = reverse("digid:logout")
-            except:  # nosec
-                pass
+                digid_slo_redirect = reverse("digid:slo-redirect")
+            except NoReverseMatch:
+                # temporary fix to make tests pass in case reverse fails
+                digid_logout = "/digid/logout/"
+                digid_slo_redirect = "/digid/slo/redirect/"
             if (
                 not request.path.startswith(
-                    (necessary_fields_url, reverse("logout"), digid_logout)
+                    (
+                        necessary_fields_url,
+                        reverse("logout"),
+                        digid_logout,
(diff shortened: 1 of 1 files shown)


src/open_inwoner/accounts/middleware.py
```diff
@@ -29,15 +29,16 @@ def __call__(self, request):
             # If the user is currently not editing their information, but it is required
             # redirect to that view.
 
-            # DigiD can be disabled, in which case the digid app isn't available
             digid_logout = "/digid/logout/"
-            try:
-                digid_logout = reverse("digid:logout")
-            except:  # nosec
-                pass
+            digid_slo_redirect = "/digid/slo/redirect/"
```
Bartvaderkin, line 33: These URL's need to use `reverse()` (eg: reverse('digid:logout') and reverse('digid:slo-redirect'), but in a try/except for a NoReverseMatch, in case the digid app is not loaded (like in development). 

Note the original code did this but was slightly messy from some earlier hotfix.

Bartvaderkin (changes requested):
Looks decent, my main question would be how confident are we these are the exact URL's we need to whitelist?

src/open_inwoner/accounts/middleware.py
```diff
@@ -29,18 +29,23 @@ def __call__(self, request):
...
-            if (
-                not request.path.startswith(
-                    (necessary_fields_url, reverse("logout"), digid_logout)
-                )
-                and request.user.require_necessary_fields()
-            ):
-                return HttpResponseRedirect(necessary_fields_url)
+                digid_slo_redirect = reverse("slo-redirect")
+            except NoReverseMatch:
+                return self.get_response(request)
```
Bartvaderkin, line 36: This could have been a `pass`, since there is a fall-through at the end of the method after the try/except/else.

But it is not a blocking matter so I approve the PR, and @pi-sigma can mark this as ready.

Bartvaderkin (dismissed):
Ready when you are.

Bartvaderkin:
@pi-sigma sadly CI is failing on some general check.

Bartvaderkin (approved):
Approving it to get it landed, I'm still looking at how to properly test this but it is complicated with the conditional URL pattern and some import/test/override issues

PR #690, opened 2023-06-22
[#1570] Speed up CI

pi-sigma:
Various tweaks to speed up the tests:
- run tests in parallel
- skip old data migration tests
- use MD5 for password hashing
- optimize postgres
- disable logging for WARNING and below


pi-sigma:
There are two issues with running our tests in parallel:

1. Elasticsearch uses persistent connections for their search client which doesn't play well with multiprocessing. Unfortunately, there is no public API to tweak things. One option would be to mess with the internals to ensure that each Django test worker is coupled with its own Elastic client. I've looked into this but decided not to pursue this avenue. The solution is bound to be messy and will likely break at some point. The second option was to split the test suite into two, run the "ordinary" tests in parallel, and do a second run for the tests that resist being run in parallel.

2. Our playwright tests don't play nice with multiprocessing either as Django can't serialize the relevant test classes. For now I simply lumped them together with the Elastic test cases, but we should look into the source of this eventually.

src/open_inwoner/conf/test.py
```diff
@@ -2,7 +2,12 @@
 Staging environment settings module.
 This *should* be nearly identical to production.
 """
+import logging
 import os
+import sys
+
+if len(sys.argv) > 1 and sys.argv[1] == "test":
+    logging.disable(logging.CRITICAL)
```
Bartvaderkin, line 11: I'm not sure if this is what we want. Supposedly we still want to see something when a real problem occurs?

src/open_inwoner/accounts/tests/test_actions_expire.py
```diff
@@ -7,6 +8,8 @@
 
 from open_inwoner.accounts.tests.factories import ActionFactory, UserFactory
 
+logging.disable(logging.CRITICAL)
```
Bartvaderkin, line 11: This line looks out of place.

Bartvaderkin (changes requested):

.coverage.fedora.1089256.271783
alextreme, line 1: can be removed?

src/open_inwoner/pdc/tests/test_data_migrations.py
```diff
@@ -1,3 +1,5 @@
+from unittest import skip
+
```
alextreme, line 2: I'd recomment to remove instead of skip these tests if they're no longer necessary

src/open_inwoner/conf/test.py
pi-sigma: I changed it so we keep logs for level >= `ERROR`

src/open_inwoner/search/tests/test_search_products_boost.py
```diff
@@ -7,6 +7,7 @@
 from .utils import ESMixin
 
 
+@tag("no-parallel")
```
Bartvaderkin, line 10: I don't see this tag used anywhere else.

src/open_inwoner/pdc/tests/test_data_migrations.py
```diff
@@ -1,106 +0,0 @@
-from django.contrib.gis.geos import Point
```
Bartvaderkin, line 1: Was there a specific reason we had to let these go?

If the amount of migrations is an issue we could look into squashing the migrations. IIRC it also had some complications with data migrations.

src/open_inwoner/conf/ci.py
```diff
@@ -13,7 +13,7 @@
     {
         "django": {
             "handlers": ["django"],
-            "level": "WARNING",
+            "level": "ERROR",
```
Bartvaderkin, line 16: I'm having doubts about the logging levels. 

Maybe we should move that to another issue en get the config changes and parallel test merged.

Bartvaderkin (changes requested):

src/open_inwoner/conf/ci.py
alextreme: I agree

src/open_inwoner/conf/test.py
```diff
@@ -2,8 +2,11 @@
 Staging environment settings module.
 This *should* be nearly identical to production.
 """
+import logging
 import os
 
+logging.disable(logging.WARNING)
```
Bartvaderkin, line 8: Lingering logging change we decided we would skip for now.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #692, opened 2023-06-23
[#1562] Update password policy

pi-sigma:
Added validator to check that passwords contain at least one upper-case letter, one lower-case letter, one digit.

Taiga: [#1562](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1562)

### DIFF
--- a/src/open_inwoner/accounts/tests/test_auth.py
+++ b/src/open_inwoner/accounts/tests/test_auth.py
@@ -74,11 +74,11 @@ class TestRegistrationFunctionality(WebTest):
 
     def test_registration_fails_with_invalid_first_name_characters(self):
         invalid_characters = '<>#/"\\,.:;'
+        register_page = self.app.get(reverse("django_registration_register"))
+        form = register_page.forms["registration-form"]
 
         for char in invalid_characters:
             with self.subTest(char=char):
-                register_page = self.app.get(reverse("django_registration_register"))
-                form = register_page.forms["registration-form"]
                 form["email"] = self.user.email
                 form["first_name"] = char
                 form["last_name"] = self.user.last_name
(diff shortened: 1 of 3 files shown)


pi-sigma:
I modified two related tests by moving the Get request and the form creation outside of the loops, which speeds up the tests a bit.

alextreme:
Would like to get this into the v1.8 release if possible

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -118,6 +118,29 @@ def test_registration_fails_with_invalid_last_name_characters(self):
                 }
                 self.assertEqual(response.context["form"].errors, expected_errors)
 
+    def test_registration_fails_with_non_diverse_password(self):
+        passwords = ["pass_word-123", "PASS_WORD-123", "NoDigits"]
```
Bartvaderkin, line 122: I miss a `'password123` and `'PASSWORD123'` (single case and digits)

src/open_inwoner/utils/validators.py
```diff
@@ -94,3 +94,20 @@ def __call__(self, image_id):
 
     def __eq__(self, other):
         return self.width == other.width and self.height == other.height
+
+
+class DiversityValidator:
+    def validate(self, password, user=None):
+        if password.isalpha() or password.isupper() or password.islower():
```
Bartvaderkin, line 101: This check is not correct and doesn't catch single case and digits (eg `"password123"`)

Bartvaderkin (changes requested):

src/open_inwoner/utils/validators.py
pi-sigma: `"password123".islower()` returns `True`. It's not obvious from the name of the method, but Python is smart enough to check only the letters.

Bartvaderkin: I'm sorry, I'm an idiot and didn't pay attention. 

I added `password123` to the test case and it failed, but it wasn't because this validator (but because the generic validator).

src/open_inwoner/accounts/tests/test_auth.py
pi-sigma: Both of these are rejected by the `CommonPasswordValidator`.

Bartvaderkin: You might have missed I edited the message, anyway let's add cases that fail this validator that have 

- lowercase and digits
- uppercase and digits
- only digits
- only lower
- only upper

without anything else a test reader needs to think about.

pi-sigma: Yeah, I missed the edit. Will incorporate the specified cases.

Bartvaderkin (approved):
Thanks, sorry for the confusion.

PR #693, opened 2023-06-23
[#1568] Process pr feedback of #1455

vaszig:
This is created in order to fix the problems according to the suggestions in #671


Bartvaderkin (approved):

Bartvaderkin:
@vaszig Thanks!

PR #695, opened 2023-06-23
[#1558] Add phonenumber validation

vaszig:


vaszig:
> I'm not 100% sure this is accurate and has usability:
> 
>     * Do we need the distinction for international numbers? Can we still enter local dutch numbers (eg: without country code) if a field has international validator?
> 
>     * Do we support input of humanized numbers? Eg, with dashes and spaces for readability? Many people write their numbers in groups (06 1234 5678 or 020 123 45 67 or 020-1234567)

If we want make use of international numbers (imo we maybe only need them in the contact form) we need the distinction. When we enter such a number we have to use the country code indeed. Should I remove it and use only dutch numbers? @alextreme what do you think?

alextreme:
I would recommend to only allow Dutch numbers (10 digits, no spaces, dashes etc). I'd prefer being too strict when it comes to data sent through to a secondary system.

vaszig:
Removed international phone validator and now there is only the dutch one. I also added a method for spaces and dashes.

src/open_inwoner/utils/validators.py
```diff
@@ -1,12 +1,75 @@
...
+
+class ParsePhoneNumber(Protocol):
+    def __call__(self, value: str) -> "PhoneNumber":
+        ...  # pragma: nocover
+
+
+@deconstructible
+class DutchPhoneNumberValidator:
+    country = "NL"
+    country_name = _("Dutch")
```
pi-sigma, line 25: "Dutch" is the name of the language. Either change the label to "language" or the field to _("Netherlands")

src/open_inwoner/utils/validators.py
```diff
@@ -1,12 +1,75 @@
...
+            raise ValidationError(
+                self.error_message,
+                params={"country": self.country_name},
+                code="invalid",
+            )
+
+        # this additional check is needed because while .parse() does some checks on country
+        #   is_possible_number() and is_valid_number() do not
+        #   eg: country=NL would accept "+442083661177"
+        if self.country:
```
pi-sigma, line 48: You specify a default above (`country = "NL"`), so why the conditional?

pi-sigma (approved):

PR #698, opened 2023-06-25
[#1571] Ensuring DigiD adminconfig doesn't throw a 500 when mock is enabled

alextreme:


vaszig (approved):
I assume we have to test this in acceptance, so I approve it.

alextreme:
It's the other way around actually, it should only be an issue if DigiD Mock is active (test and certain acceptance environments)

PR #700, opened 2023-06-26
[#1566] Add email registration to case detail contact form

vaszig:
FYI, currently we have the contact form in the case detail view but we only support sending a question via the api configuration. With this task we want to add the email configuration as well. 

It will be easier to review this if you take a look at [https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openklant/views/contactform.py](url), where Bart had already implemented this for the basic contact form page.

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -5,8 +5,7 @@ from typing import List
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.http import Http404, StreamingHttpResponse
-from django.shortcuts import redirect
-from django.urls import reverse, reverse_lazy
+from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from django.views import View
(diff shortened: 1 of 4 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -414,11 +414,26 @@ class CaseContactFormView(CaseAccessMixin, CaseLogMixin, FormView):
...
-            return self.register_by_api(form)
+            config = OpenKlantConfig.get_solo()
+
+            success = True
+            if config.register_email:
+                form.cleaned_data[
+                    "question"
+                ] += f"\n\nCase number: {self.case.identificatie}"
+                success = (
+                    self.register_by_email(form, config.register_email) and success
```
pi-sigma, line 427: Is `and success` necessary here, seeing that it is set to `True` at L421?

pi-sigma (approved):
Looking good (the suggested change is cosmetic).

src/open_inwoner/cms/cases/views/status.py
vaszig: We used this with Bart and if I remember right we did it for readability purposes. The variable is modified a lot of times so it is better for the user to understand what's going on.

PR #701, opened 2023-06-26
Fixed lingering problem in tests after merges of conflicting branches

Bartvaderkin:
Also upgraded the assertion of mock matchers called to a mixin


Bartvaderkin:
@vaszig Can you review and merge this so we get green builds again?

vaszig (approved):
I am sorry, I was waiting for the tests to pass and I forgot about it.

PR #705, opened 2023-06-29
:lipstick: [#1594] Improving onderwerp selection and dropdowns

jiromaykin:
- [x] set Categories 'interesses' desktop grid to 4 columns on desktop, and less for multiple small-screen media queries,
- [x] remove borders around dropdown icons and correct the names of certain icons named 'settings' to 'horizontal' dots.
- [x]  add  small triangle on top of border of dropdown content
- [x] make sure styling doesn't impact _everything everywhere all at once_
- [x] EXTRA: i also added correct modal background color: https://projects.invisionapp.com/share/XP134RWRK3MG#/screens/471522251

https://taiga.maykinmedia.nl/project/open-inwoner/issue/1594

Screenshot 2023-07-10 at 11 37 26

### DIFF
--- a/src/open_inwoner/components/templates/components/Action/Actions.html
+++ b/src/open_inwoner/components/templates/components/Action/Actions.html
@@ -20,7 +20,7 @@
                 <div class="actions__actions">
                     {% button icon="file-pdf" text=_("Export to PDF") hide_text=True href="profile:action_export" uuid=action.uuid icon_outlined=True transparent=True %}
                     {% if action|is_connected:request.user %}
-                        {% dropdown icon="settings" %}
+                        {% dropdown icon="more_horiz" %}
                             <div class="dropdown__item">
                                 {% get_action_edit_url action=action plan=plan as action_url %}
                                 {% button icon="edit" text=_("Bewerken") href=action_url icon_outlined=True transparent=True %}
--- a/src/open_inwoner/scss/components/Dropdown/Dropdown.scss
+++ b/src/open_inwoner/scss/components/Dropdown/Dropdown.scss
@@ -16,6 +16,34 @@
     background-color: var(--color-white);
     border: 1px solid var(--color-mute);
     z-index: 1;
+
+    &::before {
+      border-bottom: 6px solid var(--color-mute);
+      border-left: 6px solid var(--color-white);
+      border-right: 6px solid var(--color-white);
+      box-sizing: border-box;
+      content: '';
+      display: block;
+      height: 6px;
+      right: 0.4em;
+      position: absolute;
+      top: -7px;
+      width: 12px;
+    }
+
+    &::after {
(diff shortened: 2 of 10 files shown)


jiromaykin:
@pi-sigma This dropdown/category PR is easy to review and won't break the site when it's merged. I hope I have targeted all possible dropdowns with this one, but maybe I've missed one, or one of them might look off with the added little triangle; feel free to check on the front-end 
(needs NPM build and watch).

src/open_inwoner/scss/components/Dropdown/Dropdown.scss
```diff
@@ -16,6 +16,34 @@
...
+      border-right: 6px solid transparent;
+      box-sizing: border-box;
+      content: '';
+      display: block;
+      height: 6px;
+      right: 0.4em;
+      position: absolute;
+      top: -6px;
+      width: 11px;
+    }
```
pi-sigma, line 46: Why the 1px difference between before/after for `top`/`width`?

src/open_inwoner/components/templates/components/Action/Actions.html
```diff
@@ -20,7 +20,7 @@
                 <div class="actions__actions">
                     {% button icon="file-pdf" text=_("Export to PDF") hide_text=True href="profile:action_export" uuid=action.uuid icon_outlined=True transparent=True %}
                     {% if action|is_connected:request.user %}
-                        {% dropdown icon="settings" %}
+                        {% dropdown icon="more_horiz" %}
                             <div class="dropdown__item">
```
pi-sigma, line 24: What is the rationale for replacing the wheel icon with dots? It has the same/similar functionality as the other settings (e.g. "Bewerken").

jiromaykin: @pi-sigma the reason: because it's in the UX design... and previously someone just picked the wrong icons. in Taiga you can see a link to the design files, but they re also here: https://taiga.maykinmedia.nl/project/open-inwoner/wiki/home - i will mark this comment as resolved.

src/open_inwoner/scss/components/Dropdown/Dropdown.scss
jiromaykin: @pi-sigma Ha! Well seen - I will remove both widths  - i just copied my own code from the navigation menus in order to create the two overlapping triangles.

jiromaykin:
> .....not sure about is the change from wheel icon to horizontal dots, since we're now using different icons for the same (?) functionality."

@pi-sigma The reason is: it's in the design + and these UX elements do actually have subtly different functions. 
The cogwheel is for settings, where you can **_change_** options, 
the dots dropdown is for _retrieving_ stuff without being able to change much.

pi-sigma (approved):

PR #710, opened 2023-07-07
[#1534] Implement download of monthly + yearly benefits reports

pi-sigma:
- create SOAP service for retrieving social benefits reports
- parse XML from SOAP response
- generate PDFs for download
- integrate with CMS
- make settings configurable in the admin


jiromaykin:
@pi-sigma @alextreme Here I am attaching what I see when I generate the PDF's on my Mac locally (should look the same on every machine?):
[Jaaropgave2022APPLE.pdf](https://github.com/maykinmedia/open-inwoner/files/12230833/Jaaropgave2022APPLE.pdf)
[MaandspecificatieJuly2023APPLE.pdf](https://github.com/maykinmedia/open-inwoner/files/12230864/MaandspecificatieJuly2023APPLE.pdf)

src/open_inwoner/ssd/xml.py
```diff
@@ -0,0 +1,409 @@
...
+#
+def calculate_loon_zvw(fiscalloon: str, vergoeding_premie_zvw: str) -> str:
+    return str(int(fiscalloon) - int(vergoeding_premie_zvw))
+
+
+def format_float_repr(value: str) -> str:
+    return "{:.2f}".format(float(value) / 100).replace(".", ",")
+
+
+def format_address(street_name: str, housse_nr: str, house_letter: str) -> str:
```
alextreme, line 23: house_nr instead of housse_nr?

src/open_inwoner/ssd/xml.py
```diff
@@ -0,0 +1,409 @@
...
+            "sign": get_sign(
+                specificatiejaaropgave_spec,
+                "WerkgeversheffingPremieZVW.CdPositiefNegatief",
+            ),
+            "value": glom(
+                specificatiejaaropgave_spec,
+                "WerkgeversheffingPremieZVW.WaardeBedrag",
+            ),
+        },
+        # TODO: figure out where this is used; relation to `arbeitskorting`?
```
alextreme, line 361: Create a task for this, seems like a TODO that still needs to be tackled

alextreme (approved):
Looks good for an initial implementation. This will have to be validated together with a live SSD testenvironment, as I expect that there will be cases that we can't forsee atm. 

Remarks I made can be picked up or noted as tasks for the next sprint.

src/open_inwoner/ssd/xml.py
pi-sigma: That's part of what we discussed with Rob and should be resolved when we hear back. I'll create a task with this and the other points.

PR #713, opened 2023-07-11
:wheelchair: [#1597] Fixing a11y tags for non-textual content

jiromaykin:
Issues (some resolved in the past or other PR's): https://taiga.maykinmedia.nl/project/open-inwoner/task/1597


jiromaykin:
@pi-sigma I think and I hope I corrected this one for accessibility well enough... someone with a broad knowledge about accessibility should probably review this too - apart from that: I am never sure when to use double or single quotes in the Django translation tags...

src/open_inwoner/templates/pages/category/detail.html
```diff
@@ -4,7 +4,7 @@
 {% block header_image %}
 {% if object.image %}
     <aside class="banner" aria-hidden="true">
-        <img class="main-image main-image--start-after" src="{{ object.image.url }}" alt="{{ object.image.default_alt_text|default:'Header afbeelding' }}">
+        <img class="main-image main-image--start-after" src="{{ object.image.url }}" alt="{{ object.image.default_alt_text|default:'' }}">
     </aside>
```
pi-sigma, line 8: See previous comment/question.

pi-sigma:
@jiromaykin About the quotes: it doesn't matter for translation tags. I use double quotes and that's the norm in our projects, but it's just a convention.

pi-sigma (approved):

src/open_inwoner/templates/digid_eherkenning/mock/base.html
```diff
@@ -25,7 +25,7 @@
             <div>
                 {% block service_menu %}{% endblock %}
                 <a class="digidmock-anchor-placeholder">
-                    <img alt="Logo Rijksoverheid" class="rijksoverheid-logo" title="Logo Rijksoverheid"
+                    <img alt="Rijksoverheid" class="rijksoverheid-logo" title="Link naar Rijksoverheid"
                         src="{% static 'digid-mock/assets/RO_DigiD_Logo_Homepage.svg' %}"/>
```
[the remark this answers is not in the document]
jiromaykin: @pi-sigma I don't really have a reason for not **translating** the titles/alts :-) in the project I see we usually don't seem to translate them - which is a bit of an odd choice, given that the ALT attribute is definitely the one giving most accessibility and feedback (if it is not decorative). Perhaps I just need to create a new issue to handle all of the untranslated alt/title attributes.

src/open_inwoner/templates/pages/category/detail.html
jiromaykin: @pi-sigma this is a bit different: this one is either empty by default, or will actually have content if the admin sets this up in the 'mappen' configuration for images: http://127.0.0.1:8000/admin/filer/folder/. I am not sure if those config fields are translatable.
The ALT attribute must stay empty **if** they are decorative images, unless the admin/webeditor feels the image is not decorative and actually conveys important information.

PR #714, opened 2023-07-17
:wheelchair: [#1611] Improving robust mobile menus with aria-expanded

jiromaykin:
issue https://taiga.maykinmedia.nl/project/open-inwoner/task/1611

The aria-expanded attribute has to be applied to the focusable, interactive control that toggles the visibility of the object - needs to be set to False initially because everything starts off in a closed state.

Also: corrects this for the dropdown... where I set the aria-expanded attribute to the wrong node.

This will set this attribute to these 2 toggle menus (3 & 4):

Screenshot 2023-07-18 at 16 17 36
 

The aria-expanded attribute should NOT be added to the desktop menus (1 & 2),
because they need to be continuously accessible with the Tab keys: so their state is not 'invisible' for screenreaders (even though they are invisible for people that can see).:

Screenshot 2023-07-18 at 16 15 34

### DIFF
--- a/src/open_inwoner/components/templates/components/Button/Button.html
+++ b/src/open_inwoner/components/templates/components/Button/Button.html
@@ -8,6 +8,7 @@
         aria-label="{% firstof title text %}"
         {% if href|startswith:"http" %}target="_blank"{% endif %}
         {% as_attributes extra_attributes %}
+        {% if ariaExpanded %} aria-expanded="{{ ariaExpanded }}" {% endif %}
     >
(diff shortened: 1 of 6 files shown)


pi-sigma (approved):

pi-sigma (commented):
The `aria-expanded` on the menu button does not seem to get toggled (see screenshot). 

Also, it's a bit odd that the menu button gets hidden behind the search bar when expanded, as this makes it difficult to collapse it by clicking on it again. This is not caused your code, but perhaps something that could be addressed in a separate issue.

Screenshot from 2023-07-20 15-59-04

PR #716, opened 2023-07-25
:art: [#1617] Fixing labels for correct HTML syntax

jiromaykin:
Labels should not be surrounding other labels = incorrect HTML that also makes some accessibility tests fail.


pi-sigma (commented):
What about [search.html](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/templates/pages/search.html#L69) 69ff.? Are these nested labels appropriate there?

jiromaykin:
> What about [search.html](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/templates/pages/search.html#L69) 69ff.? Are these nested labels appropriate there?

@pi-sigma those I already fixed in the older PR you still have to review :-D (this one: https://github.com/maykinmedia/open-inwoner/pull/709) That's the older PR I worked on first - if that one is okay, then those changes should be merged first and will also improve this new one.

pi-sigma (approved):

PR #718, opened 2023-07-31
[#1624] escape html of product content field

pi-sigma:
Taiga: [#1624](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1624)


src/open_inwoner/pdc/models/product.py
```diff
@@ -201,6 +202,10 @@ def get_absolute_url(self, category=None):
     def has_cta_tag(self):
         return "\[CTABUTTON\]" in self.content
 
+    def save(self, *args, **kwargs):
+        self.content = html.escape(self.content).replace("\\&", "&")
+        super().save()
+
```
Bartvaderkin, line 208: As discussed in person: we established the field stores markdown syntax, but markdown allows HTML tags.

I though about it again: I think the html-escape should happen later, at render-time just before we render the markdown (eg: not here in save like we talked about at some point). What we could do here is `strip_tags()` to get rid of html tags, and leave the entities to the later escape.

This way the code is still editable (users `&`'s won't be` &amp;`'s). Possibly the markdown render function should be updated to do this.

This leaves the mystery of figuring out where the random slashes come from, and if they also show up back in the editor.

Bartvaderkin (changes requested):

Bartvaderkin:
The double slashes come from CKeditor or its markdown plugin trying to a some sort of html escape by putting double backslash in front of the angle bracket. If I use the same javascript libraries to render the string it does escape the html-tags, so this feels like a marked/gfmarkdown thing. But I cannot find where this happens.

So I appended a commit with a slightly different approach to clean this up: we undo the weird escape and then strip_tags.

Bartvaderkin (approved):

alextreme:
@Bartvaderkin please doublecheck your solution together with the ckeditor-table option.

I've checked on production where I could find HTML-tags, and noticed that this was the case with 4 products on MG:

https://mijn.groningen.nl/onderwerpen/producten/bijstand-voor-tweedehands-computer-aanvragen/

If we strip tags I'd prefer not to remove (from the editor) valid content. I now also suspect that these tables are the reason why HTML-content is allowed...

Bartvaderkin:
@alex I updated this PR: we no longer do html escape or strip tags, and only remove the weird slashes at render time as they are needed to be able to edit the html again (I guess the slashes are escaping html in ckeditor).

alextreme (approved):
Prima voor nu, graag merge conflicts verhelpen en een aparte issue maken voor ckeditor+tables

PR #722, opened 2023-08-02
[#1642] Fix SSD

pi-sigma:
- fix XML parsing
- add endpoints for constructing request URL's to SSDConfig
- improve error handling
- clean up type annotations


src/open_inwoner/ssd/models.py
```diff
@@ -22,6 +22,18 @@ class SSDConfig(SingletonModel):
         null=True,
         blank=True,
     )
+    maandspecificatie_endpoint = models.CharField(
```
alextreme, line 25: Having proper defaults (eg. 'UitkeringsSpecificatieClient-v0600') would make sense here. Due to the version number I guess making the endpoints configurable provides more flexibility, however as we don't know the differences with other versions hardcoding the endpoints would avoid confusion. Fine for now.

alextreme (approved):

PR #724, opened 2023-08-03
[#1535] Added subheading anchors to product page sidebar

pi-sigma:
- add subheadings to product page sidebar
- remove anchors for 'links' and 'see further' ('zie ook') 

Taiga: [#1535](https://taiga.maykinmedia.nl/project/open-inwoner/task/1535)

### DIFF
--- a/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
+++ b/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
@@ -6,13 +6,13 @@
         {% button href="#" icon="expand_more" icon_position="after" extra_classes="anchor-menu--mobile__title anchor-menu__toggle" bordered=False text=_("Op deze pagina") %}
         <ul class="anchor-menu__list">
             {% for anchor in anchors %}
-                <li class="anchor-menu__list__item">
+                <li class="anchor-menu__list-item">
                     <a class="link" href="{{ anchor.0 }}">{{ anchor.1 }}</a>
                     <!-- subheadings -->
                     {% if anchor.2 %}
                         <ul class="anchor-menu__sublist anchor-menu__sublist--mobile">
                             {% for subheading in anchor.2 %}
-                                <li class-"anchor-menu__sublist__item">
+                                <li class="anchor-menu__list-item">
                                     <a class="link" href="#{{ subheading.1 }}">{{ subheading.0 }}</a>
                                 </li>
                             {% endfor %}
@@ -28,13 +28,13 @@
 
     <ul class="anchor-menu__list">
         {% for anchor in anchors %}
-            <li class="anchor-menu__list__item">
+            <li class="anchor-menu__list-item">
                 <a class="link" href="{{ anchor.0 }}">{{ anchor.1 }}</a>
                 <!-- subheadings -->
                     {% if anchor.2 %}
                         <ul class="anchor-menu__sublist">
                             {% for subheading in anchor.2 %}
-                                <li>
+                                <li class="anchor-menu__list-item">
                                     <a class="link" href="#{{ subheading.1 }}">{{ subheading.0 }}</a>
                                 </li>
                             {% endfor %}
--- a/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
+++ b/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
@@ -6,8 +6,18 @@
         {% button href="#" icon="expand_more" icon_position="after" extra_classes="anchor-menu--mobile__title anchor-menu__toggle" bordered=False text=_("Op deze pagina") %}
         <ul class="anchor-menu__list">
             {% for anchor in anchors %}
-                <li class="anchor-menu__list-item">
+                <li class="anchor-menu__list__item">
                     <a class="link" href="{{ anchor.0 }}">{{ anchor.1 }}</a>
+                    <!-- subheadings -->
+                    {% if anchor.2 %}
+                        <ul class="anchor-menu__sublist anchor-menu__sublist--mobile">
+                            {% for subheading in anchor.2 %}
+                                <li class-"anchor-menu__sublist__item">
+                                    <a class="link" href="#{{ subheading.1 }}">{{ subheading.0 }}</a>
+                                </li>
+                            {% endfor %}
+                        </ul>
+                    {% endif %}
                 </li>
             {% endfor %}
             {{ contents }}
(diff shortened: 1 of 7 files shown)


src/open_inwoner/pdc/tests/test_product.py
```diff
@@ -254,3 +254,33 @@ def test_sidemenu_button_is_rendered_when_no_cta_inside_product_content(self):
...
+        product = ProductFactory(
+            content="##First subheading\rlorem ipsum...\r##Second subheading",
+            link="http://www.example.com",
+        )
+
+        response = self.app.get(
+            reverse("products:product_detail", kwargs={"slug": product.slug})
+        )
+
+        links = response.pyquery(".anchor-menu__sublist").find("a")
```
Bartvaderkin, line 271: Minor detail: you can expand the CSS selector instead of calling another `.find()`:

```py
links = response.pyquery(".anchor-menu__sublist a")
```

src/open_inwoner/pdc/utils.py
```diff
@@ -1 +1,25 @@
...
+    md = markdown.Markdown()
+    html_string = md.convert(content)
+
+    soup = BeautifulSoup(html_string, "html.parser")
+
+    subs = []
+    for tag in soup.find_all("h2"):
+        subheading = tag.text
+        slug = slugify(subheading)
+        subs.append((subheading, slug))
```
Bartvaderkin, line 23: See note elsewhere.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -52,6 +53,9 @@ def get_product_rendered_content(product):
 
             element.attrs["class"] = class_name
 
+            if tag == "h2":
+                element.attrs["id"] = slugify(element.text)
```
Bartvaderkin, line 57: I think we should namespace the ID's this generates so we don't generate collisions with whatever ID's we might be using ourselves. Like with a prefix:

```py
element.attrs["id"] = f"subheader-{slugify(element.text)}"
```

(also theoretically this could generate duplicates if titles would be repeated, but let's not worry about that)

Bartvaderkin (changes requested):

src/open_inwoner/scss/components/Header/AnchorMenu.scss
```diff
@@ -82,6 +82,30 @@
...
+    line-height: var(--spacing-large);
+
+    li {
+      border-left: none;
+    }
+    li:not(:last-child) {
+      margin-bottom: var(--spacing-large);
+    }
+    li > a {
+      font-size: small;
```
jiromaykin, line 99: Please use var(--font-size-body-small) for this - becasue 13px is too small here.

src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
```diff
@@ -8,6 +8,16 @@
             {% for anchor in anchors %}
                 <li class="anchor-menu__list-item">
                     <a class="link" href="{{ anchor.0 }}">{{ anchor.1 }}</a>
+                    <!-- subheadings -->
+                    {% if anchor.2 %}
+                        <ul class="anchor-menu__sublist anchor-menu__sublist--mobile">
+                            {% for subheading in anchor.2 %}
+                                <li>
```
jiromaykin, line 15: In order to make this work more specifically with the JS Gumshoe library you need to give each HTML tag its own class, preferably with BEM; so this LI could have an 'anchor-menu__sublist--item class' or something; so you can target them separately.

src/open_inwoner/scss/components/Header/AnchorMenu.scss
```diff
@@ -82,6 +82,30 @@
     }
   }
 
+  // nested list
+  &__sublist {
```
jiromaykin, line 86: You can use this Sublist class also by setting it as having a default styling of "font-weight" normal when the parent is active/open - and _**only**_ set it to "font-weight: bold" if the heading is actually scrolled to (we do not want everythign to be bold when the user hits only the top-heading).

src/open_inwoner/scss/components/Header/AnchorMenu.scss
```diff
@@ -82,6 +82,30 @@
...
+    list-style-type: '- ';
```
jiromaykin, line 87: I would like to show this to a designer actually - because I think this should be 'list-style-type:none; so it looks cleaner and ore in line with the rest of the website, but i'm really not sure about this one. So perhaps leave it as is.

pi-sigma: After playing around a bit with different styles, I think you're right. Changed it accordingly.

pi-sigma:
@Bartvaderkin Why do you keep changing the titles of my PRs? I use the imperative mood for a reason. See here:

[https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181)

jiromaykin (changes requested):
Nice work - just one more detail to improve: on mobile there should not be a left-margin in the dropdown for the H1 level; adding a bit of a margin to the H2 level is OK for now (look at https://open-inwoner-test.maykin.nl/onderwerpen/zorg-en-ondersteuning/producten/logeeropvang/ if you need an example):

left-margin

src/open_inwoner/scss/components/Header/AnchorMenu.scss
```diff
@@ -63,22 +63,47 @@
...
-      border-left: var(--border-width) solid;
-      border-color: var(--color-gray-light);
-      padding: var(--spacing-large);
+      margin: 0;
     }
   }
 
-  &__list-item--active .link {
+  &__list-item .link {
+    padding: 0.75rem 0 0.75rem 1.5rem;
```
jiromaykin, line 77: Also: no need for magic numbers here. Use the spacing that is already used in the anchor-menu; so something a little more like this: "var(--spacing-medium) 0 var(--spacing-medium) var(--spacing-medium)"

pi-sigma:
@jiromaykin I replaced the magic numbers with variables. The spacing to the left of the `h1`, however, is no longer `margin` but `padding` (in order to make the continuous border with coloring work). Removing that altogether will push the `h1` all the way to the left, so that it is no longer aligned with the menu button. I'm not 100% sure that the current alignment on mobile is correct either. Feel free to double-check.

Bartvaderkin:
> @Bartvaderkin Why do you keep changing the titles of my PRs? I use the imperative mood for a reason. See here:
> 
> https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181

We we're using capitalized Verbs in this project. Describe what you did, not the task: "Added this", "Updated that"

jiromaykin (approved):

Bartvaderkin (approved):

PR #727, opened 2023-08-08
[#1536] Refactored Haalcentraal to use a dataclass and update to 2.0-final

Bartvaderkin:
The refactoring and update commits are split for convenience.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_profile_views.py
+++ b/src/open_inwoner/accounts/tests/test_profile_views.py
@@ -1,3 +1,5 @@
+from dataclasses import asdict
+from datetime import date
 from unittest.mock import patch
 
 from django.test import override_settings
@@ -21,6 +23,7 @@ from open_inwoner.utils.tests.helpers import AssertTimelineLogMixin, create_imag
 from ...cms.profile.cms_apps import ProfileApphook
 from ...cms.tests import cms_tools
 from ...openklant.tests.data import MockAPIReadPatchData
+from ...haalcentraal.api_models import BRPData
 from ...questionnaire.tests.factories import QuestionnaireStepFactory
 from ..choices import ContactTypeChoices, LoginTypeChoices
 from ..forms import BrpUserForm, UserForm
@@ -610,20 +613,22 @@ class ProfileDeleteTest(WebTest):
 @requests_mock.Mocker()
 @override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
 class MyDataTests(HaalCentraalMixin, WebTest):
-    expected_response = {
-        "first_name": "Merel",
-        "initials": "M.",
-        "last_name": "Kooyman",
-        "prefix": "de",
-        "birthday": "10-04-1982",
-        "birthday_place": "Leerdam",
-        "gender": "vrouw",
-        "street": "King Olivereiland",
-        "house_number": 64,
-        "postcode": "2551JV",
-        "place": "'s-Gravenhage",
-        "land": None,
-    }
+    maxDiff = None
+
+    expected_response = BRPData(
+        first_name="Merel",
+        initials="M.",
+        last_name="Kooyman",
+        infix="de",
+        birthday=date(1982, 4, 10),
+        birthday_city="Leerdam",
+        gender="vrouw",
+        street="King Olivereiland",
+        housenumber="64",
+        postcode="2551JV",
+        city="'s-Gravenhage",
+        country="",
+    )
 
     def setUp(self):
         self.user = UserFactory(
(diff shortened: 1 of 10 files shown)


alextreme:
Ziet er in principe netjes uit. Je houdt volgens mij hiermee nog wel rekening ermee dat Groningen nog HC BRP v1.3 hanteert, toch?

Bartvaderkin:
@alextreme Ja die switch op `settings.BRP_VERSION` zit er nog in.

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -0,0 +1,37 @@
...
+    # extra fields
+    # TODO rename birthday_city
+    initials: str = ""
+    birthday_city: str = ""
+    gender: str = ""
+    # TODO rename postcode to postal_code
+    postcode: str = ""
+    # TODO rename land to country
+    country: str = ""
+
```
pi-sigma, line 28: - Remove comments
- `birthday_city` should really be `birth_place` or `place_of_birth`. See above.

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -610,20 +613,22 @@ def test_delete_staff_user_via_frontend_does_not_work(self):
...
+    maxDiff = None
+
+    expected_response = BRPData(
+        first_name="Merel",
+        initials="M.",
+        last_name="Kooyman",
+        infix="de",
+        birthday=date(1982, 4, 10),
+        birthday_city="Leerdam",
+        gender="vrouw",
```
pi-sigma, line 625: `birthday_city` should really be `birth_place` or `place_of_birth`. See below.

src/open_inwoner/haalcentraal/tests/test_logging.py
```diff
@@ -13,16 +13,16 @@
 
 @freeze_time("2021-10-18 13:00:00")
 @requests_mock.Mocker()
+@override_settings(
+    LOG_OUTGOING_REQUESTS_ENABLED=True,
+    LOG_OUTGOING_REQUESTS_DB_SAVE=True,
+)
 class TestPreSaveSignal(ClearCachesMixin, HaalCentraalMixin, TestCase):
```
pi-sigma, line 20: - I don't think there is a setting `LOG_OUTGOING_REQUESTS_ENABLED`
- `LOG_OUTGOING_REQUESTS_DB_SAVE=True` is the defaul, so override shouldn't be necessary

src/open_inwoner/templates/pages/profile/mydata.html
```diff
@@ -33,15 +33,15 @@ <h2 class="h2">{% trans "Mijn BRP gegevens" %}</h2>
...
                 <div class="tabled__row">
                     <div class="tabled__item tabled__item--bold">{% trans "Geboortedatum" %}</div>
-                    <div class="tabled__item">{{my_data.birthday|default:"-"}}</div>
+                    <div class="tabled__item">{{my_data.birthday|date:"d-m-Y"|default:"-"}}</div>
                 </div>
                 <div class="tabled__row">
                     <div class="tabled__item tabled__item--bold">{% trans "Geboorteplaats" %}</div>
-                    <div class="tabled__item">{{my_data.birthday_place|default:"-"}}</div>
+                    <div class="tabled__item">{{my_data.birthday_city|default:"-"}}</div>
                 </div>
```
pi-sigma, line 45: See previous comments about `birth_place`...

src/open_inwoner/haalcentraal/utils.py
```diff
@@ -83,24 +91,49 @@ def update_brp_data_in_db(user, brp_version, initial=True):
```
pi-sigma, line 91: Oddly the line I want to comment doesn't show up in this view (only if you follow "show file").

Instead of
```
 if data.get("personen"):
     data = data.get("personen", [])[0]
 else:
     data = []
```
perhaps the following:
```
if data := data.get("personen", []):
    data = data[0]
```
then you can ditch the `else` clause.

pi-sigma (commented):
Couple of small remarks/suggestions.

src/open_inwoner/haalcentraal/tests/test_logging.py
Bartvaderkin: Good catch, I see this got updated since I last had overrides in my local.py

I'm keeping the `@override_settings(LOG_OUTGOING_REQUESTS_DB_SAVE=True)` here because settings like these are often modified in local.py (as I do).

Bartvaderkin:
Thanks for the feedback, I updated the PR with new commits.

pi-sigma (approved):

PR #728, opened 2023-08-09
[#1647] Add configurable warning banner

pi-sigma:
Taiga: [#1647](https://taiga.maykinmedia.nl/project/open-inwoner/task/1647)


src/open_inwoner/components/templates/components/Header/WarningHeader.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n link_tags static %}
+
+<style>
+	.warning-header {
+		background-color: {{ warning_banner_background_color }};
```
alextreme, line 5: I don't see the warning_banner_font_color option used here?

Also, as we have CSP headers restricting inline styles this won't work out of the box on non-dev environments.

See the use of request.csp_nonce in templates/master.html, this will be the only way your style can be picked up.

src/open_inwoner/utils/templatetags/utils.py
```diff
@@ -98,3 +100,8 @@ def readable_size(value):
 @register.filter
 def cookies_accepted(request: HttpRequest) -> bool:
     return request.COOKIES.get("cookieBannerAccepted") == "true"
+
+
+@register.simple_tag
+def warning_banner_enabled() -> bool:
```
alextreme, line 106: Don't do this via a templatetag, they are already overused in the project. See utils/context_processors.py

alextreme (changes requested):

alextreme (approved):
Looks fine to me, @jiromaykin can give this a double-check just in case

src/open_inwoner/scss/components/WarningHeader/WarningHeader.scss
```diff
@@ -0,0 +1,20 @@
+.warning-header {
+  display: flex;
+  box-sizing: border-box;
+  padding: 1rem;
```
jiromaykin, line 4: We have a whole bunch of nice margins and distances to chose from. Instead of 1rem, i would use "padding: var(--spacing-large);" here. 
This also makes it easier to change ALL spaces at once if ever the "huisstijl" needs to change.

src/open_inwoner/scss/components/WarningHeader/WarningHeader.scss
```diff
@@ -0,0 +1,20 @@
...
+
+  .warning-header__container {
+    display: flex;
+    align-items: center;
+    padding: 0 4rem;
```
jiromaykin, line 9: '4rem' is a rather odd magic number 🙂, and will behave differently on different types of screens (like retina, and non-retina), so I'd use "padding: 0;" here, and set the max- widths for different screen-widths, because this one should have any padding on mobile anyway. And: we use a '**mobile first**' CSS styling in the entire project so we start with the mobile styling first and then add media-queries for wider screens:
So add something like this and check to see if it looks okay, maybe you need extra styling for screenwidths between 500px & 768px but probably not: 
`.warning-header__container {
    padding: 0; @media (min-width: 768px) {
      max-width: 90%;
    }`

jiromaykin (changes requested):
I have added front-end feedback and requested changes that will be quite informative about how we use icons etc. in this project 🙂

src/open_inwoner/configurations/migrations/0047_auto_20230809_0833.py
```diff
@@ -0,0 +1,52 @@
...
+                max_length=18,
+                verbose_name="Warning banner background",
+            ),
+        ),
+        migrations.AddField(
+            model_name="siteconfiguration",
+            name="warning_banner_enabled",
+            field=models.BooleanField(
+                default=False,
+                help_text="Wether the warning banner should be displayed",
```
jiromaykin, line 29: Should be spelled as "whether".

alextreme:
Creating test database for alias 'default'...
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0047_auto_20230809_0833, 0048_auto_20230816_1255 in configurations).
To fix them run 'python manage.py makemigrations --merge'

jiromaykin (approved):

alextreme:
django.db.migrations.exceptions.NodeNotFoundError: Migration configurations.0049_siteconfiguration_extra_css dependencies reference nonexistent parent node ('configurations', '0048_auto_20230816_1255')

jiromaykin (approved):

PR #729, opened 2023-08-09
clean up implementation of ssd

pi-sigma:
- Add missing translations of month names


alextreme (approved):
Still a draft PR so won't merge, but this PR looks fine

pi-sigma:
I've removed the code bits I used for local testing/development, as these are no longer needed.

PR #730, opened 2023-08-11
[#1645] improve design of product page

pi-sigma:
- make cta button text configurable
- remove mobile-view button from product detail template (unnecessary duplication)
- break up overflowing links in text field

Taiga: [#1645](https://taiga.maykinmedia.nl/project/open-inwoner/task/1645)


alextreme:
A test that verifies that a modified button_text is shown on the product page would be appreciated. Might feel a bit simple though...

Bartvaderkin (commented):
Code is decent but wouldn't it be nicer if this could've been edited through the placeholder? 

Like:
```
[CTABUTTON "click me!"]
```

alextreme:
> Code is decent but wouldn't it be nicer if this could've been edited through the placeholder?
> 
> Like:
> 
> ```
> [CTABUTTON "click me!"]
> ```

Would indeed be more flexible, but I'd prefer to stick with Paul's simpler approach as making the markdown-content parsing more complicated may be tricky. I'm not sure if having multiple buttons with different names would be necessary, especially as all buttons point to the same URL/form...

Bartvaderkin (approved):

alextreme:
Have made this PR ready-for-review, but allowing @pi-sigma to double-check just in case it was on Draft for a reason

pi-sigma:
@alextreme Good to go, I forgot to mark it for review.

PR #732, opened 2023-08-14
[#1657] add options to hide categories and search from anonymous users

pi-sigma:
- added model fields (bool) for hiding categories, search bar + search page
- added `LoginMaybeRequiredMixin` to categories and search views in order to conditionally restrict access to authenticated users

Taiga: [#1657](https://taiga.maykinmedia.nl/project/open-inwoner/task/1657)


alextreme (approved):
Looks good, however considering the refactor is a bit bigger than only the requested change @jiromaykin should also take a look (and he's back tomorrow)

jiromaykin:
> ....considering the refactor is a bit bigger than only the requested change @jiromaykin should also take a look...

I will check it out and will need to update everything locally and see what I can do here.

src/open_inwoner/configurations/migrations/0048_auto_20230816_1255.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="siteconfiguration",
+            name="hide_search_from_anonymous_users",
+            field=models.BooleanField(
+                default=False,
+                help_text="If checked, only authenticated users will be able to search the page.",
+                verbose_name="Hide search from anonymouns users",
```
jiromaykin, line 22: 1st: Spelling mistake: "anonymouns" should be "anonymous"

src/open_inwoner/configurations/models.py
```diff
@@ -445,6 +445,20 @@ class SiteConfiguration(SingletonModel):
         default=True,
         help_text=_("Whether file sharing via the messages is allowed or not"),
     )
+    hide_categories_from_anonymous_users = models.BooleanField(
+        verbose_name=_("Hide categories from anonymouns users"),
```
jiromaykin, line 449: 2nd occurence: Spelling mistake: "anonymouns" should be "anonymous"

src/open_inwoner/configurations/models.py
```diff
@@ -445,6 +445,20 @@ class SiteConfiguration(SingletonModel):
...
+        default=False,
+        help_text=_(
+            "If checked, categories will be hidden from users who are not logged in."
+        ),
+    )
+    hide_search_from_anonymous_users = models.BooleanField(
+        verbose_name=_("Hide search from anonymouns users"),
```
jiromaykin, line 456: 3rd: Spelling mistake: "anonymouns" should be "anonymous"

jiromaykin (changes requested):
All else is looking good, apart from the spellign mistakes; so when logged all have become invisible, both on mobile as well as on desktop.

jiromaykin (approved):

PR #733, opened 2023-08-15
Upgrading django-log-outgoing-requests to 0.5.0

alextreme:


pi-sigma (approved):
Settings had to be updated to accommodate latest changes. Should be fine now.

PR #734, opened 2023-08-17
[#1669] fix data type in in xml parsing

pi-sigma:
+ modify test data to match response from acceptance environment

Taiga: [#1669](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1669)


pi-sigma:
The patch treats the symptom, not the cause. It turns out that the response from the SOAP service sometimes contains one `DossierHistorie`, sometimes more (the Python datatype from parsing the response being `list` and `dict`). The 'additional' `DossierHistorie` data doesn't show up in the PDF that was provided, so we need to reconsider the structure of the data we're dealing with.

Related to Taiga user story [#1670](https://taiga.maykinmedia.nl/project/open-inwoner/us/1670)

Bartvaderkin:
> the response from the SOAP service sometimes contains one DossierHistorie, sometimes more (the Python datatype from parsing the response being list and dict)

@pi-sigma This issue demonstrates why `xml_to_dict()` is a bit of a kludge and fundamentally in-accurate. 

Ideally we'd want tools that handle XML correctly, hence why I initially was going for the XSLT route. I had another library that parses the XSD and WSDL's and used correct dataclasses, but it wasn't complete enough.

src/open_inwoner/ssd/tests/test_views.py
```diff
@@ -52,7 +52,7 @@ def test_get(self):
 
     @patch(
         "open_inwoner.ssd.client.UitkeringClient.get_report",
-        return_value=mock_report(str(FILES_DIR / "uitkering_response.xml")),
+        return_value=mock_report(str(FILES_DIR / "uitkering1.xml")),
```
Bartvaderkin, line 55: What is the difference between the different uitkering files? Lets give these files a name with something more readable then just a number. 

If the language supports it (XML does) you could also add a comment on top inside the file and at the location of any interesting changes.

src/open_inwoner/ssd/tests/test_xml_parsing.py
```diff
@@ -8,7 +8,8 @@
 
 class XMLParsingTests(TestCase):
     def test_uitkering_response_parsing(self):
-        uitkering_file = str(FILES_DIR / "uitkering_response.xml")
+        uitkering_file = str(FILES_DIR / "uitkering1.xml")
+        # uitkering_file = str(FILES_DIR / "uitkering2.xml")
```
Bartvaderkin, line 12: Lingering comment, and I think you won't need the `str()` cast but keep it a Path?

Bartvaderkin (changes requested):
I'm principe I'm not a huge fan of big fixture changes like the one in `uitkering.xml`.

Also the naming of the uitkering files is not good:. Note the jaaropgave has `jaaropgave_reponse.xml` so it is clear at a glance it is a response.

So I'd go with something like:

`uitkering_response_basic.xml`
`uitkering_response_extra_dossierhistorie.xml`

Bartvaderkin (approved):

PR #736, opened 2023-08-18
[#1672] Enabled non-unique emails for digid users

pi-sigma:
Taiga: [#1672](https://taiga.maykinmedia.nl/project/open-inwoner/task/1672)

- digid users can register with with email addresses that are already in use
- non-digid users cannot register with existing email addresses
- all users can edit their profile

I refactored the auth tests and divided the suite as follows:
1. Registration tests concerning DigiD users
2. Registration tests concerning users who register with email + password
3. DuplicateEmail tests, which concern both digid and non-digid users


alextreme:
To check:
- Chat/messages -> messages are linked to a user and not an emailadres
- Validating of emailadres -> is currently not done and not in scope of the user story. Would be good to add this as a new user story to the backlog

src/open_inwoner/accounts/models.py
```diff
@@ -216,6 +219,52 @@ class Meta:
...
+        ):
+            return
+
+        # some account does not have login_type digid
+        raise ValidationError(
+            {
+                "email": ValidationError(
+                    _(
+                        "A user with this Email already exists. If you need to register "
+                        "with an Email addresss that is already in use, both users of the "
```
alextreme, line 261: Instead of "both... DigiD" -> contact us

src/open_inwoner/accounts/tests/test_user_manager.py
```diff
@@ -21,6 +21,13 @@ def test_create_user(self):
         self.assertFalse(user.is_staff)
         self.assertFalse(user.has_usable_password())
 
+    def test_create_user_with_non_unique_email_and_digid_login(self):
+        user1 = User.objects.create_user("test@example.com", password="12345")
```
alextreme, line 25: TODO :)

src/open_inwoner/accounts/models.py
```diff
@@ -216,6 +219,52 @@ class Meta:
...
+        # no duplicates
+        if not existing_users:
+            return
+
+        # the curent user is editing their profile
+        if self in existing_users:
+            return
+
+        # no active user with duplicate email
+        if not any((user.is_active for user in existing_users)):
```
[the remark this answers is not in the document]
pi-sigma: Then we can no longer distinguish between the case with no duplicates (-> user can register) and the case with duplicates which have all been deactivated (-> user cannot register).

src/open_inwoner/accounts/views/contacts.py
```diff
@@ -76,17 +77,23 @@ def get_form_kwargs(self):
...
         cleaned_data = form.cleaned_data
-        email = cleaned_data["email"]
+
         user = self.request.user
-        contact_user = User.objects.filter(email__iexact=email)
+        first_name = cleaned_data["first_name"]
+        last_name = cleaned_data["last_name"]
+        email = cleaned_data["email"]
+
+        contact_users = User.objects.filter(
```
alextreme, line 86: I would recommend to first do this based on only email, and only if there are duplicates to use first+last name to differentiate between them.

Reason being that otherwise I have to enter my contact's name exactly as they entered theirs.

src/open_inwoner/accounts/models.py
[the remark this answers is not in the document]
alextreme: Makes sense, otherwise we ignore inactive users but then it isn't clear that the user can't log in using an inactivate account. Lets keep it as-is.

src/open_inwoner/accounts/views/contacts.py
alextreme: Discussed, this is done as described in forms.py. However we don't want to duplicate this code.

Recommendation is to keep the existing_user from the find_user() method and if not None then set this to the cleaned_data, this so the User can be reused from within form_valid method in the view.

src/open_inwoner/conf/base.py
```diff
@@ -457,6 +457,7 @@
 # Allow logging in with email+password
 AUTHENTICATION_BACKENDS = [
     "open_inwoner.accounts.backends.CustomAxesBackend",
+    "open_inwoner.accounts.backends.CustomDigiDBackend",
```
alextreme, line 460: To be removed

src/open_inwoner/conf/test.py
```diff
@@ -12,11 +12,12 @@
 
 # Allow logging in with both username+password and email+password
 AUTHENTICATION_BACKENDS = [
-    "open_inwoner.accounts.backends.CustomAxesBackend",
-    "open_inwoner.accounts.backends.UserModelEmailBackend",
-    "django.contrib.auth.backends.ModelBackend",
-    "digid_eherkenning.mock.backends.DigiDBackend",
-    "open_inwoner.accounts.backends.CustomOIDCBackend",
+    # "open_inwoner.accounts.backends.CustomAxesBackend",
+    "open_inwoner.accounts.backends.CustomDigiDBackend",
```
alextreme, line 16: To be reverted

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -312,10 +386,11 @@ def test_user_cannot_delete_other_users_contact(self):
         other_user = UserFactory()
         response = self.app.post(self.delete_url, user=other_user, status=404)
 
+    # TODO: fix
```
alextreme, line 389: TODO can be removed

pi-sigma:
Django gives the following warning on `runserver`:

`accounts.User: (auth.W004) 'User.email' is named as the 'USERNAME_FIELD', but it is not unique.
        HINT: Ensure that your authentication backend(s) can handle non-unique usernames.`

I think we are good, since we (i) have separate authentication backends for digid and non-digid users and (ii) allow non-unique emails only for digid users. Needs to be double-checked.

src/open_inwoner/accounts/models.py
```diff
@@ -316,6 +356,7 @@ def get_contact_type_display(self) -> str:
         choice = ContactTypeChoices.get_choice(self.contact_type)
         return choice.label
 
+    @property
     def is_not_active(self):
         return not self.is_active
```
Bartvaderkin, line 361: This function is weird, why does it exist instead of `not user.is_active` like normal python?

src/open_inwoner/accounts/models.py
```diff
@@ -216,6 +219,47 @@ class Meta:
         verbose_name = _("User")
         verbose_name_plural = _("Users")
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._old_bsn = self.bsn
+
+    def __str__(self):
+        return f"{self.first_name} {self.last_name} ({self.email})".strip()
```
Bartvaderkin, line 227: This isn't fully correct: because of legacy reasons we have `example.org` addresses as placeholders (see `User.get_contact_email`) that we don't want to display.

src/open_inwoner/accounts/tests/test_contact_views.py
```diff
@@ -275,10 +290,69 @@ def test_adding_contact_with_invalid_last_name_chars_fails(self):
...
+        form = response.forms["contact-form"]
+        form["first_name"] = existing_user.first_name
+        form["last_name"] = existing_user.last_name
+        form["email"] = existing_user.email
+
+        response = form.submit(user=self.digid_user)
+        pending_invitation = self.digid_user.contacts_for_approval.first()
+
+        self.assertEqual(response.status_code, 302)
+        self.assertContains(response.follow(), existing_user.email)
```
Bartvaderkin, line 325: Webtest checks response codes for the thing we do, so a `.follow()` only works for a redirecting 3xx response. Similarly does a .get() or post() checks for a 2xx etc (you can specify an exact code as argument).

Bartvaderkin (changes requested):

src/open_inwoner/accounts/forms.py
```diff
@@ -317,26 +303,71 @@ def __init__(self, user, *args, **kwargs):
...
+        if existing_users.count() > 1:
+            raise ValidationError(
+                _(
+                    "We're having trouble finding an account with this information."
+                    "Please contact us for help."
                 )
+            )
+
+        # exactly 1 user
+        existing_user = existing_users.get()
```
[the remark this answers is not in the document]
pi-sigma: The `User.DoesNotExist` shouldn't be necessary, though, since we're returning early if the queryset is empty, no?

src/open_inwoner/accounts/tests/test_logging.py
```diff
@@ -67,7 +67,7 @@ def test_registration_is_logged(self):
             {
                 "message": _("user was created"),
                 "action_flag": list(LOG_ACTIONS[4]),
-                "content_object_repr": user.email,
+                "content_object_repr": f"{user.first_name} {user.last_name} ({user.email})",
```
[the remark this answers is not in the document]
pi-sigma: Some use the `user.first_name`, others the `self.user.first_name` (similarlly for the `last_name`). Still, refactoring sounds like a good idea.

Bartvaderkin:
I rebased and pushed this, but my changes will be in another branch: so we can PR on the PR and have a diff of what I did.

PR #739, opened 2023-08-21
:lipstick: [#1492] Adding notification dot in login menu

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/task/1492

show just a dot IF there are any notifications for messages OR plans.
we do not have a design for this so I just made it look like this (dot has the same size as the other one and overlaps the down-arrow a bit):

open-dicht-dot

### DIFF
--- a/src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
+++ b/src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
@@ -9,6 +9,12 @@
             <li class="primary-navigation__list-item">
             {% button text=_('Welkom ')|addstr:request.user.get_short_name type="button" icon="expand_more" icon_position="after" icon_outlined=True transparent=True extra_classes="primary-navigation--toggle" %}
 
+            {% with messages_notification=request.user.get_new_messages_total plans_notification=request.user.get_plan_contact_new_count %}
+                {% if messages_notification or plans_notification %}
+                    <span class="indicator"><span class="indicator__dot"></span></span>
+                {% endif %}
+            {% endwith %}
+
                 <ul class="primary-navigation__list subpage-list">
                     {% show_menu_below_id "home" 0 100 100 100 "cms/menu/primary.html" %}
 
--- a/src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
+++ b/src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
@@ -9,14 +9,13 @@
             <li class="primary-navigation__list-item">
             {% button text=_('Welkom ')|addstr:request.user.get_short_name type="button" icon="expand_more" icon_position="after" icon_outlined=True transparent=True extra_classes="primary-navigation--toggle" %}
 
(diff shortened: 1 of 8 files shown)


src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
```diff
@@ -8,6 +8,11 @@
         <ul class="primary-navigation__list">
             <li class="primary-navigation__list-item">
             {% button text=_('Welkom ')|addstr:request.user.get_short_name type="button" icon="expand_more" icon_position="after" icon_outlined=True transparent=True extra_classes="primary-navigation--toggle" %}
+            {% for child in children %}
+                {% if child.indicator %}
+                    <span class="indicator"><span class="indicator__dot"></span></span>
+                {% endif %}
+            {% endfor %}
```
Bartvaderkin, line 15: This for-loop is probably not what we want, as there could be multiple menu items with indicators. 

Also there is a weird situation because the `children` context variable shouldn't be present at this point, it supposedly gets generated by `show_menu_below_id` a little further down.

I'll investigate what is going on and create a snippet of code for this.

Bartvaderkin (changes requested):

src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
pi-sigma: This seems to be a pattern. Cf. e.g. `templates/cms/menu/primary.html`: a loop over `children`, even though the context is updated by `show_menu` further down.

Bartvaderkin: It seems like a bug or flaw in django-cms where it doesn't use the context stack the way template tags are supposed to. I filled an issue https://github.com/django-cms/django-cms/issues/7628 but I don't expect anything from it.

Bartvaderkin:
@jiromaykin We were a bit too early to merge my fix branch, but at least the test did catch the issue you spotted as well.

src/open_inwoner/pdc/tests/test_category.py
```diff
@@ -86,7 +86,7 @@ def test_only_published_subcategories_exist_in_detail_page_when_logged_in(self):
     def test_only_published_categories_exist_in_my_categories_page(self):
         response = self.app.get(reverse("profile:categories"), user=self.user)
         self.assertEqual(
-            list(response.context["categories"]),
+            list(response.context["menu_categories"]),
```
Bartvaderkin, line 89: This test didn't test what it supposed to test (it is not about the menu but about a form), so this fix is not what we want. I'm rewriting it right now.

jiromaykin: @Bartvaderkin The bad test is no longer bad - all seems good now! + and @pi-sigma also the "configure Onderwerpen to become invisible for logged out users" still seems to be working this way. May have something to do with the 'context'.

Bartvaderkin (approved):

pi-sigma (approved):

PR #740, opened 2023-08-21
[#1675] Added extra_css field to SiteConfiguration and master template

Bartvaderkin:


jiromaykin (changes requested):
Nice! And also: nicely hidden and hard to find, only for advanced users :-)

src/open_inwoner/configurations/admin.py
```diff
@@ -198,10 +205,42 @@ class SiteConfigurarionAdmin(OrderedInlineModelAdminMixin, SingletonModelAdmin):
...
     inlines = [SiteConfigurationPageInline]
     form = SiteConfigurarionAdminForm
 
+    readonly_fields = [
+        "extra_css_allowed",
+    ]
+
+    @admin.display(
+        description="",
+    )
```
pi-sigma, line 228: `description="Allowed CSS properties"` will fix the issue brought up by Jiro

src/open_inwoner/configurations/admin.py
```diff
@@ -198,10 +205,42 @@ class SiteConfigurarionAdmin(OrderedInlineModelAdminMixin, SingletonModelAdmin):
...
+                    "extra_css_allowed",
+                ),
+            },
+        ),
     )
     inlines = [SiteConfigurationPageInline]
     form = SiteConfigurarionAdminForm
 
+    readonly_fields = [
+        "extra_css_allowed",
```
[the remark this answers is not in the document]
pi-sigma: See comment below

src/open_inwoner/configurations/tests/test_extra_css.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.config = SiteConfiguration.get_solo()
+        self.config.extra_css = "body {color: red;}"
+        self.config.save()
+
+        response = self.app.get(reverse("pages-root"))
+        extra_css = response.context.get("extra_css")
+
+        self.assertEquals(extra_css, self.config.extra_css)
+        self.assertContains(response, "body {color: red;}")
```
pi-sigma, line 32: Perhaps a more specifc assert (similar to the next test): 

```
expected = "body {color: red;}"
actual = response.pyquery("#extra-css")[0].text.strip()
self.assertEquals(expected, actual)
```

src/open_inwoner/scss/admin/_admin_theme.scss
```diff
@@ -330,3 +330,20 @@ form div.help {
...
+    font-size: 95%;
+  }
+}
+.css-properties-table {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
+  grid-column-gap: 2em;
+  &__column {
+    font-family: monospace;
+    font-size: 95%;
```
pi-sigma, line 347: What's the reason for this 5% decrease? I can barely tell the difference from looking at the screen.

src/open_inwoner/utils/css.py
```diff
@@ -0,0 +1,64 @@
...
+ALLOWED_CSS_PROPERTIES = _ALLOWED_CSS_PROPERTIES
+ALLOWED_SVG_PROPERTIES = _ALLOWED_SVG_PROPERTIES
+
+
+def allowed_properties() -> List[str]:
+    # used for admin etc
+    parts = set()
+    parts.update(ALLOWED_CSS_PROPERTIES)
+    parts.update(ALLOWED_SVG_PROPERTIES)
+    return list(sorted(parts))
```
pi-sigma, line 19: The properties from bleach come in the form of a `frozenset`, which is assigned to `ALLOWED_CSS_PROPERTIES` and `ALLOWED_SVG_PROPERTIES` above. In this function you first build a `set` and then return a `list` Why?

If we don't intend to mutate these structures, it would be better to stick to `frozenset`, which is immutable.

src/open_inwoner/utils/itertools.py
```diff
@@ -0,0 +1,24 @@
+from itertools import islice
+from math import ceil
+
+
+def batched(iterable, chunk_size: int):
+    """
+    split `iterable `in parts of `chunk_size` length
+
+    implements Python 3.12 itertools.batched()
```
pi-sigma, line 9: Cool, but is there a reason for not simply copying over the function definition from Python 3.12?

Also, if we do type annotations, they should be consistent I think. So either `def batched(iterable: Iterable, chunk_size: int)` or `def batched(iterable, n)`, letting the `n` speak for itself (see the docs).

src/open_inwoner/utils/itertools.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    iterator = iter(iterable)
+    while chunk := tuple(islice(iterator, chunk_size)):
+        yield chunk
+
+
+def split(iterable, num_parts: int = 2):
```
pi-sigma, line 16: See previous comment about type annotation

src/open_inwoner/utils/tests/test_css.py
```diff
@@ -0,0 +1,53 @@
...
+        css = "body { width:10px; color: #fff;}"
+        expected = "body {width:10px; }"
+
+        self.assertEqual(
+            expected, clean_stylesheet(css, allowed_css_properties=allowed_props)
+        )
+
+    def test_clean_stylesheet_allows_any_string(self):
+        """
+        proofs we need to escape this if we print in html
```
pi-sigma, line 48: Do you mean "proves"? 

It's good to know that we need to do this, hence it should be documented. But what's the point of proving it via a test? I would prefer to have this information included in the docstring of `utils.clean_stylesheet`, where it's more prominent.

src/open_inwoner/scss/admin/_admin_theme.scss
Bartvaderkin: To make it slightly smaller because monospace is so chonky.

src/open_inwoner/utils/css.py
Bartvaderkin: I'm going to strip this list and remove some of the weird properties nobody would use.

src/open_inwoner/utils/itertools.py
Bartvaderkin: - This seems good enough. We should update python in general but that is out of scope.
- I updated the type hint, with generics
- I renamed the module to `iteration`

src/open_inwoner/utils/tests/test_css.py
Bartvaderkin: Yes. 

The problem with comments about code behaviour is they go stale.

pi-sigma (approved):

Bartvaderkin:
@jiromaykin You've marked yourself as 'requested changes' here but I don't see any requests. Did you meant to just add a comment?

jiromaykin:
>"You've marked yourself as 'requested changes' here but I don't see any requests"

Well I did :-) I  asked for the unused colon to be removed (but putting in a useful title in there instead is also acceptable), so now I can approve.

src/open_inwoner/configurations/admin.py
```diff
@@ -198,10 +205,42 @@ class SiteConfigurarionAdmin(OrderedInlineModelAdminMixin, SingletonModelAdmin):
...
+    ]
+
+    @admin.display(
+        description=_("Allowed CSS properties"),
```
jiromaykin, line 227: This is what solved my issue in the request; approved.

jiromaykin (approved):

PR #742, opened 2023-08-23
rename variable in test

pi-sigma:
A variable was renamed in the main code but not in the test during the recent Haalcentral refactor


alextreme:
Seems now another test is failing... but at least the BRP one is resolved thanks

alextreme (approved):

PR #746, opened 2023-08-24
[#1689, #1671] Fixed SSD maandspecificatie

pi-sigma:
- Fixed missing 'Voorvoegsel` and improved error-handling in monthly report
- changed month names in pdf header (abbreviation -> full name)
- added tests for xml parsing utility functions

Taiga [#1689](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1689) and Taiga [#1671](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1671)


src/open_inwoner/ssd/xml.py
```diff
@@ -45,11 +47,13 @@ def format_name(first_name: str, voorvoegsel: str, last_name: str):
     first_name_initials = [name[0] + "." for name in first_names]
     first_name_formatted = " ".join(first_name_initials)
```
Bartvaderkin, line 48: Old feedback: for single-use iterables it's nicer to use generators, ex: a generator expression instead of a list comprehension. There is no need to allocate memory for all the new items at once if we're only looking at one at a time.

```py
    # not square brackets but parenthesis
    first_name_initials = (name[0] + "." for name in first_names)
    first_name_formatted = " ".join(first_name_initials)

    # the parenthesis aren't actually part of the expression itself but for syntax reasons so sometimes can be left out like when calling a function, although in practice Black will probably do a linebreak
    first_name_formatted = " ".join(name[0] + "." for name in first_names)
```

src/open_inwoner/ssd/xml.py
```diff
@@ -22,7 +22,9 @@ def format_float_repr(value: str) -> str:
 
 
 def format_address(street_name: str, house_nr: str, house_letter: str) -> str:
-    return f"{street_name} {house_nr} {house_letter}"
+    if house_letter:
+        return f"{street_name} {house_nr} {house_letter}"
+    return f"{street_name} {house_nr}"
 
 
 def format_date(date_str) -> str:
```
Bartvaderkin, line 30: Looks like this PR adds a lot of glom() empty string default arguments (below), but these  `format_`and `calculate` functions don't handle the empty string values.

For example this `datetime.strptime(date_str, "%Y%m%d").strftime("%d-%m-%Y")` raises `ValueError: time data '' does not match format '%Y%m%d'` when `date_str` is empty.

Bartvaderkin (changes requested):
Looks like this PR adds a lot of glom() empty string default arguments (below), but the  `format_`and `calculate` functions don't handle the empty string values.

For example:
```py
datetime.strptime(date_str, "%Y%m%d").strftime("%d-%m-%Y")
``` 
when `date_str` is empty  raises:
```py
ValueError: time data '' does not match format '%Y%m%d'
```

Additionally joining strings by using a template is inaccurate if some parts are empty:

```py
f"{street_name} {house_nr} {house_letter}"
```
If `house_nr` is empty you have a double space. it's not the end of the world here but conceptually not correct. (I'd probably push the parts in a list and filter+join it with a space)

src/open_inwoner/ssd/xml.py
```diff
@@ -43,17 +47,17 @@ def format_date_month_name(date_str) -> str:
...
+    first_names = first_name.strip().split(" ")
+    first_name_initials = (name[0] + "." for name in first_names)
     first_name_formatted = " ".join(first_name_initials)
 
-    if voorvoegsel:
-        return f"{first_name_formatted} {voorvoegsel} {last_name}"
-    return f"{first_name_formatted} {last_name}"
+    elements = (first_name_formatted, voorvoegsel, last_name)
+
+    return " ".join((item.strip() for item in elements if item))
```
Bartvaderkin, line 56: This is not correct. The conditions is applied before the left side of the loop:

```py
ll = ['a', 'b', '', '   ', 'c']
[a.strip() for a in ll if a]
-> ['a', 'b', '', 'c']
```

So strings that .strip() to an empty string are not seen and filtered by the condition.

src/open_inwoner/ssd/tests/test_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    format_date,
+    format_date_month_name,
+    format_name,
+    get_sign,
+    get_column,
+)
+
+
+class ParsingUtilsTest(TestCase):
+    @parameterized.expand([
```
Bartvaderkin: We should probably discus it if we want to add new infrastructure. Possibly there already experience to take advantage of. I guess this one is fine for now.

src/open_inwoner/ssd/xml.py
```diff
@@ -22,18 +22,22 @@ def format_float_repr(value: str) -> str:
...
-        return f"{street_name} {house_nr} {house_letter}"
-    return f"{street_name} {house_nr}"
+    elements = (street_name, house_nr, house_letter)
+    return " ".join((item.strip() for item in elements if item != ""))
 
 
-def format_date(date_str) -> str:
+def format_date(date_str: str) -> str:
+    if not date_str:
+        return date_str
```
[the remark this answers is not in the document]
Bartvaderkin: > the glom that returns the date_str returns "" by default, so this is covered.

The glom happens outside this function so that dependency needs to be remembered. It is much safer to encapsulate and make every part behave correctly. Eg: if the input is not useful return an empty string instead of whatever was put in.

src/open_inwoner/ssd/tests/test_utils.py
pi-sigma: Let's talk about it on Thursday; this stuff might not be needed if we decide to refactor.

src/open_inwoner/ssd/xml.py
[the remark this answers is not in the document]
Bartvaderkin: The `calculate_loon` and `format_float_repr` also need some of this: both parse the possibly empty string without catching value error:

```py
ValueError: could not convert string to float: ''
ValueError: invalid literal for int() with base 10: ''
```

src/open_inwoner/ssd/xml.py
```diff
@@ -170,32 +170,36 @@ def get_uitkering_dict(xml_data) -> Optional[dict]:
...
+            "value": glom(dossier_dict, "Regeling", default=""),
         },
     }
 
     #
     # Details
     #
     details = {}
-    details_list = glom(dossier_dict, "Componenthistorie")
+    details_list = glom(dossier_dict, "Componenthistorie", default="")
```
Bartvaderkin, line 191: Not sure if this is wrong but it works: the default value is a string for a what is usually a list.

Bartvaderkin (changes requested):

src/open_inwoner/ssd/xml.py
pi-sigma: Ugh, good catch. It works because both are iterable, but I'll change it.

PR #747, opened 2023-08-24
:lipstick: [#1690] Aligned authenticated- and login nav buttons

jiromaykin:
Login button on the right (on desktop) should be aligned to right-side of container, even when Onderwerpen are made invisible.
https://taiga.maykinmedia.nl/project/open-inwoner/task/1690


jiromaykin:
Adding @pi-sigma as a reviewer because I'm wondering if this will still look good on Linux and/or weird browsers.

pi-sigma (approved):

alextreme (approved):

PR #748, opened 2023-08-25
[#1427] Implemented OpenZaak status translation

Bartvaderkin:
This touches a lot of files, but basically just adds a simple lookup model and helper code, an admin with import/export and tests for the lookup/translate code.


src/open_inwoner/components/templatetags/status_tags.py
Bartvaderkin, line 1: :partying_face:

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -29,7 +29,19 @@ <h1 class="h1" id="title">{{ case.description }}</h1>
             {#  Status history. #}
             {% if case.statuses %}
                 <h2 class="h2" id="statuses">{% trans 'Status' %}</h2>
-                {% status_list case.statuses %}
```
Bartvaderkin, line 32: Begone!

src/open_inwoner/components/templatetags/status_tags.py
pi-sigma:
```
while template_tags:
    template_tags.pop()
```

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -221,6 +224,19 @@ def get_initiator_display(self, case: Zaak) -> str:
             return ""
         return ", ".join([get_role_name_display(r) for r in roles])
 
+    def get_statuses_data(
+        self, statuses: List[Status], lookup: TranslationLookup
+    ) -> List[dict]:
```
pi-sigma, line 229: You can use the built-in type `list` for the annotation since we don't need to support Python < 3.9

src/open_inwoner/utils/translate.py
```diff
@@ -0,0 +1,38 @@
...
+
+from glom import glom
+
+
+class TranslationLookup:
+    """
+    simple key value lookup seeded from queryset.values_list(key, value)
+    """
+
+    def __init__(self, key_value_iterable: Iterable[Tuple[str, str]]):
```
pi-sigma, line 11: `Tuple` > `tuple` and delete import

alextreme (approved):

PR #749, opened 2023-08-28
:lipstick: [#1688] Changing primary color into secondary for text-links

jiromaykin:
in order to make the use of secondary color more consistent
(primary only used for buttons and primary actions, accentkleur only used for 'huisstijl)

https://taiga.maykinmedia.nl/project/open-inwoner/task/1688


jiromaykin:
@Bartvaderkin @pi-sigma so I THINK I have changed all remaining 'text-like' elements and gave them a secondary color - but let me know if you found more that I forgot.

pi-sigma:
Looking good for the most part. However I don't get the background color for the content block consistently. It seems to work for some colors but not for others.

Screenshot 2023-08-30 at 11-32-47 Logeeropvang - Open Inwoner Deventer

pi-sigma (changes requested):
Please double-check the background color for the contact box to see if it works as intended.

jiromaykin:
Aloha @pi-sigma : I have used the current secondary colors our current municipalities are using but ofcourse you can use more. Plus: I have also added a pseudo-element, to deal with the background, that needs to be based on the secondary color and only changes its opacity (turns out HSL _only_ deals with brightness/brilliancy and saturation, and _not_ with opacity).

Test colors:
Groningen: rgb(0, 101, 189) / accent: rgb(224, 0, 26)
Zwolle: rgb(70, 107, 180)
Enschede: rgb(223, 7, 21)
Deventer: rgb(77, 77, 77) / primary: rgb(0, 116, 173)
Zaanstad: rgb(168, 168, 163) / primary: rgb(0, 166, 199)
Leeuwarden: rgb(0, 150, 219)
Rotterdam: rgb(0, 129, 31)

src/open_inwoner/templates/pages/product/detail.html
```diff
@@ -129,7 +129,7 @@ <h3 class="h3" id="see">{% trans "Zie ook" %}</h3>
 
             {% render_column span=9 %}
                 {% if object.contacts.exists %}
-                    {% render_notification type='success' icon=False %}
```
Bartvaderkin, line 132: Looks like this was the only (remaining) use of this template-tag so we could remove it.

Bartvaderkin (approved):
We could remove that template tag (either in this PR, or quick next one after merge)

PR #750, opened 2023-08-29
:lipstick: [#1697] Hiding menu-button on mobile if Onderwerpen is hidden

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/task/1697


jiromaykin:
@pi-sigma I wonder if my simple solution is correct here.

pi-sigma (approved):

PR #756, opened 2023-09-05
[#1707] Remove double search bar from search page

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/task/1707

### DIFF
--- a/src/open_inwoner/components/templatetags/filter_tags.py
+++ b/src/open_inwoner/components/templatetags/filter_tags.py
@@ -4,12 +4,12 @@ register = template.Library()
 
 
 @register.inclusion_tag("components/Filter/Filter.html")
-def filter(field, **kwargs):
+def filters(field, **kwargs):
     """
     Building the filter options for the search page.
 
     Usage:
-        {% filter field=field %}
+        {% filters field=field %}
 
     Variables:
         + field: Field | This is a form field.
--- a/src/open_inwoner/search/tests/test_page.py
+++ b/src/open_inwoner/search/tests/test_page.py
@@ -295,7 +295,7 @@ class SearchPagePlaywrightTests(
                 expect(checkbox).to_be_enabled()
                 expect(checkbox).not_to_be_checked()
 
-        def _click_checkbox_for_name(name):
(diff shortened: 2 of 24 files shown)


jiromaykin:
@Bartvaderkin with this issue the Search bar inside the main content on the Search page needs to be removed; but... tests have been built for that particular Search bar - we now have to use the Desktop search bar and the Mobile Search bar in the navigation and test those instead 
(plus also: some municipalities want the search to be invisible for unauthorized users as well).

Bartvaderkin:
As discussed I'm working on this as we have to rework the various search forms and controls.

src/open_inwoner/components/templatetags/filter_tags.py
```diff
@@ -4,12 +4,12 @@
 
 
 @register.inclusion_tag("components/Filter/Filter.html")
-def filter(field, **kwargs):
+def filters(field, **kwargs):
```
Bartvaderkin, line 7: This is just a cosmetic change so it doesn't shadow the global `filter` function.

src/open_inwoner/search/tests/test_page.py
```diff
@@ -129,3 +129,193 @@ def test_pagination_links(self):
             self.assertEqual(
                 link["href"], f"?query=content&tags={self.tag.slug}&page=2"
             )
+
+
+@tag("e2e")
+@tag("elastic")
```
Bartvaderkin, line 135: @pi-sigma This test is both e2e and elastic.. this might be a problem in our CI setup.

Bartvaderkin: I think I fixed it by adding `--exclude-tag=e2e` to the `elastic` test run, and adding the elasticsearch service to the `e2e` phase (which works because they also aren't run in parallel).

pi-sigma: In hindsight, the configuration with two exclusion tags was bound to create this issue, but I was too focused on making CI work for the tests we had at that time. Good to see you got it working though...

pi-sigma (approved):

PR #758, opened 2023-09-07
[#1709] Add info for when there are zero search results

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/task/1709

### DIFF
--- a/src/open_inwoner/templates/pages/search.html
+++ b/src/open_inwoner/templates/pages/search.html
@@ -5,9 +5,7 @@
     {% render_form form=form method="GET" id="search-form" %}
         {#  zoek button section #}
         <div class="grid">
-            {% spaceless %}<div class="grid__sidebar">
-                &nbsp;
-            </div>{% endspaceless %}
+            <div class="grid__sidebar"></div>
             <div class="grid__main">
                 <h1 class="h1">Zoeken</h1>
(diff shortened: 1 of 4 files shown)


src/open_inwoner/templates/pages/search.html
```diff
@@ -48,10 +48,29 @@ <h3 class="search-results__item-title">{{ hit.name }}</h3>
                     {% pagination page_obj=page_obj paginator=paginator request=request %}
                 </div>
             </div>
+        {% else %}
+            <div class="grid">
+                {% spaceless %}
+                <div class="grid__sidebar">&nbsp;</div>
+                {% endspaceless %}
```
Bartvaderkin, line 55: Does this/theses `spaceless` tags around a single tag actually do anything? I see similar on line 75 but it looks strange, like there would still be space before/after the template tags?

jiromaykin: @Bartvaderkin Ah yes, I thought it was a styling tag, but it's not. Spaceless tag is a standard Django template tag, used because newlines/spaces in templates behave like a character. Only whitespaces between successive opening HTML tags and successive closing tags are removed, so not effective for single elements.

Bartvaderkin (approved):

PR #762, opened 2023-09-11
[#1739] :bug: Fix infinite loop for search page filters

jiromaykin:
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1739

+ Preferably use '`let` 'instead of '`var`' for scope
+ Replaced `setInterval` with `setTimeout` to avoid continuous looping. Code inside setTimeout will execute once after the specified delay (250 milliseconds) 
+ Setting `timerId` to null instead of zero is a common practice in JavaScript when using variables to represent timers or intervals + makes it clear that the timerId variable is not currently associated with any active timer - a way to indicate that there is no ongoing timer operation.

### DIFF
--- a/src/open_inwoner/js/components/search/index.js
+++ b/src/open_inwoner/js/components/search/index.js
@@ -11,15 +11,17 @@ const radioButtons = document.querySelectorAll(
   })
 })
 
-var timerId = 0
+let timerId = null
 
 const searchForm = document.getElementById('search-form')
 
 const filterButtons = document.querySelectorAll('.filter .checkbox__input')
 ;[...filterButtons].forEach((checkbox) => {
   checkbox.addEventListener('change', (event) => {
-    clearInterval(timerId)
-    timerId = setInterval(() => {
+    clearTimeout(timerId)
+
+    // Set a new timeout
+    timerId = setTimeout(() => {
       searchForm.submit()
     }, 250)
   })


src/open_inwoner/js/components/search/index.js
```diff
@@ -11,15 +11,17 @@ const radioButtons = document.querySelectorAll(
...
 
-var timerId = 0
+let timerId = null
 
 const searchForm = document.getElementById('search-form')
 
 const filterButtons = document.querySelectorAll('.filter .checkbox__input')
 ;[...filterButtons].forEach((checkbox) => {
   checkbox.addEventListener('change', (event) => {
     clearInterval(timerId)
```
Bartvaderkin, line 21: This should be `clearTimeout(timerId)`. Although this works, from MDN:

> It's worth noting that the pool of IDs used by [setInterval()](https://developer.mozilla.org/en-US/docs/Web/API/setInterval) and [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout) are shared, which means you can technically use clearInterval() and [clearTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout) interchangeably. However, for clarity, you should avoid doing so.

Bartvaderkin (changes requested):
Wow, an interval.. :confused: :shrug:

src/open_inwoner/js/components/search/index.js
jiromaykin: Oops, let's completely get rid of that interval....

src/open_inwoner/js/components/search/index.js
```diff
@@ -11,15 +11,17 @@ const radioButtons = document.querySelectorAll(
...
 const searchForm = document.getElementById('search-form')
 
 const filterButtons = document.querySelectorAll('.filter .checkbox__input')
 ;[...filterButtons].forEach((checkbox) => {
   checkbox.addEventListener('change', (event) => {
-    clearInterval(timerId)
-    timerId = setInterval(() => {
+    clearTimeout(timerId)
+
+    // Set a new interval
```
pi-sigma, line 23: Should be "Set a new timeout", since you've replaced `clearInterval` with `clearTimeout` following Bart's suggestion.

pi-sigma (changes requested):
The code comment is misleading, other than that looks fine.

pi-sigma (approved):

Bartvaderkin (approved):

PR #763, opened 2023-09-11
[#1711] [#1710] Make search filters conditional and add mobile toggle

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1711
and also tackling issue  #1710, because they are dependent: https://taiga.maykinmedia.nl/project/open-inwoner/task/1710

- Not sure if I should add all 4 options with the first one being a checkbox to turn ALL filtering OFF immediately, or if I should just show the 3 options for Categories/Tags/Organizations, so this is what I built,

EDIT: removed the first filter-option and just left the other 3.

Screenshot 2023-09-14 at 18 54 30


src/open_inwoner/js/components/search/filter-options.js
```diff
@@ -0,0 +1,26 @@
...
+
+const searchForm = document.getElementById('search-form')
+
+const filterButtons = document.querySelectorAll('.filter .checkbox__input')
+;[...filterButtons].forEach((checkbox) => {
+  checkbox.addEventListener('change', (event) => {
+    clearInterval(timerId)
+    timerId = setInterval(() => {
+      searchForm.submit()
+    }, 250)
```
Bartvaderkin, line 24: Interval :wink:

Bartvaderkin:
`{% spaceless %}` is not a free operation and a bit of a kludge so I'm having doubts it is a good idea let's not slap that everywhere for cosmetics on html source nobody looks at. You can get a long way with a bit of attention to the template and how to nest raw html and template tags.

Also note in existing cases `{% spaceless %}`  was also used in the same vein as the `<p class="p">` cargo cult: it was a kludge for a particular misunderstood problem which then gets replicated but according to spec and everything shouldn't have to be a problem.

src/open_inwoner/configurations/models.py
```diff
@@ -316,6 +316,38 @@ class SiteConfiguration(SingletonModel):
...
     )
+
+    # search filters
+    search_filters = models.BooleanField(
+        verbose_name=_("Use filters"),
+        default=True,
+        help_text=_(
+            "Whether to show any checkboxes in order to filter the search results. If this is unchecked, none of the options below will work"
+        ),
+    )
```
Bartvaderkin, line 327: I don't think we need the first boolean as we already cover all options with the other three. A global toggle would make sense if there where a lot of options and there was need to disable them all at once.

src/open_inwoner/templates/pages/search.html
```diff
@@ -20,10 +20,26 @@ <h1 class="h1">Zoeken</h1>
         {#  facets and results section #}
         {% if results %}
             <div class="grid">
-                <aside class="grid__sidebar" aria-label="{% trans "Zoekfilters" %}">
-                    {% filter field=form.categories %}
-                    {% filter field=form.tags %}
-                    {% filter field=form.organizations %}
+                <aside class="grid__sidebar grid__filters" aria-label="{% trans "Zoekfilters" %}">
+                    <h2 class="h2">{% trans "Zoekfilters" %}</h2>
```
Bartvaderkin, line 24: This title stays even when I disable all the search options, not sure if that is intented?

src/open_inwoner/js/components/search/filter-mobile.js
```diff
@@ -0,0 +1,35 @@
+export class FilterMobile {
+  static selector = '.filter__mobile--toggle'
+
+  constructor(node) {
+    this.node = node
+    this.filterList = node.closest('.filter') // Find the closest parent with class 'filter'
+    this.isSmallscreen = window.innerWidth < 768 // Check if it's a mobile screen
```
Bartvaderkin, line 7: In general it is a bad idea to put magic numbers (and decisions based on them) in random files. We don't want every component all have their own definitions of what is a small screen (mobile screen?).

src/open_inwoner/js/components/search/filter-mobile.js
```diff
@@ -0,0 +1,35 @@
+export class FilterMobile {
+  static selector = '.filter__mobile--toggle'
+
+  constructor(node) {
+    this.node = node
+    this.filterList = node.closest('.filter') // Find the closest parent with class 'filter'
```
Bartvaderkin, line 6: `.closest()` is often a kludge.

In general component hydration should work from the outside in: the selector selects the outer element, and the constructor grabs sub-elements chained deeper in that node. 

The idea of components is they're self contained, so when using closest() you're just grabbing and attaching to random parts outside the components responsibility. Sometimes it is necessary when there is a shared thing outside the component but is it here?

src/open_inwoner/js/components/search/filter-mobile.js
```diff
@@ -0,0 +1,35 @@
...
+      this.filterList.classList.remove('filter--open')
+    }
+  }
+}
+
+/**
+ * Controls the toggling of filter lists on mobile to view more
+ */
+const filterToggles = document.querySelectorAll(FilterMobile.selector)
+;[...filterToggles].forEach((filterToggle) => new FilterMobile(filterToggle))
```
Bartvaderkin, line 35: Let's start removing this cargo-culted code pattern and just use forEach on the nodeList:

````js
document.querySelectorAll(FilterMobile.selector).forEach((filterToggle) => new FilterMobile(filterToggle))
````

Later we'll move them to the index.

This goes for every `querySelectorAll()` & `;[...xxx]` wart.

File: src/open_inwoner/components/templates/components/Filter/Filter.html [hunk 7]
```diff
@@ -1,15 +1,18 @@
-{% load i18n filter_tags icon_tags form_tags %}
+{% load i18n filter_tags icon_tags button_tags form_tags %}
 
+{% spaceless %}
```
Bartvaderkin, line 3: See comment in the Conversation.

Putting spaceless here would just turn this whole structure into a oneliner, why would that be good?

src/open_inwoner/templates/pages/search.html
```diff
@@ -20,10 +20,26 @@ <h1 class="h1">Zoeken</h1>
...
+                    {% if search_filters %}
+                        {% if search_filter_categories %}
+                            {% if form.categories|length > 0 %}
```
Bartvaderkin, line 27: This `|length > 0` looks unusual. 

If you evaluate `form.somefield` in a template then Django will in fact render the html output, eg: ``{{ form.somefield }}``. 

So what this line does is render the default widget, then check if the html string has any length, and if so render it again using our custom widgets. That is not efficient and probably not what we really want to do here.

Bartvaderkin (changes requested):

Bartvaderkin:
I'm missing tests for this but I'll have to do more work on this and possibly restructure anyway.

src/open_inwoner/templates/pages/search.html
jiromaykin: Yes, with this weirdness I was trying to not display the ugly empty list after search (checkboxes are currently completely removed if there are no results for filters that need to exist) - but with Alex decided to move this behaviour to a new issue
(https://taiga.maykinmedia.nl/project/open-inwoner/task/1749).

src/open_inwoner/js/components/search/filter-mobile.js
jiromaykin: @Bartvaderkin Okay, that would mean that, in future, we would need a bunch of imported Classes + initialized constants (with DOMContentLoaded event listener?) in the index file?

Bartvaderkin: Yes, in the future we'll add them all to the list in `open_inwoner/js/components/index.js` so we have a good overview of all the components, and they activate on page load as well as in htmx fragements.

Bartvaderkin (approved):
I'll approve it and leave the `filter-options.js` for later when we address the components javascript.

Bartvaderkin:
I'm merging this for an interesting rebase.

PR #764, opened 2023-09-12
[#1735] Fix cookiebanner position for mobile devices

jiromaykin:
This issue only occurs on mobile devices (margins on desktop are not affected by scroll or size of URL bar)
PR can only be reviewed on mobile devices, so either use Gronk or use Browserstack.

issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1735


alextreme (approved):
Let Marie do a double-check on the testenvironment with Browserstack

jiromaykin:
@alextreme Yes this one is a bit difficult to check and I'm not sure if Browserstack is truly 100% native - I may need to buy an android so I can use NGROK
(https://ngrok.com/)

PR #767, opened 2023-09-18
[#1696] Added registration_text to SiteConfiguration and registration_necessary page

Bartvaderkin:


src/open_inwoner/accounts/templates/accounts/registration_necessary.html
```diff
@@ -1,13 +1,15 @@
...
 
 {% block content %}
 
 <div class="registration-grid">
 {% render_grid %}
     {% render_column span=9 %}
         {% render_card tinted=True  %}
+            {% get_solo 'configurations.SiteConfiguration' as config %}
             <h1 class="h1">{% trans "Registratie voltooien" %}</h1><br>
+            {% if config.registration_text %}<p class="p">{{ config.registration_text|urlize|linebreaksbr }}</p><br>{% endif %}
```
pi-sigma, line 12: Perhaps add an `id` or something specific to the `<p>` element so you can target it directly in the test. See comment below.

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -964,13 +964,27 @@ def test_non_digid_user_can_edit_profile(self):
...
+        config.save()
+
+        user = UserFactory(
+            first_name="",
+            last_name="",
+            login_type=LoginTypeChoices.digid,
+        )
+        response = self.app.get(self.url, user=user)
+        self.assertContains(response, "Hello registration text")
+        self.assertContains(response, ' href="http://foo.bar/" ')
```
pi-sigma, line 986: I tend to follow the rule that blanket assertions against an entire content are OK when they are negative ("Assert that xyz is not in content"), but a positive assertion should target specific elements in the HTML (using pyquery or something like it). Not sure if this is right or if it's sometimes OK to assert against the entire content...

src/open_inwoner/accounts/templates/accounts/registration_necessary.html
Bartvaderkin: Ideally we'd have locators and their assertions like in Playwright so we can target by role or CSS rules.

src/open_inwoner/accounts/tests/test_auth.py
Bartvaderkin: It'd be more accurate to use the structure to test but the strings are unique enough and I don't feel like updating the whole file with playwright.

PR #768, opened 2023-09-18
[#1703] Hidden BRP country from My Data table when not set

Bartvaderkin:


pi-sigma:
Changed the title of the PR: "hidden" is the adjective, not the verb in the imperfect.

pi-sigma (approved):

Bartvaderkin:
Changed the title back because I want the first word in the commit message to be meaningful so we can scan a list of commit messages and quickly determine what is interesting for the task at hand,

PR #769, opened 2023-09-18
[#1701] Changed User.get_full_name() to prefer .display_name over .first_name

Bartvaderkin:
Ability to change this centrally shows why DRY code (like a method) is preferred over spreading ad-hoc code.

### DIFF
--- a/src/open_inwoner/accounts/models.py
+++ b/src/open_inwoner/accounts/models.py
@@ -313,8 +313,9 @@ class User(AbstractBaseUser, PermissionsMixin):
         return self._seed
 
     def get_full_name(self):
-        first_name = self.display_name or self.first_name
-        parts = (first_name, self.infix, self.last_name)
+        # validator allowed spaces as values
+        first_name = self.display_name.strip() or self.first_name.strip()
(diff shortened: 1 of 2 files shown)


src/open_inwoner/accounts/models.py
```diff
@@ -313,7 +313,8 @@ def seed(self):
         return self._seed
 
     def get_full_name(self):
-        parts = (self.first_name, self.infix, self.last_name)
+        first_name = self.display_name or self.first_name
```
pi-sigma, line 316: The `display_name` field uses `CharFieldValidator`, which allows strings containing only spaces. We probably don't want to prefer a display_name containing only space. We could should `strip()` and use `isspace()` to check for that or change the validator on the model field to prevent this from arising in the first place.

pi-sigma (changes requested):

src/open_inwoner/accounts/models.py
Bartvaderkin: Ugh that's not great. It is unlikely we have only-spaces values because form-fields use .strip() by default but I modified the code to be safe.

pi-sigma (approved):

PR #770, opened 2023-09-18
[#1705] Changed profile PDF export to not show headers of empty sections

Bartvaderkin:


src/open_inwoner/templates/export/profile/profile_export.html
```diff
@@ -36,22 +36,29 @@ <h1 class="h1">{{ object.get_full_name }}</h1>
...
+    {% with object.actions.visible as actions %}
+    {% if files %}
+        <h2 class="h2 h2--extra">{% trans "Acties" %}</h2>
+        {% include "export/include/actions.html" with actions=actions %}
+    {% endif %}
+    {% endwith %}
 
-    <h2 class="h2 h2--extra">{% trans "Bestanden" %}</h2>
-    {% if object.get_all_files %}
-        {% file_table files=object.get_all_files download_view="profile:documents_download" %}
```
Bartvaderkin, line 54: @pi-sigma Existing example of querying the database for a result-list but only using it for a boolean check, then discarding the results, and then calling the same exact function again (with all the queries en IO and everything) to actually process.

PR #772, opened 2023-09-19
:white_check_mark: [#1640] Add playwright tests for cookie banner

stevenbal:


src/open_inwoner/configurations/tests/test_analytics.py
```diff
@@ -111,3 +115,57 @@ def test_siteimprove_analytics_not_rendered(self):
...
+    def setUp(self):
+        from open_inwoner.accounts.tests.factories import DigidUserFactory
+
+        super().setUp()
+
+        self.user = DigidUserFactory.create()
+        self.user_login_state = self.get_user_bsn_login_state(self.user)
+
+    def test_cookie_accept(self):
+        config = SiteConfiguration.get_solo()
```
Bartvaderkin, line 132: If you use a Solo object like this it is safer to also use the `ClearCachesMixin` so it doesn't leak data across tests. 

Alternately you can use the patch/mock solution (eg: `@patch("open_inwoner.configurations.models.SiteConfiguration.get_solo")` etc) to get a clean instance (although I had weird issue with text fields using this so used the regular get/save).

Example: it is not clear if there would be a `config.cookie_info_text` set here even though it is the default of the model.

src/open_inwoner/configurations/tests/test_analytics.py
```diff
@@ -111,3 +115,57 @@ def test_siteimprove_analytics_not_rendered(self):
...
+
+        context = self.browser.new_context(storage_state=self.user_login_state)
+
+        page = context.new_page()
+        page.goto(self.live_reverse("profile:action_list"))
+
+        expect(page.get_by_text(config.cookie_info_text)).to_be_visible()
+
+        accept_button = page.get_by_text(_("Alles toestaan"))
+        accept_button.click()
```
Bartvaderkin, line 142: We have an ongoing discussion if we should do a more specific check to see if a page contains a certain string instead of just checking the global page. 

So maybe we can locate the actual popup element, check if it visible and then check the text and buttons chained from that (you can call `.locator()`/get_by_xyz etc on a Locator to make more specific locations).

Bartvaderkin (changes requested):
Cool, thanks! I've added some small notes relating to the oddities of the project.

src/open_inwoner/configurations/tests/test_analytics.py
stevenbal: You're right about the second approach (I'd have to explicitly set the value of `config.cookie_info_text`), so I'll stick with the mixin

alextreme (approved):

PR #773, opened 2023-09-19
[#1713, #1714] Added basic video model, player & cms-plugin

Bartvaderkin:
What is in here?

- basic Video model + admin (from Anne F.)
- CMS plugin + model (to put a Video on CMS pages)
- Change to Product model + template to add Video on Product detail view

For clarity you can view the separate commits.

Possibly @jiromaykin want's to do some CSS or template work so the player adjusts size to available space? If so either amend the PR now, or create a ticket for this sprint.


jiromaykin:
@Bartvaderkin Yes, cool, the video frame seems to have a set width to 640px + could use a top-margin; and I'll need to make it responsive somehow :)
I've created a new issue (#1756) for this and will probably branch off from this one.

src/open_inwoner/pdc/models/product.py
```diff
@@ -101,6 +101,15 @@ class Product(models.Model):
             "Product content with build-in WYSIWYG editor. By adding '[CTABUTTON]' you can embed a cta-button for linking to the defined form or link"
         ),
     )
+    video = models.ForeignKey(
```
pi-sigma, line 104: Is this what we want? A `ForeignKey` from `A` to `B` defines a many-to-one relationship. I this case, many products can be associated with a single video (as in: one video covers different products), but each product can have only one video. My first thought is that, if anything, it should be the other way around (you could have several videos for a single product, highlighting different aspects perhaps), or perhaps even have  a many-to-many relationship.

src/open_inwoner/scss/components/Video/Video.scss
```diff
@@ -0,0 +1,14 @@
+.video {
+  &__container {
```
pi-sigma, line 2: Check with @jiromaykin if he wants to fill this in, else delete the empty classes

pi-sigma (changes requested):

src/open_inwoner/media/models/video.py
```diff
@@ -0,0 +1,76 @@
...
+            middle_truncate(self.title, 50),
+            self.player_type,
+            self.link_id,
+            self.language,
+        )
+
+    @property
+    def external_url(self):
+        if self.player_type == VideoPlayerChoices.youtube:
+            url = "https://www.youtube.com/watch?v={link_id}&enablejsapi=1"
```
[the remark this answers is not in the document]
Bartvaderkin: This code was from an older project so used old `.format()`, but I replaced it with f-strings as I should've done before.

src/open_inwoner/pdc/models/product.py
Bartvaderkin: We specifically went for this because it was the simplest solution, until ultimately we convert the pages to be CMS pages (where the editor can add and positions videos freely).

src/open_inwoner/media/choices.py
```diff
@@ -0,0 +1,8 @@
+from django.utils.translation import ugettext_lazy as _
+
+from djchoices import ChoiceItem, DjangoChoices
+
+
+class VideoPlayerChoices(DjangoChoices):
```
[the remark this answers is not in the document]
Bartvaderkin: @pi-sigma I made Taiga 1769 issue to swap the old choices for modern native ones.

pi-sigma (approved):

PR #775, opened 2023-09-20
[#1740] Removed old data migration tests

pi-sigma:
Taiga [#1740](https://taiga.maykinmedia.nl/project/open-inwoner/task/1740)


alextreme (approved):
Old migration-related tests, can indeed be removed

PR #778, opened 2023-09-22
[#1749] Changed search to always show facet checkboxes, including non-matching choices

Bartvaderkin:


src/open_inwoner/search/views.py
```diff
@@ -63,7 +63,7 @@ def search(self, form):
         # update form fields with choices
         for facet in results.facets:
             if facet.name in form.fields:
-                form.fields[facet.name].choices = facet.choices()
+                form.fields[facet.name].choices = facet.total_choices()
```
jiromaykin, line 66: the requirement is to show (0) if the total is zero, so the UX communication about the results is more clear - is is possible to add that?

Screenshot 2023-09-25 at 11 19 10

src/open_inwoner/templates/pages/search.html
```diff
@@ -11,30 +11,29 @@ <h1 class="h1">{% trans "Zoeken naar " %} "{% firstof search_form.query.value ""
...
-                    {% if search_filter_tags %}
-                         {% filters field=search_form.tags form_id="search-form" %}
-                    {% endif %}
-                    {% if search_filter_organizations %}
-                        {% filters field=search_form.organizations form_id="search-form" %}
-                    {% endif %}
-            {%  endif %}
-            {#  end search filters #}
-            </aside>
+    <div class="grid">
```
jiromaykin, line 14: Ah nice - does this mean it's no longer dependent on the pagination?

jiromaykin (changes requested):
Only needs zeros to make zero results more clear. All else neat.

src/open_inwoner/templates/pages/search.html
Bartvaderkin: Not 100% sure what you mean but with this change the facet checkboxes always show even when zero results.

The reason the "do we have results" check used the paginator is because the paginator was conveniently available in the context.

src/open_inwoner/search/views.py
Bartvaderkin: Of course. I appended a change to the PR.

jiromaykin (approved):

PR #780, opened 2023-09-25
[#1745] Add default styles for some HTML tags

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1745


jiromaykin:
Closing this because it should be part of a continuous refactoring and will also be corrected by implementing NL Design-System.

PR #781, opened 2023-09-25
:memo: Add install instructions for ElasticSearch

stevenbal:


Bartvaderkin (approved):
I wasn't even aware we had a docker script for this, I was using an ubuntu service.

PR #783, opened 2023-09-25
:sparkles: [#1750] Add meta tags for social media sharing

stevenbal:


src/open_inwoner/pdc/views.py
```diff
@@ -216,6 +216,7 @@ def get_context_data(self, **kwargs):
         if product.contacts.exists():
             anchors.append(("#contact", _("Contact")))
 
+        context["meta_description"] = product.summary
```
stevenbal, line 219: @alextreme I'm not entirely sure if this is the right field to use as description in the meta tags, what do you think?

alextreme: Using the product summary is fine

src/open_inwoner/templates/master.html
```diff
@@ -4,6 +4,14 @@
         <meta charset="utf-8">
         <title>{% block title %}{% if page_title %}{{ page_title }} - {% endif %}{{ site_name }}{% endblock %}</title>
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+        <meta property="og:title" content="{% if page_title %}{{ page_title }} - {% endif %}{{ site_name }}">
```
alextreme, line 8: I would expect `<meta name="description" content="..."/>` here too

alextreme (changes requested):
After adding name="description" this can be merged

Bartvaderkin (dismissed):
I miss a few basic tags that matter and we have data available for, like og:image and twitter:image (every product has an image), og:url (product.get_absolute_url() eg without random query params) and og:type (hardcoded as 'article')

alextreme (approved):

alextreme:
Mag gemerged worden zodra deze groen is

PR #784, opened 2023-09-25
[#1761] Make sidebar anchor headings consistent

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1753 

in product anchor sidebar
- remove H1 title 
- make all H2's same font-size


jiromaykin:
@pi-sigma This one will look familiair to you - I'm not sure if this still needs the same tests.

src/open_inwoner/pdc/utils.py
```diff
@@ -19,7 +19,7 @@ def extract_subheadings(content: str, tag: str) -> list[tuple[str, str]]:
     subs = []
     for tag in soup.find_all("h2"):
         subheading = tag.text
-        slug = f"subheader-{slugify(subheading)}"
-        subs.append((subheading, slug))
+        slug = f"#{slugify(subheading)}"
```
pi-sigma, line 22: Is there a reason you removed the "subheader"? This was intended as a namespace for the slug.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -58,7 +58,7 @@ def get_product_rendered_content(product):
             element.attrs["class"] = class_name
 
             if tag == "h2":
-                element.attrs["id"] = f"subheader-{slugify(element.text)}"
+                element.attrs["id"] = f"{slugify(element.text)}"
```
pi-sigma, line 61: See previous comment

pi-sigma (changes requested):

src/open_inwoner/pdc/utils.py
jiromaykin: Well, now they aren't "subheaders" anymore - all these **headings** (not _headers_, headers are different things) are all on the same level now, with the exact same functionality (see how this is all removed from the HTML template too). All headings are unique anyway, so they do not need extra's to clog their slug?

Instead of subheadings they are now more like 'article/editor' headings I'd say.
It _could_, however, be nice to differentiate these 'article/editor' headings from the 'block' headings (=like Locations/Contact etc.) _only_ in the list-items in `AnchorMenu.html,` for later use in PR #785  - so I would like to wrap them differently in the HTML and change that in my PR, but then they would still not need a highly specific slug for the front-end, so I would still remove that.

Let me know what you think, or if this is unclear.

pi-sigma: The reason to introduce the namespace was not the distinction between headers and "subheaders", but the fact that the slug was used as `id` for the `h2` tag (`element.attrs["id"] = f"{slugify(element.text)}`). The point of the namespace is to exclude possible collisions.

jiromaykin:
@pi-sigma @Bartvaderkin Hopefully this is better now - for some reason the scrolling wasn't neat anymore now all the anchors are on the same level. I had to change the components that had their ID attached to their surrounding wrappers and set those ID's in their Headings in order to make the scrolling to the anchors look consistent (so: for the FAQ and Locations components). 
Feel free to give any feedback.

pi-sigma (approved):

PR #785, opened 2023-09-25
[#1748] Add option to make product-detail content collapse

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1748

- [x] Set Boolean false by default: show CKeditor content visible as normal.
- [x] If Readmore checkbox is set: show the collapse button and make collapsing true by default (=toggle button).
- [x] also: If admin checks collapse to true, then: hide subheadings in Sidebar (=hide CKeditor content headings there so scroll-anchors only work for appended anchors)

Note: the Anchorscrolling won't work properly if the Django-CMS Toolbar is visible/edit-mode.
Preview in this screencast: Approved by designer Jacki.

edit-product

⬆️ how to edit product, turns content into a collapse:

https://github.com/maykinmedia/open-inwoner/assets/118177951/d1f9ae94-0907-467a-bf3a-d6e232030c5a


jiromaykin:
Note to myself for later translations: 
English is the basis and after translations it will all be like this:
`<a class="readmore__toggle button button--textless button--icon button--icon-after button--secondary button--transparent" href="#" title={% trans "Klik hier om de tekst uit te klappen" %} aria-label={% trans "Klik hier voor meer informatie" %} aria-expanded="false">
                {% icon icon="expand_more" icon_position="after" icon_outlined=True %}
                <span class="readmore__button-text">{% trans "Klik hier voor meer informatie" %}</span>
                <span class="readmore__button-text--hidden">{% trans "Toon minder informatie" %}</span>
            </a>`

src/open_inwoner/pdc/models/product.py
```diff
@@ -59,6 +59,11 @@ class Product(models.Model):
         max_length=300,
         help_text=_("Short description of the product, limited to 300 characters."),
     )
+    read_more = models.BooleanField(
+        verbose_name=_("Read more"),
```
alextreme, line 63: Collapse content

jiromaykin:
@pi-sigma This one has finally turned 'green' again - so if you want and have the time, you can also review this one (I changed the checkbox labeltext into  'collapse content' now)

src/open_inwoner/pdc/views.py
```diff
@@ -217,6 +220,7 @@ def get_context_data(self, **kwargs):
         context["anchors"] = anchors
         context["related_products_start"] = 6 if product.links.exists() else 1
         context["product_links"] = product.links.order_by("pk")
+        context["readmore"] = readmore
```
alextreme, line 223: You don't really need 'readmore' as intermediate variable:

`context["readmore"] = product.read_more`

And adding this variable manually to the context is also not necessary. You can simply use:

`{% if object.read_more %}` 

in the template. Every DetailView automatically has `object` in the context.

pi-sigma (changes requested):
Looking good overall. As we discussed, the variable for expanding/collapsing the content should be renamed, though I'll leave it to you if you prefer `content_is_collapsable`, `content_is_expandable`, or sth. like it.

alextreme (approved):

PR #787, opened 2023-09-26
[#1715] Added management command to delete expired contact invitations

pi-sigma:
Taiga [#1715](https://taiga.maykinmedia.nl/project/open-inwoner/task/1715)

### DIFF
--- a/src/open_inwoner/accounts/management/commands/delete_invitations.py
+++ b/src/open_inwoner/accounts/management/commands/delete_invitations.py
@@ -1,0 +1,38 @@
+from collections import defaultdict
+from datetime import timedelta
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+from open_inwoner.utils.logentry import system_action
+
+from ...models import Invite
+
+
+class Command(BaseCommand):
+    help = (
+        "Delete contact invitations which are older than a predefined period "
+        "(the default is defined in `settings.INVITE_EXPIRY_DAYS`)"
+    )
+
+    def handle(self, *args, **kwargs):
+        now = timezone.now()
+        expired_invitations = Invite.objects.select_related("inviter").filter(
+            created_on__lt=now - timedelta(days=settings.INVITE_EXPIRY_DAYS)
+        )
+
+        grouped = defaultdict(list)
+
+        for invite in expired_invitations:
+            grouped[invite.inviter].append(invite)
+
+        for inviter, invites in grouped.items():
+            invites_info = [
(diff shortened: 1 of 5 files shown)


Bartvaderkin (changes requested):

src/open_inwoner/accounts/management/commands/delete_invitations.py
```diff
@@ -0,0 +1,34 @@
...
+        for invite in expired_invitations:
+            invitee_email = invite.invitee_email
+            created_on = invite.created_on.strftime("%Y-%m-%d")
+            invites_info.append((invitee_email, created_on))
+
+        TimelineLog.objects.create(
+            timestamp=now.strftime("%Y-%m-%dT%H:%M:%S"),
+            extra_data={
+                "deleted_invitations": invites_info,
+            },
```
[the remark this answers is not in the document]
pi-sigma: When I use `system_action` I get an error when passing the `content_object` as argument (it's not expected by `TimeLineLog`). Is this actually used anywhere?

src/open_inwoner/accounts/management/commands/delete_invitations.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        inviters = set(invite.inviter for invite in expired_invitations)
+
+        for user in inviters:
+            my_expired_invitations = expired_invitations.filter(inviter__id=user.id)
+            invites_info = [
+                f"{invite.invitee_email} (invited on {invite.created_on.strftime('%Y-%m-%d')})"
+                for invite in my_expired_invitations
+            ]
+            message = f"Expired contact invitations from {user.first_name} {user.last_name} deleted"
```
Bartvaderkin, line 31: *If* we print names we use the methods so we don't forget infixes etc (eg: `user.get_full_name()`).

In this case the name might not be needed as we already set the user on the record. For triage it'd also be helpful to include a count in the message (eg: "Expired N contact invitations')

Bartvaderkin (changes requested):

src/open_inwoner/utils/logentry.py
```diff
@@ -163,5 +168,6 @@ def system_action(
             "action_flag": LOG_ACTIONS[5],
             "message": message,
             "log_level": log_level,
-        },
+        }
+        | kwargs,
```
Bartvaderkin, line 172: Note the existing codebase doesn't really use overloaded operators so I'm not convinced we should sprinkle them around in random places. (eg: it is no longer simplistic code)

Bartvaderkin (approved):

PR #788, opened 2023-09-26
[#1567] Modified contact-forms to send new `subject_code/onderwerp` field to e-Suite

Bartvaderkin:
This covers both contact-forms: the configurations each get a new field with a code that gets sent to e-Suite/OpenKlant


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -507,6 +501,11 @@ def register_by_email(self, form, recipient_email):
     def register_by_api(self, form, config: OpenKlantConfig):
         assert config.has_api_configuration()
 
+        try:
+            ztc = ZaakTypeConfig.objects.filter_case_type(self.case.zaaktype).get()
+        except ObjectDoesNotExist:
```
pi-sigma, line 506: Does `filter_case_type` always return a single object?

Bartvaderkin: It does. It returns a queryset so it can be chained with other filters and situations that want a queryset (eg: `filter_enabled_for_case_type()` and `filter_questions_enabled_for_case_type()`)

alextreme (approved):
Discussed and this looks good, will need testing with the Enschede eSuite Contactmomenten API

PR #789, opened 2023-09-26
[#1656] Added CKEditor widget to category description field

pi-sigma:
Taiga [#1656](https://taiga.maykinmedia.nl/project/open-inwoner/us/1656)


Bartvaderkin (changes requested):
I'd expect this would also require some changes in the place this field is rendered.

If this widget produces/stores HTML it'd needs to be marked safe for output (or it'd get auto-escaped), and if it produces/stores  Markdown then it needs to be rendered.

Maybe a simple test for that would be good?

pi-sigma:
I've used a template tag to render the category description. Tried to avoid this approach by adding the rendered content to the context inside `CategoryDetailView`, but for some reason it doesn't render correctly. Instead of `<p>some <em>italic</em> text</p>` we get something like `[<p>some <em>italic</em> text</p>, <em>italic</em>]`. The parts of the description are rendered, but they show up as comma-separated elements inside a list. Debugging the template didn't yield any results: the `rendered_content` variable has the correct value.

Possibly related to the issue with escape characters we came across earlier (Taiga [#1624](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1624)).

src/open_inwoner/templates/pages/category/detail.html
```diff
@@ -17,7 +17,7 @@ <h1 class="h1">
                 {% button icon="edit" text=_("Open in admin") hide_text=True href="admin:pdc_category_change" object_id=object.pk %}
             {% endif %}
         </h1>
-        <p class="p">{{ object.description|linebreaksbr }}</p>
+				{{ object.description|category_ckeditor_content|safe }}
```
alextreme, line 20: Please attempt to move this either to the Category model or into the view, ask be in case this doesn't render properly

PR #794, opened 2023-10-02
[#1779] Fix alignment of search results

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1779
needs filled sidebar or else checkboxes will align weirdly


pi-sigma (approved):
I find the spacing around the feedback part strange. In particular, the spacing between the question ("Heeft u gevonden wat u zocht?") and the buttons is much larger than that between "Feedback" and the the paginator.  See screenshot It would be nice if the elements that belong together logically would also be grouped together visually. However, since this is not the result of your PR, feel free to address this in a separate ticket (or not at all, I'll leave it to you).

Screenshot from 2023-10-02 16-14-39

PR #795, opened 2023-10-02
:lipstick: [#1556] Styling for selects with empty option selected

stevenbal:


src/open_inwoner/templates/django/forms/widgets/select.html
```diff
@@ -0,0 +1,5 @@
+<select name="{{ widget.name }}" {% if widget.value.0 == "" %}class="{{widget.attrs.class}} muted"{% endif %}{% include "django/forms/widgets/attrs.html" %}>{% for group_name, group_choices, group_index in widget.optgroups %}{% if group_name %}
```
stevenbal, line 1: It doesn't seem to be possible to style a <select> based on the selected <option> using CSS and I didn't want to create a custom widget that would have to be set on each instance of a `...ChoiceField`, so I opted for this approach instead

PR #796, opened 2023-10-02
:lipstick: [#1556] Styling for selects with empty option selected

stevenbal:


src/open_inwoner/templates/django/forms/widgets/select.html
```diff
@@ -0,0 +1,5 @@
+<select name="{{ widget.name }}" {% if widget.value.0 == "" %}class="{{widget.attrs.class}} muted"{% endif %}{% include "django/forms/widgets/attrs.html" %}>{% for group_name, group_choices, group_index in widget.optgroups %}{% if group_name %}
```
stevenbal, line 1: It doesn't seem to be possible to style a `<select>` based on the selected `<option>` using CSS and I didn't want to create a custom widget that would have to be set on each instance of `...ChoiceField`, so I opted for this approach instead

jiromaykin: Note: We put all our Form Component templates in this project in this particular directory here: _src/open_inwoner/components/templates/components/Form_
So I don't think we should put anything into the Django templates Dir.

Also: with this approach we need to create a 1st option inside all Selects that should start with some kind of empty default, like here:
http://127.0.0.1:8000/mijn-uitkeringen/  
So I guess you should add one there :-) 
Not sure what to call it, perhaps "Select month".
Same for:
http://127.0.0.1:8000/mijn-uitkeringen/jaaropgaven/ "Select year"

More pages with selects, that are probably allright: 
http://127.0.0.1:8000/mijn-profiel/contacts/
http://127.0.0.1:8000/mijn-profiel/actions/
And ">> Mijn profiel >> Mijn acties >> Bewerk [Actie title]"

benefits-select

jiromaykin (changes requested):
Not sure if we are allowed to change project hierarchy by putting stuff in a Django template folder, so probably better to move them to our Components Templates 
+ with this general approach we need to add empty options for selects that don't have them set properly yet.

src/open_inwoner/templates/django/forms/widgets/select.html
stevenbal: > Also: with this approach we need to create a 1st option inside all Selects that should start with some kind of empty default, like here:

Just to be sure: do you mean that having `Sep 2023` **not** muted as the initial value would be okay as well? I could modify the widget to check if the initial value is empty or not and apply the class with the muted styling based on that

> More pages with selects, that are probably allright:
http://127.0.0.1:8000/mijn-profiel/contacts/

I checked the contacts page, the change I mentioned above would cause "Alle" to be displayed muted, because it's value is `""`. Is this correct?
image

> Note: We put all our Form Component templates in this project in this particular directory here: src/open_inwoner/components/templates/components/Form
So I don't think we should put anything into the Django templates Dir.

I put the template in `templates/django/forms/widgets/select.html` to override the default widget that Django uses under the hood, which makes this change available for all selects without having to manually add classes to each occurring select. I'm not sure if there is another way to override a default Django widget without having to monkey patch stuff, which seems like it would be unnecessary in this case in my opinion.

If it turns out we do not want this styling applied to each select or if there is no generic solution, the only solution is probably to add the muted class to all selects manually (which could be kind of a chore, I'm not sure how many selects there are :stuck_out_tongue:)

jiromaykin: Indeed: the first option in the selects in the Maandopgaven/Jaaropgaven should **not** be muted because it's already showing a real option - the muted style is a UX indication, so should only be applied to a default option to indicate to the user something can be selected/filtered.
I'm not sure anymore but I tried to tackle this with **CSS-only** a couple of months ago and I think I found that some selects weren't using the value attribute at all, which made them empty, which meant that too many of them had a zero value so it would show the mute style too often.

In my previous answer I posted all of the places that have a Select; so currently i'd think there's only a maximum of 6 selects in the entire project.
And yes: I'd say 'alle' for Contacts should be muted because it means nothing has been filtered by the  select yet.

As for the location of the templates and if we just simply wish to override: I am not sure - I'd have to ask @alextreme or @Bartvaderkin.

stevenbal: > In my previous answer I posted all of the places that have a Select; so currently i'd think there's only a maximum of 6 selects in the entire project.

Ah alright, in that case I think all of them are displayed correctly:

image

image

image

image

image

jiromaykin: If that's true then i will approve this :-) for new selects we'll need to make a correct model with values I guess.

jiromaykin (approved):

PR #801, opened 2023-10-09
:sparkles: [#1992] Link mijn vraag and aanvraag

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/us/1992?no-milestone=1

### DIFF
--- a/src/open_inwoner/apimock/apis/openklant-read/contactmomenten/objectcontactmomenten/aaaaaaaa-aaaa-aaaa-aaaa-ddddddddddd.json
+++ b/src/open_inwoner/apimock/apis/openklant-read/contactmomenten/objectcontactmomenten/aaaaaaaa-aaaa-aaaa-aaaa-ddddddddddd.json
@@ -1,0 +1,5 @@
+{
+    "contactmoment": "https://contactmomenten.nl/api/v1/contactmoment/aaaaaaaa-aaaa-aaaa-aaaa-bbbbbbbbbbbb",
+    "object": "https://example.com/foo",
+    "objectType": "zaak"
+}
--- a/src/open_inwoner/components/templatetags/table_tags.py
+++ b/src/open_inwoner/components/templatetags/table_tags.py
@@ -1,6 +1,7 @@
 from typing import List, TypedDict
 
 from django import template
+from django.urls import reverse
 from django.utils.translation import gettext as _
 
 from open_inwoner.accounts.views.contactmoments import KCMDict
@@ -72,7 +73,7 @@ def case_table(case: dict, **kwargs) -> dict:
 
 
 @register.inclusion_tag("components/table/table.html")
-def contactmoment_table(kcm: KCMDict, **kwargs) -> dict:
+def contactmoment_table(kcm: KCMDict, zaak=None, **kwargs) -> dict:
     """
     Renders a table below the dashboard for additional values related to a Zaak (case).
 
@@ -97,6 +98,23 @@ def contactmoment_table(kcm: KCMDict, **kwargs) -> dict:
         row = {"headers": [{"text": label}], "cells": [{"text": value}]}
 
         rows.append(row)
+
+    if zaak:
+        rows.append(
+            {
+                "headers": [{"text": _("Zaakidentificatie")}],
+                "cells": [
+                    {
+                        "text": zaak.identificatie,
+                        "url": reverse(
+                            "cases:case_detail", kwargs={"object_id": str(zaak.uuid)}
+                        ),
+                        "hyperlink_text": _("Ga naar zaak"),
+                        "class": "zaak-hyperlink",
+                    }
+                ],
+            }
+        )
 
     add_row_if_not_empty("identificatie", _("Identificatie"))
     add_row_if_not_empty("type", _("Type"))
--- a/src/open_inwoner/components/templatetags/table_tags.py
+++ b/src/open_inwoner/components/templatetags/table_tags.py
@@ -1,10 +1,11 @@
-from typing import List, TypedDict
+from typing import List, Optional, TypedDict
 
 from django import template
 from django.urls import reverse
 from django.utils.translation import gettext as _
 
 from open_inwoner.accounts.views.contactmoments import KCMDict
+from open_inwoner.openzaak.api_models import Zaak
 
 register = template.Library()
 
@@ -73,7 +74,7 @@ def case_table(case: dict, **kwargs) -> dict:
 
 
 @register.inclusion_tag("components/table/table.html")
-def contactmoment_table(kcm: KCMDict, zaak=None, **kwargs) -> dict:
+def contactmoment_table(kcm: KCMDict, zaak: Optional[Zaak] = None, **kwargs) -> dict:
     """
     Renders a table below the dashboard for additional values related to a Zaak (case).
 
@@ -110,7 +111,7 @@ def contactmoment_table(kcm: KCMDict, zaak=None, **kwargs) -> dict:
                             "cases:case_detail", kwargs={"object_id": str(zaak.uuid)}
                         ),
                         "hyperlink_text": _("Ga naar zaak"),
-                        "class": "zaak-hyperlink",
+                        "class": "case-detail__link",
                     }
                 ],
             }
--- a/src/open_inwoner/openklant/tests/data.py
+++ b/src/open_inwoner/openklant/tests/data.py
@@ -1,3 +1,4 @@
+from requests.exceptions import RequestException
 from zgw_consumers.constants import APITypes
 from zgw_consumers.test import generate_oas_component, mock_service_oas_get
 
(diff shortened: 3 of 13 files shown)


src/open_inwoner/openklant/wrap.py
```diff
@@ -87,6 +90,59 @@ def fetch_klantcontactmoment(
...
+
+    return ret or []
+
+
+def fetch_objectcontactmoment(
+    contactmoment: ContactMoment, object_type: str
+) -> Optional[ObjectContactMoment]:
+    ocms = fetch_objectcontactmomenten_for_object_type(contactmoment, object_type)
+    ocms = ocms or [None]
+    return ocms[0]
```
stevenbal, line 143: This implementation does assume that a `ContactMoment` can only have one `ObjectContactMoment` with `objectType=zaak`

src/open_inwoner/openklant/tests/data.py
```diff
@@ -26,10 +29,17 @@ def setUpServices(cls):
...
+        oz_config.save()
+
     @classmethod
     def setUpOASMocks(self, m):
         mock_service_oas_get(m, KLANTEN_ROOT, "kc")
         mock_service_oas_get(m, CONTACTMOMENTEN_ROOT, "cmc")
+        mock_service_oas_get(m, ZAKEN_ROOT, "zrc")
 
 
 class MockAPIReadPatchData(MockAPIData):
```
Bartvaderkin, line 45: Let's swap all the `generate_oas_component` for `generate_oas_component_cached` in these MockAPIData classes.

src/open_inwoner/openklant/wrap.py
```diff
@@ -87,6 +90,59 @@ def fetch_klantcontactmoment(
...
+
+    object_contact_momenten = factory(ObjectContactMoment, response)
+
+    # resolve linked resources
+    for ocm in object_contact_momenten:
+        assert ocm.contactmoment == contactmoment.url
+        ocm.contactmoment = contactmoment
+        if ocm.object_type == "zaak":
+            object_path = urlparse(ocm.object).path
+            ocm.object = fetch_single_case(object_path.split("/")[-1])
```
Bartvaderkin, line 118: How likely is it we have multiple contact-moments with the same object? Maybe it would be better to de-duplicate the requests as the API's are slow.

Also it is not good pattern to parse the URLs for UUID's; they are supposed to be opaque. Just make a fetch_single_case() variant that uses a URL.

src/open_inwoner/openklant/wrap.py
Bartvaderkin, line 143: These lines look weird? I had to think an extra second :)
```
if ocms:
   return ocms[0]
```

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -178,6 +180,48 @@ def test_show_detail_for_bsn(self, m):
...
+                ),
+                "channel": data.contactmoment["kanaal"].title(),
+                "text": data.contactmoment["tekst"],
+                "onderwerp": self.contactformsubject.subject,
+                "antwoord": data.contactmoment["antwoord"],
+                "identificatie": data.contactmoment["identificatie"],
+                "type": data.contactmoment["type"],
+                "status": _("Afgehandeld"),
+                "url": detail_url,
+            },
```
Bartvaderkin, line 211: It'd be a hint more readable and black-friendly if we use short variables so the structures don't become too hairy to read (eg: how it takes `kcm` to not repeat `response.context["contactmoment"]` )

src/open_inwoner/components/templatetags/table_tags.py
```diff
@@ -72,7 +73,7 @@ def add_row_if_not_empty(key, label):
 
 
 @register.inclusion_tag("components/table/table.html")
-def contactmoment_table(kcm: KCMDict, **kwargs) -> dict:
+def contactmoment_table(kcm: KCMDict, zaak=None, **kwargs) -> dict:
```
Bartvaderkin, line 76: We have a type here we can use.

src/open_inwoner/openklant/wrap.py
```diff
@@ -87,6 +90,59 @@ def fetch_klantcontactmoment(
...
+    client = build_client("contactmomenten")
+
+    moments = _fetch_objectcontactmomenten_for_contactmoment(
+        contactmoment, client=client
+    )
+
+    # eSuite doesn't implement a `object_type` query parameter
+    ret = [moment for moment in moments if moment.object_type == object_type]
+
+    return ret or []
```
Bartvaderkin, line 135: You don't need the `or []` here.

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -30,6 +30,18 @@
       overflow: hidden;
       word-break: break-word;
       white-space: normal;
+
+      .zaak-hyperlink {
+        float: right;
```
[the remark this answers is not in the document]
stevenbal: I removed the `float` and added a wrapper, I tried using `flex` on the `td`, but that broke the table styling

Bartvaderkin (changes requested):
Some loose stuff.

src/open_inwoner/apimock/apis/openklant-read/contactmomenten/objectcontactmomenten/aaaaaaaa-aaaa-aaaa-aaaa-ddddddddddd.json
```diff
@@ -0,0 +1,5 @@
+{
+    "contactmoment": "https://contactmomenten.nl/api/v1/contactmoment/aaaaaaaa-aaaa-aaaa-aaaa-bbbbbbbbbbbb",
```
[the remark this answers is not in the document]
stevenbal: I think it's because the `objectcontactmomenten` mock endpoint currently always returns the same response, which only appears to work for the other example vraag. Locally I run Open Klant as a docker container (in the `docker` directory) instead of using the mocks

src/open_inwoner/openklant/wrap.py
stevenbal: > How likely is it we have multiple contact-moments with the same object? Maybe it would be better to de-duplicate the requests as the API's are slow.

I think that it will occur sometimes, I'm not entirely sure what you mean by deduplicating though. This function retrieves the objectcontactmomenten for one specific contactmoment (it's called from the Contactmoment detail view)

Bartvaderkin: But in this loop, multiple `ocm`'s could link to the same `ocm.object` right? 

If you create a dict with the URI's as key (and None as value) it naturally de-duplicates the URI's, so then you can load all the unique URI's with `.keys()` and assign each result back to the dict (replacing the None). 

Then go through the `ocm`'s again and grab the resource by URI from the dict.

stevenbal: Ahh like that, I'm not sure if it is possible for a single contactmoment to have multiple relations through objectcontactmoment with the same `object` (zaak), but I could add it just to be sure

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -30,6 +30,24 @@
...
       white-space: normal;
+
+      &.table__item--status {
+        .table__item--wrapper {
+          display: flex;
+          justify-content: space-between;
+        }
+      }
+
+      .zaak-hyperlink {
```
jiromaykin, line 41: Nice! Although instead of 'hyperlink' i'd just call this a '`contactmoment__link`' in BEM style, since this is more specific for contactmoment and not so much for Tables in this project and we decided not to use Dutch ('zaak') if possible + this link needs color-primary as a color. This way it will not interfere with the "contactmomenten__link" class + In future I'll be restyling this page dramatically, but if you could add these tiny changes then I will approve :-)

jiromaykin (changes requested):

src/open_inwoner/components/templatetags/table_tags.py
```diff
@@ -98,6 +100,23 @@ def add_row_if_not_empty(key, label):
...
+        rows.append(
+            {
+                "headers": [{"text": _("Zaakidentificatie")}],
+                "cells": [
+                    {
+                        "text": zaak.identificatie,
+                        "url": reverse(
+                            "cases:case_detail", kwargs={"object_id": str(zaak.uuid)}
+                        ),
+                        "hyperlink_text": _("Ga naar zaak"),
```
jiromaykin, line 113: Oops, I missed this one - we use 'link' in the entire project for hyperlinks, and just the one underscore would not be BEM-like, so at least use two underscores with the blockname first - probably something like `.case-detail__link`

jiromaykin (changes requested):

jiromaykin (approved):

Bartvaderkin (approved):

PR #802, opened 2023-10-09
:lipstick: [#1790] Alter empty my cases overview design

stevenbal:


alextreme:
@pi-sigma please double-check, I noticed that the number of cases isn't shown anymore with this change.

pi-sigma (approved):

PR #804, opened 2023-10-09
[#1788] Refactor case list view

pi-sigma:
Taiga [#1788](https://taiga.maykinmedia.nl/project/open-inwoner/task/1788)

- removed sidebar (we no longer distinguish between open and closed cases)
- included submissions in case list view and created uniform interface for zaken + submissions
- refactored case retrieval
- refactored logging of case access


src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -1,37 +1,39 @@
...
         self.log_user_action(
             self.request.user,
-            _("Zaak bekeken: {case}").format(case=case_identificatie),
+            _("Zaaken bekeken: {cases}").format(cases=", ".join(case_ids)),
+        )
+
+    def log_access_case_detail(self, case: UniformCase):
+        """
+        Log access to a particular case
+        """
```
Bartvaderkin, line 33: This comment doesn't add any information.

src/open_inwoner/cms/cases/views/cases.py
```diff
@@ -6,116 +6,81 @@
...
+        context["cases"] = case_dicts
+        context.update(paginator_dict)
 
-    def get_anchors(self) -> list:
-        return [
-            (reverse("cases:open_submissions"), _("Openstaande aanvragen")),
-            (reverse("cases:open_cases"), _("Lopende aanvragen")),
-            ("#cases", _("Afgeronde aanvragen")),
-        ]
+        self.log_access_cases(case_dicts)
```
Bartvaderkin, line 81: I note this now logs both cases and submissions.

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -340,13 +337,19 @@ def test_list_open_cases(self, m):
                     "description": self.zaaktype["omschrijving"],
                     "current_status": self.status_type1["omschrijving"],
                 },
+                {
+                    "uuid": self.zaak3["uuid"],
```
Bartvaderkin, line 341: Can we give the (numbered) test resources a more meaningful name so the test stay readable: `self.zaak1`, `self.zaak3` don't mean a lot. 

Ideally something short that shows the intent of the resource (eg: `self.zaak_intern` is an internal case)

Ex: `status_type1` should be `status_type_initial`, and `status_type2` like  `status_type_final`

src/open_inwoner/cms/cases/urls.py
```diff
@@ -70,7 +54,7 @@
         CaseDocumentUploadFormView.as_view(),
         name="case_detail_document_form",
     ),
-    path("open/", RedirectView.as_view(pattern_name="cases:open_cases"), name="index"),
-    path("content/", InnerOpenCaseListView.as_view(), name="open_cases_content"),
-    path("", OuterOpenCaseListView.as_view(), name="open_cases"),
+    path("open/", RedirectView.as_view(), name="redirect"),
```
Bartvaderkin, line 57: Do we still need this redirect view? It was to switch between open/closed IIRC.

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -367,153 +370,108 @@ def test_list_open_cases(self, m):
...
         response = self.app.get(
-            self.inner_url_closed, user=self.user, headers={"HX-Request": "true"}
+            self.inner_url, user=self.user, headers={"HX-Request": "true"}
         )
         self.assertNotContains(response, st1.status)
         self.assertContains(response, st1.translation)
 
-    def test_list_closed_cases_logs_displayed_case_ids(self, m):
+    def test_zaak_proecess_data(self, m):
+        pass
```
Bartvaderkin, line 440: Empty test.

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -367,153 +370,108 @@ def test_list_open_cases(self, m):
...
+        self.assertEqual(self.user, case_log[0].user)
+        self.assertEqual(self.user, case_log[0].content_object)
+
+        # no logs for internal, hence non-displayed cases
         for log in logs:
-            self.assertNotIn(self.zaak1["identificatie"], log.extra_data["message"])
-            self.assertNotIn(self.zaak2["identificatie"], log.extra_data["message"])
+            self.assertNotIn(
+                self.zaak_intern["identificatie"], log.extra_data["message"]
+            )
```
Bartvaderkin, line 472: This could be less hairy with `assertTimelineLog()`

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -563,49 +521,97 @@ def test_list_cases_paginated(self, m):
         self.assertNotContains(response_2, self.zaak2["identificatie"])
         self.assertContains(response_2, "?page=1")
 
-    @patch.object(CaseListMixin, "paginate_by", 1)
+    @patch.object(InnerCaseListView, "paginate_by", 1)
     def test_list_cases_paginated_logs_displayed_case_ids(self, m):
```
Bartvaderkin, line 525: This function could be less hairy with `assertTimelineLog()`

Bartvaderkin (dismissed):

stevenbal (changes requested):
Looks good :+1:, just some minor remarks

src/open_inwoner/openzaak/tests/test_cases.py
pi-sigma: This needs to be coordinated across all tests that make use of this data. In this context we could refactor the relevant tests to make the use of zaken data uniform. I took a stab at encapsulating the zaken test data in a mixin as part of [PR 815](https://github.com/maykinmedia/open-inwoner/pull/815). Would be happy to take this refactor further and rename the individual zaken, as you suggested. Let's discuss this.

src/open_inwoner/openzaak/api_models.py
```diff
@@ -32,6 +37,52 @@ class Zaak(ZGWModel):
...
+        identificatie: str,
+        zaak_config: "OpenZaakConfig",
+    ):
+        if zaak_config.reformat_esuite_zaak_identificatie:
+            return self._reformat_esuite_zaak_identificatie(self.identificatie)
+        return identificatie
+
+    def process_data(self):
+        """
+        Prepare data for template
```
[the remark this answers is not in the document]
pi-sigma: What do you mean "what is the process"? There's a docstring, it's pretty clear.  I don't agree that the `dict` was easier to modify before the refactor. First, it was burried in a convoluted class that was hard to make sense of. Secondly, there was not a single case where "other things from the view" were added to the dict, which includes only attributes from the data object which are reformatted for the template. We can still add things to the context in the view. If anything, there is now better and clearer separation of concerns: the dictionaries for `Zaak` and `OpenSubmission` contain data from the objects, the view grabs this data and adds additional stuff to the context.

src/open_inwoner/cms/cases/views/cases.py
pi-sigma: Yes, submissions are now cases too. I created a `Protocol` to document this but only added the type hints above (lines 61-62) after Steven's suggestion.

PR #807, opened 2023-10-12
:lipstick: [#1801] Only make explicit case link clickable in list

stevenbal:


stevenbal:
No longer required, closing this

PR #813, opened 2023-10-16
:bug: [#1809] Fix CTA button on product page

stevenbal:


src/open_inwoner/js/components/readmore/index.js
```diff
@@ -8,7 +8,6 @@ export class ReadMore {
   }
 
   toggleReadMore(event) {
-    event.preventDefault()
```
stevenbal, line 11: @alextreme this change was committed three months ago, so this must have been broken for a while?

alextreme: Readmore was added in v1.11, so this would make sense

alextreme (approved):

PR #815, opened 2023-10-18
[#1799] Implement CMS plugin for mijn aanvragen

pi-sigma:
Taiga [#1799](https://taiga.maykinmedia.nl/project/open-inwoner/task/1799)

### DIFF
--- a/src/open_inwoner/cms/cases/tests/test_plugin_cases.py
+++ b/src/open_inwoner/cms/cases/tests/test_plugin_cases.py
@@ -2,8 +2,12 @@ from unittest.mock import patch
 
 from django.test import TestCase, override_settings
 
+import lxml
 import requests_mock
+from pyquery import PyQuery as pq
 
+from open_inwoner.accounts.choices import LoginTypeChoices
+from open_inwoner.accounts.tests.factories import UserFactory
 from open_inwoner.openzaak.tests.mixins import ZakenTestMixin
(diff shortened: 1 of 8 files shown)


alextreme:
Rebase on PR 804, after that discuss this with @Bartvaderkin

pi-sigma:
Side note: I was tired of copy-pasting the boilerplate code required for tests that involve zaken, so I created a re-usable mixin. If we want to keep this we should modify other tests involving zaken as well. Otherwise the mixin should be removed.

src/open_inwoner/cms/cases/tests/test_plugin_cases.py
```diff
@@ -0,0 +1,86 @@
...
+        # anonymous user
+        html, context = cms_tools.render_plugin(CasesPlugin)
+
+        self.assertIsNone(context["cases"])
+
+        # check that html empty
+        with self.assertRaises(lxml.etree.ParserError) as ctx:
+            pq(html)
+
+        self.assertEqual(str(ctx.exception), "Document is empty")
```
Bartvaderkin, line 40: @pi-sigma We don't really need lxml and pyquery to check if a string is empty :smile:

pi-sigma:
Closing for now because the PR has been on hold for quite a while. Can be re-opened or scavenged if the issue is taken up again.

PR #818, opened 2023-10-19
:whale: Fix NPM install in Docker

stevenbal:
upgrading FontAwesome caused dependency issues


stevenbal:
@alextreme I'm not sure if we run Open Inwoner in docker, if not then this would have to be modified in the deployment as well

alextreme:
@stevenbal thanks, we use maykin-deployment/projects/open-inwoner for SaaS deployment, I see that this fix was already applied for that.

alextreme (approved):

PR #819, opened 2023-10-19
:sparkles: [#1802] Allow search for cases via searchbar

stevenbal:
in case of exact match on identificatie

### DIFF
--- a/src/open_inwoner/search/views.py
+++ b/src/open_inwoner/search/views.py
@@ -2,11 +2,12 @@ from django.contrib import messages
 from django.core.paginator import InvalidPage, Paginator
 from django.http import Http404
 from django.http.response import HttpResponseRedirect
-from django.urls import reverse_lazy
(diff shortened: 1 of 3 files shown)


src/open_inwoner/search/views.py
```diff
@@ -49,6 +50,17 @@ def search(self, form):
...
+            cases = fetch_cases(self.request.user.bsn, identificatie=query)
+            if cases and len(cases) == 1:
+                return HttpResponseRedirect(
+                    reverse(
+                        "cases:case_detail", kwargs={"object_id": str(cases[0].uuid)}
+                    )
+                )
+
         if not query:
             return self.render_to_response(context)
```
pi-sigma, line 65: This should come before the previous check for zaken search.

pi-sigma (changes requested):

pi-sigma (approved):

PR #823, opened 2023-10-23
[#1814] Upgraded digid-eherkenning to 0.9

pi-sigma:
The new version validates BSN numbers during the mock login flow.

This will prevent issues like Taiga [#1814](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1814) by requiring users to enter a BSN that satifies pattern validation as well as the 11-check.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_auth.py
+++ b/src/open_inwoner/accounts/tests/test_auth.py
@@ -1,4 +1,5 @@
 from datetime import date
+from unittest.mock import patch
 from urllib.parse import urlencode
 
 from django.contrib.sites.models import Site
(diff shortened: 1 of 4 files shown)


src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -68,7 +69,11 @@ def test_registration_page_only_digid_with_invite(self):
             furl(reverse("digid:login")).add({"next": necessary_url}).url,
         )
 
-    def test_digid_fail_without_invite_redirects_to_login_page(self):
+    @patch("digid_eherkenning.validators.Proef11ValidatorBase.__call__")
+    def test_digid_fail_without_invite_redirects_to_login_page(self, m):
+        # disable mock form validation to check redirect
+        m.return_value = True
```
Bartvaderkin, line 75: Why do we need to mock the validator? Couldn't we just use a valid 11-proef test value?

Bartvaderkin (changes requested):

src/open_inwoner/accounts/tests/test_auth.py
pi-sigma: We are testing the redirect for invalid BSN numbers. However, a BSN can be invalid for two reasons: (a) it fails formal validation, (b) it is formally valid but doesn't correspond to an actual BSN. If we pass a valid 11-proef value, we won't be able to test the redirect as we can't test actual BSN's.

Bartvaderkin (approved):

PR #824, opened 2023-10-23
[#1794] Redesign case detail-page header

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1793
The newly. designed Dashboard component requires restructuring of the detail page grid + has impact on Contactmoment detail page.
Also: Anchormenu references need to be removed for Case detail 
(and from Case list in other PR).

Note: unsure which variable is needed for "Verwachte uitslag" (now set to `end_date_legal`)

### DIFF
--- a/src/open_inwoner/scss/components/Container/Container.scss
+++ b/src/open_inwoner/scss/components/Container/Container.scss
@@ -14,10 +14,9 @@
   }
 
   @media (min-width: 768px) {
-    $hm: max(
-      calc(((100vw - var(--container-width)) / 2) - 7px),
-      var(--spacing-large)
-    );
+    // $hm has a minimum of var(--spacing-large), and a maximum of $hlargest
+    $hlargest: calc(((100vw - var(--container-width)) / 2) - 7px);
+    $hm: max($hlargest, var(--spacing-large));
     margin: var(--spacing-extra-large) $hm calc(var(--row-height) * 2);
 
     &--no-margin {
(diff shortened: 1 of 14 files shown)


pi-sigma (changes requested):
The spinner on the detail page is no longer centered after the removal of the sidebar anchors (see screenshot). Is this something that will be fixed in a separate PR?

Screenshot from 2023-10-26 11-01-32

jiromaykin:
> The spinner on the detail page is no longer centered after the removal of the sidebar anchors"

I've fixed it in this PR for the Detail page only - but it should actually be simplified and homogenised in another PR for the List view after this is merged (because the sidebar is going to be removed in all Aanvragen pages), so I'll have remember that.

pi-sigma:
Looking good (flaky tests but seems alright now).

jiromaykin:
> Looking good (flaky tests but seems alright now).

Strangely enough the tests suddenly passed through this morning (after 3 days) - odd!

pi-sigma (approved):

src/open_inwoner/scss/components/Container/Container.scss
```diff
@@ -14,7 +14,9 @@
   }
 
   @media (min-width: 768px) {
```
Bartvaderkin, line 16: Are we sure about these magic numbers as breakpoints? I see `@media (min-width: 768px) {` here but in the file above there is `@media (min-width: 767px) {`

Bartvaderkin (approved):

src/open_inwoner/scss/components/Container/Container.scss
jiromaykin: Ah: **_min_**-width should be 768px everywhere, but **_max_**-width should be 767px

PR #825, opened 2023-10-25
[#1798] Refactored document list in case detail view

pi-sigma:
Taiga [#1798](https://taiga.maykinmedia.nl/project/open-inwoner/task/1798)

- added creation date to documents
- added indicator for recently added documents ("recently added" made configurable)

TODO:
- the size of the document icon on the left needs to adjust dynamically, depending on whether the "recently added" indicator is present or not (currently the size is hardwired via padding)

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -1,8 +1,11 @@
 import dataclasses
+import datetime as dt
 import logging
 from collections import defaultdict
+from datetime import datetime
 from typing import List, Optional
 
+from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.http import Http404, StreamingHttpResponse
@@ -50,6 +53,7 @@ from open_inwoner.openzaak.models import (
     ZaakTypeStatusTypeConfig,
 )
 from open_inwoner.openzaak.utils import get_role_name_display, is_info_object_visible
+from open_inwoner.utils.time import has_new_elements
 from open_inwoner.utils.translate import TranslationLookup
 from open_inwoner.utils.views import CommonPageMixin, LogMixin
 
@@ -64,6 +68,7 @@ class SimpleFile:
     name: str
     size: int
     url: str
+    created: Optional[datetime] = None
 
 
 class OuterCaseDetailView(
@@ -201,6 +206,11 @@ class InnerCaseDetailView(
                 "end_statustype_data": end_statustype_data,
                 "documents": documents,
                 "allowed_file_extensions": sorted(config.allowed_file_extensions),
+                "new_docs": has_new_elements(
+                    documents,
+                    "created",
+                    dt.timedelta(days=settings.DOCUMENT_RECENT_DAYS),
+                ),
             }
             context["case"].update(self.get_upload_info_context(self.case))
             context["anchors"] = self.get_anchors(statuses, documents)
@@ -285,7 +295,7 @@ class InnerCaseDetailView(
                 zt_statustype_config = ztc.zaaktypestatustypeconfig_set.get(
                     statustype_url=case.status.statustype.url
                 )
-            # case has no status, or status type not found
+            # case has no status, or statustype config not found
             except (AttributeError, ObjectDoesNotExist):
                 pass
             else:
@@ -395,8 +405,10 @@ class InnerCaseDetailView(
                             "info_id": info_obj.uuid,
                         },
                     ),
+                    created=case_info_obj.registratiedatum,
                 )
             )
+
         return documents
 
     def get_form_kwargs(self):
--- a/src/open_inwoner/components/templates/components/File/File.html
+++ b/src/open_inwoner/components/templates/components/File/File.html
@@ -1,48 +1,57 @@
-{% load i18n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
+{% load i18n l10n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
 
 <aside class="file" aria-label="{% trans "Bestand" %} {{ name }}">
-    <div class="file__file">
-    {% if is_image %}
-        {% icon icon="image" outlined=True %}
-    {% else %}
-        {% icon icon="description" outlined=True %}
-    {% endif %}
-    <p class="p">
-        {{ name }}
-        {% if extension and size %}
-            ({{extension}}, {{size|readable_size}})
-        {% elif extension %}
-            ({{extension}})
-        {% elif size %}
-            ({{size|readable_size}})
-        {% endif %}
-    </p>
-    {% if allow_delete %}
-        {% dropdown icon="more_horiz" %}
-            <div class="dropdown__item">
-                {% url 'inbox:start' as inbox_url %}
-                {% with inbox_url|addstr:"?file="|addstr:uuid as share_url %}
-                    {% button icon="send" text=_("Deel") href=share_url icon_outlined=True transparent=True %}
-                {% endwith %}
-            </div>
-            {% if show_download %}
-                <div class="dropdown__item">
-                    {% button icon="file_download" text=_("Download") href=url icon_outlined=True transparent=True %}
-                </div>
+    <div class="file__container {% if recently_added %}file__container--recent{% endif %}">
+        <div class="file__file">
+            <p class="file__symbol">
+                {% if is_image %}
+                    {% icon icon="image" outlined=True %}
+                {% else %}
+                    {% icon icon="insert_drive_file" outlined=True %}
+                {% endif %}
+            </p>
+
+            <p class="p file__data">
+                {% if recently_added %}
+                    <span class="file__file--recent">{% icon icon="fiber_manual_record" outlined=False extra_classes="file__file--recent" %}{% trans "Nieuw" %}</span>
+                {% endif %}
+                {{ name }}
+                {% if extension and size %}
+                    ({{extension}}, {{size|readable_size}})
+                {% elif extension %}
+                    ({{extension}})
+                {% elif size %}
+                    ({{size|readable_size}})
+                {% endif %}
+                <span class="file__date">{{ created|date:'d.m.Y' }}</span>
+            </p>
+            {% if allow_delete %}
+                {% dropdown icon="more_horiz" %}
+                    <div class="dropdown__item">
+                        {% url 'inbox:start' as inbox_url %}
+                        {% with inbox_url|addstr:"?file="|addstr:uuid as share_url %}
+                            {% button icon="send" text=_("Deel") href=share_url icon_outlined=True transparent=True %}
+                        {% endwith %}
+                    </div>
+                    {% if show_download %}
+                        <div class="dropdown__item">
+                            {% button icon="file_download" text=_("Download") href=url icon_outlined=True transparent=True %}
+                        </div>
+                    {% endif %}
+                    <div class="dropdown__item">
+                        {% url 'profile:documents_delete' uuid=uuid as delete_url %}
+                        {% render_form form=None form_action=delete_url method="POST" spaceless=True id="deactivate-form" extra_classes="confirm" data_confirm_title=_("Weet je het zeker dat je dit bestand wilt verwijderen?") data_confirm_cancel=_("Nee") data_confirm_default=_("Ja") %}
+                            {% csrf_token %}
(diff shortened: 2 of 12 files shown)


Bartvaderkin:
Please fix the commit message and PR title. "Feature/1798 document list" doesn't mean a lot.

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
+{% load i18n l10n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
 
 <aside class="file" aria-label="{% trans "Bestand" %} {{ name }}">
     <div class="file__file">
```
jiromaykin, line 4: In order to correct the not-stretchy behaviour, I would surround the first flex-column with it's own class here so the icon can then be BOTH centered and stretched (so needs to be a nested Flexbox), but make sure to not use the 'icon' class or else CSS will start thinking this column is also an actual material-icon :-/
`<p class="file__symbol">.....`

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -7,7 +7,14 @@
     {% else %}
         {% icon icon="description" outlined=True %}
     {% endif %}
+
```
jiromaykin, line 10: here you can close the new file__symbol paragraph, and in File.css add this: 
`.file__symbol {
    display: flex;
    flex-direction: column;
    align-items: center;
    background-color: var(--color-gray-lightest);
    padding: var(--spacing-large) 0;
    margin: 0;
    box-sizing: content-box;

    *[class*='icon'],
    *[class*='Icon'] {
      align-self: stretch;
      padding: var(--spacing-large);
    }
  }`
 
**NB:** This means that this can be **removed** there: 
 
`&__file > *[class*='icon']:first-child,
  &__file > *[class*='Icon']:first-child {
    background-color: var(--color-gray-lightest);
    padding: var(--spacing-large);
    box-sizing: content-box;
  }`

src/open_inwoner/components/templates/components/File/File.html
jiromaykin, line 4: A flexbox element cannot _both_ be stretched AND centered so you need to wrap it in another flex element, could be a paragraph (but can also be something else) so you could add `<p class="file__symbol">` here.

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -46,7 +46,7 @@ <h2 class="h2" id="statuses">{% trans 'Status' %}</h2>
 
             {#  Documents. #}
             {% if case.documents %}
-                <h2 class="h2" id="documents">{% trans 'Documenten' %}</h2>
+                <h2 class="h2" id="documents">{% trans 'Documenten' %}{% if case.new_docs %}{% icon icon="fiber_manual_record" outlined=False extra_classes="file__file--recent" %}{% endif %}</h2>
```
jiromaykin, line 49: I would maybe not use an icon here (because it takes space) - for example: I made a CSS circle for the notifications in our menu so it could be positioned absolutely- but for Absolute positioning you need a parent with Position:relative anyway.
So: look at the design and think of changing the structure to help you position the dot Absolute 🙂 or use a CSS border-radius circle if that makes it easier: https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=8-15743&mode=design&t=CuHG1L2dDE2kGhX1-0

example-notificationDot

jiromaykin (changes requested):
Requesting quite a few changes 🙂 - if some of it is too difficult, I can help, but this week I'm very busy with other things.

src/open_inwoner/templates/pages/cases/status_inner.html
jiromaykin: Oh ignore my CSS circle comment - if you use `vertical-align:middle` it should look more like in the design:

Screenshot 2023-10-31 at 14 49 26

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,13 +1,22 @@
...
+        <p class="file__symbol">
+            {% if is_image %}
+                {% icon icon="image" outlined=True %}
+            {% else %}
+                {% icon icon="description" outlined=True %}
+            {% endif %}
+        </p>
+
+
     <p class="p">
```
jiromaykin, line 14: Remove all `<BR>'s` and set this in CSS as:
`display: flex;      flex-direction: column;`

jiromaykin (changes requested):
Preferably do not use BR for styling, it has no structural meaning and cannot easily be overwritten by CSS. 
In this case it causes a problem: we cannot change some element's padding/margin, so this does not conform to the design. 

The structure that can use Flexbox should look a bit more more like this (see attachment), without any BR's, but make sure it still looks good on mobile + some columns need to be set to `align-items: center;` when needed, like: depending on wether they actually have a 'recently_added' element or not.

recently_added-Flexbox

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,13 +1,23 @@
...
+        {% if recently_added %}
+            <span class="file__notification">
+                {% icon icon="fiber_manual_record" outlined=False extra_classes="file__file--recent" %}
+                <span class="file__file--recent">{% trans "Nieuw" %}</span>
+            </span>
+        {% endif %}
         {{ name }}
```
jiromaykin, line 21: Your future life will be easier when you also wrap a Span around the name + size, but this is okay for now.

src/open_inwoner/scss/components/File/File.scss
```diff
@@ -7,11 +7,20 @@
...
   }
 
-  &__file > *[class*='icon']:first-child,
-  &__file > *[class*='Icon']:first-child {
-    background-color: var(--color-gray-light);
+  &__symbol {
+    display: flex;
+    align-items: center;
+    background-color: var(--color-gray-lightest);
     padding: var(--spacing-large);
```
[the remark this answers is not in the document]
jiromaykin: Please change the old padding by setting the side-padding to **zero**/0 🙂 and like i said the the earlier comment
  
Screenshot 2023-11-06 at 12 10 35

src/open_inwoner/scss/components/File/File.scss
```diff
@@ -7,11 +7,35 @@
     display: flex;
   }
 
-  &__file > *[class*='icon']:first-child,
-  &__file > *[class*='Icon']:first-child {
-    background-color: var(--color-gray-light);
+  &__notification {
+    display: flex;
+    flex-direction: row;
```
jiromaykin, line 12: Now that you've wrapped this element correctly, you should give it a different line-height, conforming to the design:
`line-height: var(--spacing-medium);`

jiromaykin (changes requested):
I know flexbox wrapping is hard to 'wrap' your mind around; just a few more changes and this is done!

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,13 +1,22 @@
...
+                {% icon icon="description" outlined=True %}
+            {% endif %}
+        </p>
+
+
     <p class="p">
+        {% if recently_added %}
+            {% icon icon="fiber_manual_record" outlined=False extra_classes="file__file--recent" %}
+            <span class="file__file--recent">{% trans "Nieuw" %}</span>
+            <br>
```
jiromaykin, line 18: Remove all BR's and give the surrounding paragraph a display Flex + flex-direction column so you can control the spacing.

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -16,6 +25,8 @@
         {% elif size %}
             ({{size|readable_size}})
         {% endif %}
+        <br>
```
jiromaykin, line 28: Again: remove BR's and just separate elements with Spans so you can actually control the spacing.
Surround elements that need to be next to each other with another span (like the red dot and the 'Nieuw' text.

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,13 +1,22 @@
...
-    {% if is_image %}
-        {% icon icon="image" outlined=True %}
-    {% else %}
-        {% icon icon="description" outlined=True %}
-    {% endif %}
+        <p class="file__symbol">
+            {% if is_image %}
+                {% icon icon="image" outlined=True %}
+            {% else %}
+                {% icon icon="description" outlined=True %}
```
jiromaykin, line 9: Just noticed in the new design: different icon needs to be used for other file types: insert_drive_file. 
`{% icon icon="insert_drive_file" ......`

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -1,13 +1,22 @@
-{% load i18n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
+{% load i18n l10n link_tags utils icon_tags form_tags dropdown_tags button_tags string_tags %}
 
 <aside class="file" aria-label="{% trans "Bestand" %} {{ name }}">
     <div class="file__file">
```
jiromaykin, line 4: This needs a new class, so you can add the Red line on the left of this div, something like `{% if recently_added %}file__container--recent{% endif %}` and add a left-border as in the design.

src/open_inwoner/scss/components/File/File.scss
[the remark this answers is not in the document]
jiromaykin: This is still not correct.
Please add the zero to remove unnecessary padding to the left/right sides here.
(padding shorthand notation here = top/bottom right/left)

See the differences:
Screenshot 2023-11-20 at 11 53 01

jiromaykin (changes requested):
I understand the flex-box thing is tricky, but please do not use BR's.

pi-sigma:
@jiromaykin  Styling is fixed, thanks for the input. The design doesn't have the text ("Download") in the download link, but I was wondering if it's better to keep it for consistency and accessibility. What do you think?


Screenshot from 2023-11-21 09-21-01

jiromaykin:
> "....The design doesn't have the text ("Download") in the download link, but I was wondering if it's better to keep it for consistency and accessibility. What do you think?...."

For accessibility this is not quite an issue: any text can be made invisible while remaining in the structure (with CSS you can use transform/opacity/positioning to make elements visible for screenreader-software only). Also: aria-label and title attribute can be used for tooltips.

svenvandescheur (changes requested):

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -201,6 +206,11 @@ def get_context_data(self, **kwargs):
                 "end_statustype_data": end_statustype_data,
                 "documents": documents,
                 "allowed_file_extensions": sorted(config.allowed_file_extensions),
+                "new_docs": has_new_elements(
```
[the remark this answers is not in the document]
svenvandescheur: @alextreme @pi-sigma as discussed: the template tag implementation might not be the most favorable approach and we might want to move away from that pattern. This does raises some issues in terms of reusability and complexity which might needs to be discusses in a separate team-wide meeting.

Some concerns are:

**How can we guarantee consistency of UI elements across different pages.**

Previously the template tag would take a predefined "(API) interface" and would render a specialized, reusable template for that component. This allowed Django templates to be built much like other modern component based frontend tooling. The template tag approach was a more developed leftover from porting a pure frontend (Vite?) implementation but felt confusing to developers less familiar with such concepts. 

Currently, some pages move away from this strategy and use a more classic Django view approach. As a result, certain (HTML) code paths might act different on different pages causing potential UI inconsistencies and DRY violations.

**How can we keep context processing maintainable.**

In order to facilitate all the individual frontend components, quite a lot of context processing is necessary. When moving away from template tags,  this typically ends up in the `get_context_data` function in the views.

Looking at https://github.com/maykinmedia/open-inwoner/blob/38ced6be235376c50c07a4e7a04ae1ba1b282781/src/open_inwoner/cms/cases/views/status.py#L117 I have some concerns about the complexity and maintainability of this processing, I't very unclear to me:

- What certain processing is for.
- If and where certain context is used.
- What the expected shape of data is within templates

**How can we keep developers happy about this setup.**

All approaches have advantages and drawback but we need to find a solution that works for all team members.

svenvandescheur (approved):

PR #826, opened 2023-10-26
:sparkles: [#1792] Add ZaakTypeResultaatType config

stevenbal:


pi-sigma (approved):
Looks good to me. @stevenbal should double-check, since I've made changes to fix the tests.

PR #827, opened 2023-10-27
Fixed case filtering in list view and AttributeError in detail view

pi-sigma:
Fixes two problems with the recent refactor of the case list ([PR 804](https://github.com/maykinmedia/open-inwoner/pull/804/files))
- the filtering of non-visible zaken was performed in the wrong place, which slowed down page load
- the addition of info_context to upload documents needs to check for `AttributeError` to account for zaken without status


Bartvaderkin (dismissed):
Please fix your commit messages and PR title. "Fix case list" doesn't mean anything next month.

stevenbal (approved):

PR #830, opened 2023-11-02
:lipstick: Fix case list styling issues

stevenbal:
Noticed that the heading was duplicated after the refactor of the case list
image

Also noticed that the status indicator header was always added, even if the status indicator is empty, resulting in unnecessary spacing at the top of the card
image


jiromaykin:
I also noticed it's now missing its Breadcrumbs. The title duplicate and missing Breadcrumbs issue is also in the Detail view.

PR #831, opened 2023-11-03
[#1827] Show second status for cases

pi-sigma:
Taiga [#1827](https://taiga.maykinmedia.nl/project/open-inwoner/task/1827)


jiromaykin (approved):
Seems to work, I tested with ISBN 111111110 with [case 190010238](http://127.0.0.1:8000/mijn-aanvragen/adaddaa3-eccd-4885-bae5-17ca53219be0/status/).

But perhaps rebase this one on develop first? (we may have DB migrations merges)

PR #832, opened 2023-11-03
[#1832] Add breadcrumbs back to case list view

pi-sigma:
Taiga [#1832](https://taiga.maykinmedia.nl/project/open-inwoner/task/1832)


jiromaykin (approved):
Looking good on the front-end :-) (I don't know about the back-end)

PR #835, opened 2023-11-06
:sparkles: [#1821] Add visibility flags for Category model

stevenbal:
Implemented the basics for the visibility flags, no DigiD/eHerkenning or Zaaktype checks yet, I created separate tasks for that

@alextreme I did not apply the visibility logic for `/mijn-profiel/onderwerpen/` since this is not mentioned in the issue, should this logic apply to this view as well?

Taiga [#1821](https://taiga.maykinmedia.nl/project/open-inwoner/us/1821)


alextreme:
> @alextreme I did not apply the visibility logic for `/mijn-profiel/onderwerpen/` since this is not mentioned in the issue, should this logic apply to this view as well?

Lets give Bram and Robert time to respond to my suggestion, and leave 'Mijn Onderwerpen' out-of-scope for the time being.

pi-sigma (approved):
~~The category detail pages are still accessible via the urls. Is this intended? If so, looking good.~~

I just looked up the ticket and saw that this is indeed intended, so good to go.

@stevenbal Could you include a link to the Taiga task in the first comment next time? Sometimes entering the task id in Taiga's url bar doesn't yield a result and then you have to search for the ticket.

stevenbal:
@pi-sigma will do :+1:

PR #838, opened 2023-11-07
[#1839] Enable/disable notifications per status type

pi-sigma:
Taiga [#1839](https://taiga.maykinmedia.nl/project/open-inwoner/task/1839)


src/open_inwoner/openzaak/utils.py
```diff
@@ -140,3 +141,14 @@ def get_retrieve_resource_by_uuid_url(
...
     )
     return url
+
+
+def get_statustype_url_from_status_url(status_url: str) -> str:
+    """
+    Get statustype url by appending the "rpath" of `status_url` (the part
+    after the last '/') to the root url of the catalogi API
+    """
+    CATALOGI_ROOT = "https://catalogi.nl/api/v1/"
```
stevenbal, line 151: I'm not sure why this is hardcoded, also if I understand it correctly I don't think this function would give me the correct statustype url, because the UUID for a `Status` is not the same as the UUID for its `StatusType`.

I think you can leave this function out and do something like this in `src/open_inwoner/openzaak/notifications.py`:

```python
from open_inwoner.openzaak.cases import resolve_status

...

# check status notification setting on `ZaakTypeStatusTypeConfig`
resolve_status(case)
statustype_url = case.status.statustype
```

stevenbal (changes requested):

src/open_inwoner/openzaak/utils.py
```diff
@@ -6,6 +6,7 @@
 from zgw_consumers.api_models.constants import RolTypes, VertrouwelijkheidsAanduidingen
 
 from open_inwoner.openzaak.api_models import InformatieObject, Rol, Zaak, ZaakType
+from open_inwoner.utils.url import get_rpath
```
stevenbal, line 9: This is now unused, so it can be removed

src/open_inwoner/utils/url.py
```diff
@@ -6,3 +8,11 @@ def build_absolute_url(path: str) -> str:
...
     protocol = "https" if settings.IS_HTTPS else "http"
     return f"{protocol}://{domain}{path}"
+
+
+def get_rpath(url: str) -> str:
+    """
+    Return the part of `url` after the last '/'
+    """
+    p = urlparse(url)
+    return p.path.rsplit("/", 1)[-1]
```
stevenbal, line 18: This is now unused, so it can be removed

src/open_inwoner/openzaak/models.py
```diff
@@ -362,6 +362,7 @@ class ZaakTypeStatusTypeConfig(models.Model):
             "Description that will be shown above the document upload widget in a case detail page"
         ),
     )
+    notify_status_change = models.BooleanField(default=True)
```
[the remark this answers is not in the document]
stevenbal: This comment was resolved, but I think you forgot to add these attributes?

stevenbal (changes requested):

stevenbal:
@pi-sigma could you fix the merge conflict as well?

pi-sigma:
@stevenbal Sorry for the confusion, I must have made a mistake during rebasing.

stevenbal (approved):

PR #839, opened 2023-11-07
[#1817] Show errors for selected documents

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1817
depends on #833 
so branched off from that.


jiromaykin:
Thought I was rebasing on the wrong branch, now reopening.

PR #841, opened 2023-11-09
[#1843] Add description field to ZaakTypeStatusTypeConfig

pi-sigma:
Taiga [#1843](https://taiga.maykinmedia.nl/project/open-inwoner/task/1843)


pi-sigma:
As with other recent additions, our tabular inlines for zaak-related configurations are getting quite wide. @jiromaykin asked if it was possible to at least restrict the size of the widget. For textarea widgets, that's easily done. For the `CKEditorWidget` it's different. The width is set in `scss/admin/_ck_editor.scss` to `624px !important;` ruling out programmatic overrides. I leave the styling it as is for now. Perhaps we can discuss during the next weekly if this should be addressed.

src/open_inwoner/openzaak/models.py
```diff
@@ -362,6 +362,14 @@ class ZaakTypeStatusTypeConfig(models.Model):
             "Description that will be shown above the document upload widget in a case detail page"
         ),
     )
+    description = models.TextField(
+        blank=True,
+        default="",
+        verbose_name=_("Frontend description"),
+        help_text=_(
+            "The text displayed in the list view for cases with this statustype"
```
stevenbal, line 370:
```suggestion
            "The text displayed in the case detail page for the status with this statustype"
```

stevenbal (changes requested):
Regarding the wide inlines, I think it's better to switch to from `admin.TabularInline` to `admin.StackedInline` for `ZaakTypeStatusTypeConfigInline` (and I think this would be better for `ZaakTypeResultaattypeConfigInline` as well

alextreme:
@pi-sigma please do

stevenbal (approved):

PR #843, opened 2023-11-10
[#1830] Prevent deletion of last website

pi-sigma:
Taiga [#1830](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1830)


stevenbal (changes requested):
Looks good overall, though I tested this locally and it doesn't work when doing a delete with the "bulk delete" action from the list view (see https://docs.djangoproject.com/en/3.2/ref/contrib/admin/actions/). I think it would be good if deleting the last `Site` via the list view could be blocked as well

image

PR #845, opened 2023-11-13
[Draft] :sparkles: [#1838] Add zaaktypen select widget for Category admin

stevenbal:


src/open_inwoner/pdc/managers.py
```diff
@@ -33,6 +37,49 @@ def visible_for_user(self, user: User):
...
+        if not getattr(user, "bsn", None):
+            return self
+
+        cases = fetch_cases(user.bsn)
+
+        months_since_last_zaak_per_zaaktype = {}
+        for case in cases:
+            resolve_zaak_type(case)
+
+            duration_since_start = relativedelta(date.today(), case.startdatum)
```
stevenbal, line 53: @alextreme just to double check, but I assumed that the `relevante zaakperiode` is used with respect to the `startdatum` of the case

src/open_inwoner/pdc/managers.py
```diff
@@ -33,6 +37,49 @@ def visible_for_user(self, user: User):
...
+                or months_since_last_zaak_per_zaaktype[case.zaaktype.identificatie]
+                > duration_since_start.months
+            ):
+                months_since_last_zaak_per_zaaktype[
+                    case.zaaktype.identificatie
+                ] = duration_since_start.months
+
+        zaaktype_ids = list(months_since_last_zaak_per_zaaktype.keys())
+
+        qs = self.filter(zaaktypen__overlap=zaaktype_ids)
```
stevenbal, line 65: @alextreme I assume that only categories linked to zaaktypen will be shown in the highlighted category plugin for DigiD authenticated users

src/open_inwoner/cms/products/cms_plugins.py
```diff
@@ -34,8 +34,17 @@ def render(self, context, instance, placeholder):
...
+        if request.user.is_authenticated and request.user.bsn:
+            categories = (
+                Category.objects.published()
+                .filter_for_user_with_zaken(request.user)
+                .order_by("name")[: self.limit]
+            )
+
+            # Fallback for if the user has no cases with case types that are linked to
+            # categories
+            if not categories:
```
stevenbal, line 46: @alextreme I added this because I assumed it wouldn't be right to show no categories at all in this case

src/open_inwoner/pdc/managers.py
alextreme: That is correct

src/open_inwoner/pdc/managers.py
alextreme: That is correct, however note that not all gemeentes will use this feature (for those the categories shown should be as-is). We could make this a featureflag if necessary in the algemene configuratie.

src/open_inwoner/cms/products/cms_plugins.py
alextreme: Falling back to the current behaviour is fine

src/open_inwoner/pdc/managers.py
stevenbal: Good point, I'll add a feature flag for this, otherwise the other visibility flags are overridden by this one

src/open_inwoner/pdc/managers.py
```diff
@@ -33,6 +37,49 @@ def visible_for_user(self, user: User):
...
+        Returns the categories linked to ZaakTypen for which the user has Zaken.
+        """
+        if not getattr(user, "bsn", None):
+            return self
+
+        cases = fetch_cases(user.bsn)
+
+        months_since_last_zaak_per_zaaktype = {}
+        for case in cases:
+            resolve_zaak_type(case)
```
stevenbal, line 51: @alextreme this approach does mean that we have to do a fetch to get the zaaktype for each zaak, which isn't very efficient. I'm not sure how we could avoid this without something like an `include` parameter.

stevenbal: Discussed: simplify this by storing the zaaktype URL on zaaktypeconfig and use that instead of identificatie on category, to avoid having to resolve here

stevenbal:
Marked as draft again until performance issues have been resolved

src/open_inwoner/configurations/admin.py
```diff
@@ -258,6 +258,10 @@ class SiteConfigurarionAdmin(OrderedInlineModelAdminMixin, SingletonModelAdmin):
                 )
             },
         ),
+        (
+            _("Display options for authenticated users"),
+            {"fields": ("enable_categories_filtering_with_zaken",)},
```
alextreme, line 263: Graag verplaatsen naar de OpenZaak solo config model

src/open_inwoner/pdc/models/category.py
```diff
@@ -72,14 +72,6 @@ class Category(MP_Node):
             "Zaaktypen waarvoor bij aanwezigheid dit onderwerp getoond moet worden."
```
alextreme, line 72: Graag valideren of de featureflag aan staat, alleen het vullen van zaaktypen toestaan indien deze is geactiveerd.

alextreme (approved):

PR #846, opened 2023-11-13
[#1769] Replace deprecated djchoices with native Django choices

pi-sigma:
Taiga [#1769](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1769)

### DIFF
--- a/src/open_inwoner/accounts/choices.py
+++ b/src/open_inwoner/accounts/choices.py
@@ -1,46 +1,55 @@
+from django.db import models
 from django.utils.translation import gettext_lazy as _
 
-from djchoices import ChoiceItem, DjangoChoices
 
-
-class LoginTypeChoices(DjangoChoices):
-    default = ChoiceItem("default", _("E-mail en Wachtwoord"))
-    digid = ChoiceItem("digid", _("DigiD"))
-    eherkenning = ChoiceItem("eherkenning", _("eHerkenning"))
-    oidc = ChoiceItem("oidc", _("OpenId connect"))
-
-
-# Created because of a filter that needs to happen. This way the form can take the empty choice and the modal is still filled.
-class AllEmptyChoice(DjangoChoices):
-    empty = ChoiceItem("", _("Alle"))
-
-
-class ContactTypeChoices(DjangoChoices):
-    contact = ChoiceItem("contact", _("Contactpersoon"))
-    begeleider = ChoiceItem("begeleider", _("Begeleider"))
-    organization = ChoiceItem("organization", _("Organisatie"))
-
-
-class EmptyContactTypeChoices(AllEmptyChoice, ContactTypeChoices):
-    pass
+class LoginTypeChoices(models.TextChoices):
+    default = "default", _("E-mail en Wachtwoord")
+    digid = "digid", _("DigiD")
+    eherkenning = "eherkenning", _("eHerkenning")
+    oidc = "oidc", _("OpenId connect")
 
 
 # Created because of a filter that needs to happen. This way the form can take the empty choice and the modal is still filled.
-class StatusEmptyChoice(DjangoChoices):
-    empty = ChoiceItem("", _("Status"))
+class AllEmptyChoice(models.TextChoices):
+    empty = "", _("Alle")
 
 
-class StatusChoices(DjangoChoices):
-    open = ChoiceItem("open", _("Open"), icon="format_list_bulleted")
-    approval = ChoiceItem("approval", _("Accordering"), icon="question_mark")
-    closed = ChoiceItem("closed", _("Afgerond"), icon="check")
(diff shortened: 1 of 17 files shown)


src/open_inwoner/accounts/models.py
```diff
@@ -397,8 +397,8 @@ def get_contact_message_url(self) -> str:
         return f"{url}#messages-last"
 
     def get_contact_type_display(self) -> str:
-        choice = ContactTypeChoices.get_choice(self.contact_type)
-        return choice.label
+        choices_mapping = dict(ContactTypeChoices.choices)
+        return choices_mapping[self.contact_type]
```
stevenbal, line 401: Minor, but I think it's better to do something like this

```suggestion
        choice = getattr(ContactTypeChoices, self.contact_type)
        return choice.label
```

```python
>>> timeit.timeit(lambda: dict(FooChoices.choices)["foo"], number=100000)
0.2956498809999175
>>> timeit.timeit(lambda: getattr(FooChoices, "foo").label, number=100000)
0.05368091000127606
```

src/open_inwoner/accounts/tests/test_action_views.py
```diff
@@ -354,17 +354,14 @@ def test_action_status(self, mock_solo):
         dropdown_button = action_element.locator(".actions__status-selector")
         dropdown_content = action_element.locator(".dropdown__content")
 
-        # check state
-        expect(dropdown_button).to_contain_text(
-            str(StatusChoices.labels[StatusChoices.open])
-        )
+        approval_label = str(StatusChoices.approval.label)
+        closed_label = str(StatusChoices.closed.label)
```
stevenbal, line 358: Does this need to be cast to a string?

src/open_inwoner/accounts/tests/test_action_views.py
```diff
@@ -354,17 +354,14 @@ def test_action_status(self, mock_solo):
...
 
-        # grab buttonss
-        status_closed_button = dropdown_content.get_by_role(
-            "button", name=str(StatusChoices.labels[StatusChoices.closed])
-        )
+        expect(dropdown_button).to_contain_text("Open")
```
stevenbal, line 360:
```suggestion
        expect(dropdown_button).to_contain_text(StatusChoices.open.label)
```

src/open_inwoner/accounts/tests/test_action_views.py
```diff
@@ -394,10 +389,10 @@ def test_action_status(self, mock_solo):
         # click on status-closed button
         status_closed_button.click()
 
+        # import pdb;pdb.set_trace()
```
stevenbal, line 392: Could you remove this?

src/open_inwoner/management/commands/generate_native_django_choices.py
```diff
@@ -0,0 +1,90 @@
+from textwrap import indent
```
stevenbal, line 1: I think this file can be removed?

src/open_inwoner/openklant/constants.py
```diff
@@ -1,19 +1,18 @@
...
+    in_behandeling = "in_behandeling", _("In behandeling")
+    afgehandeld = "afgehandeld", _("Afgehandeld")
 
     @classmethod
     def safe_label(cls, value, default=""):
         if not value:
             return default
         try:
-            return cls.labels[value]
+            return dict(cls.choices)[value]
```
stevenbal, line 15: Same as before, I would prefer this since it's more efficient
```suggestion
            return getattr(cls, value).label
```

stevenbal (changes requested):
Minor comments, looks good overall

src/open_inwoner/accounts/choices.py
```diff
@@ -1,46 +1,55 @@
...
-    organization = ChoiceItem("organization", _("Organisatie"))
+class ContactTypeChoices(models.TextChoices):
+    contact = "contact", _("Contactpersoon")
+    begeleider = "begeleider", _("Begeleider")
+    organization = "organization", _("Organisatie")
 
 
-class EmptyContactTypeChoices(AllEmptyChoice, ContactTypeChoices):
-    pass
+class EmptyContactTypeChoices(models.TextChoices):
```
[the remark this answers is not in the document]
pi-sigma: Annoying, but no: `EmptyContactTypeChoices: cannot extend enumeration 'AllEmptyChoice'`

src/open_inwoner/accounts/tests/test_action_views.py
pi-sigma: The labels are lazy strings, which will cause the test to fail. I'll add a comment to make this explicit.

stevenbal (approved):

PR #848, opened 2023-11-15
[#1856] Integrate KVK API

pi-sigma:
Taiga [#1856](https://taiga.maykinmedia.nl/project/open-inwoner/task/1856)


src/open_inwoner/contrib/kvk/models.py
```diff
@@ -0,0 +1,59 @@
+from typing import Optional, Union
+
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+from simple_certmanager.models import Certificate
+from solo.models import SingletonModel
+
+
+class KvKConfig(SingletonModel):
```
stevenbal, line 10: Could you add this model to the `django-admin-index.json` fixture? That way it shows up in the admin

src/open_inwoner/contrib/kvk/models.py
```diff
@@ -0,0 +1,59 @@
...
+    api_root = models.CharField(
+        verbose_name=_("API root"),
+        max_length=128,
+        blank=True,
+        help_text=_("The root of the API (e.g. https://api.kvk.nl/api/v1)"),
```
stevenbal, line 15: Just to double check: does it matter if this URL has a trailing slash `/` or not? I remember that this did matter for the ZGW APIs.

stevenbal (changes requested):
Looks good :+1:

src/open_inwoner/contrib/kvk/models.py
pi-sigma: Good catch! I've introduced `urljoin` in `client.py` to make sure it doesn't matter how the root is specified in the admin.

stevenbal (approved):

PR #849, opened 2023-11-16
Issues/2023 11 16

svenvandescheur:


alextreme:
Not sure why this would be necessary, please do a review with a colleague first

pi-sigma (changes requested):
@svenvandescheur Could you add a description that explains what the issue is? If this is about Mac-specific library paths, you should put the code in your `conf/local.py`, which is intended for machine-specific settings. Depending on the issue, we could update the installation instructions with a subsection for Mac users (in this case, feel free to include any other Mac-related problems you encountered), but the general settings shouldn't contain machine-specific stuff.

svenvandescheur:
This is equal to: https://github.com/open-zaak/open-zaak/pull/1472

This adds the possibility to configure the [`GEOS_LIBRARY_PATH`](https://docs.djangoproject.com/en/4.2/ref/contrib/gis/install/geolibs/#geos-library-path) and the [`GDAL_LIBRARY_PATH`] Django settings using a .env (dotenv) file. It does not alter the default setting values.

On Mac system, the relevant library may not be installed on the location expected by default. Django provides an escape hatch to alter the path using a setting and point it towards the correct files.

Settings this using a `local.py` might work (and support wont change), however for modern setups using a `.env` is more favorable and since most relevant settings (like those for database connectivity) already support environment variables across various projects, I think it's more sensible to support using tis this way.

pi-sigma:
@svenvandescheur The database settings (among other things) are configurable via `env` variables because they differ in production (different clients use different settings). There is not point in making these paths configurable for Mac users because all production servers run Debian. That's what `local.py` is for, to adapt the settings to your personal environment. I get that it's annoying to stumble over this with no indication on how to solve it; hence the suggestion to update the installation instructions with a section for Mac users.

svenvandescheur:
> @svenvandescheur The database settings (among other things) are configurable via `env` variables because they differ in production (different clients use different settings). There is not point in making these paths configurable for Mac users because all production servers run Debian. That's what `local.py` is for, to adapt the settings to your personal environment. I get that it's annoying to stumble over this with no indication on how to solve it; hence the suggestion to update the installation instructions with a section for Mac users.

I don't see why we should use `local.py` over `.env`. The latter is more flexible (and modern/standars based) over the older local.py which is a leftover from historic setups. They are not mutually exclusive and this patch will not prevent anyone from configuring this, or anything else using a local.py file. 

At tis point I don't see any reason not to support this and I do see reasons why we do want to support this (setting the envar globally for instance) is only to keep in sync with other open-X projects.

pi-sigma:
@alextreme Discussed with @svenvandescheur. The library paths are indeed relevant for Mac/Windows users, but the bigger issue is how we structure our settings, in particular, how we use the settings in `base.py` vis-a-vis development settings. Do we want as few settings as possible in `base.py` and extend/override in `local.py`, which means every developer needs to do that for themselves; or do we want to include these options in `base.py` and make them configurable via `env` variables? I don't mind either way, but it would be good to make a principled decision about how we want to handle this sort of thing going forward.

pi-sigma (approved):

alextreme:
@pi-sigma in principle I'd adhere to using `local.py` for development settings and only use envvars for relevant deployment-time settings (either dockerized or not). The reasoning behind this is mostly due to envvars in Kubernetes setups being shared between services on the same node which can cause surprises ( https://github.com/kubernetes/kubernetes/issues/89033 ), you only want to listen to envvars if you need to.

PR #851, opened 2023-11-17
[#1706] Redesign personal info page

pi-sigma:
Taiga [#1706](https://taiga.maykinmedia.nl/project/open-inwoner/task/1706)

### DIFF
--- a/src/open_inwoner/accounts/forms.py
+++ b/src/open_inwoner/accounts/forms.py
@@ -196,7 +196,7 @@ class NecessaryUserForm(forms.ModelForm):
         self.fields["infix"].required = False
         self.fields["last_name"].required = True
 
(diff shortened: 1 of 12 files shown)


jiromaykin (changes requested):
I don't think I have even been able to review all the changes necessary - it's a lot of work, but here is a start.
Also: for content that is too wide for the card you would _either_ need to make an ellipsis CSS style (with this combo: `white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;`) _OR_ you could change the number of columns for tablet size screens,
Perhaps get some inspiration here: https://github.com/maykinmedia/open-inwoner/pull/840/files
 
 
Screenshot 2023-11-28 at 17 58 47

alextreme (approved):

PR #854, opened 2023-11-21
:sparkles: [#1855] Add mock eHerkenning login

stevenbal:
~~Based on https://github.com/maykinmedia/open-inwoner/pull/852, so that PR should be merged first~~

task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1855

I decided to implement the mock in `src/eherkenning`, because I didn't take the time to create a pretty mock login interface for eHerkenning (like it exists for DigiD in the `django-digid-eherkenning` package). I think we could move this to the library later on if necessary


alextreme:
Doing it in src/foo is fine (this is the way django-log-outgoing-requests was started within OIP by Vasileios).

As discussed we have an envvar that must be set to have the eHerkenning mock active, just like with DigiD.

@pi-sigma please do a first review

src/eherkenning/backends.py
```diff
@@ -0,0 +1,37 @@
...
+class eHerkenningBackend(_eHerkenningBackend):
+    """
+    Custom backend to identify users based on the KvK number instead of RSIN
+    """
+
+    def get_or_create_user(self, request, saml_response, saml_attributes):
+        kvk = self.get_kvk_number(saml_attributes)
+        if kvk == "":
+            error_message = "Login failed due to no KvK being returned by eHerkenning."
+            raise eHerkenningError(error_message)
```
pi-sigma, line 19: `raise eHerkenningError("Login failed due to no KvK being returned by eHerkenning.")`

(The variable is not used elsewhere; might as well use the string directly)

src/eherkenning/mock/backends.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def authenticate(self, request, kvk=None):
+        if kvk is None:
+            return
+
+        if kvk == "":
+            self.log_error(request, self.error_messages["eherkenning_no_kvk"])
+            return
+
+        elif not re.match(r"^[0-9]+$", kvk):
```
pi-sigma, line 57: In the last step you continued with `if`, here (and in the next step) you do `elif`, even though there's no difference. I tend to prefer `if` in these situations b/c the implied "else" is not significant.

src/eherkenning/mock/backends.py
```diff
@@ -0,0 +1,67 @@
...
+        **{
+            "eherkenning_no_kvk": _(
+                "Login failed due to no KvK being returned by eHerkenning."
+            ),
+            "eherkenning_len_kvk": _(
+                "Login failed due to no KvK having more then 8 digits."
+            ),
+            "eherkenning_num_kvk": _("Login failed due to no KvK not being numerical."),
+        }
+    )
```
pi-sigma, line 28: Messages for logging shouldn't be translated. There's no use case, and it's confusing when they end up in log aggregating systems translated into the default language of the project. I know the `digid-eherkenning` library does this too, but that's a mistake and I filed an issue about it ([link](https://github.com/maykinmedia/django-digid-eherkenning/issues/49)).

src/eherkenning/mock/views/eherkenning.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+        return f"{url}?{urlencode(params)}"
+
+
+class eHerkenningAssertionConsumerServiceMockView(View):
+    """
+    mock replacement for eherkenning:acs
+    """
+
+    def get_login_url(self):
```
pi-sigma, line 64: Isn't the name misleading, as the function returns either the `login_url` or the `cancel_url`, depending on the case? Perhaps `get_fail_url` or something like it. This would also highlight the contrast with `get_success_url`

src/eherkenning/templates/eherkenning/mock/eherkenning_base.html
```diff
@@ -0,0 +1,79 @@
...
+        <header class='header'>
+            <div>
+                {% block service_menu %}{% endblock %}
+                <a class="digidmock-anchor-placeholder">
+                    <img alt="Logo Rijksoverheid" class="rijksoverheid-logo" title="Logo Rijksoverheid"
+                        src="{% static 'digid-mock/assets/RO_DigiD_Logo_Homepage.svg' %}"/>
+                </a>
+                <nav class='header-navigation'></nav>
+            </div>
+            <div style='margin-top: 15px'></div>
```
pi-sigma, line 32: Styling for this and the next template is partly or completely (?) inline. Is this preliminary or the intended result for these templates?

src/eherkenning/templates/eherkenning/mock/eherkenning_login.html
```diff
@@ -0,0 +1,43 @@
...
+                <a class="language-menu-item digidmock-anchor-placeholder" data-locale="en">EN</a>
+            </li>
+            <li>
+                <a class="language-menu-item active digidmock-anchor-placeholder" data-locale="nl">NL</a>
+            </li>
+        </ul>
+    </nav>
+{% endblock %}
+
+
```
pi-sigma, line 16: Extra empty line

src/eherkenning/tests/test_mock_views.py
```diff
@@ -0,0 +1,184 @@
...
+        # verify no links to eHerkenning remain in template
+        self.assertNotContains(response, "://eherkenning.nl")
+        self.assertNotContains(response, "://www.eherkenning.nl")
+
+
+OVERRIDE_SETTINGS = dict(
+    EHERKENNING_MOCK_APP_TITLE="FooBarBazz-MockApp",
+    EHERKENNING_MOCK_RETURN_URL=RETURN_URL,  # url to redirect to after success
+    EHERKENNING_MOCK_CANCEL_URL=CANCEL_URL,  # url to navigate to when users clicks 'cancel/annuleren'
+    # ROOT_URLCONF="tests.project.mock_urls",
```
pi-sigma, line 25: Remove or uncomment?

src/open_inwoner/accounts/managers.py
```diff
@@ -29,16 +29,15 @@ class eHerkenningManager(BaseeHerkenningManager):
...
 
-    def get_by_rsin(self, rsin):
-        return self.get_queryset().get(rsin=rsin)
+    def get_by_kvk(self, kvk):
+        return self.get_queryset().get(kvk=kvk)
 
-    def eherkenning_create(self, rsin, **kwargs):
-        raise NotImplementedError("old code, please verify before use")
+    def eherkenning_create(self, kvk, **kwargs):
         # TODO what is this @rsin.com email hostname? @example.org is bad enough but this actually exists
```
pi-sigma, line 36: You can remove the comment now, Bart will be happy :)

src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -27,6 +27,18 @@ <h1 class="h1">{% trans 'Welkom' %}</h1>
                     {% endrender_card %}
 		{% endif %}
 
+        {% if eherkenning_enabled %}
+                    {% render_card direction='horizontal' tinted=True %}
```
pi-sigma, line 31: Indentation seems off.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -139,3 +144,86 @@ def get_success_url(self):
             del session["invite_url"]
 
         return super().get_success_url()
+
+
+class CustomeHerkenningAssertionConsumerServiceMockView(
+    eHerkenningAssertionConsumerServiceMockView
+):
+    def get_login_url(self):
```
pi-sigma, line 152: Could be `get_fail_url` or sth. similar, as in a previous comment.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -139,3 +144,86 @@ def get_success_url(self):
...
+            and reverse("profile:registration_necessary") in next_url
+        ):
+            # If user logs in via the invitation flow redirect to the invitation
+            # accept view if login fails
+            absolute_invite_url = self.request.build_absolute_uri(invite_url)
+            return absolute_invite_url
+        elif (
+            invite_url
+            and next_url
+            and not reverse("profile:registration_necessary") in next_url
```
pi-sigma, line 171: With `elif`, you can remove the last clause.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -139,3 +144,86 @@ def get_success_url(self):
...
```
pi-sigma, line 214: See previous comment.

src/open_inwoner/conf/production.py
```diff
@@ -29,6 +29,12 @@
 else:
     DIGID_ENABLED = False
 
+EHERKENNING_MOCK = config("EHERKENNING_MOCK", default=False)
+if not DEBUG:
+    AUTHENTICATION_BACKENDS += ["eherkenning.backends.eHerkenningBackend"]
+elif EHERKENNING_MOCK:
+    AUTHENTICATION_BACKENDS += ["eherkenning.mock.backends.eHerkenningBackend"]
```
pi-sigma, line 36: This means that `DEBUG` could be true in production? I would expect that perhaps in `conf.base` we conditionally add `eherkenning.mock.backends.eHerkenningBackend`, depending on the value of `DEBUG` and `EHERKENNING_MOCK`, but `production.py` simply has `AUTHENTICATION_BACKENDS += ["eherkenning.backends.eHerkenningBackend"]`. Perhaps I misunderstand how the settings are used.

src/open_inwoner/configurations/models.py
```diff
@@ -485,6 +485,15 @@ class SiteConfiguration(SingletonModel):
         ),
     )
 
+    # authentication options
+    eherkenning_enabled = models.BooleanField(
+        verbose_name=_("eHerkenning authentication enabled"),
+        default=False,
+        help_text=_(
+            "Whether we want the users to be able to log in with eHerkenning or not."
```
pi-sigma, line 493: `"Whether users can log in with eHerkenning or not."`

pi-sigma (dismissed):
Looks good to me. Couple minor remarks, and there's no link to the eHerkenning mock when you follow to `accounts/login`, only the DigiD mock.

pi-sigma:
`MyProfileView` adds the following policy for changing user passwords to the context ([link](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/views/profile.py#L100)):

`context["can_change_password"] = user.login_type != LoginTypeChoices.digid`

I suspect this should be amended to

`context["can_change_password"] = user.login_type != (LoginTypeChoices.digid or LoingTypeChoices.eherkenning)`

src/open_inwoner/conf/production.py
alextreme: I'd recommend to do this the same as how this is done with the DigiD backends

src/eherkenning/templates/eherkenning/mock/eherkenning_base.html
stevenbal: I copied the template from the DigiD mock flow from the `digid-eherkenning` library, to not spend too much time on the looks of it, since I think this template in particular probably needs attention later on if we want to get this into the library

stevenbal:
@pi-sigma 
> there's no link to the eHerkenning mock when you follow to accounts/login, only the DigiD mock.

that's probably because the feature is disabled by default, if you enable it via the General configuration in the admin, it should show up:

image

stevenbal:
@alextreme I did also notice that the unique constraint for `email` on `User` is not applied to `DigiD` (https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/models.py#L236), do we want to exempt `eHerkenning` from this as well? We do use the KvK number in the email for users logging in with `eHerkenning`, so I think the `email`s would be unique anyway?

src/eherkenning/templates/eherkenning/mock/eherkenning_base.html
alextreme: Styling this form is out-of-scope as there isn't a single eHerkenning look-and-feel. A plain html form was the minimum, and this looks a lot neater.

src/open_inwoner/conf/production.py
stevenbal: Discussed with Alex: I'll remove the check for `DEBUG` and simply check if `EHERKENNING_MOCK` is true

alextreme:
> @alextreme I did also notice that the unique constraint for `email` on `User` is not applied to `DigiD` (https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/models.py#L236), do we want to exempt `eHerkenning` from this as well? We do use the KvK number in the email for users logging in with `eHerkenning`, so I think the `email`s would be unique anyway?

Email shouldn't be unique for DigiD or eHerkenning. For the latter, I suspect there will be cases where different related companies want to use a single emailadres for notifications.

src/open_inwoner/accounts/views/auth.py
stevenbal: I modified this to be `get_failure_url` in the mock views in `src/eherkenning`. Because the `eHerkenningAssertionConsumerServiceMockView` relies on `get_login_url` in `handle_error` (https://github.com/maykinmedia/django-digid-eherkenning/blob/master/digid_eherkenning/views/eherkenning.py#L91), I would have to override that method as well, which seems a bit excessive if we're just changing a method name :/

src/eherkenning/backends.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_or_create_user(self, request, saml_response, saml_attributes):
+        kvk = self.get_kvk_number(saml_attributes)
+        if kvk == "":
+            raise eHerkenningError(
+                "Login failed due to no KvK being returned by eHerkenning."
+            )
+
+        created = False
+        try:
+            user = UserModel.eherkenning_objects.get_by_kvk(kvk)
```
alextreme, line 24: zou nog fout kunnen gaan bij meerdere users met hetzelfde kvk nummer. Maar dat is nu ook het geval voor DigiD en een 500 in deze situatie lijkt mij prima.

src/eherkenning/mock/idp/forms.py
```diff
@@ -0,0 +1,19 @@
...
+from django.utils.translation import gettext_lazy as _
+
+from eherkenning.validators import KVKValidator
+
+
+class eHerkenningPasswordLoginForm(forms.Form):
+    auth_name = forms.CharField(
+        max_length=255,
+        required=True,
+        label=_("eHerkenning gebruikersnaam"),
```
alextreme, line 11: Eigenlijk vind ik de DigiD (en eHerkenning) mock net teveel lijken op de echte DigiD. Gewoon 1 veld met "KvK nummer" vind ik duidelijker en eerlijker. Maar voor nu voldoende.

src/open_inwoner/conf/production.py
```diff
@@ -29,6 +29,12 @@
 else:
     DIGID_ENABLED = False
 
+EHERKENNING_MOCK = config("EHERKENNING_MOCK", default=False)
```
alextreme, line 32: Gechecked, dit zal dan gelden voor test+staging+production en lijkt me in orde

src/open_inwoner/configurations/models.py
```diff
@@ -485,6 +485,13 @@ class SiteConfiguration(SingletonModel):
         ),
     )
 
+    # authentication options
+    eherkenning_enabled = models.BooleanField(
+        verbose_name=_("eHerkenning authentication enabled"),
+        default=False,
```
alextreme, line 491: Akkoord, is opt-in

PR #855, opened 2023-11-21
[#1795] Add backend support for multiple file upload

pi-sigma:
Taiga: [#1795](https://taiga.maykinmedia.nl/project/open-inwoner/task/1795)

Front-end: [PR #833](https://github.com/maykinmedia/open-inwoner/pull/833)

### DIFF
--- a/src/open_inwoner/cms/cases/tests/test_htmx.py
+++ b/src/open_inwoner/cms/cases/tests/test_htmx.py
@@ -433,7 +433,7 @@ class CasesPlaywrightTests(
         expect(upload_form).to_be_visible()
 
         file_input = upload_form.get_by_label(
-            _("Sleep of selecteer bestand"), exact=True
+            _("Sleep of selecteer bestanden"), exact=True
         )
         file_input.set_input_files(
             files=[
--- a/src/open_inwoner/cms/cases/tests/test_htmx.py
+++ b/src/open_inwoner/cms/cases/tests/test_htmx.py
@@ -504,17 +504,21 @@ class CasesPlaywrightTests(
             Mock GET "zaakinformatieobjecten" endpoint (schemas/ZaakInformatieObject array).
             Creates schemas/ZaakInformatieObject dict for each item in uploads.
             """
-            items = [generate_oas_component(
-                "zrc",
-                "schemas/ZaakInformatieObject",
-                url=f"{ZAKEN_ROOT}zaakinformatieobjecten/e55153aa-ad2c-4a07-ae75-15add57d6",
-                informatieobject=upload["url"],
-                zaak=self.zaak["url"],
-            ) for upload in uploads]
+            items = [
+                generate_oas_component(
+                    "zrc",
+                    "schemas/ZaakInformatieObject",
+                    url=f"{ZAKEN_ROOT}zaakinformatieobjecten/e55153aa-ad2c-4a07-ae75-15add57d6",
+                    informatieobject=upload["url"],
+                    zaak=self.zaak["url"],
+                )
+                for upload in uploads
+            ]
             return json.dumps(items)
 
         m.get(
-            f"{ZAKEN_ROOT}zaakinformatieobjecten?zaak={self.zaak['url']}", text=mock_list
+            f"{ZAKEN_ROOT}zaakinformatieobjecten?zaak={self.zaak['url']}",
+            text=mock_list,
         ),
 
         # Upload mock.
@@ -529,7 +533,7 @@ class CasesPlaywrightTests(
 
             # Create a UUID based on a seed derived from the file name.
             # This makes sure the two test cases have unique entries.
-            seed = file_name.ljust(16, '0').encode("utf-8")
+            seed = file_name.ljust(16, "0").encode("utf-8")
             uuid = UUID(bytes=seed)
 
             uploaded_informatie_object = generate_oas_component(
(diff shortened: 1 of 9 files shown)


src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -432,7 +432,7 @@ def test_cases(self, m):
         upload_form = page.locator("#document-upload")
         expect(upload_form).to_be_visible()
 
-        file_input = upload_form.get_by_text(_("Sleep of selecteer bestand"))
+        file_input = page.locator("[name='file']")
```
svenvandescheur, line 435: I think we should prefer the [get_by_label](https://playwright.dev/python/docs/api/class-page#page-get-by-label) where possible (or any other selector that testes a11y).

src/open_inwoner/cms/cases/views/status.py
svenvandescheur, line 1: If I understand correctly, a single failed upload will not stop the other uploads?

svenvandescheur (changes requested):
Would recommend using get_by_label, other than that, based on a quick look this seems fine to me (didn’t do ann actual test yet though).

src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -3,7 +3,7 @@
 <div class="form__control file-input">
     {% render_card direction="vertical" %}
         {% icon icon="upload" icon_position="before" outlined=True %}
-        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %}>
+        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %} multiple>
```
svenvandescheur, line 6: Should this be dynamic based on the field type, and shouldn't we (dynamically) change the labels into plurals accordingly?

pi-sigma: `multiple` should be set dynamically and has been fixed. I'm not sure about the labels. There's no change in the backend if the client drops a file or two into the widget, so I'm not sure what could decide if the label should be singular or plural. Passing that info front to back seems overkill. I tend towards leaving it as is, or perhaps use an "optional plural" form like "Selecteer bestand(en)"

src/open_inwoner/cms/cases/views/status.py
pi-sigma: The entire upload flow is interrupted if a single upload fails. Discussed with Alex. This is the intended behavior.

src/open_inwoner/cms/cases/tests/test_htmx.py
pi-sigma: I changed the `id` and `for` attributes for the `input` and `label` elements to make the test work with `get_by_label`. Let me know if this is not desirable.

svenvandescheur (approved):

src/open_inwoner/cms/cases/tests/test_htmx.py
svenvandescheur: I did do a small change, instead of using a hardcoded id I used the "auto_id" for both the input and the label.

src/open_inwoner/components/templates/components/Form/FileInput.html
svenvandescheur: I changed the label simply based on multiple. I don't think we need to have plural forms based on the length of the selection.

src/open_inwoner/templates/pages/cases/document_form.html
```diff
@@ -4,6 +4,6 @@
 {% render_form id="document-upload" form=form method="POST" hxencoding="multipart/form-data" hxpost=hxpost_document_action hxtarget="#form_upload" submit_text=_("Bestand uploaden") extra_classes="case-detail-form" %}
     {% csrf_token %}
     {% input form.type no_label=True no_help=True class="label input" id="id_type" extra_classes="file-type__select" %}
```
jiromaykin, line 6: `no_label=True` was already set to True on Develop so you may experience a merge conflict here with the previous commit - All else looking good!

jiromaykin (approved):
Nice work - the Success notification does have the annoying habit of becoming sticky and it should get into Focus after a successful upload, but that's not part of this task.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -505,9 +505,14 @@ def handle_document_upload(self, request, form):
...
             )
-            created_relationship = connect_case_with_document(
-                self.case.url, getattr(created_document, "url", None)
-            )
+
+            try:
+                created_relationship = connect_case_with_document(
+                    self.case.url, created_document.get("url")
+                )
+            except AttributeError:
```
svenvandescheur, line 513: @pi-sigma I've added the AttributeError handling here because created_document could be `None`, however I think we might want to separate the error handling for the both fields only create the relationship if the document is actually created.

This might requires some small refactoring here.

svenvandescheur (changes requested):

src/open_inwoner/cms/cases/views/status.py
pi-sigma: @svenvandescheur I don't think we need to expose this to the user; they only care about whether the upload worked or not.

Relatedly, I tested the code again and it now fails with `getattr` for me too, as it should. So I think we're good to go.

svenvandescheur (approved):

PR #856, opened 2023-11-21
:art: [#1822] Final corrections for case pages

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1822 

- [x] remove double status sections
- [x] add missing configurable Status/Result content texts (Status content in current + in completed/past paragraphs, Result content only in final status paragraph)
- [x] checking: min-width should be 768px everywhere, but max-width should be 767px (**breakpoints**)
- [x] center Spinner the same in all Aanvragen pages

+part of issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1870

- [x]  decrease vertical spacing between components


src/open_inwoner/templates/pages/cases/statuses.html
```diff
@@ -21,7 +21,16 @@ <h2 class="h3 h2-statuses" id="statuses">{% trans 'Status' %}</h2>
                             {% if case.end_statustype_data %}{% icon icon="expand_more" icon_position="after" outlined=True %}{% endif %} </button>
                         <div class="status-list__notification-content {% if case.end_statustype_data %}status-content--open{% endif %} {% if not case.end_statustype_data %}status-content--open{% endif %}" id="{{ content_id }}">
                             <p class="p p-text p--compact">
-                                {{ case_type_result_type_config.description|default:"" }}</p>
+                                {# Terrible experiments to see what configured texts can be retreived to the front-end #}
```
jiromaykin, line 24: @pi-sigma Since you worked on [the PR#1843 here](https://github.com/maykinmedia/open-inwoner/pull/841/files), I was hoping you coul tell me how to retrieve that text to this template.

jiromaykin: Same question for @stevenbal : for case that are behandeld/afgehandeld, I don't see any of the configured texts appear with {{case_type_result_type_config.description}} so I am wondering if that is correct code to retrieve it to the front-end.

stevenbal: I don't see anything being passed to the frontend relating to result_type_config (https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/cms/cases/views/status.py#L307-L316 / https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/cms/cases/views/status.py#L192-L223), so I think we have to add this first

PR #857, opened 2023-11-22
[#1860] Prefixing subject with "Onderwerp"

alextreme:


src/open_inwoner/openklant/views/contactform.py
```diff
@@ -163,7 +163,7 @@ def register_by_api(self, form, config: OpenKlantConfig):
         subject = form.cleaned_data["subject"].subject
         subject_code = form.cleaned_data["subject"].subject_code
         question = form.cleaned_data["question"]
-        text = f"{subject}\n\n{question}"
+        text = f"Onderwerp: {subject}\n\n{question}"
```
pi-sigma, line 166: `text = _("Onderwerp: {subject}\n\n{question}").format(subject=subject, question=question)`

pi-sigma (changes requested):

src/open_inwoner/openklant/views/contactform.py
alextreme: Feel free to make this translated if you want (I agree that this was decided), but note that this would also require making the other labels in this form/view translated as there are quite a few hardcoded.

pi-sigma: Will do (might as well, while we're at it)

pi-sigma: @alextreme take a quick look to see if I missed something (the remaining untranslated strings should all be logging-related)

alextreme: This seems fine to me, thanks for cleaning it up!

PR #858, opened 2023-11-23
:lipstick: [#1817] Show errors for selected documents 

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1817


pi-sigma (changes requested):
A lingering log statement, otherwise looking good!

src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -1,15 +1,23 @@
...
+        <div class="file-input__selected">
+            <label class="button button--borderless" for="{{ field.auto_id }}">
+            {% icon icon="add_circle_outlined" icon_position="before" outlined=True %}
+                {% if multiple %}{% trans 'Selecteer meer bestanden' %}{% else %}{% trans 'Selecteer ander bestand' %}{% endif %}
+            </label>
+        </div>
     {% endrender_card %}
 
-    {% if field.help_text %}<p class="p p--muted p--small">{{ field.help_text }}</p>{% endif %}
+    {% if field.help_text %}<p class="p p--small p--upload-info">{{ field.help_text }}</p>{% endif %}
```
svenvandescheur, line 20: Is there any reason why the help text styling should be changes?

src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -1,15 +1,23 @@
...
-<div class="form__control file-input">
+{% get_solo 'openzaak.OpenZaakConfig' as openzaak_config %}
+
+<div class="form__control file-input" aria-live="polite">
     {% render_card direction="vertical" %}
         {% icon icon="upload" icon_position="before" outlined=True %}
-        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %}{% if multiple %} multiple{% endif %}>
-        <label class="button button--primary" for="{{ field.auto_id }}">
+        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %}{% if multiple %} multiple{% endif %} data-max-size="{{ openzaak_config.max_upload_size }}">
+        <label class="button button--primary file-input__empty" for="{{ field.auto_id }}">
```
svenvandescheur, line 9: Can we either:

- Set the label(s) to be a an element (`file-input__label`) with a modifier set to is's meaning (`file-input__label file-input__label--empty`).
- Or make the individual names more descriptive (`file-input__label-empty`)

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -23,6 +31,22 @@ export class FileInput extends Component {
     return this.node.querySelector(`${FileInput.selector}__input`)
   }
 
+  /**
+   * Return the label when zero files are selected.
+   * @return {HTMLInputElement}
```
svenvandescheur, line 36: The return type for a label element is `HTMLLabelElement`.

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -137,22 +169,22 @@ export class FileInput extends Component {
...
+   * @param {Event} e - The event.
    */
   noop(e) {
     e.preventDefault()
   }
 
   /**
    * Adds files in dataTransfer to input, only the first item is added if not `[multiple]`.
-   * @param {File[]} files
+   * @param {File[]} files - Array of files to add.
```
svenvandescheur: Technically this is a `FileList`, not an Array (even though the jsdoc type annotation is incorrect here as well).

src/open_inwoner/scss/components/Form/FileInput.scss
```diff
@@ -49,12 +64,42 @@
     text-align: center;
   }
 
+  .p--upload-error {
```
svenvandescheur, line 67: Modifiers should have names that make sense in a very generic way. "upload-error" is not a generic style for a piece of text. 

- If the styling is very specific to a component than the "upload-error" should be a part of that component (and the classname should look something like `file-input__upload-error`.
- If however, the styling a generic way to demonstrate errors (or "danger texts") `p--danger` might be better here (but then it should be defined in the scss file for the "P" component.
- If we need to reuse the "upload error" and it's to complex to be composed out of other components with modifiers it can be a block on it's own. In which case it should have it's own file/folder in at least in the scss structure. 

Based on the code my guess is that first scenario applies here. 

BEM in short:

- Block: Component name for something reusable, e.g. `card`.
- Element: A part of a component that cannot be (re)used without a block, e.g. `card__body`.
- Modifier: An additional class that indicates a certain "state", "theme", "type" etc. of either a block or an element, e.g. `card card--info` or `card__body card__body--direction-vertical`.

src/open_inwoner/templates/pages/cases/document_form.html
```diff
@@ -5,5 +5,6 @@
     {% csrf_token %}
     {% input form.type no_label=True no_help=True class="label input" id="id_type" extra_classes="file-type__select" %}
     {% file_input form.files %}
-    {% form_actions primary_text=_("Upload documenten") primary_icon="arrow_forward" enctype="multipart/form-data" %}
+    {% form_actions primary_text=_("Upload documenten") enctype="multipart/form-data" %}
+    <p class="p--upload-error p--small p--upload-info p--upload-help">{% trans "Verwijder eerst bestanden die niet voldoen aan de voorwaarden" %}</p>
```
svenvandescheur, line 9: A class `p--upload-error` is being referenced here which should be (see previous comment):

- An actual modifier of "P", which could be correct but in which case it's specified in the wrong file and with an to broad modifier name.
- An element of file-input in which case it can/should not be used outside the template for the file input.
- A component (block) with it's own scss file (and possbly template) in which case it's interface and meaning should be properly defined.

svenvandescheur (commented):
On the right track, some remarks about naming convention and coding style though.

pi-sigma:
This looks good to me. As far as I can see, Sven's comments have been addressed.
@svenvandescheur Can you take a brief look at this and confirm or disconfirm?

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -23,6 +31,22 @@ export class FileInput extends Component {
     return this.node.querySelector(`${FileInput.selector}__input`)
   }
 
+  /**
+   * Return the label when zero files are selected.
+   * @return HTMLInputElement
```
svenvandescheur, line 36: `@return {HTMLInputElement}`

Don't forget the brackets ;)

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -23,6 +31,22 @@ export class FileInput extends Component {
...
+   */
+  getLabelEmpty() {
+    return this.node.querySelector(`${FileInput.selector}__label-empty`)
+  }
+
+  /**
+   * Return the label if more than 0 files are selected.
+   * @return HTMLInputElement
```
svenvandescheur, line 44: Same, also this is probably returns an: `HTMLLabelElement`.

src/open_inwoner/scss/components/Form/NonFieldError.scss
```diff
@@ -0,0 +1,6 @@
+.form {
+  &__non-field-error {
```
svenvandescheur, line 2: This is incorrect, if a non field error is a component (or) block on it's own it should not be an element of something else (form). This is mixture happens on other places as well IIRC (form__control?) but lets not increase the usage of that pattern here.

The correct block name IMO should be "non-field-error" and should not contain the form class name.

src/open_inwoner/templates/pages/cases/document_form.html
```diff
@@ -5,5 +5,8 @@
     {% csrf_token %}
     {% input form.type no_label=True no_help=True class="label input" id="id_type" extra_classes="file-type__select" %}
     {% file_input form.files %}
-    {% form_actions primary_text=_("Upload documenten") primary_icon="arrow_forward" enctype="multipart/form-data" %}
+    {% form_actions primary_text=_("Upload documenten") enctype="multipart/form-data" %}
+    <div class="form__non-field-error" hidden>
```
svenvandescheur, line 9: See remark about class name.

svenvandescheur (changes requested):
Small last remarks.

alextreme:
Discussed with Paul and Sven because of the remarks and the draft-PR status, and I choose to merge this in due to the pending release. The remaining remarks are small enough that they can be resolved at a later moment.

PR #859, opened 2023-11-23
:sparkles: [#1866] Enable category personalization based on KvK number

stevenbal:
Based on https://github.com/maykinmedia/open-inwoner/pull/854, so that should be merged first

task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1866

### DIFF
--- a/src/open_inwoner/accounts/tests/factories.py
+++ b/src/open_inwoner/accounts/tests/factories.py
@@ -34,6 +34,12 @@ class DigidUserFactory(UserFactory):
     is_prepopulated = True
 
 
+class eHerkenningUserFactory(UserFactory):
+    login_type = LoginTypeChoices.eherkenning
+    kvk = "12345678"
+    is_prepopulated = True
+
+
 class TokenFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = "authtoken.Token"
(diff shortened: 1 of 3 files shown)


src/open_inwoner/pdc/tests/test_views.py
```diff
@@ -121,12 +125,11 @@ def test_category_list_view_visibility_for_digid_user(self):
             list(response.context["object_list"]), [self.category2, self.category3]
         )
 
-    @skip("eHerkenning is not implemented yet")
     def test_category_list_view_visibility_for_eherkenning_user(self):
         url = reverse("products:category_list")
 
         # TODO should be eHerkenningUserFactory
```
pi-sigma, line 131: Remove comment

src/open_inwoner/accounts/tests/factories.py
```diff
@@ -34,6 +34,12 @@ class DigidUserFactory(UserFactory):
     is_prepopulated = True
 
 
+class eHerkenningUserFactory(UserFactory):
+    login_type = LoginTypeChoices.eherkenning
+    kvk = "12345678"  # TODO ideally this should be random (but still valid)
```
pi-sigma, line 39: I would remove the comment. The `DigidUserFactory` has the same comment, which is from December last year. Who is going to implement this and when?

pi-sigma (changes requested):
Looking good! I would remove some comments.

pi-sigma (approved):

PR #861, opened 2023-11-27
:sparkles: [#1878] Post-login registration form for eHerkenning

stevenbal:
Based on https://github.com/maykinmedia/open-inwoner/pull/859, so that should be merged first

task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1878

Aside from adapting the necessary registration form, the following was also changed/added:
- added eHerkenning link to registration page
- made DigiD/eHerkenning links on registration page conditional based on `settings.DIGID_ENABLED` and `SiteConfiguration.eherkenning_enabled`

### DIFF
--- a/src/open_inwoner/accounts/forms.py
+++ b/src/open_inwoner/accounts/forms.py
@@ -208,6 +208,9 @@ class NecessaryUserForm(forms.ModelForm):
                 del self.fields["last_name"]
             if not user.infix:
                 del self.fields["infix"]
+        elif user.login_type == LoginTypeChoices.eherkenning:
+            for field_name in ["first_name", "infix", "last_name"]:
+                del self.fields[field_name]
 
 
 class CustomPasswordResetForm(PasswordResetForm):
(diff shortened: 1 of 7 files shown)


stevenbal:
I removed the first and last name fields from this form, I did notice however that the generated email is prefilled in the form:

image

I'm not sure if this is intentional, do we want this? @alextreme

alextreme:
> I removed the first and last name fields from this form, I did notice however that the generated email is prefilled in the form:
> 
> image
> 
> I'm not sure if this is intentional, do we want this? @alextreme

How we deal with emailadresses and DigiD is a bit odd, but as it's a required field it is auto-generated. I'd recommend to do the same with eHerkenning, and if it's auto-generated that we hide this value just like with DigiD (it is overwritten anyway).

stevenbal:
> and if it's auto-generated that we hide this value just like with DigiD (it is overwritten anyway).

Ah I see, I'll implement it the same for eHerkenning

alextreme (approved):
Looks good to me

PR #869, opened 2023-12-01
[#1867] Prevent retrieval of zaken with missing rsin

pi-sigma:
Prevents retrieval of Zaken if fetching with RSIN is enabled but RSIN is missing

Taiga [#1867](https://taiga.maykinmedia.nl/project/open-inwoner/task/1867)


alextreme (approved):
I'll create a task to add a regressiontest

PR #875, opened 2023-12-04
[#1898] Design profile page for businesses

pi-sigma:
Taiga [#1895](https://taiga.maykinmedia.nl/project/open-inwoner/task/1898)

### DIFF
--- a/src/open_inwoner/templates/pages/profile/me.html
+++ b/src/open_inwoner/templates/pages/profile/me.html
@@ -17,7 +17,7 @@
     </p>
 </section>
 
-{% if user.kvk %}
+{% if user.is_eherkenning_user %}
     {# Business information #}
     <section class="tabled tabled--flexible profile-section profile-section--bordered">
         <div class="tabled__section">
--- a/src/open_inwoner/templates/pages/profile/me.html
+++ b/src/open_inwoner/templates/pages/profile/me.html
@@ -10,49 +10,78 @@
 {% with user=request.user %}
 
 <div class="tabled tabled--flexible">
-    <h1 class="h1 tabled__title">{% trans "Welkom" %}, {{ user.first_name }}</h1>
-    <span class="tabled__key">{%trans "voor het laatst ingelogd op" %}: <span class="tabled__value">{{ user.last_login|date:"j F Y" }} om {{ user.last_login|date:"H:i" }} uur</span></span>
+    <h1 class="h1" id="title">{% trans "Welkom" %}{% if not user.is_eherkenning_user %}, {{ user.first_name }}{% endif %}</h1>
+    <p class="tabled__key">
+    {%trans "voor het laatst ingelogd op" %}: <span class="tabled__value">{{ user.last_login|date:"j F Y" }} om {{ user.last_login|date:"G:i" }} uur</span>
+    </p>
 </div>
 
-{# Personal information #}
-<div class="tabled tabled--flexible profile-section" id="personal-info">
-    <div class="tabled__section">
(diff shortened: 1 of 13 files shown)


src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -10,49 +10,70 @@
 {% with user=request.user %}
 
 <div class="tabled tabled--flexible">
-    <h1 class="h1 tabled__title">{% trans "Welkom" %}, {{ user.first_name }}</h1>
-    <span class="tabled__key">{%trans "voor het laatst ingelogd op" %}: <span class="tabled__value">{{ user.last_login|date:"d-m-Y" }} om {{ user.last_login|date:"H:i" }} uur</span></span>
+    <h1 class="h1" id="title">{% trans "Welkom" %}{% if not user.is_digid_user %}, {{ user.first_name }}{% endif %}</h1>
```
jiromaykin, line 13: The optional comma might show for companies that do not have a first-name.

jiromaykin (approved):
I am approving this for the front-end at least.

src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -9,228 +9,249 @@
...
-                {% endif %}
+<section class="tabled tabled--flexible">
+    <h1 class="h1" id="title">{% trans "Welkom" %}{% if not user.is_eherkenning_user %}, {{ user.first_name }}{% endif %}</h1>
+    <p class="tabled__key">
+    {% trans "voor het laatst ingelogd op" %}: <span
+    class="tabled__value">{{ user.last_login|date:"j F Y" }} om {{ user.last_login|date:"G:i" }} uur</span>
+    </p>
+</section>
+
+{% if user.kvk %}
```
stevenbal, line 20: Can't we use `user.is_eherkenning_user` here as well?

stevenbal (approved):
Looks good, I do have some questions (might be the case that this is by design, but I just want to double check):
- There is a `bewerk` button above `voorkeur voor meldingen`, bus only `ontvang meldingen over can be edited` and `communicatiekanaal` cannot be edited? 
image

- `bekijk alle gegevens` is present in the design but I don't think it's implemented? image
- I couldn't find `email` anywhere in the profile?

jiromaykin:
> * `bekijk alle gegevens` is present in the design but I don't think it's implemented? 

That's a good question @stevenbal ; I am not sure there _is_ actually any more data to show for Bedrijven, so perhaps @pi-sigma can answer this.

> * I couldn't find `email` anywhere in the profile?

In future there is the wish to also add SMS/textmessaging but this functionality doesn't exist yet; not sure if this is too much of a problem but perhaps it does feel like people are given a choice to edit something here and then they are let down because they actually can't.
What do you think @alextreme ?

alextreme:
> > * I couldn't find `email` anywhere in the profile?
> 
> In future there is the wish to also add SMS/textmessaging but this functionality doesn't exist yet; not sure if this is too much of a problem but perhaps it does feel like people are given a choice to edit something here and then they are let down because they actually can't. What do you think @alextreme ?

I discussed this and simply stating E-mail as per the design is fine.

pi-sigma:
@jiromaykin We could retrieve and display various companay data via the kvk api, but this is not implemented and wasn't part of the spec, as far as I remember. If this is desired, we should create a new ticket and discuss what data should be retrieved.

@stevenbal You're right about both points; will fix and push final changes.

stevenbal:
Looks good to me :+1:

PR #876, opened 2023-12-04
:bug: [#1904] Fix Case list crashing due to reused Statustypen

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1904

if the ZGW APIs reuse StatusTypen for different ZaakTypen, we end up with multiple `ZaakTypeStatusTypeConfig` that have the same `statustype_url`, which led to errors when retrieving these configs. To fix this we explicitly filter for the correct `ZaakTypeConfig` as well


pi-sigma (changes requested):
I would document the use of the second status type in the test; otherwise this will likely appear mysterious in the future. Otherwise this looks good to me.

alextreme (approved):

pi-sigma (approved):

PR #880, opened 2023-12-05
:alembic: [#1910] Add NLDS test-component as experiment

jiromaykin:
Just a test PR to see if designtoken works together with Django, issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1910


jiromaykin:
Closing this experiment, because this is not the correct way to use Style-Dictionary :-) the correct way can be seen in the new git-module here: https://github.com/maykinmedia/open-inwoner-design-tokens - however this experiment does show the correct way to set a wrapper theme class and set the correct classnames for an Utrecht heading.

PR #881, opened 2023-12-05
:recycle: [#1889] Remove null=True from User CharFields

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1889


alextreme (commented):
Tests mogen nog wat aandacht krijgen maar prima om dit te corrigeren

alextreme (approved):
Looks fine, but a doublecheck from @pi-sigma would be appreciated

PR #884, opened 2023-12-07
[#1874] Enabled selection of company branch for eHerkenning users

pi-sigma:
- Let eHerkenning users choose company branch on login
- Refactored project structure and move KvK app out of contrib

Taiga [1874](https://taiga.maykinmedia.nl/project/open-inwoner/task/1874)


pi-sigma:
@Bartvaderkin `test_any_page_for_digid_user_redirect_to_necessary_fields` is failing locally for unknown reasons. I've tried disabling the kvk middleware, like I did for some of the other tests, but to no avail.

src/open_inwoner/kvk/views.py
```diff
@@ -0,0 +1,90 @@
...
+            request,
+            self.template_name,
+            {
+                "company_branches": company_branches,
+                "form": form,
+            },
+        )
+
+    def post(self, request):
+        form = self.get_form()
```
stevenbal, line 77: I tested this locally and it's possible to (after logging in with a KvK number) perform a request with the same session ID (with postman for example) and a branch number that does not belong to the KvK number you logged in with. I think it would be good to validate that the `branch_number` present in the data belongs to the KvK number, to avoid security issues

src/open_inwoner/kvk/views.py
```diff
@@ -0,0 +1,90 @@
...
+            company_branches=company_branches,
+            form=form,
+            redirect=redirect,
+        )
+
+        return context
+
+    def get(self, request):
+        if not getattr(request.user, "kvk", None):
+            return HttpResponse(_("Unauthorized"), status=401)
```
stevenbal, line 48: I would expect the same check for the `.post`

src/open_inwoner/scss/components/Card/Card.scss
```diff
@@ -16,6 +16,33 @@
...
+    }
+  }
+
+  .left {
+    float: left;
+  }
+  .right {
+    float: right;
+  }
+  .radio {
```
stevenbal, line 39: Alignment seems a bit off, I'm not sure if we want to fix this in this PR or if Jiro will take a look afterwards

image

src/open_inwoner/kvk/views.py
```diff
@@ -0,0 +1,90 @@
...
+        return render(
+            request,
+            self.template_name,
+            {
+                "company_branches": company_branches,
+                "form": form,
+            },
+        )
+
+    def post(self, request):
```
stevenbal, line 76: I noticed that clicking `Log in` without selecting a branch returns you to the same page, but doesn't show an error message. I think it would be nice show some kind of error message

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -1253,6 +1392,7 @@ def test_page_show_config_text(self):
         self.assertContains(response, "Hello registration text")
         self.assertContains(response, ' href="http://foo.bar/" ')
 
+    @skip("Debug")
```
stevenbal, line 1395: Should be removed probably?

stevenbal (changes requested):

src/open_inwoner/scss/components/Card/Card.scss
alextreme: Discussed with Steven, please re-use the login-button component instead of modifying the cards. Jiro can indeed make this more in line with the design later and I don't want to modify components which are used allover the place

src/open_inwoner/kvk/views.py
alextreme: Please improve this

stevenbal:
@pi-sigma I'll pick up the requested changes for this PR

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -384,6 +394,9 @@ def _setUpMocks(self, m):
         ]:
             m.get(resource["url"], json=resource)
 
+        # TODO
+        # m.get(reverse("kvk:branches"))
```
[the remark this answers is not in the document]
stevenbal: Removed it, because the `KvKLoginMiddleware` is skipped in these tests

src/open_inwoner/scss/components/Card/Card.scss
stevenbal: Modified it and used the existing templatetags, which removes the need for this extra CSS

image

src/open_inwoner/accounts/tests/test_auth.py
stevenbal: Fixed this, the `NecessaryFieldsMiddleware` was essentially disabled due to this change: https://github.com/maykinmedia/open-inwoner/pull/884/commits/68ca2fc755dcc3d145ceabfd1abe283bfd9c90e3#diff-09a70cba3ae30157873d6826d98ae3cf634d9f0c764558ef37e98b4620443217R18. I removed that line and switched the order of the `KvKLoginMiddleware` and `NecessaryFieldsMiddleware` to fix other tests

src/open_inwoner/kvk/views.py
stevenbal: image

src/open_inwoner/kvk/views.py
stevenbal: image

alextreme:
This looks fine for now, I've added #1996 for further styling

alextreme (approved):

PR #887, opened 2023-12-11
[#1885] Changed document admin to be readonly

Bartvaderkin:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1885

In the pentest you could upload here, but this is user content so no need to create/edit it through the admin.

Fields:
- [x] document.file -> frontend widget only allows `settings.UPLOAD_FILE_TYPES`
- [x] action.file -> frontend widget only allows `settings.UPLOAD_FILE_TYPES`
- [x] message.file -> frontend widget only allows `settings.UPLOAD_FILE_TYPES`
- [x] ~~siteconfig.theme_stylesheet~~ -> .css extension validation
- [x] ~~CaseUploadForm.files~~ -> configurable extension validation in admin
- [x] ProductFile.file -> filer already has validation: https://github.com/django-cms/django-filer/blob/master/filer/validation.py#L30
- [x] PlanTemplate.file -> filer already has validation: https://github.com/django-cms/django-filer/blob/master/filer/validation.py#L30
- [x] QuestionnaireStepFile.file -> filer already has validation: https://github.com/django-cms/django-filer/blob/master/filer/validation.py#L30

### DIFF
--- a/src/open_inwoner/accounts/admin.py
+++ b/src/open_inwoner/accounts/admin.py
@@ -142,7 +142,36 @@ class _UserAdmin(ImageCroppingMixin, UserAdmin):
 
 @admin.register(Action)
 class ActionAdmin(UUIDAdminFirstInOrder, PrivateMediaMixin, admin.ModelAdmin):
-    readonly_fields = ("uuid",)
+    fields = [
+        "uuid",
+        "name",
+        "description",
+        "status",
+        "type",
+        "end_date",
+        "file",
+        "is_for",
+        "created_on",
+        "updated_on",
+        "created_by",
+        "plan",
+        "is_deleted",
+    ]
+    readonly_fields = (
+        "uuid",
+        "name",
+        "description",
+        "status",
+        "type",
+        "end_date",
+        "file",
+        "is_for",
+        "created_on",
+        "updated_on",
+        "created_by",
+        "plan",
+        "is_deleted",
+    )
     list_display = ("name", "status", "plan", "created_on", "created_by", "is_deleted")
     list_filter = (
         "is_deleted",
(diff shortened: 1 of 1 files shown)


alextreme:
Note my remark: "This is to be corrected, the default document-file extensions from the Open Zaak configuration can be used as the default whitelist. This validation is to be added on a model-level, and other filefields in Open Inwoner can be checked and if necessary the same validation can be added to those filefields as well."

As Users can upload documents (via a plan) this will need validation of some kind. And the other FileFields will likely also require attention.

pi-sigma (approved):

stevenbal:
@alextreme in the PR description I made a list of the file fields that I could find. 

- The `FilerFileField` actually already has validation that blocks you from uploading .svg, .html or other files that can allow XSS. 
- `document.file`, `action.file` and `message.file` all use `settings.UPLOAD_FILE_TYPES` in the frontend and the admins are readonly (added in this PR)
- `siteconfig.theme_stylesheet`: admin only and only allows .css
- `CaseUploadForm.files`: already checks the extensions based on the configured extensions in OpenZaakConfig

src/open_inwoner/accounts/admin.py
```diff
@@ -141,8 +161,39 @@ class _UserAdmin(ImageCroppingMixin, UserAdmin):
...
+        "type",
+        "end_date",
+        "display_file_url",
+        "is_for",
+        "created_on",
+        "updated_on",
+        "created_by",
+        "plan",
+        "is_deleted",
+    )
```
Bartvaderkin, line 196: It is not very DRY to repeat the same list? You could assign then same:

`readonly_fields = fields`

Also even more robust and less manual configuration to use `has_change_permission()` instead of explicitly listing everything. For example, if a new field gets added we need to remember it has to be added here (twice!) instead of being readonly by default. It is also more correct to make the model as a whole readonly instead of manually marking all fields.

This goes for all the changes that do similar.

Bartvaderkin (commented):
See note in the diff.

alextreme (approved):
The updated PR looks fine to me, retriggered the tests

PR #891, opened 2023-12-11
[#1928] Check missing volgnummer for second status preview

pi-sigma:
`InnerCaseDetailView.get_second_status_preview` occasionally failed due to a `TypeError` because case status types retrieved via the eSuite don't always contain a `volgnummer`.

I added a check to the function and updated the annotation of our `StatusType` api model

Taiga [#1928](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1928)

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -251,6 +251,10 @@ class InnerCaseDetailView(
         """
         statustype_numbers = [s.volgnummer for s in statustypen]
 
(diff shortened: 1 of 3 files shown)


stevenbal (approved):
Code looks good :+1: 

@alextreme not entirely sure if this solution is approved in the comments in taiga yet, so I'll wait with merging

alextreme:
Pushing this through in spite of concerns, it's not like we have an option here.

PR #892, opened 2023-12-11
[#1932] Skipping zaaktype in case identificatie doesn't exist in zaaktype list

alextreme:
https://taiga.maykinmedia.nl/project/open-inwoner/issue/1932


stevenbal:
Apparently if Zaaktypen are linked to the Category for which the user has no Zaken, this error occurred. Not sure how I never ran into this :sweat_smile: 

Updated the tests so this is tested as well

stevenbal (approved):

PR #893, opened 2023-12-12
[#1894] Updated django-filer dependency (and dropped mptt) to make dependabot happy

Bartvaderkin:


pi-sigma (approved):

alextreme:
Filer upgrades have been known to cause migration issues in the past, so :crossed_fingers:

PR #894, opened 2023-12-12
[#1894] Updated various networking related dependencies

Bartvaderkin:


alextreme (approved):
Approved. Pip does complain about python3-saml though, but this is unrelated.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
python3-saml 1.14.0 requires lxml<4.7.1, but you have lxml 4.9.1 which is incompatible.

PR #907, opened 2023-12-18
:recycle: [#1930] Category personalisation fixes

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1930


stevenbal:
@alextreme I noticed the following:
- To determine the categories shown in the menu, I only use the root nodes and then do the permission filter for the user
- To determine the categories shown in the homepage plugin, I take all nodes (including child nodes), then do permission filtering and then `uitgelicht`/zaaktypen filtering. This way subcategories can show up in the plugin on the homepage if uitgelicht, but not in the menu, is this what we want? Or do we want to show child nodes in the menu / list directly as well?

alextreme:
> @alextreme I noticed the following:
> 
>     * To determine the categories shown in the menu, I only use the root nodes and then do the permission filter for the user
> 
>     * To determine the categories shown in the homepage plugin, I take all nodes (including child nodes), then do permission filtering and then `uitgelicht`/zaaktypen filtering. This way subcategories can show up in the plugin on the homepage if uitgelicht, but not in the menu, is this what we want? Or do we want to show child nodes in the menu / list directly as well?

We don't want to show child nodes in the menu, the idea is that from the category detailpage you can click through and browse the subcategories (if there are any).

I would expect that we also need to update CategoryDetailView in order to make use of the  visible_for_user() manager for the subcategories. I don't see that in your PR or on develop, so double-check this just in case.

alextreme (changes requested):
Double-check CategoryDetailView

Bartvaderkin (approved):
Nice, and more test code then functional code :laughing:

PR #909, opened 2023-12-18
:sparkles: [#1910] Add first NL-DesignSystem components

jiromaykin:
- add our [designtokens-package submodule](https://github.com/maykinmedia/open-inwoner-design-tokens)
- add Utrecht CSS NPM package + Utrecht Components NPM package
- add actions for submodules to CI yml's
- add Utrecht classes in some heading templates
- let Utrecht design variable **Keys** point to OIP variable **Values**

issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1910

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "deployment/maykin-deployment"]
-	path = deployment/maykin-deployment
-	url = git@bitbucket.org:maykinmedia/maykin-deployment.git
-[submodule "design-tokens/open-inwoner-design-tokens"]
-	path = design-tokens/open-inwoner-design-tokens
+[submodule "open-inwoner-design-tokens"]
+	path = open-inwoner-design-tokens
 	url = git@github.com:maykinmedia/open-inwoner-design-tokens.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "deployment/maykin-deployment"]
 	path = deployment/maykin-deployment
 	url = git@bitbucket.org:maykinmedia/maykin-deployment.git
+[submodule "design-tokens/open-inwoner-design-tokens"]
+	path = design-tokens/open-inwoner-design-tokens
+	url = git@github.com:maykinmedia/open-inwoner-design-tokens.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,8 +46,8 @@ WORKDIR /app
 COPY ./build /app/build/
 COPY ./*.json ./*.js ./.babelrc /app/
 
-# Clone design token submodule with specific Tag version (cloning normally done using git submodule update --init)
-RUN git clone --depth 1 --branch 0.0.2 https://github.com/maykinmedia/open-inwoner-design-tokens.git
+# Clone design token submodule (normally done using git submodule update --init)
+RUN git clone https://github.com/maykinmedia/open-inwoner-design-tokens.git
 
 # install WITH dev tooling
 RUN npm ci --legacy-peer-deps
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,8 +46,8 @@ WORKDIR /app
 COPY ./build /app/build/
 COPY ./*.json ./*.js ./.babelrc /app/
 
-# Clone design token submodule (normally done using git submodule update --init)
-RUN git clone https://github.com/maykinmedia/open-inwoner-design-tokens.git
+# Clone design token submodule with specific Tag version (cloning normally done using git submodule update --init)
+RUN git clone --depth 1 --branch 0.0.2 https://github.com/maykinmedia/open-inwoner-design-tokens.git
 
 # install WITH dev tooling
 RUN npm ci --legacy-peer-deps
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,9 @@ WORKDIR /app
 COPY ./build /app/build/
 COPY ./*.json ./*.js ./.babelrc /app/
 
+# Clone design token submodule (normally done using git submodule update --init)
+RUN git clone https://github.com/maykinmedia/open-inwoner-design-tokens.git
+
 # install WITH dev tooling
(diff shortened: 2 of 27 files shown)


jiromaykin:
Note to self: De design-tokens git-submodule genereert lokaal CSS die nu niet gevonden wordt;  ci kan het bestand niet vinden omdat de git submodule waarschijnlijk niet meegebouwd wordt. 
ToDo om uit te zoeken: ervoor zorgen dat `@import 'design-tokens/open-inwoner-design-tokens/dist/css/index.css';` werkt door het package te symlinken, vanuit de root te NPM installeren, en er een lokale NPM module van te maken die vanuit de node-modules makkelijk door alle SCSS geimporteerd kan worden (zonder naar NPM te publiceren) en deze dan later toe te voegen bij `.github/workflows/ci.yml`
https://docs.npmjs.com/cli/v8/commands/npm-link

jiromaykin:
Before reviewing, please read why we are using 'utrecht' here:
https://taiga.maykinmedia.nl/project/open-inwoner/wiki/design-tokens-nl-design-system
Also this explains how in future we can add new properties and other CSS properly in the design-tokens.
In future we may even use classes with the name 'denhaag' in it, so don't be alarmed.

pi-sigma (changes requested):
Looking good! You should update the installation instruction for Open Inwoner itself (`INSTALL.rst` under "Get the code") to make sure the submodules are also cloned. Otherwise the folder for the design tokens is there but empty. There are two ways: 

1.
```
git clone --recursive git@github.com:maykinmedia/open-inwoner.git
cd open-inwoner
```

2. 
```
git clone git@github.com:maykinmedia/open-inwoner.git
cd open-inwoner

# initialize submodules
git submodule update --init --recursive
```
These are equivalent, the difference is a matter of style. I kind of prefer the second, as it's more explicit (it's not like you have to type this again and again). Perhaps include a comment below to explain what those submodules are.

While you're at it, you could remove the `$` signs from the code blocks in `INSTALL.rst` so that people can actually copy/paste the code snippets.

.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "deployment/maykin-deployment"]
-	path = deployment/maykin-deployment
-	url = git@bitbucket.org:maykinmedia/maykin-deployment.git
```
stevenbal, line 3: I'm not sure if we want to remove `maykin-deployment`?

Dockerfile
```diff
@@ -46,6 +46,9 @@ WORKDIR /app
 COPY ./build /app/build/
 COPY ./*.json ./*.js ./.babelrc /app/
 
+# Clone design token submodule (normally done using git submodule update --init)
+RUN git clone https://github.com/maykinmedia/open-inwoner-design-tokens.git
```
stevenbal, line 50: I'm not sure how often this repo will be updated, but I think it would be a good idea to pin this to a specific tag. Otherwise every pipeline of Open Inwoner will always pull in the `master` branch of the package's repo.

We could probably do something like this: https://stackoverflow.com/a/21699307 (also makes us skip any history, since we're not going to use that in the image anyway)

package.json
```diff
@@ -7,7 +7,10 @@
     "doc": "doc"
   },
   "scripts": {
-    "build": "npm run collect && npm run bundle",
+    "build": "npm run postinstall && npm run collect && npm run bundle",
+    "build-design-tokens": "npm --prefix open-inwoner-design-tokens run build",
+    "watch-design-tokens": "npm --prefix open-inwoner-design-tokens run start",
+    "postinstall": "npm --prefix ./open-inwoner-design-tokens install && npm --prefix ./open-inwoner-design-tokens run build",
```
stevenbal, line 13: Just curious: is there a difference between using `-- prefix ./open-inwoner-design-tokens` and `--prefix open-inwoner-design-tokens`?

stevenbal (changes requested):

stevenbal:
Looks good overall, just some remarks regarding docker setup/submodules

.gitmodules
jiromaykin: This was done as proposed by @alextreme :-)

package.json
jiromaykin: Ah, well spotted - yes there is a difference here: 

- **_with_** the dot (absolute path) it specifies an **absolute** path to the directory where NPM should look for the package's package.json file. In this case, it points to the current directory (./) and then to the open-inwoner-design-tokens directory. 
- **_without_** the dot (relative path) it assumes that the open-inwoner-design-tokens directory is present in the same directory where the NPM command is run, in our case in the root; 

I'll make this consistent.

Dockerfile
jiromaykin: I've just created a Tag (v0.0.1) but I wonder: if this will actually still be updated a lot, isn't it better to keep on pulling the Main branch - in these beginning stages? Every commit means a new Tag since they are big changes every time (I think).
In a far(?) future that repo could become the single source of truth (when they will also be used by the designers).

jiromaykin: I already turned it into this "`RUN git clone --depth 1 --branch 0.0.2 https://github.com/maykinmedia/open-inwoner-design-tokens.git`" with the tag, but how will we remember the Tag needs to be updated too? Add to both repo's documentations?

alextreme: It's indeed quite a chore to keep these kinds of relations in-sync between repositories, this is why for maykin-deployment I also avoid pinning to a specific branch (and why I'd prefer to go for a single combined repo, this would avoid the problem altogether).

For now I approve and prefer pulling from main/master instead of a release-branch. I'll set a reminder to re-evaluate this in half a year, if changes have become incidental then we'll pin this again to a release-branch

jiromaykin: @stevenbal @alextreme So should I remove the clone-from-Tag thing and revert to what it was?

alextreme: yes

pi-sigma (approved):

alextreme (approved):

PR #911, opened 2023-12-19
[#1938] Added theme stylesheet field to SiteConfiguration

Bartvaderkin:


src/open_inwoner/configurations/admin.py
```diff
@@ -136,6 +136,7 @@ class SiteConfigurarionAdmin(OrderedInlineModelAdminMixin, SingletonModelAdmin):
                     "primary_font_color",
                     "secondary_font_color",
                     "accent_font_color",
+                    "theme_stylesheet",
```
jiromaykin, line 139: Works really nice! I do wonder if it is wise to put this button in this section here? We decided to 'hide' the other CSS field much lower in the Admin config at 'Geavanceerde opties' - and I guess that's the one that overrides everything as a final last load? So this uploaded file is not the last one loaded? (which is fine actually)

Bartvaderkin: @jiromaykin Good point, I moved it in the admin to the same part as the extra CSS.

And you are correct this uploaded file is placed before the extra CSS field, seems most sensible but we can change it if necessary.

jiromaykin (approved):
Works well!

PR #912, opened 2023-12-20
[#1956] Return None from SSD client on RequestsException

pi-sigma:
Taiga [#1956](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1956)

### DIFF
--- a/src/open_inwoner/ssd/tests/test_client.py
+++ b/src/open_inwoner/ssd/tests/test_client.py
@@ -6,6 +6,7 @@ from django.test import TestCase
 
 import requests_mock
 from lxml import etree
+from requests.exceptions import ConnectionError
 
 from ..client import JaaropgaveClient, UitkeringClient
 from .factories import ConcreteSSDClient, SSDConfigFactory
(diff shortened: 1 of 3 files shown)


src/open_inwoner/ssd/tests/test_client.py
```diff
@@ -125,6 +126,22 @@ def test_tsl_client_cert_and_private_key(self, mock_request_body, mock_request):
...
+        ssd_client.config = SSDConfigFactory.build(
+            service__url="https://example.com/soap-service",
+        )
+
+        context = {"bsn": "dummy", "period": "dummy"}
+        response = ssd_client.templated_request(**context)
+
+        self.assertIsNone(response)
+
+        patch.stopall()
```
Bartvaderkin, line 143: The general pattern for mocking is to use `patch()` as a method decorator or context manager, so the stop/unpatch is always executed.

Example from the project:

    @patch("open_inwoner.pdc.models.mixins.geocode_address", side_effect=IndexError)
    def test_exception_is_handled_when_city_and_postcode_are_not_provided(
        self, mock_geocoding
    ):

Bartvaderkin (changes requested):

PR #913, opened 2023-12-21
[#1963] Sort documents in case detail view

pi-sigma:
Taiga [#1963](https://taiga.maykinmedia.nl/project/open-inwoner/task/1963)

### DIFF
--- a/src/open_inwoner/cms/cases/tests/test_htmx.py
+++ b/src/open_inwoner/cms/cases/tests/test_htmx.py
@@ -202,6 +202,7 @@ class CasesPlaywrightTests(
             url=f"{ZAKEN_ROOT}zaakinformatieobjecten/e55153aa-ad2c-4a07-ae75-15add57d6",
             informatieobject=f"{DOCUMENTEN_ROOT}enkelvoudiginformatieobjecten/014c38fe-b010-4412-881c-3000032fb812",
             zaak=self.zaak["url"],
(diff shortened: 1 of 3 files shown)


pi-sigma:
@alextreme I encountered failing tests which were missing the `registratiedatum` attribute that I use to sort the document list. This makes me wonder about the assumptions we are making w.r.t. information objects.

The schema for `ZaakInformatieObject` does not require a `registratiedatum`, but our corresponding dataclass in `openzaak/api_models.py` assumes that the requirement is satisfied. Similarly, the file creation in our case detail view sets the `created` attribute on the file based on the `registratiedatum` of the `ZaakInformatieObject`. Can we assume that instances of `ZaakInformatieObject` have a `registratiedatum`. If not, how do we want to handle the sorting?

alextreme (changes requested):
Consider fallback on title and ensure we have tests which show behaviour without registratiedatum

pi-sigma:
@alextreme A creation date is indeed automatically inserted for documents without registratiedatum. Comparing them however is not straightforward because one is timezone-aware the other naive. I went with the solution you suggested, fallback on document title/name and unordered list, if necessary.

alextreme:
@pi-sigma Good solution, however please add a regressiontest verifying the original issue that the newest document should be the first one, I didn't see this from your changes (but may have overlooked it)

alextreme (approved):

PR #917, opened 2023-12-29
[#1991] StatusTypeConfig configuration not refreshing 

alextreme:
__init__ is only called once upon app start, not per pageview

https://stackoverflow.com/questions/46599145/django-class-view-init

And we also need to filter the statustypeconfig/resulttypeconfigs on the zaaktypeconfig, as the eSuite has one statustype linked to multiple zaaktypes. Otherwise we most likely are using the statustypeconfig from a different zaaktype

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -98,7 +98,7 @@ class InnerCaseDetailView(
     contact_form_class = CaseContactForm
     case: Optional[Zaak] = None
 
-    def __init__(self):
+    def dispatch(self, request, *args, **kwargs):
         self.statustype_config_mapping = {
             zaaktype_statustype.statustype_url: zaaktype_statustype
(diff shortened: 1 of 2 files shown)


alextreme:
Mostly FYI

alextreme:
Failing tests: (open_inwoner.openzaak.tests.test_case_detail.TestCaseDetailView)
- test_pass_endstatus_type_data_if_endstatus_not_reached
- test_status_is_retrieved_when_user_logged_in_via_digid 

I suspect that the additional filtering of statustype/resultaattypeconfig on zaaktype.identificatie means that the mocked-responses need to be expanded to ensure the statustypeconfig's are properly picked up

alextreme:
Leaving this for now due to timeconstraints, either of you can pick this up to complete the PR by fixing the failing tests. I will cherry-pick these commits for in the next v1.12 release as these caused blocking issues

stevenbal:
I'll pick this up

stevenbal (approved):

PR #918, opened 2023-12-29
[#1972] Remove 'Scroll omlaag' button for previous statusses

alextreme:


jiromaykin (commented):
Approved :-) Don't know why this was there.

PR #919, opened 2023-12-29
[#1976] Ensuring notifications are no longer sticky

alextreme:


src/open_inwoner/scss/components/Notification/_Notifications.scss
```diff
@@ -1,5 +1,4 @@
 .notifications {
-  position: sticky;
```
jiromaykin, line 2: This does mean ALL notifications are no longer sticky, like the approval for newly added contact etc. Do we want this?

alextreme: I don't see a value in having the notifications sticky (I think the only reason we did this was to ensure notifications wouldn't interfere with the rest of the design). Consensus with Groningen is non-sticky and a 'wegklik kruisje' for the post-upload MijnAanvragen notification, however your call if you want to make this change overall or not

jiromaykin: This PR will need to either be hijacked by me or put on hold for the future as part of researching what the effect will be: https://taiga.maykinmedia.nl/project/open-inwoner/task/1976

alextreme: hijack away or decline, my ego can take it :+1:

jiromaykin (approved):

PR #922, opened 2024-01-02
[#1977] Showing calltoaction statusbutton on the detailview if config…

alextreme:
…ured (seems to be missing due to an incorrect merge)


stevenbal (approved):
@alextreme just to double check, is the call to action button supposed to be mutually exclusive with the `Scroll omlaag` button for uploads? If that is the case, I'll approve

alextreme:
> @alextreme just to double check, is the call to action button supposed to be mutually exclusive with the `Scroll omlaag` button for uploads? If that is the case, I'll approve

That is indeed the case: show the calltoaction button if configured, otherwise show the scroll-omlaag button if internal document uploads are configured

PR #923, opened 2024-01-02
[#1971] Double-check if our current status is an eindstatus due to eS…

alextreme:
…uite having multiple eindstatusses

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -172,6 +172,15 @@ class InnerCaseDetailView(
             # The end status data is not passed if the end status has been reached,
             # because in that case the end status data is already included in `statuses`
             end_statustype = next((s for s in statustypen if s.is_eindstatus), None)
+            # The following check is a eSuite-specific workaround to deal with multiple statustypes per zaaktype having isEindstatus: true
+            # In the case when we have reached a statustype with isEindstatus set we assume this is our eindstatus
+            if (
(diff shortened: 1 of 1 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -172,6 +172,15 @@ def get_context_data(self, **kwargs):
             # The end status data is not passed if the end status has been reached,
             # because in that case the end status data is already included in `statuses`
             end_statustype = next((s for s in statustypen if s.is_eindstatus), None)
```
stevenbal, line 174: Wouldn't this line become irrelevant after the changes you made? If I understand it correctly, the conditional below would always be triggered and override `end_statustype` to be the statustype of the last status (if `is_eindstatus` is `True`) right?

alextreme: I indeed went for the most minimal change for my hotfix for v1.12. I suspect you're correct and have updated this branch to remove this line and avoid future confusion

stevenbal (approved):

src/open_inwoner/cms/cases/views/status.py
stevenbal: Hmm okay I may have been wrong, I glanced over that fact that in this line the `statustypen` are used directly, so that way we will still get the `end_statustype` even if this status hasn't been reached yet, so I think we still need this line :sweat_smile:

PR #928, opened 2024-01-04
:sparkles: [#1949] Mijn aanvragen filtering by vestigingsnummer

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1949


alextreme (changes requested):
Looks good, however could we move `request.session.get("KVK_BRANCH_NUMBER")` to a utility function? We're going to need this session-value more often so it feels worthwhile to avoid duplicating this constant

alextreme (approved):

PR #933, opened 2024-01-05
[#1760] Implemented user feed

Bartvaderkin:
Review notes:

It's a lot of files, but to get an idea of how it works first look at:

- `open_inwoner/userfeed/feed.py` for the `get_feed(user)`  that is the main entry for retrieval and display (in the plugin etc)
- `open_inwoner/userfeed/hooks/*.py` for the 'hook' functions that create items of a specific type, the function to mark some as completed, and their optional 'adapter' class that generalizes the type for display by `get_feed()` (and the template etc)

Everything else is just glue to make it work or to tie it into the rest of the application. Note the liberal use of patch/mock in the tests to keep the user feed specific tests out of the other parts of the code.

In the future we'll add more item types to the choices/hooks/adapters.

---

To test the plugin put it on the homepage, and then run this command to add a test message:

```
python src/manage.py add_feed_message --user 1 --title "Helo World" --message "Waddup?" --url "http://www.maykinmedia.nl"
```
These never expire, so delete from `/admin/userfeed/feeditemdata/`. To find valid `--user` User id's use the command without arguments and it prints staff users:
```
python src/manage.py add_feed_message
```


src/open_inwoner/cms/plugins/models/userfeed.py
```diff
@@ -0,0 +1,6 @@
+from cms.models import CMSPlugin
+
+
+class UserFeed(CMSPlugin):
+    # TODO add options
```
stevenbal, line 5: Is this something that will be done in later PRs?

src/open_inwoner/userfeed/adapters.py
```diff
@@ -0,0 +1,40 @@
...
+
+def get_item_adapter_class(type: str) -> type[FeedItem]:
+    return feed_adapter_map.get(type, feed_adapter_default)
+
+
+feed_adapter_default = FeedItem
+feed_adapter_map = dict()
+
+
+def register_item_adapter(adapter_class: type[FeedItem], feed_type: str):
```
stevenbal, line 12: Probably a bit nitpicky, but would we want to enforce that `feed_type` is one of the options in the `FeedItemType` choices?

stevenbal (changes requested):
Looks good overall :+1:, just a few minor remarks/questions

src/open_inwoner/userfeed/apps.py
```diff
@@ -0,0 +1,29 @@
...
+
+    this expects things calling their register_xyz() function at module level
+    """
+
+    hooks_dir = Path(__file__).parent / "hooks"
+    for f in os.listdir(hooks_dir):
+        name, ext = os.path.splitext(f)
+        if ext == ".py" and name != "__init__":
+            path = f"open_inwoner.userfeed.hooks.{name}"
+            import_module(path)
```
pi-sigma, line 29: If you follow the pattern suggested earlier, you could simply `import_module("open_inwoner.userfeed.hooks")`

src/open_inwoner/userfeed/feed.py
```diff
@@ -0,0 +1,82 @@
...
+
+ACTION_COMPLETED_HISTORY_RANGE = timedelta(minutes=10)
+
+
+@dataclasses.dataclass()
+class Feed:
+    items: list[FeedItem] = dataclasses.field(default_factory=list)
+    summary: list[str] = dataclasses.field(default_factory=list)
+
+    def has_display(self) -> int:
```
pi-sigma, line 27: Return value is `bool`

src/open_inwoner/userfeed/feed.py
```diff
@@ -0,0 +1,82 @@
...
+        return self.total_items > 0
+
+    def action_required(self) -> int:
```
pi-sigma, line 30: See previous

src/open_inwoner/userfeed/management/commands/add_feed_message.py
```diff
@@ -0,0 +1,51 @@
...
+            "--url",
+            help="Action URL",
+            default="",
+        )
+
+    def handle(self, *args, **options):
+        try:
+            user = User.objects.get(pk=options["user"])
+        except User.DoesNotExist:
+            self.stdout.write("user_id not found, use one off:")
```
pi-sigma, line 36: Typo: "off" > "of"

src/open_inwoner/userfeed/models.py
```diff
@@ -0,0 +1,172 @@
...
+from open_inwoner.accounts.models import User
+from open_inwoner.userfeed.choices import FeedItemType
+
+
+class FeedItemManager(models.Manager):
+    def mark_uuid_completed(
+        self,
+        user: User,
+        type: str,
+        ref_uuid,
```
pi-sigma, line 21: type annotation

src/open_inwoner/userfeed/models.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        if ref_key is not None:
+            qs = qs.filter(ref_key=ref_key)
+        return qs.mark_completed(force=force)
+
+    def mark_object_completed(
+        self,
+        user: User,
+        type: str,
+        ref_object,
```
pi-sigma, line 38: type annotation

src/open_inwoner/userfeed/summarize.py
```diff
@@ -0,0 +1,61 @@
...
+                "In {count} case the status has changed",
+                "In {count} cases the status has changed",
+                num_items,
+            ),
+            num_items,
+        )
+
+
+def summarize_action_required(items: Iterable[FeedItem]) -> Optional[SummaryLine]:
+    num_items = sum(1 for item in items if item.action_required)
```
pi-sigma, line 32: `num_items = len([item for item in items if item.action_required])` is faster than performing `sum`. It requires space for the list, but that shouldn't be an issue.

src/open_inwoner/userfeed/summarize.py
```diff
@@ -0,0 +1,61 @@
...
+                "In {count} case your action is required",
+                "In {count} cases your action is required",
+                num_items,
+            ),
+            num_items,
+        )
+
+
+def summarize_simple_message(items: Iterable[FeedItem]) -> Optional[SummaryLine]:
+    num_items = sum(1 for item in items if item.type == FeedItemType.message_simple)
```
pi-sigma, line 45: See previous

src/open_inwoner/userfeed/tests/test_feed.py
```diff
@@ -0,0 +1,64 @@
...
+from open_inwoner.userfeed.models import FeedItemData
+from open_inwoner.userfeed.tests.factories import FeedItemDataFactory
+
+
+class FeedTests(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        cls.user = UserFactory()
+
+    def test_get_feed__and__simple_message(self):
```
pi-sigma, line 21: double underscore deliberate?

pi-sigma (changes requested):

src/open_inwoner/userfeed/adapters.py
Bartvaderkin: Yes we would, but I'll have to look into how we'd do that (and if python 3.9(

src/open_inwoner/userfeed/summarize.py
Bartvaderkin: I don't think speed is an argument if you're fine allocating throwaway lists.

 This is more descriptive so I'm keeping it.

src/open_inwoner/userfeed/tests/test_feed.py
Bartvaderkin: Yess

src/open_inwoner/userfeed/models.py
```diff
@@ -0,0 +1,172 @@
...
+    ref_key = models.CharField(
+        _("External object de-duplication"), blank=True, max_length=64
+    )
+    ref_object_field = GenericForeignKey(
+        ct_field="ref_object_type", fk_field="ref_object_id"
+    )
+    ref_object_type = models.ForeignKey(
+        ContentType, on_delete=models.CASCADE, blank=True, null=True
+    )
+    ref_object_id = models.PositiveIntegerField(blank=True, null=True)
```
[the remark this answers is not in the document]
Bartvaderkin: `ref_key` is for extra filtering without having to query into the JSON. (like for the zaakinfoobject the zaak has the uuid but we need to deduplicate the zaakinfoobject) 

`ref_object_field` is a regular django GenericForeignKey (together with `ref_object_type`) so we can find the actual object in the database (in like the Plan object). I called it `ref_object_field` so the `ref_object` attrribute can be a property that does some convenience

src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -0,0 +1,46 @@
+{% load i18n icon_tags %}
+
+
+<section class="userfeed">
```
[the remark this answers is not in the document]
Bartvaderkin: Good point, I'll fix that.

src/open_inwoner/cms/plugins/models/userfeed.py
Bartvaderkin: Yea, they mentioned they want to enable/disable and configure stuff so I added the options but then it was not part of this ticket.

src/open_inwoner/userfeed/adapters.py
Bartvaderkin: Ugh, apparently we can just use the type of the choices class instead of `str`.

src/open_inwoner/plans/management/commands/plans_expire.py
```diff
@@ -48,6 +49,10 @@ def handle(self, *args, **options):
                 f"The email was sent to the user {receiver} about {plans.count()} expiring plans"
             )
 
+            # hook into userfeed
+            for p in plans:
+                plan_expiring(receiver, p)
```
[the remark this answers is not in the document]
Bartvaderkin: OK this is also how some of the timelinelogs works so let's do that.

src/open_inwoner/plans/tests/test_plans_expire.py
```diff
@@ -55,7 +56,8 @@ def test_notify_about_expiring_plan_inactive_user(self):
         call_command("plans_expire")
         self.assertEqual(len(mail.outbox), 0)
 
-    def test_notify_about_expiring_plan(self):
+    @patch("open_inwoner.plans.management.commands.plans_expire.plan_expiring")
+    def test_notify_about_expiring_plan(self, mock_plan_expiring: Mock):
```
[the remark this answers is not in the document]
Bartvaderkin: Weird didn't notice the warning, I renamed and moved stuff so it makes more sense.

Bartvaderkin:
I applied the PR fixes.

Also: there was a mistake in the logic where the incoming-notification handler would check and bail early for emailable Users and would never reach the userfeed hook if the users had a bad email or email-notifications disabled.

stevenbal (approved):

pi-sigma (approved):

PR #935, opened 2024-01-05
:recycle: [#1999] Use fixed unique ID fieldname for OIDC

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1931

to allow the ID claim name to differ from the fieldname on our model

### DIFF
--- a/src/digid_eherkenning_oidc_generics/backends.py
+++ b/src/digid_eherkenning_oidc_generics/backends.py
@@ -17,6 +17,7 @@ logger = logging.getLogger(__name__)
 class OIDCAuthenticationBackend(_OIDCAuthenticationBackend):
     config_identifier_field = "identifier_claim_name"
     callback_path = None
+    unique_id_user_fieldname = ""
 
     def authenticate(self, request, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


src/digid_eherkenning_oidc_generics/backends.py
```diff
@@ -68,6 +67,7 @@ class OIDCAuthenticationDigiDBackend(SoloConfigDigiDMixin, OIDCAuthenticationBac
 
     login_type = LoginTypeChoices.digid
     callback_path = reverse_lazy("digid_oidc:callback")
+    unique_id_user_fieldname = "bsn"
```
stevenbal, line 70: @alextreme I figured since these names aren't gonna change, I'd keep them fixed and allow the `identifier_claim_name` to be configurable, that way `identifier_claim_name` (e.g. `sub`) will always be mapped to `unique_id_user_fieldname` (`bsn`)

alextreme (approved):

src/digid_eherkenning_oidc_generics/backends.py
alextreme: Sure makes sense

PR #936, opened 2024-01-08
🐛 [#1975] Keep previous selection when selecting new files to upload

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1975

### DIFF
--- a/src/open_inwoner/js/components/form/FileInput.js
+++ b/src/open_inwoner/js/components/form/FileInput.js
@@ -163,6 +163,7 @@ export class FileInput extends Component {
     const files = [...input.files].filter((_, i) => i !== index)
 
     this.addFiles(files, true)
+    this.files = files
 
     // We need to render manually since we're not making state changes.
     this.render()
@@ -197,6 +198,7 @@ export class FileInput extends Component {
 
     _files.filter((v) => v).forEach((file) => dataTransfer.items.add(file))
     input.files = dataTransfer.files
+    this.files = _files
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -179,26 +180,44 @@ export class FileInput extends Component {
...
+    let { files } = this.getInput()
+
+    // Bugfix for https://taiga.maykinmedia.nl/project/open-inwoner/task/1975
+    // Selecting files using the select window triggers a change event and would normally
+    // overwrite the previously selected files, so the previously selected files need to
+    // be stored and joined with the newly uploaded files, to avoid the selection from being overwritten
+    if (e && e.type === 'change') {
+      files = [...(this.files || []), ...files]
+      this.addFiles(files, true)
+    }
```
stevenbal, line 219: Couldn't really find another way to fix this bug for selecting files using the button/window, because it seems to be default behavior to immediately override the input's files, so the previously selected files need to be stored somewhere.

jiromaykin: Yes this is an incredibly difficult aspect of the default behaviour for the HTML file-input type 😢

jiromaykin (approved):
It works! Just a thought: instead of adding this to "issue" branch/dir, it could also have been added to our "fix" branch in this project?

stevenbal:
> Just a thought: instead of adding this to "issue" branch/dir, it could also have been added to our "fix" branch in this project?
Do you mean using `fix/1975-multiple-doc-upload` as the branch name, instead of `issue/1975-multiple-doc-upload`? I've always used `feature/...` or `issue/...` (out of habit), though if we want to use `fix` instead of `issue` for future branches, that's fine by me

alextreme (approved):

PR #942, opened 2024-01-09
[#2004] Fix order of company branches

pi-sigma:
Make sure that the main company branch ("hoofdvestiging") comes before any subsidiary ("nevenvestiging")

Taiga: [#2004](https://taiga.maykinmedia.nl/project/open-inwoner/task/2004)

### DIFF
--- a/src/open_inwoner/accounts/tests/test_auth.py
+++ b/src/open_inwoner/accounts/tests/test_auth.py
@@ -1036,8 +1036,16 @@ class DuplicateEmailRegistrationTest(WebTest):
         """Assert that eHerkenning users can register with duplicate emails"""
 
         mock_kvk.return_value = [
-            {"kvkNummer": "12345678", "vestigingsnummer": "1234"},
-            {"kvkNummer": "12345678", "vestigingsnummer": "5678"},
+            {
+                "kvkNummer": "12345678",
+                "vestigingsnummer": "1234",
+                "handelsnaam": "Mijn bedrijf",
+            },
+            {
+                "kvkNummer": "12345678",
+                "vestigingsnummer": "5678",
+                "handelsnaam": "Mijn bedrijf",
+            },
         ]
 
         test_user = eHerkenningUserFactory.create(
--- a/src/open_inwoner/kvk/client.py
+++ b/src/open_inwoner/kvk/client.py
@@ -105,10 +105,23 @@ class KvKClient:
 
     def get_all_company_branches(self, kvk: str, **kwargs) -> list[Optional[dict]]:
         """
-        Get data about all branches ("hoofdvestiging" + "nevenvestigingen") of a company
+        Get data about all branches ("hoofdvestiging" + "nevenvestigingen") of a company.
+
+        Filter response from KvK API (remove elements which are not specific branches) and
+        sort the results (move the main branch ("hoofdvestiging") to the front of the list)
         """
         kwargs.update({"kvkNummer": kvk})
-        return self.search(**kwargs).get("resultaten", [])
+        branches = self.search(**kwargs).get("resultaten", [])
+
+        for branch in branches.copy():
+            if not (
+                branch["type"] == "hoofdvestiging" or branch["type"] == "nevenvestiging"
+            ):
+                branches.remove(branch)
+            elif branch["type"] == "hoofdvestiging":
+                branches.insert(0, branches.pop(branches.index(branch)))
+
+        return branches
 
     def retrieve_rsin_with_kvk(self, kvk, **kwargs) -> Optional[str]:
         basisprofiel = self._request(
--- a/src/open_inwoner/kvk/tests/test_api.py
+++ b/src/open_inwoner/kvk/tests/test_api.py
@@ -39,7 +39,7 @@ class KvKAPITest(TestCase):
     def test_generic_search_by_kvk(self):
         self.mocked_requests.return_value.json.return_value = mocks.simple
 
-        company = self.kvk_client.search(kvk="69599084")
+        company = self.kvk_client.search(kvk="68750110")
 
         self.assertEqual(
             company,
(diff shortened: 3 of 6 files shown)


stevenbal (changes requested):
The code looks good :+1: 

I'm not entirely sure if we're doing the right thing here though. I'm testing locally with the test KVK API and now it looks like this for kvk number `68750110`

image

My assumption was that if you selected the `hoofdvestiging`, you would be able to see all Aanvragen / vragen for all vestigingen. It seems however that the `hoofdvestiging` does have a `vestigingnummer` (different from the KVK number), which would mean it would only show the data related to that `vestigingsnummer` .

The second choice in my screenshot was what I initially thought was the one you'd select if you want to see all data for all branches and that one happens to be of `type=rechtspersoon`:

```json
"resultaten": [
    {
        "kvkNummer": "68750110",
        "handelsnaam": "Test BV Donald",
        "type": "rechtspersoon",
        "links": [
            {
                "rel": "basisprofiel",
                "href": "https://api.kvk.nl/test/api/v1/basisprofielen/68750110"
            }
        ]
    },
    {
        "kvkNummer": "68750110",
        "vestigingsnummer": "000037178598",
        "handelsnaam": "Test BV Donald",
        "adresType": "bezoekadres",
        "straatnaam": "Hizzaarderlaan",
        "plaats": "Lollum",
        "type": "hoofdvestiging",
        "links": [
            {
                "rel": "basisprofiel",
                "href": "https://api.kvk.nl/test/api/v1/basisprofielen/68750110"
            },
            {
                "rel": "vestigingsprofiel",
                "href": "https://api.kvk.nl/test/api/v1/vestigingsprofielen/000037178598"
            }
        ]
    },
    {
        "kvkNummer": "68750110",
        "vestigingsnummer": "000037178601",
        "handelsnaam": "Test BV Donald Nevenvestiging",
        "adresType": "bezoekadres",
        "straatnaam": "Brinkerinckbaan",
        "plaats": "Diepenveen",
        "type": "nevenvestiging",
        "links": [
            {
                "rel": "basisprofiel",
                "href": "https://api.kvk.nl/test/api/v1/basisprofielen/68750110"
            },
            {
                "rel": "vestigingsprofiel",
                "href": "https://api.kvk.nl/test/api/v1/vestigingsprofielen/000037178601"
            }
        ]
    }
]
```

I think we should probably do the following:
1. regardless of whether we see a `hoofdvestiging` in the KvK data, the first choice should be the one that allows you to see all data (simply the KVK number)
2. the second choice should be the `hoofdvestiging`
3. the rest should be the remaining `nevenvestiging`en (if they exist) and we should not show any other types like `rechtspersoon`

@alextreme thoughts?

alextreme:
>     1. regardless of whether we see a `hoofdvestiging` in the KvK data, the first choice should be the one that allows you to see all data (simply the KVK number)
> 
>     2. the second choice should be the `hoofdvestiging`
> 
>     3. the rest should be the remaining `nevenvestiging`en (if they exist) and we should not show any other types like `rechtspersoon`
> 
> 
> @alextreme thoughts?

That seems like a better approach yes. Sorry, I mixed up the fact that the 'KvK number without branch selection' (so for acting as the company as a whole) is something different than selecting the hoofdvestiging, which is a branch

pi-sigma:
@alextreme @stevenbal What about the case where the response only contains branches with a `vestigingsnummer`? For example

```yaml
[
  {
    "kvkNummer": "69599084",
    "vestigingsnummer": "000038509504",
    "handelsnaam": "Test EMZ Dagobert",
    "adresType": "bezoekadres",
    "straatnaam": "Abebe Bikilalaan",
    "plaats": "Amsterdam",
    "type": "hoofdvestiging",
    "links": [
      {
        "rel": "basisprofiel",
        "href": "https://api.kvk.nl/test/api/v1/basisprofielen/69599084"
      },
      {
        "rel": "vestigingsprofiel",
        "href": "https://api.kvk.nl/test/api/v1/vestigingsprofielen/000038509504"
      }
    ]
  },
  {
    "kvkNummer": "69599084",
    "vestigingsnummer": "000038509520",
    "handelsnaam": "Test EMZ Nevenvestiging Govert",
    "adresType": "bezoekadres",
    "straatnaam": "Geneinde",
    "plaats": "Maastricht",
    "type": "nevenvestiging",
    "links": [
      {
        "rel": "basisprofiel",
        "href": "https://api.kvk.nl/test/api/v1/basisprofielen/69599084"
      },
      {
        "rel": "vestigingsprofiel",
        "href": "https://api.kvk.nl/test/api/v1/vestigingsprofielen/000038509520"
      }
    ]
  }
]
```
Given our current setup, the user won't be able to log in and retrieve information about the entire company. Should we check for this case and create an additional option (i.e. without vestiging restriction), or display only the options returned by the KvK API?

stevenbal:
@pi-sigma I would do it like this:
- modify `get_all_company_branches` to only return entries with type `hoofdvestiging` or `nevenvestiging`
- use those branches in the form used for branch selection and always add an option that simply uses the KVK number (so no branch number) as the first option. This way we don't have to check the options returned by the KVK API and we're sure that the entire company can always be selected

Bartvaderkin (approved):

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -1036,8 +1036,16 @@ def test_eherkenning_user_success(self, mock_kvk):
         """Assert that eHerkenning users can register with duplicate emails"""
 
         mock_kvk.return_value = [
-            {"kvkNummer": "12345678", "vestigingsnummer": "1234"},
-            {"kvkNummer": "12345678", "vestigingsnummer": "5678"},
+            {
+                "kvkNummer": "12345678",
+                "vestigingsnummer": "1234",
+                "hanelsnaam": "Mijn bedrijf",
```
stevenbal, line 1042:
```suggestion
                "handelsnaam": "Mijn bedrijf",
```

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -1036,8 +1036,16 @@ def test_eherkenning_user_success(self, mock_kvk):
...
+            },
+            {
+                "kvkNummer": "12345678",
+                "vestigingsnummer": "5678",
+                "hanelsnaam": "Mijn bedrijf",
```
stevenbal, line 1047:
```suggestion
                "handelsnaam": "Mijn bedrijf",
```

src/open_inwoner/kvk/client.py
```diff
@@ -101,10 +101,23 @@ def get_company_headquarters(self, kvk: str, **kwargs) -> dict:
...
+        Get data about all branches ("hoofdvestiging" + "nevenvestigingen") of a company.
+
+        Filter response from KvK API (remove elements which are not specific branches) and
+        sort the results (move the main branch ("hoofdvestiging") to the front of the list)
         """
         kwargs.update({"kvkNummer": kvk})
-        return self.search(**kwargs).get("resultaten", [])
+        branches = self.search(**kwargs).get("resultaten", [])
+
+        for branch in branches:
```
stevenbal, line 112: I'm generally wary of modifying lists that are being looped over, since this can lead to unexpected behaviour. For example, code like this runs infinitely because the list is modified during the loop:

```python
my_list = [1, 2, 3, 4, 5]

for item in my_list:
    print(item)
    if item == 2:
        my_list.insert(0, 6)
```

When testing with the test KvK API locally I didn't run into errors, but let's copy the branches list we loop over just to be sure.

```suggestion
        for branch in branches.copy():
```

src/open_inwoner/kvk/tests/test_api.py
```diff
@@ -36,126 +36,82 @@ def test_search_endpoint(self):
 
         self.assertEqual(endpoint1, endpoint2)
 
-    def test_generic_search_by_kvk(self):
```
stevenbal, line 39: Why is this removed?

src/open_inwoner/kvk/views.py
```diff
@@ -64,6 +64,13 @@ def get(self, request):
             request.session.save()
             return HttpResponseRedirect(redirect.url)
 
+        # create pseudo-branch representing the company as a whole
+        master_branch = {
+            "kvkNummer": company_branches[0]["kvkNummer"],
```
stevenbal, line 69: Nitpicking, but I would use `request.user.kvk` here, in case the KVK nummer in the API in missing for some reason, but you can leave it as is

stevenbal (changes requested):

src/open_inwoner/kvk/tests/test_api.py
pi-sigma: I removed it because it doesn't actually test anything: request calls are mocked, the bare search function contains no logic, hence the result is trivially equivalent to the mock. However, perhaps it's useful to keep even such a trivial test, in case someone further down the road tempers with the search method.

src/open_inwoner/kvk/client.py
```diff
@@ -105,19 +101,37 @@ def get_company_headquarters(self, kvk: str, **kwargs) -> dict:
...
+                branch["type"] == "hoofdvestiging" or branch["type"] == "nevenvestiging"
+            ):
+                branches.remove(branch)
+            elif branch["type"] == "hoofdvestiging":
+                branches.insert(0, branches.pop(branches.index(branch)))
+
+        return branches
 
     def retrieve_rsin_with_kvk(self, kvk, **kwargs) -> Optional[str]:
-        basisprofiel = self._request(
```
stevenbal, line 114: I think something went wrong while rebasing? This is the fix implemented in https://github.com/maykinmedia/open-inwoner/pull/952

stevenbal (commented):
Looks good now, the only remaining thing is the incorrect rebase (which also causes the failing tests)

stevenbal (approved):

PR #945, opened 2024-01-10
[#1859] Replace e-suite subject with OIP subject in contactmomenten

pi-sigma:
Taiga: [#1859](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1859)


stevenbal (changes requested):

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -77,10 +77,20 @@ def get_kcm_data(self, kcm: KlantContactMoment) -> KCMDict:
...
+
+        # replace e_suite_subject_code with OIP configured subject, if applicable
+        e_suite_subject_code = kcm.contactmoment.onderwerp
+
+        try:
+            subject = ContactFormSubject.objects.get(subject_code=e_suite_subject_code)
+        except ContactFormSubject.DoesNotExist:
+            data["onderwerp"] = ""
+        else:
+            data["onderwerp"] = subject.subject
```
stevenbal, line 92: This will raise `MultipleObjectsReturned` in case there are multiple subjects with the same code

```suggestion
        contactformsubject = ContactFormSubject.objects.filter(subject_code=e_suite_subject_code).first()
        data["onderwerp"] = getattr(contactformsubject, "subject", "")
```

stevenbal: I ran into this locally, because Open Klant does not have `onderwerp` in the `Contactmomenten` model, which caused the `e_suite_subject_code` to be an empty string.

On our test environment the ContactFormSubjects are configured like this, which caused a `MultipleObjectsReturned`:

image

@alextreme this does have to be configured properly for this to be of use, otherwise every contactmoment will have `Over Open Inwoner` as subject

stevenbal: I also think that if `kcm.contactmoment.onderwerp` is empty, `data["onderwerp"]` should probably just be empty as well, since Open Klant doesn't implement this `onderwerp` field

stevenbal: 1. I think catching the `MultipleObjectsReturned` and setting the value to `None` is good. I do also think it'd be good if we explicitly check if `kcm.contactmoment.onderwerp` is empty and set the value to `None` in that case as well, otherwise if there is a single `ContactFormSubject` that maps `""` to something, it would always do that for Open Klant.
2. Adding logging is a good idea, this also makes it more relevant to add the check I mentioned before, because otherwise this log would be spammed all the time in case Open Klant is used

alextreme: Not showing the Onderwerp if the onderwerp-value is empty indeed seems like the way to go

stevenbal (approved):

PR #949, opened 2024-01-11
:bug: [#2003] Fix Klant creation

stevenbal:
TODO:
- [ ] use proper `subject_identificatie` for the three types
- [ ] ...


alextreme:
We'll keep this as draft for the time being. OpenKlant v2 is being rounded up, and we'll have to support eSuite Klanten API for the time being, so improving things for OpenKlant v0.1 isn't necessary at the moment

stevenbal:
@alextreme we can probably close this PR, since OpenKlant v1 isn't used and @pi-sigma fixed klant creation for eSuite

PR #957, opened 2024-01-15
[#1911] Invalid file removal to any order

jiromaykin:
When trying to remove invalid files and there are multiple, deletion is only possible from the top, 
not from any arbitrary index = needs to be possible anywhere.
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1911


jiromaykin:
This issue was solved by #978

PR #958, opened 2024-01-15
Fix/2018 zgw performance

stevenbal:


stevenbal:
replaced by https://github.com/maykinmedia/open-inwoner/pull/1011

PR #959, opened 2024-01-15
[#1979] Change date captions for cases list as design

jiromaykin:
issue https://taiga.maykinmedia.nl/project/open-inwoner/task/1979 
Dates in case list were not put in the positions they needed to be, so now put conforming to design (closely aligned to the Case title) 
And: also improved some of the Card paddings according to design + changed the name of the gray variant to a "contextual" name that is more commonly used by designers and with design-tokens ('900' instead of '90').

Link to design: https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=1%3A273&mode=design&t=PfzsswiRMAp5biaL-1

### DIFF
--- a/src/open_inwoner/templates/pages/cases/list_inner.html
+++ b/src/open_inwoner/templates/pages/cases/list_inner.html
@@ -16,38 +16,42 @@
             <div class="card__body">
                 <!-- submission cases -->
                 {% if case.case_type == "OpenSubmission" %}
-                    <a href="{{ case.vervolg_link }}"  class="cases__link">
-                    <p class="h4"><span class="link link__text">{{ case.naam }}</span></p>
+                    <a href="{{ case.vervolg_link }}" class="cases__link">
+                    <h2 class="card__heading">{{ case.naam }}</h2>
                     {% render_list %}
                     <span class="case-list">
+                        <li class="list-item list-item--compact">
+                            {% with end_date=case.eind_datum_geldigheid|default:"Geen" %}
+                            <p class="card__caption card__text--small"><span>{% trans "Aanvraag verloopt op:" %}</span><span class="card__text--dark">{{ end_date|date }}</span></p>
+                            {% endwith %}
+                        </li>
                         {% list_item text="Openstaande aanvraag" caption=_("Status") compact=True strong=False %}
-                        {% list_item text=case.datum_laatste_wijziging caption=_("Laatste wijziging") compact=True strong=False %}
-                        {% with end_date=case.eind_datum_geldigheid|default:"Geen" %}
-                            {% list_item text=end_date caption=_("Verloopt op") compact=True strong=False%}
-                        {% endwith %}
                     </span>
                     {% endrender_list %}
-                    <span class="link link--icon link--secondary" aria-label="{% trans "Bekijk aanvraag" %}" title="{% trans "Ga naar aanvraag" %}">
-                        <span class="link__text">{% trans "Bekijk aanvraag" %}</span>
+                    <span class="link link--icon link--secondary">
+                        <span class="link__text">{% trans "Aanvraag afronden" %}</span>
                         {% icon icon="arrow_forward" icon_position="after" primary=True outlined=True %}
                     </span>
                     </a>
                 <!-- other cases -->
                 {% else %}
                     <a href="{% url 'cases:case_detail' object_id=case.uuid %}" class="cases__link">
-                    <p class="h4"><span class="link link__text">{{ case.description }}</span></p>
+                    <h2 class="card__heading">{{ case.description }}</h2>
                     {% render_list %}
                         <span class="case-list">
+                        <li class="list-item list-item--compact">
+                            <p class="card__caption card__text--small"><span>{% trans "Aanvraag ingediend op:" %}</span><span class="card__text--dark">{{ case.start_date }}</span></p>
+                        </li>
                         {% if case.current_status %}
                             {% list_item case.current_status caption=_("Status") compact=True strong=False %}
                         {% else %}
                             {% list_item "No status" caption=_("Status") compact=True strong=False %}
                         {% endif%}
-                        {% list_item case.start_date caption=_("Ontvangstdatum") compact=True strong=False %}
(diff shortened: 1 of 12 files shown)


src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -423,7 +423,7 @@ def test_cases(self, m):
         page.goto(self.live_reverse("cases:index"))
 
         # case title
-        case_title = page.get_by_role("link", name=self.zaaktype["omschrijving"])
+        case_title = page.get_by_text("Coffee zaaktype")
```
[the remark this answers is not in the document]
pi-sigma: This is not right, given that `self.zaaktype["omschrijving"]` = `"Coffee zaaktype"`.

The problem is `aria-label="{% trans "Bekijk aanvraag" %}"` in the link of the case. The accessible name for the link is given by the content of the `h4` element, hence Playwright gets confused and thinks the `name` for the `link` role is "Bekijk aanvrag", the content of `aria-label`. If you replace `aria-label` with `aria-labelledby` and link it to the `h4`, the test works. Two more comments:
- Strictly speaking, the content of the link is much more than "Coffee zaaktype"; it contains the information of all the list elements inside `<a>....</a>`. I don't know if it's still appropriate to use `aria-labelledby` in this case, linking it to only part of the content, though my intuition is that it's OK.
- If you change the html as suggested, you should also change it for the open submission, which has the same structure. We don't have a separate test for those, but the situation is the same and should be handled accordingly.

src/open_inwoner/templates/pages/cases/list_inner.html
```diff
@@ -16,38 +16,41 @@ <h1 class="utrecht-heading-1" id="cases">{{ page_title }} ({{ paginator.count }}
             <div class="card__body">
                 <!-- submission cases -->
                 {% if case.case_type == "OpenSubmission" %}
-                    <a href="{{ case.vervolg_link }}"  class="cases__link">
-                    <p class="h4"><span class="link link__text">{{ case.naam }}</span></p>
+                    <a href="{{ case.vervolg_link }}" class="cases__link" aria-label="{% trans "Bekijk aanvraag" %}" title="{% trans "Bekijk aanvraag" %}">
```
pi-sigma, line 19: Since you changed the `span` element below to "Voltooi aanvraag" (which is a nice way to capture the distinction between cases and open submissions), you should change this too. Also, see previous comment about `aria-label` and `aria-labelledby`.

src/open_inwoner/templates/pages/cases/list_inner.html
```diff
@@ -16,38 +16,41 @@ <h1 class="utrecht-heading-1" id="cases">{{ page_title }} ({{ paginator.count }}
...
+                    <span class="link link--icon link--secondary">
+                        <span class="link__text">{% trans "Voltooi aanvraag" %}</span>
                         {% icon icon="arrow_forward" icon_position="after" primary=True outlined=True %}
                     </span>
                     </a>
                 <!-- other cases -->
                 {% else %}
-                    <a href="{% url 'cases:case_detail' object_id=case.uuid %}" class="cases__link">
-                    <p class="h4"><span class="link link__text">{{ case.description }}</span></p>
+                    <a href="{% url 'cases:case_detail' object_id=case.uuid %}" class="cases__link" aria-label="{% trans "Bekijk aanvraag" %}" title="{% trans "Bekijk aanvraag" %}">
```
pi-sigma, line 38: This is the `aria-label` that causes the problem.

pi-sigma (changes requested):

src/open_inwoner/cms/cases/tests/test_htmx.py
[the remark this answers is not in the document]
jiromaykin: Ah yes, I actually need to rethink this. An aria-label is not needed in case of an anchor-surrounding/overriding-everything but we probably still want a separate label for the button inside the anchor.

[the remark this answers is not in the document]
jiromaykin: @pi-sigma After deeper consideration: I've removed the title attributes since those are actually very bad for accessibility - also the aria-labels are not necessary here because the screenreaders have access to all texts and can read everything (except for the purposefully hidden icons).
(I should start removing unnecessary labels throughout the entire project).
If the component follows correct HTML you don't need aria-labels, unless you wish to override something.

src/open_inwoner/templates/pages/cases/list_inner.html
```diff
@@ -16,38 +16,42 @@ <h1 class="utrecht-heading-1" id="cases">{{ page_title }} ({{ paginator.count }}
             <div class="card__body">
                 <!-- submission cases -->
                 {% if case.case_type == "OpenSubmission" %}
-                    <a href="{{ case.vervolg_link }}"  class="cases__link">
-                    <p class="h4"><span class="link link__text">{{ case.naam }}</span></p>
+                    <a href="{{ case.vervolg_link }}" class="cases__link">
+                    <h2 class="card__heading--h4"><span class="link link__text">{{ case.naam }}</span></h2>
```
pi-sigma, line 20: Maybe remove the `h4` from the class and just use a descriptive class name: `car__heading`, `card__date-caption`, or sth. like this.

pi-sigma (changes requested):

src/open_inwoner/templates/pages/cases/list_inner.html
jiromaykin: @pi-sigma Yes, it looks like in the new design ALL headings are going to be 18px, so I'll need to make this a generic heading class, mostly set on H2.

pi-sigma (commented):
Now that the date has been moved closer to the title, the distance between two lines of a multi-line title is bigger than the distance between title and date. Should we change this?

Screenshot from 2024-01-30 10-33-37

jiromaykin:
> Now that the date has been moved closer to the title, the distance between two lines of a multi-line title is bigger...

I have changed the line-height and changed the default text for open submissions button.

pi-sigma (approved):

PR #960, opened 2024-01-15
:bug: [#1961] Implement redirect after DigiD/eHerkenning OIDC login

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1961

### DIFF
--- a/src/open_inwoner/accounts/views/registration.py
+++ b/src/open_inwoner/accounts/views/registration.py
@@ -146,6 +146,8 @@ class NecessaryFieldsUserView(LogMixin, LoginRequiredMixin, InviteMixin, UpdateV
 
     def get_success_url(self):
         messages.add_message(self.request, messages.SUCCESS, "Registratie is voltooid")
+        if "next" in self.request.GET:
+            return self.request.GET["next"]
         return reverse("pages-root")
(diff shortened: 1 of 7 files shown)


stevenbal:
@alextreme I'll leave this for now, the whole middleware redirect thing is getting too complicated (hence the failing tests)

alextreme:
Agreed, lets take a look together at a later stage and discuss with @pi-sigma what a better approach would be

src/open_inwoner/accounts/views/registration.py
```diff
@@ -146,6 +146,8 @@ def get_object(self, queryset=None):
 
     def get_success_url(self):
         messages.add_message(self.request, messages.SUCCESS, "Registratie is voltooid")
+        if "next" in self.request.GET:
+            return self.request.GET["next"]
```
Bartvaderkin, line 150: Here we need a `is_safe_url` (or in Django 4.0 `url_has_allowed_host_and_scheme`) to protect against unsafe redirects

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #961, opened 2024-01-16
[#2007] Disable upload message if there are recently uploaded documents

pi-sigma:
Taiga [#2007](https://taiga.maykinmedia.nl/project/open-inwoner/task/2007)

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -383,6 +383,11 @@ class InnerCaseDetailView(
                     zt_statustype_config.document_upload_description
                 )
 
+        # disable document upload message
+        if self.request.session.get("uploads", ""):
+            case_type_document_upload_description = ""
+            del self.request.session["uploads"]
+
         return {
             "case_type_config_description": case_type_config_description,
             "case_type_document_upload_description": case_type_document_upload_description,
@@ -655,6 +660,8 @@ class CaseDocumentUploadFormView(CaseAccessMixin, LogMixin, FormView):
             success_message,
             extra_tags="as_markdown local_message",
         )
+
+        self.request.session["uploads"] = True
 
         return HttpResponseClientRedirect(
             reverse("cases:case_detail", kwargs={"object_id": str(self.case.uuid)})


open-inwoner-design-tokens
stevenbal, line 1: I think your `open-inwoner-design-tokens` submodule is outdated

stevenbal (changes requested):
Code looks good and I tested this locally and it works :+1: 

Only something seems to go wrong with the submodule reverting to an older commit

open-inwoner-design-tokens
pi-sigma: Just realized this while testing locally. I'll update the tokens and push again.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -222,6 +222,13 @@ def get_context_data(self, **kwargs):
             hooks.case_status_seen(self.request.user, self.case)
             hooks.case_documents_seen(self.request.user, self.case)
 
+            # used in template + for disabling document upload message
+            self.has_recently_uploaded_documents = has_new_elements(
+                documents,
+                "created",
+                dt.timedelta(days=settings.DOCUMENT_RECENT_DAYS),
```
stevenbal, line 229: @alextreme I see this is now a configurable number of days, would it be possible to instead check if there are new documents since the current status was added? Or could there be a scenario where a user has to upload documents at different points in the time while the same zaakstatus is active?

alextreme: @stevenbal that is an option, however a case could be in the same state for a few weeks or quickly toggle between multiple statusses. We discussed a few alternatives but this one was the easiest and avoids surprises

stevenbal (approved):

pi-sigma:
Submodules are fixed

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -383,6 +386,10 @@ def get_upload_info_context(self, case: Zaak):
                     zt_statustype_config.document_upload_description
                 )
 
+        # disable document upload message
+        if self.has_recently_uploaded_documents:
```
alextreme, line 390: This isn't how this should be solved. The upload-message should only be hidden directly after you've uploaded your documents, and not be related to if you have recent documents or not.

Reason is that the municipality can also upload documents, so ensure we only hide the upload description after returning from CaseDocumentUploadFormView. I'd personally do this based on a GET-argument after a succesful upload

pi-sigma: Makes sense, I didn't consider this case.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -383,6 +383,10 @@ def get_upload_info_context(self, case: Zaak):
                     zt_statustype_config.document_upload_description
                 )
 
+        # disable document upload message after upload
+        if self.request.session.get("uploads", ""):
+            case_type_document_upload_description = ""
+
```
alextreme, line 389: This doesn't take into account that you can view multiple cases during your session, you are now hiding the message for all cases of the user. Unset this flag after you see it has been set

PR #962, opened 2024-01-16
[#1980] Add configuration field for buttons in cases listview

jiromaykin:
Make 'Bekijk aanvraag' text configurable for ordinary Cases, but make it a default text for open-submissions: https://taiga.maykinmedia.nl/project/open-inwoner/task/1980


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -213,6 +213,11 @@ def get_context_data(self, **kwargs):
                         "call_to_action_text",
                         None,
                     ),
+                    "case_link_text": getattr(
+                        self.statustype_config_mapping.get(end_statustype.url),
+                        "case_link_text",
+                        None,
```
jiromaykin, line 219: @stevenbal Do you think this is missing a translation? Not sure if this needs to be 'none' here... Should I add the default value here?

stevenbal: I think this is missing a translation yes, `label` does use a translation for example: https://github.com/maykinmedia/open-inwoner/pull/962/files#diff-b85163f9f163b6780c498ce8ef19dd75145591086f469fa2b67cdfaec753a1e0R193

jiromaykin: Translation would only be needed for the label and the default value (?) since (on the fornt-end) this is a configurable text that will be changed by the admins - it works in the admin, but the label looks 'bold' on the Admin when I view it so not sure what is causing that.

stevenbal: The label looks bold in the admin because the field is required right now

@alextreme we currently don't support translations for these admin configurable texts, right?

src/open_inwoner/openzaak/models.py
```diff
@@ -442,6 +442,14 @@ class ZaakTypeStatusTypeConfig(models.Model):
         verbose_name=_("Call to action text"),
         help_text=_("The text displayed on the call-to-action button"),
     )
+    case_link_text = models.CharField(
+        max_length=40,
+        default="Bekijk aanvraag",
```
stevenbal, line 447: I think I'd prefer it if this field is optional (`blank=True`) and have the default `"Bekijk aanvraag"` in `status.py`. That way admins can simply make the `case_link_text` empty if they want to use the default again

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -450,6 +455,11 @@ def get_statuses_data(
                 "description": getattr(
                     statustype_config_mapping.get(s.statustype.url), "description", None
                 ),
+                "case_link_text": getattr(
+                    statustype_config_mapping.get(s.statustype.url),
+                    "case_link_text",
+                    None,
```
stevenbal, line 461:
```suggestion
                    _("Bekijk aanvraag"),
```

stevenbal (changes requested):

src/open_inwoner/cms/cases/views/status.py
alextreme: @stevenbal admin and frontend are Dutch-only, if that is what you mean

stevenbal: I meant frontend indeed :+1:

stevenbal (approved):

PR #963, opened 2024-01-17
[#2030] Add option to upload and use custom fonts

pi-sigma:
Taiga [#2030](https://taiga.maykinmedia.nl/project/open-inwoner/task/2030)


src/open_inwoner/scss/views/App.scss
```diff
@@ -221,7 +221,7 @@
 
   --font-size-heading-card: 18px;
 
-  --font-family-body: 'TheSans C5';
+  --font-family-body: 'Body';
```
jiromaykin, line 224: Yeah, I guess there is no other way :-)

src/open_inwoner/scss/views/App.scss
```diff
@@ -253,6 +253,26 @@
   --form-width: 500px;
 }
 
+// Custom fonts with fallback
+@font-face {
+  font-family: 'Body';
+  src: local('Body'),
+    url('/media/custom_fonts/text_body_font.otf') format('opentype'),
```
jiromaykin, line 260: This won't work for TTF and WOFF files. Or: ⚠️ we would have to explicitly tell the user in the admin to only use OTF files. It IS possible to convert TTF files to OTF online but for some fonts this could lead to conversion problems. 
Fontsquirrel turns fonts into WOFF2 by default, but with Fontforge, i think, you can use a Python/PIP script to turn them into OTF and any other type for multiple compatibility: https://github.com/fontforge/fontforge

You can download test font files from Fontsquirrel,
Like for example: https://www.fontsquirrel.com/fonts/work-sans where you'll see TTF is common,
Different font-file types would need a different specification in CSS, it is possible to use font-face like this: 
```url('/media/custom_fonts/text_body_font.otf') format('opentype'),
url('/media/custom_fonts/text_body_font.woff') format('woff'),
url('/media/custom_fonts/text_body_font.eot')
      format('embedded-opentype'),
url('/media/custom_fonts/text_body_font.ttf') format('truetype'),
url('/media/custom_fonts/text_body_font.svg') format('svg');```

**Review by jiromaykin — COMMENTED** [review]
This is not a final review yet - these are some comments for deeper research.

File: src/open_inwoner/scss/views/App.scss (hunk 2 above)
Comment (pi-sigma) [reply src/open_inwoner/scss/views/App.scss:260]: Yeah I thought we would only support OTF (you're right this should be made explicit in the help text if we take this route). The issue with supporting alternatives is that this will lead to a bunch of `404`s with browsers failing to find the specified fonts. Let's discuss this in the office.

File: src/open_inwoner/scss/views/App.scss [hunk 3]
```diff
@@ -192,7 +192,7 @@
   /// Font.
 
   --font-color-heading: var(--color-black);
-  --font-family-heading: 'TheMix C5';
+  --font-family-heading: 'Heading';
```
jiromaykin, line 195: Since right now (almost)( all H1's are defined by the design-tokens submodule, they are not overwritten with the font-file upload, so add a fallback font to the H1 fiel in the .utrecht-heading-1 class, like `font-family: var(--font-family-heading-1), var(--font-family-heading-1) ;` - which hopefully is enough to overwrite this.

jiromaykin (changes requested):
When i look in the admin, I see this odd title appear, would be nicer to have an explicit explanation there on using TTF files :-)

font-admin

pi-sigma:
The title on the `StackedInline` should disappear after running `npm run build`.

src/open_inwoner/configurations/models.py
```diff
@@ -584,6 +588,81 @@ def get_help_text(self, request) -> Optional[str]:
...
+            path="custom_fonts/",
+        )
+
+    def update_filename_heading(self, filename: str) -> str:
+        return CustomFontSet.update_filename(
+            self,
+            filename,
+            new_name=CustomFontName.heading,
+            path="custom_fonts/",
+        )
```
Bartvaderkin, line 617: - Model methods go below the model fields.

- Also this code is really unusual: it calls instance methods with the class and explicit self? 

````python
    # why this?
    return CustomFontSet.update_filename(
            self,
            filename,
            new_name=CustomFontName.heading,
            path="custom_fonts/",
        )
    # and not this like normal?
    return self.update_filename(
            filename,
            new_name=CustomFontName.heading,
            path="custom_fonts/",
        )
````

File: src/open_inwoner/configurations/constants.py [hunk 5]
```diff
@@ -0,0 +1,7 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class FontFileName(models.TextChoices):
+    body = _("text_body_font"), _("Text body font")
+    heading = _("heading_font"), _("Heading font")
```
Bartvaderkin, line 7: Do we need this file? The same choices are defined just above this in ./choices.py

src/open_inwoner/configurations/models.py
```diff
@@ -584,6 +588,81 @@ def get_help_text(self, request) -> Optional[str]:
...
+
+    site_configuration = models.OneToOneField(
+        SiteConfiguration,
+        verbose_name=_("Configuration"),
+        related_name="custom_fonts",
+        on_delete=models.CASCADE,
+    )
```
Bartvaderkin, line 624: I assume the OneToOneField is to get an inline for the admin? That's ok.

Bartvaderkin (dismissed):
Cool feature. As this is overwriting the same file/URL I foresee caching issues, where someone uploads new fonts and then nobody sees them because browser cache.

If that happens we might have to fix cache headers (like etag+revalidate instead of just age)

src/open_inwoner/configurations/models.py
pi-sigma: - I can't put the methods below the fields because the method names are referenced in `upload_to`, and I can't use strings referencing the methods (as you would when referencing a model that's defined further down in the code) because Django will interpret the string as the destination of `upload_to`. Is there any other way to reference methods in model fields?
- Class/explicit `self`: confusion on my part

Bartvaderkin: Oh, I missed these are the upload_to callback. Usually callbacks are put above the class (they're not really methods but Django just calls them with the instance as first argument so it looks like it).

src/open_inwoner/scss/views/App.scss
jiromaykin, line 260: This should be 'ttf' not 'otf' (or else it won't render properly).

src/open_inwoner/scss/views/App.scss
```diff
@@ -253,6 +253,26 @@
   --form-width: 500px;
 }
 
+// Custom fonts with fallback
+@font-face {
+  font-family: 'Body';
+  src: local('Body'),
+    url('/media/custom_fonts/text_body_font.otf') format('opentype'),
+    url('/static/fonts/TheSansC5/DesktopFonts/TheSansC5-5_Plain.otf')
```
jiromaykin, line 261: Here also : 'otf' -> 'ttf'

src/open_inwoner/scss/views/App.scss
```diff
@@ -253,6 +253,26 @@
...
+      format('opentype');
+  font-display: swap;
+}
+
+@font-face {
+  font-family: 'Heading';
+  src: local('Heading'),
+    url('/media/custom_fonts/heading_font.otf') format('opentype'),
```
jiromaykin, line 269: And here also : 'otf' -> 'ttf'

src/open_inwoner/scss/views/App.scss
```diff
@@ -253,6 +253,26 @@
...
+    url('/static/fonts/TheSansC5/DesktopFonts/TheSansC5-5_Plain.otf')
```
jiromaykin, line 270: Finally here: 'otf' -> 'ttf'

src/open_inwoner/configurations/migrations/0058_customfontset.py
```diff
@@ -0,0 +1,68 @@
...
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                (
+                    "text_body_font",
+                    open_inwoner.configurations.models.CustomFontField(
+                        blank=True,
+                        help_text="Regular font for the text body. Supported are files with the .ttf extension.",
```
jiromaykin, line 33: Minor detail and not disapproved, but the message here may need to be more in 'active' language, something like "Upload body text font here. TTF font types only" - same as the one for the Headings.
Also I am wondering if we want this font upload to be all the way in the Admin bottom, instead of higher up, closer to the 'custom colors' section?

jiromaykin (changes requested):

src/open_inwoner/configurations/migrations/0058_customfontset.py
pi-sigma: I changed the messages, but moving the admin inline seems to require modifications to the admin template. Let's discuss and perhaps create a separate task for this.

Bartvaderkin: @pi-sigma to add to that: the amount of fields on the configuration singleton is getting a bit out of hand, so for that discussion we could also consider splitting the model, like in technical system configuration, content options, layout stuff and texts. 

This could also make progress to getting rid of that crazy context-processor that just copies a pile of data from the singleton into a dictionary.

pi-sigma: Sounds good. We should come up with a good division though, lest we end up proliferating configuration types and creating something worse.

alextreme:
Jiro wants to check the weird thing with italic stuff

src/open_inwoner/configurations/admin.py
```diff
@@ -60,6 +60,9 @@ def delete_model(self, request, obj):
         else:
             super().delete_model(request, obj)
 
+    class Media:
+        css = {"all": ("css/admin/admin_overrides.css",)}
```
jiromaykin, line 64: I'm not sure why (as it was working before) but when I upload any kind of TTF file, it does correctly end up in the Media directory, but I do not see anything happening on the front-end (perhaps we need a fallback font somewhere?).
Here you can download clearly different TTF fonts for testing: https://www.fontsquirrel.com/fonts/acme

pi-sigma: The font faces definitions were missing the custom fonts after rebasing. Also, I had to specify the font-family for our `utrecht-heading`. Should work now.

jiromaykin (changes requested):
Unfortunately the rendering does not work properly for Bold/Italic font styles if they aren't uploaded as well - the 'faux' rendering makes font-styling _barely_ visible in Chrome and even strips the Bold styling from Card titles, so I'd say: only for the Body font you'll need to add upload options for ""regular"/"italic"/"bold"/"bold-italic". But keep the `  font-weight: bold;` and `font-style: italic;` definitions for each individual font as needed.

In general: "every browser is doing its own thing in font rendering. ... Every browser seems to take its own approach to rendering font weight and anti-aliasing." Using a native bold-face and italic-face is the best way to counter browser differences.

I don't think it is necessary to add "italic" or "bold" uploads for Headings since this should be rendered as a different font anyway, so For Headings you can remove my original bold/italic definitions and probably only use:

`@font-face {
  font-family: 'Heading';
  src: url('/media/custom_fonts/heading_font.ttf') format('truetype'),
    url('/static/fonts/ubuntu/Ubuntu-M.ttf') format('truetype');
}`

alextreme:
FAILED (failures=3)

jiromaykin (approved):
Congrats, it is working! 
Now Admins can do a lot of things...

bold-italic-combined

PR #964, opened 2024-01-17
[#1772] Upgrade django-log-outgoing-requests to 0.6.0

pi-sigma:
Taiga [#1772](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1772)


alextreme (approved):
Should be okay, can be merged after tests pass. I will have to add this cronjob though but that's done via maykin-deployment

PR #965, opened 2024-01-18
[#2032] Make submodule update from main branch

jiromaykin:
issue: also connected to #909  maybe connecting to a tag in future


jiromaykin:
@alextreme would this be the correct way to do it? What if I update the design-token package? Will everyone be able to automatically update or will developers have to manually pull from Main?

alextreme:
@jiromaykin thanks this at least will avoid confusion which version of the design token package the submodule points to. Updating the submodule is done manually, afaik there isn't a way to force this to be in sync

alextreme (approved):

PR #966, opened 2024-01-18
:recycle: [#2027] Only set KVK branch sessionvar if branch is selected

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2027

previously, in case the user did not select a branch, the KVK_BRANCH_NUMBER would be set to the KVK, which meant that everywhere this sessionvariable was used, it had to be compared to the KVK number to see if it was an actual branch number or if it was set just to skip the branch selection screen


alextreme (changes requested):
Tests

alextreme (approved):

PR #967, opened 2024-01-18
[#2023] Implemented Category/Product admin access based on auth.Group

Bartvaderkin:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2023

To make this work create a Group in the admin and assign some Categories. Then assign this Group to a non-superuser, and login as this User. 

Then in the Category list view you'll only see this category, and in the Product list view you only see Product from this Category. Additionally in the Product edit view you can only assign Categories from your Groups (there is some logic to keep assigned Categories you're not allowed to manage)..


src/open_inwoner/pdc/models/category.py
```diff
@@ -104,6 +104,16 @@ class Category(MP_Node):
         ),
     )
 
+    access_groups = models.ManyToManyField(
+        "auth.Group",
+        verbose_name=_("Restrict to admin groups"),
+        blank=True,
+        help_text=_(
+            "If set only users belonging to this group can edit this category and it's products"
```
pi-sigma, line 112: Typos:

If set, only users belonging to this group can edit this category and its products

pi-sigma (changes requested):
A group user who is not part of a group with access to category X can still view and change X (or any other category) by assigning the permission `pdc | Product | Can change product` to themselves.

It would be best to just hide the permission section from the admin for group users (only show it to superusers).

Bartvaderkin:
@pi-sigma  

- Unless I'm missing something you shouldn't be able to navigate to a category or product edit-page you don't have access to, as it won't show up in the list because it is filtered at the `get_queryset()` level. Which is also why the error message reports not-found instead of bad-permission. It is out of scope to cover this with a custom message.

- I don't think this is an issue? Users with these category restrictions are of lower responsibility and won't have group management because that would create this escalation route.

src/open_inwoner/pdc/admin/category.py
```diff
@@ -189,12 +191,56 @@ class CategoryAdmin(
...
     formats = [base_formats.XLSX, base_formats.CSV]
 
+    def get_fields(self, request, obj=None):
+        fields = list(super().get_fields(request, obj=obj))
+        if not request.user.has_perm("auth.change_group"):
+            try:
+                fields.remove("access_groups")
+            except ValueError:
+                pass
+        return fields
```
stevenbal, line 218: Does this work currently? I tried to trigger this by visiting the detail page and I could still see the field:
image

The user group had these permissions:
image

src/open_inwoner/pdc/admin/product.py
```diff
@@ -40,6 +42,41 @@ class Meta:
         fields = "__all__"
         widgets = {"content": CKEditorWidget}
 
+    categories = forms.ModelMultipleChoiceField(
```
stevenbal, line 45: Should I be able to see/edit this field if I'm logged in as a user that has managed categories? I was able to remove categories from a product here and block my own access to this product after doing so

src/open_inwoner/pdc/tests/test_category_admin.py
```diff
@@ -95,6 +96,52 @@ def test_user_cannot_publish_child_category_without_root_published_on_list_page(
         updated_category = Category.objects.get(slug="bar4")
         self.assertFalse(updated_category.published)
 
+    def test_access_limited_to_linked_auth_groups(self):
```
stevenbal, line 99: I think we're missing a testcase to verify that a user with managed categories cannot see/edit the `access_groups` field on the `Category` change page

stevenbal (changes requested):

src/open_inwoner/pdc/admin/category.py
Bartvaderkin: I removed this, this was from before I managed to add the m2m field to the Group's admin and makes more sense there.

src/open_inwoner/pdc/admin/product.py
Bartvaderkin: I added a check in the form and a test that forces there to always have at least one category.

src/open_inwoner/pdc/tests/test_category_admin.py
Bartvaderkin: I added this test (even though it is now readonly)

stevenbal (approved):

pi-sigma (approved):

PR #972, opened 2024-01-19
Implemented cached version of `generate_oas_component()`

Bartvaderkin:
I tried this after-hours and it works surprisingly well so I cleaned it up and put it here:

It caches `generate_oas_component()` calls that are repeated multiple times, like when they are part of a `TestCase.setup()` or some other often re-used code like the `MockAPIData` class.

It works by using some funky python `hash()` code to get a cache key and `deepcopy()` to isolate each usage.

Also it's obviously a memory hole, but memory and `deepcopy()` are a lot cheaper then CPU cycles running the OAS/faker stack.

### DIFF
--- a/src/open_inwoner/openzaak/tests/helpers.py
+++ b/src/open_inwoner/openzaak/tests/helpers.py
@@ -2,6 +2,10 @@ import re
 from copy import deepcopy
 from uuid import uuid4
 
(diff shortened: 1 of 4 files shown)


stevenbal (approved):
Code looks good :+1:. Does it make the tests faster as well?

PR #976, opened 2024-01-19
:sparkles: [#2025] Daily email digest to inform admins about failed emails

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2025

### DIFF
--- a/src/open_inwoner/configurations/tests/test_emails.py
+++ b/src/open_inwoner/configurations/tests/test_emails.py
@@ -1,0 +1,96 @@
+from django.core import mail
+from django.core.management import call_command
+from django.test import TestCase
+
+from django_yubin.constants import RESULT_FAILED
+from django_yubin.models import Log, Message
+from freezegun import freeze_time
+
+from ..models import SiteConfiguration
+
+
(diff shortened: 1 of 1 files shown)


src/open_inwoner/configurations/management/commands/send_failed_mail_digest.py
```diff
@@ -0,0 +1,21 @@
+import logging
+
+from django.core.management.base import BaseCommand
+
+from open_inwoner.configurations.emails import inform_admins_about_failing_emails
+
+logger = logging.getLogger(__name__)
```
pi-sigma, line 7: You're defining a logger but then only write to `stdout`. You can either get rid of the logger or use it in the `handle`.

pi-sigma (changes requested):

pi-sigma (approved):

alextreme (approved):
Please also add a PR cronjobs.conf in deployment, otherwise I'll overlook it: https://bitbucket.org/maykinmedia/maykin-deployment/src/master/projects/open-inwoner/cronjobs.conf

PR #977, opened 2024-01-22
⬆️ [#2040] Upgrade python to 3.11 and zgw-consumers to 0.28.0

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2040

This PR upgrades to python3.11 and zgw-consumers to 0.28.0, in this new version of zgw-consumers we no longer do schema parsing to determine request paths, which removes latency that was caused by that

Notes:
- we should probably go over every integration manually on test/staging to see if it still works. Since a lot of stuff is mocked in tests, we won't detect problems like missing CRS headers for Zaken API until we test it on an actual environment. Also because we no longer rely on operationIds to determine request paths, we'll have to see if eSuite actually implements the same request paths as Open Zaak
- `get_paginated_results` doesn't work with the new client, created another issue for this -> https://taiga.maykinmedia.nl/project/open-inwoner/task/2060
- also created an issue to refactor the client usage to be more in line with Open Forms, that way we could eventually move common code to a package


alextreme:
Depends on upgrading cindy+sandra to Debian 12

pi-sigma (approved):

src/open_inwoner/utils/api.py
```diff
@@ -0,0 +1,84 @@
...
+            response_json = None
+
+        try:
+            response.raise_for_status()
+        except requests.HTTPError as exc:
+            if response.status_code >= 500:
+                raise
+            raise ClientError(response_json) from exc
+
+        return response_json
```
Bartvaderkin, line 84: Feels weird this is not part of either apie_pie or ZGW-consumers but heh.

Bartvaderkin (approved):
At last a newer python!

Removing the operation ID's feels brave, but if ZGW-consumers believes in it then lets go for it.

I put two notes but not blocking.

src/open_inwoner/conf/base.py
```diff
@@ -806,6 +806,7 @@
     os.path.join(DJANGO_PROJECT_DIR, "openzaak", "tests", "files"),
     os.path.join(DJANGO_PROJECT_DIR, "openklant", "tests", "files"),
 ]
+# ZGW_CONSUMERS_CLIENT_CLASS = "ape_pie.APIClient"
```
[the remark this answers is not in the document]
stevenbal: Ah I missed that one, that setting is a legacy setting from `zgw-consumers`, so I'll remove it

src/open_inwoner/utils/api.py
stevenbal: I agree, I think we'll want this for more projects

alextreme:
Mag gemerged worden zodra de conflicts zijn geresolved

PR #978, opened 2024-01-22
[#1982] Add file-type error validation for document-upload

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/1982 

Plus: also fixed the issue where the delete button became non-functional when more than 1 faulty file is selected.

### DIFF
--- a/src/open_inwoner/components/templates/components/Card/CardContainer.html
+++ b/src/open_inwoner/components/templates/components/Card/CardContainer.html
@@ -4,29 +4,29 @@
         {% for category in categories %}
             {% url 'products:category_detail' slug=category.slug as category_url %}
             {% if category.icon %}
-                {% card src=category.icon.file.url alt=category.icon.name title=category.name href=category_url image_object_fit=image_object_fit %}
+                {% card src=category.icon.file.url alt=category.icon.name title=category.name href=category_url compact=True image_object_fit=image_object_fit %}
             {% else %}
-                {% card src=category.image.file.url alt=category.image.name title=category.name href=category_url image_object_fit=image_object_fit %}
+                {% card src=category.image.file.url alt=category.image.name title=category.name href=category_url compact=True image_object_fit=image_object_fit %}
             {% endif %}
         {% endfor %}
     {% endif %}
 
     {% if subcategories %}
         {% for subcategory in subcategories %}
-            {% category_card category=subcategory parent_category=parent_category image_object_fit=image_object_fit %}
+            {% category_card category=subcategory parent_category=parent_category compact=True image_object_fit=image_object_fit %}
         {% endfor %}
     {% endif %}
 
     {% if products %}
         {% for product in products %}
             {% get_product_url product as product_url %}
-            {% product_card title=product.name description=product.summary url=product_url image=product.icon image_object_fit=image_object_fit %}
+            {% product_card title=product.name description=product.summary url=product_url image=product.icon compact=True image_object_fit=image_object_fit %}
         {% endfor %}
     {% endif %}
 
     {% if plans %}
         {% for plan in plans %}
-            {% description_card title=plan.title description=plan.goal|truncatechars:51 url=plan.get_absolute_url elypsis=True object=plan image_object_fit=image_object_fit %}
+            {% description_card title=plan.title description=plan.goal|truncatechars:51 url=plan.get_absolute_url elypsis=True object=plan compact=True image_object_fit=image_object_fit %}
         {% endfor %}
     {% endif %}
 
--- a/src/open_inwoner/components/templates/components/Form/FileInput.html
+++ b/src/open_inwoner/components/templates/components/Form/FileInput.html
@@ -1,11 +1,9 @@
-{% load i18n solo_tags button_tags card_tags form_tags icon_tags %}
-
-{% get_solo 'openzaak.OpenZaakConfig' as openzaak_config %}
+{% load i18n button_tags card_tags form_tags icon_tags %}
 
 <div class="form__control file-input" aria-live="polite">
     {% render_card direction="vertical" %}
         {% icon icon="upload" icon_position="before" outlined=True %}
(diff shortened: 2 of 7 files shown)


src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -5,7 +5,7 @@
 <div class="form__control file-input" aria-live="polite">
     {% render_card direction="vertical" %}
         {% icon icon="upload" icon_position="before" outlined=True %}
-        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %}{% if multiple %} multiple{% endif %} data-max-size="{{ openzaak_config.max_upload_size }}">
+        <input class="file-input__input" id="{{ field.auto_id }}" name="file" type="file"{% if field.field.required %} required{% endif %}{% if multiple %} multiple{% endif %} data-max-size="{{ openzaak_config.max_upload_size }}" data-file-types="{{ openzaak_config.allowed_file_extensions }}">
```
stevenbal, line 8: I'm not sure if this `FileInput` is used anywhere other than the document upload form, but since it looks like a generic template it seems wrong to directly use `openzaak_config` here.

I'd prefer it if we pass something like `allowed_file_extensions=openzaak_config.allowed_file_extensions` here: https://github.com/maykinmedia/open-inwoner/blob/e8471d865eb52e26543b9c378d8299ec77952b18/src/open_inwoner/templates/pages/cases/document_form.html#L6 and then directly use `allowed_file_extensions` in the `FileInput` template

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -261,53 +275,57 @@ export class FileInput extends Component {
...
     const labelDelete = this.getFilesList().dataset.labelDelete || 'Delete'
     const getFormNonFieldError = this.getFormNonFieldError()
     const formSubmitButton = this.getFormSubmitButton()
 
-    // Only show errors notification if data-max-file-size is exceeded + add error class to file-list
     const maxMegabytes = this.getLimit()
+    const uploadFileTypes = this.getUploadTypes().toUpperCase()
+
+    const sizeError = sizeMB > maxMegabytes
+    const typeError = !uploadFileTypes.includes(ext)
```
stevenbal, line 289: Not sure if this is a big problem, but because `uploadFileTypes` is a string and not a list with extensions, a file with a space as extension is also allowed (because ` ` occurs in the string):

image

Similarly something like "file.pdf, doc" would also be allowed.

I'd turn it into a list to make sure this is not possible:
```suggestion
    const typeError = !uploadFileTypes.split(", ").includes(ext)
```

src/open_inwoner/js/components/form/FileInput.js
```diff
@@ -261,53 +275,57 @@ export class FileInput extends Component {
...
+            typeError || sizeError
+              ? `
+                <p class="p p--small error">
+                  <span aria-hidden="true" class="material-icons-outlined">warning_amber</span>
+                  ${
+                    typeError
+                      ? `Dit type bestand ("${ext}") is ongeldig. Geldige bestandstypen zijn: ${uploadFileTypes}.`
+                      : ''
+                  }
+                  ${sizeError ? 'Dit bestand is te groot' : ''}
```
stevenbal, line 318: Not sure if we consider this an issue, but if a file has both a `sizeError` and a `typeError` the `sizeError` does not appear as a separate error, but is added to the end instead:

image

stevenbal (changes requested):

stevenbal:
Looks good overall, few minor remarks

src/open_inwoner/js/components/form/FileInput.js
jiromaykin: If I would put them on a separate line, then I'd guess the icon should be added to the second error as well.
Or: It could be more important if the file is uploaded is of the wrong type, and then fileSize errors are not important anymore?
Let me ask the designer.

pi-sigma: I'd prefer the first option: show both errors at the same time, but on separate lines and each with its own error icon. It's annoying to resolve one error (e.g. convert to different file type), only to then find out that there's another problem.

jiromaykin: @pi-sigma I've already discussed this with the designer and the municipality; in case of a combined error we will get a (new) combined message with 1 icon only, so I will add that.

src/open_inwoner/components/templates/components/Form/FileInput.html
jiromaykin: Ok @stevenbal then I think I should also do this for the max-file-size.

src/open_inwoner/js/components/form/FileInput.js
jiromaykin: Odd: in all my browsers my front-end is actually already displaying these odd extensions as faulty and returns an error, but I will add this array method anyway because it certainly will not do harm,

src/open_inwoner/components/templates/components/Card/CardContainer.html
```diff
@@ -4,29 +4,29 @@
         {% for category in categories %}
             {% url 'products:category_detail' slug=category.slug as category_url %}
             {% if category.icon %}
-                {% card src=category.icon.file.url alt=category.icon.name title=category.name href=category_url image_object_fit=image_object_fit %}
+                {% card src=category.icon.file.url alt=category.icon.name title=category.name href=category_url compact=True image_object_fit=image_object_fit %}
```
stevenbal, line 7: I don't mind these changes, but is this related to issue [1982](https://taiga.maykinmedia.nl/project/open-inwoner/task/1982)?

stevenbal (approved):
Just a question regarding the `CardContainer` changes, the rest looks good

pi-sigma (approved):

src/open_inwoner/components/templates/components/Card/CardContainer.html
jiromaykin: It actually is, yes - The 'drag' field here was no longer conforming to the design because things changed in the Card paddings in another PR: some cards now have different vertical padding + different horizontal padding, but other Cards still have the same padding on all sides.

PR #982, opened 2024-01-23
[#2038] Remove appointment model

pi-sigma:
Taiga: [#2038](https://taiga.maykinmedia.nl/project/open-inwoner/task/2038)


Bartvaderkin (commented):
Cool. Did you check for fixtures etc? Like admin index, permissions etc?

Also I note I see no application or test code removal, was it this code this dead?

pi-sigma:
I updated the fixtures and messages. Apparently, the code was not used at all.

Bartvaderkin (approved):
:broom: :+1:

PR #983, opened 2024-01-23
Applied generate_oas_component_cached() where applicable

Bartvaderkin:
Let's use `generate_oas_component_cached()` in all the relevant setUp()'s and see what it does for test runtime.

I also refactored/restored some ZGW tests that earlier sacrificed test-isolation for performance, by using the setupTestData class-method, but this is no longer needed. This is slightly slower because the normal factories run for every tests but it is safer and more correct.


Bartvaderkin:
The CI is absolute mayhem, at first glance I see some bad factories and sloppy test setups, so I'll look into those and then see if there are actual problems.

stevenbal (approved):

pi-sigma (approved):

PR #984, opened 2024-01-23
[#2044] Remove export-to-pdf code

pi-sigma:
Taiga: [#2044](https://taiga.maykinmedia.nl/project/open-inwoner/task/2044)


stevenbal (changes requested):
I checked and saw that the following files still have profile export related code
- https://github.com/maykinmedia/open-inwoner/blob/cleanup/2044-remove-pdf-export-code/src/open_inwoner/components/templates/components/Action/Actions.html#L21
- https://github.com/maykinmedia/open-inwoner/blob/cleanup/2044-remove-pdf-export-code/src/open_inwoner/templates/pages/profile/actions/list.html#L13

I checked where this is used and the Action under `Samenwerken` now does show a PDF export button, but it doesn't work because the underlying view doesn't exist anymore. Was the intention to remove the export for Actions as well?
image

pi-sigma:
I got confused because the action export templates are placed under `profile`, so thought they should be removed as well. I missed the fact that the export could be accessed from elsewhere. I'll add the relevant code back.

stevenbal (approved):

PR #986, opened 2024-01-23
:lipstick: [#2029] Set new default open-source font

jiromaykin:
Issue: 

1. Set Ubuntu as heading font and Lato as body font: https://taiga.maykinmedia.nl/project/open-inwoner/task/2029
2. also needs this change in font variable name in this Design-token package PR: https://github.com/maykinmedia/open-inwoner-design-tokens/pull/6

This PR only works when you check out the corresponding [branch of this PR](https://github.com/maykinmedia/open-inwoner-design-tokens/pull/6) and run NPM build on both repositories.

Preview styling here: http://localhost:8000/style-demo/


src/open_inwoner/scss/views/App.scss
```diff
@@ -254,63 +254,56 @@
...
 }
 
 @font-face {
-  font-family: 'TheSans C5';
+  font-family: 'Body';
   font-weight: bold;
-  src: local('TheSans C5'),
-    url('/static/fonts/TheSansC5/DesktopFonts/TheSansC5-7_Bold.otf')
-      format('opentype');
+  src: url('/static/fonts/lato/Lato-Bold.ttf') format('truetype');
```
pi-sigma, line 278: Why do you specify `font-style: normal;` for the `Heading`, but not for the `Body`?

src/open_inwoner/scss/views/App.scss
```diff
@@ -221,10 +221,10 @@
 
   --font-size-heading-card: 18px;
 
-  --font-family-body: 'TheSans C5';
+  --font-family-body: 'Body';
   --font-color-body: var(--color-gray-dark);
   --font-size-body: 16px;
-  --font-line-height-body: 21px;
+  --font-line-height-body: 24px;
```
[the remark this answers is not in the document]
jiromaykin: No, this is the line-height used in all new designs, so it doesn't depend on the font.

src/open_inwoner/scss/views/App.scss
```diff
@@ -254,63 +254,68 @@
...
-  src: local('TheSans C5'),
-    url('/static/fonts/TheSansC5/DesktopFonts/TheSansC5-7_Bold.otf')
-      format('opentype');
+  src: url('/static/fonts/lato/Lato-Bold.ttf') format('truetype');
+}
+
+@font-face {
+  font-family: 'Body';
+  font-style: italic;
+  src: url('/static/fonts/lato/Lato-Italic.ttf') format('truetype');
```
jiromaykin, line 283: @pi-sigma Around here I am specifying separate font-files for the regular/bol/Italic/bold-italic situations, since that is the official way to prevent so-called "faux-rendering" but... I am not sure what effect this will have on your PR #963

pi-sigma:
From `Mijn profiel`:
- The "Mijn vragen" card looks a little squished, distance between title and text could be greater (though that was also the case with the previous font)

The new font looks very light to me; perhaps it just takes some getting used to...

pi-sigma (approved):

jiromaykin:
> From `Mijn profiel`:  * The "Mijn vragen" card looks a little squished

Ah, @pi-sigma  that is not because of the font, but because of the way the grid works: if there is only one card in a row it becomes very small (a different issue, not related to Font).

PR #981, opened 2024-01-22
[#2037] Workaround for missing statusses from the eSuite statuslist endpoint

alextreme:

PR #987, opened 2024-01-24
[#2045] Add regression test for e-suite missing statuses

pi-sigma:
Regression test for Taiga issue [#2037](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2037) ([PR 981](https://github.com/maykinmedia/open-inwoner/pull/981/files))

Taiga: [2045](https://taiga.maykinmedia.nl/project/open-inwoner/task/2045)

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_case_detail.py
+++ b/src/open_inwoner/openzaak/tests/test_case_detail.py
@@ -11,6 +11,7 @@ from django.utils.translation import gettext_lazy as _
 import requests_mock
 from django_webtest import WebTest
 from freezegun import freeze_time
+from pyquery import PyQuery
 from timeline_logger.models import TimelineLog
 from webtest import Upload
 from webtest.forms import Hidden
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -877,6 +877,46 @@ def test_document_ordering_by_name(self, m):
         self.assertEqual(documents[1].name, "uploaded_document_title.txt")
         self.assertEqual(documents[2].name, "yet_another_document_title.txt")
 
+    @patch("open_inwoner.cms.cases.views.status.fetch_status_history")
+    def test_e_suite_missing_current_status(self, m, mock_fetch_status_history):
```
stevenbal, line 881: Since the original issue (https://taiga.maykinmedia.nl/project/open-inwoner/issue/2037) is specifically about the upload button not showing (because `InnerCaseDetailView.is_file_upload_enabled_for_statustype` was returning `False` due to the statustype not being resolved) , I think it would be good to add checks for that here as well

stevenbal (changes requested):

stevenbal (approved):

PR #988, opened 2024-01-24
[#2048] Add regression test for eSuite statustype zaaktype mismatch

pi-sigma:
Taiga: [#2048](https://taiga.maykinmedia.nl/project/open-inwoner/task/2048)

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_case_detail.py
+++ b/src/open_inwoner/openzaak/tests/test_case_detail.py
@@ -818,6 +818,49 @@ class TestCaseDetailView(AssertRedirectsMixin, ClearCachesMixin, WebTest):
                 res = detail_view.get_second_status_preview(status_types)
                 self.assertEqual(res, result)
 
+    def test_e_suite_statustype_zaaktype_mapping(self, m):
+        """
+        Regression test for #1991
+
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -818,6 +818,57 @@ def test_second_status_preview(self, m):
...
+            case_link_text="bogus",
+        )
+        ZaakTypeStatusTypeConfigFactory(
+            zaaktype_config=self.zaaktype_config,
+            statustype_url=self.status_type_finish["url"],
+            status_indicator=StatusIndicators.success,
+            status_indicator_text="bar",
+            call_to_action_url="https://www.example.com",
+            call_to_action_text="Click me",
+            case_link_text="Bekijk aanvraag",
```
Bartvaderkin, line 860: It is better for readability to strip the factories and generators from fields that aren't relevant to the test (also a few just above). 

If they are required for the model but not relevant for the test they should be part of the factory.

Bartvaderkin (changes requested):

Bartvaderkin (approved):

PR #990, opened 2024-01-25
[#2041] Added hard database constraints on User's identifier/login_type fields

Bartvaderkin:
I tried to make it impossible to store bad identifier data (like duplicates), or have lingering data (like a bsn when the login type is not digid).

We need to check a few assumptions:

- [ ] user will always have an identifier (email, bsn, oidc_id, rsin or kvk) for their login_type
- [ ] users cannot not switch login types
- [ ] eherkenning users cannot have both rsin and kvk
- [ ] email must be unique for default logins, but not for eherkenning 
- [ ] it is intentional OIDC allows non-unique emails

I think I have all cases in the new test, see diff: src/open_inwoner/accounts/tests/test_user_constraints.py


alextreme:
Background is that it's possible (once in a blue moon and all within a few seconds) to register multiple Users via DigiD with the same bsn

Bartvaderkin:
The CI is absolute mayhem, at first glance I see some bad factories and sloppy test setups, so I'll look into those and then see if there are actual problems.

Bartvaderkin:
After discussion with Alex I unassigned @stevenbal and @pi-sigma for now because it's slightly too early (although you're welcome to have a look as we'll discuss this monday)

src/open_inwoner/accounts/migrations/0071_auto_20240125_1227.py
```diff
@@ -0,0 +1,88 @@
...
+        migrations.AddConstraint(
+            model_name="user",
+            constraint=models.CheckConstraint(
+                check=models.Q(
+                    models.Q(("kvk", ""), models.Q(("rsin", ""), _negated=True)),
+                    models.Q(models.Q(("kvk", ""), _negated=True), ("rsin", "")),
+                    models.Q(("login_type", "eherkenning"), _negated=True),
+                    _connector="OR",
+                ),
+                name="check_kvk_or_rsin_exclusive_when_login_eherkenning",
```
stevenbal, line 85: this will not work with the current implementation of eHerkenning in OIP, because we set both the KVK and RSIN for users

PR #991, opened 2024-01-25
:sparkles: [#1861] Show status in mijn vragen and use same design as aanvragen

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1861

### DIFF
--- a/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
+++ b/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
@@ -1,6 +1,27 @@
 .contactmomenten {
   margin-top: var(--spacing-giant);
 
+  & .grid {
+    grid-row-gap: var(--spacing-medium);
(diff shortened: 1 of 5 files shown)


src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -1,6 +1,27 @@
 .contactmomenten {
   margin-top: var(--spacing-giant);
 
+  & .grid {
```
stevenbal, line 4: This should probably become a generic thing, since it's the same as for `.cases`

jiromaykin: Yes, we will need to refactor this (at some point) in the entire project, because some cards are spread over 4 columns (Onderwerpen) and others over 3 (Cases, Vragen, Profiel cards) and they all need responsiveness. But this is okay for now.

jiromaykin (approved):

PR #992, opened 2024-01-25
[#2036] Add tab-design for login page

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2036
Differs a bit in design from the tab we have in 'Mijn uitkeringen'


src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -8,95 +8,153 @@
...
-                        {% link href=href_with_next text=_('Inloggen met DigiD') secondary=True icon='arrow_forward' extra_classes="link--digid" %}
-                    {% endwith %}
-                    {% endrender_card %}
-            {% else %}
-                {% render_card direction='horizontal' tinted=True compact=True %}
-                    <a href="{% url 'digid:login' %}" class="link digid-logo">
-                        <img class="digid-logo__image" src="{% static 'accounts/digid_logo.svg' %}" alt="DigiD inlogpagina">
-                    </a>
-                    {% url 'digid:login' as href %}
-                    {% with href|addnexturl:next as href_with_next %}
```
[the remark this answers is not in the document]
jiromaykin: No it's there - it has just moved to line nr. 47.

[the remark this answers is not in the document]
Bartvaderkin: (discussed this in chat)

src/open_inwoner/js/components/form/LoginForm.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  activateTabFromHash() {
+    const hash = window.location.hash
+    const particulierLink = document.querySelector(
+      '.tab__header[href="/accounts/login/#particulier"]'
+    )
+    const zakelijkLink = document.querySelector(
+      '.tab__header[href="/accounts/login/#zakelijk"]'
```
Bartvaderkin, line 57: Feedback for later: Django URL's are technically dynamic so we don't want hardcoded URL's in the Javascript (or HTML, CSS etc). This is why there are `{% url "login" %}` in the templates (to `reverse()` the URLs from whatever is in the urls.py's).
 
So these Link selectors should be different and not use the href= but a class or id (maybe chain of the Tab if they wrap the Link)

src/open_inwoner/js/components/form/LoginForm.js
```diff
@@ -0,0 +1,83 @@
...
+
+      zakelijkTab.classList.add('hide')
+      zakelijkTab.classList.remove('active')
+      zakelijkLink.classList.remove('active')
+    }
+  }
+}
+
+const loginformFocuses = document.querySelectorAll(LoginFormFocus.selector)
+;[...loginformFocuses].forEach((element) => new LoginFormFocus(element))
```
Bartvaderkin, line 83: Feedback for later: The queryselector has a forEach() so you can just chain from that:

````js
document.querySelectorAll(LoginFormFocus.selector).forEach((element) => new LoginFormFocus(element))
```

Bartvaderkin (approved):
I appended some fixes to the urls in login.html (the next url wasn't applied to both the text and icon url, and some urls were hardcoded).

I also added some comments for later (most just code style, one possible issue but it is not an important feature and doesn't seem to come up in normal use).

I also checked the conditionals so I think this is good to go for acceptance.

src/open_inwoner/js/components/form/LoginForm.js
jiromaykin: OK, @Bartvaderkin I will have to create new id's/selectors here because the tabs are not inside those panels-with-id's; 
they are elsewhere in the HTML structure.

PR #994, opened 2024-01-29
:sparkles: [#1846] Make login page text wysiwyg

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/1846

### DIFF
--- a/src/open_inwoner/accounts/templates/registration/login.html
+++ b/src/open_inwoner/accounts/templates/registration/login.html
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n static logo_tags grid_tags card_tags form_tags link_tags button_tags solo_tags cms_tags %}
+{% load i18n static utils logo_tags grid_tags card_tags form_tags link_tags button_tags solo_tags cms_tags %}
 
 
 {% block header_image %}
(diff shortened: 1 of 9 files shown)


src/open_inwoner/accounts/templates/registration/login.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n static logo_tags grid_tags card_tags form_tags link_tags button_tags solo_tags cms_tags %}
+{% load i18n static ssd_tags logo_tags grid_tags card_tags form_tags link_tags button_tags solo_tags cms_tags %}
```
alextreme, line 2: can we move the markdown filter from ssd_tags into the utils tags if we're using it in other places too?

src/open_inwoner/scss/components/Typography/Link.scss
```diff
@@ -83,3 +83,20 @@
     transform: scale(1);
   }
 }
+
+.wysiwyg {
```
jiromaykin, line 87: I think it would be nicer to add a completely new file for all of the 'wysiwyg' styles inside the Typography SCSS directory - because CKEditor probably has many unpleasant side effects,
so in src/open_inwoner/scss/components/Typography perhaps add a 'wysiwyg' file with that as a parent class and then actual tags can be added in the future (we will add Tables at some point too).

jiromaykin (changes requested):

alextreme (approved):
Dank!

PR #995, opened 2024-01-29
♻️ [#2060] Replace get_paginated_results with pagination_helper

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2060

This PR replaces `get_paginated_results` (which was broken for zgw-consumers 0.28.0) with the new `pagination_helper` and refactors the client usage such that we have an `APIClient` for each API (zaken, catalogi, klanten, etc.)


src/open_inwoner/openklant/wrap.py
```diff
@@ -88,211 +82,42 @@ def fetch_klantcontactmoment(
...
-        return []
-
-    klanten = factory(Klant, response)
+        return
 
-    return klanten
+    return client.retrieve_objectcontactmoment(contactmoment, object_type)
 
 
 def fetch_klant(
```
stevenbal, line 95: I'm not entirely sure if we still need these functions after the changes, I suppose they do save us the hassle of having to build a client when we use it in views

src/open_inwoner/openzaak/clients.py
```diff
@@ -1,28 +1,565 @@
...
+
 logger = logging.getLogger(__name__)
 
 
+class ZakenClient(APIClient):
+    @cache_result(
+        "cases:{user_bsn}:{max_requests}:{identificatie}",
+        timeout=settings.CACHE_ZGW_ZAKEN_TIMEOUT,
+    )
+    def fetch_cases(
```
stevenbal, line 47: We might want to rename this to be a bit more standardized (because now I see things like case / zaak mixed). Perhaps something like create/read/update/delete + name of the resource in the API: `retrieve_zaken`

src/open_inwoner/openklant/wrap.py
```diff
@@ -31,21 +25,27 @@ def fetch_klantcontactmomenten(
     if not user_bsn and not user_kvk_or_rsin:
         return []
 
+    client = build_client("klanten")
+    if client is None:
+        return []
```
stevenbal, line 30: I think it would be better to raise an error in `build_client` if no service was found, that way we can properly handle the situation in which configuration is missing, but that might be tackled in a separate issue

src/open_inwoner/openzaak/clients.py
```diff
@@ -1,28 +1,565 @@
...
+            params.update({"identificatie": identificatie})
+
+        try:
+            response = self.get(
+                "zaken",
+                params=params,
+                headers=CRS_HEADERS,
+            )
+            data = get_json_response(response)
+            # FIXME the previous implementation that used `get_paginated_results`
```
stevenbal, line 75: @alextreme do you know more about this? `max_cases` was set to 100, but for OZ that is the pagination limit, which means that it never actually tried to fetch the results of the next page. Is the page size for eSuite different?

I changed `max_cases` to be `max_requests` to limit the number of requests we do when retrieving paginated results

alextreme: I've checked the situation with eSuite Groningen, paging is set to 20 on the zaken endpoint. So yes the page size can be different

alextreme:
Nice, thanks for this Steven! I'll wait for Bart and/or Paul to do a review first

src/open_inwoner/openzaak/clients.py
stevenbal: Hmm okay, do you think increasing `max_requests` for pagination to 4 would be good? Or should I bring back the option to `zgw_consumers.pagination_helper` to stop once a maximum number of results is reached?

src/open_inwoner/openklant/clients.py
```diff
@@ -1,26 +1,246 @@
...
+    def retrieve_objectcontactmomenten_for_object_type(
+        self, contactmoment: ContactMoment, object_type: str
+    ) -> List[ObjectContactMoment]:
+
+        moments = self.retrieve_objectcontactmomenten_for_contactmoment(contactmoment)
+
+        # eSuite doesn't implement a `object_type` query parameter
+        ret = [moment for moment in moments if moment.object_type == object_type]
+
+        return ret
```
pi-sigma, line 222: 1. Since we're not using the list (?), you could:

- `yield from (moment for moment in moments if...)`
- (in the following function): return next(ocms, None)

2. This seems to be the only use of the function which potentially returns multiple objects, but then we're only retrieving and returning the first (in `retrieve_objectcontactmoment`). Is this to allow for possible extensions where we actually use multiple `objectcontactmomenten` from this? Otherwise this and the following function could be collapsed.

src/open_inwoner/openklant/clients.py
```diff
@@ -1,26 +1,246 @@
...
     }
-    if type_ in services:
+    if client_class := services_to_client_mapping.get(type_):
         service = getattr(config, f"{type_}_service")
         if service:
-            client = _build_client(service, client_factory=JSONParserClient)
+            client = _build_client(service, client_factory=client_class)
             return client
 
     logger.warning(f"no service defined for {type_}")
```
pi-sigma, line 246: `logger.warning("no service defined for {type}", type=type_)` (better to avoid f-string interpolation with logging)

src/open_inwoner/openzaak/api_models.py
```diff
@@ -273,3 +273,32 @@ class Notification(Model):
     actie: str
     aanmaakdatum: datetime
     kenmerken: Dict = field(default_factory=dict)
+
+
+@dataclass
+class OpenSubmission(Model):
+    url: str
```
pi-sigma, line 280: Now that open submissions are recognized as first-class citizens, do we still the `UniformCase` protocol that I introduced for the `InnerCaseListView`? It was a workaround at a time where it wasn't quite clear how we want to think about the distinction between zaken and open submissions:

- [https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openzaak/types.py](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openzaak/types.py)
- [https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/cms/cases/views/cases.py#L89](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/cms/cases/views/cases.py#L89)

pi-sigma (changes requested):
Nice work! Most of the comments are suggestions, so feel free to adopt what seems right.

src/open_inwoner/openklant/clients.py
```diff
@@ -1,26 +1,246 @@
 import logging
-from typing import Optional
+from typing import List, Optional
```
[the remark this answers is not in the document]
Viicos: Just passing by: we recently removed all these occurrences to use the builtins instead (available since 3.9 and 3.10 for `str | None` syntax) in Open Forms using ruff to do it automatically. Related PRs:
- https://github.com/open-formulieren/open-forms/pull/3802
- https://github.com/open-formulieren/open-forms/pull/3818

Just have to be careful when converting `Union/Optional` to ` | `, it might fail at runtime e.g. when doing `"SomeForwardClass" | int` (in that case, this should be `"SomeForwardClass | int"`

src/open_inwoner/openzaak/clients.py
alextreme: I guess it depends on the situation, I wouldn't mind if the pagination used in the cases listview simply uses the same page size as the ZGW API it is connected to uses but if it was limited to 100 previously then doing at most 5 requests would keep this comparable. As long as there's a decent maximum I'm okay with it

src/open_inwoner/openklant/clients.py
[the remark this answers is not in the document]
Bartvaderkin: @pi-sigma Please let's stick to one standard and not individually start new styles and patterns; it'll make it even messier. We can change it in a big sweep of we want.

src/open_inwoner/openklant/wrap.py
Bartvaderkin: I think we could remove them and just use the client instances in the views. 

These are a level of indirection and we should really be re-using clients (for connection and expensive cert/schema setups etc).

src/open_inwoner/openklant/wrap.py
Bartvaderkin: Yes, but we should also handle connection and response errors so I'd make it a new ticket.

src/open_inwoner/utils/api.py
```diff
@@ -14,71 +12,17 @@ class ClientError(Exception):
...
-
-        try:
-            response.raise_for_status()
-        except requests.HTTPError as exc:
-            if response.status_code >= 500:
-                raise
-            raise ClientError(response_json) from exc
-
-        return response_json
+    return response_json
```
Bartvaderkin, line 28: This also feels like a common thing that would be part of ZGW or the client library, like the pagination?

Bartvaderkin (changes requested):

src/open_inwoner/openklant/clients.py
Viicos: `logger.warning("no service defined for %s", type_)` is actually the correct form (see https://docs.python.org/3/howto/logging.html#optimization)

pi-sigma: I thought `logger.warning("no service defined for {type}", type=type_)` was equivalent to `logger.warning("no service defined for %s", type_)`; perhaps that's not the case.

Viicos: Yeah it's a bit misleading because the doc statement `These newer formatting options _are_ supported` only applies format options, e.g. `%(levelname)s:%(message)s` vs `{levelname}:{message}', not to the actual log messages passed to the logging methods.

Viicos: > I thought `logger.warning("no service defined for {type}", type=type_)` was equivalent to `logger.warning("no service defined for %s", type_)`; perhaps that's not the case.

log methods only allow pos. args (except for `exc_info` and similar). I also think that apart from performance (which can be meaningless is most cases), it allows easier grouping of log messages in tools like Sentry

Bartvaderkin: Are we doing log aggregation in tools like Sentry in this project? I don't think the existing logs are setup for that so that would be introducing something new.

Viicos: But it's true that with deferred string interpolation, it would only error at runtime if there's a mismatch in the number of arguments. I think linting rules exist to warn on this issue though

src/open_inwoner/openklant/clients.py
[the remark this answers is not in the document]
stevenbal: This seems like something we should do in one go indeed

src/open_inwoner/openklant/clients.py
stevenbal: > Are we doing log aggregation in tools like Sentry in this project? I don't think the existing logs are setup for that so that would be introducing something new.

We do have Sentry set up for test/acceptance/prod and if `logger.warning(f"no service defined for {type_}")` gets triggered for two different `type_`s, it will create two separate issues in Sentry, whereas `logger.warning(f"no service defined for %s", type_)` will group them together

src/open_inwoner/utils/api.py
stevenbal: I discussed this with Sergei and Open Formulieren doesn't use this pattern, so he said he won't put it in zgw-consumers. Additionally we should probably get rid of the `ClientError`s altogether

src/open_inwoner/openzaak/clients.py
stevenbal: I'll limit it to 5 requests for cases (1 by default + 4 for pagination if it exists)

src/open_inwoner/openklant/wrap.py
stevenbal: I'll remove them :+1:

src/open_inwoner/openklant/clients.py
stevenbal: I could collapse it into one function `retrieve_objectcontactmoment`, since `retrieve_objectcontactmomenten_for_object_type` is indeed only used by that function currently. Though we might need this in the future, so that's why I added it initially

src/open_inwoner/openklant/wrap.py
stevenbal: I removed these functions and now use the clients directly. I wanted to switch to using context managers (`with build_client("zaak")`, but in some cases we can't do this currently, because we do not raise errors if we can't build a client but return `None` instead. This is related to the issue that we currently simply show no results if we can't build a client

pi-sigma (approved):

Bartvaderkin (approved):

PR #997, opened 2024-01-30
[#2047] Regression test eSuite case document ordering 

pi-sigma:
Taiga: [#2047](https://taiga.maykinmedia.nl/project/open-inwoner/task/2047)

- Added test to ensure case documents are ordered by date
- Changed mocks to ensure document ordering works with time + offset (e.g. `registratiedatum="2022-12-12T00:00:00+01:00"`)

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_case_detail.py
+++ b/src/open_inwoner/openzaak/tests/test_case_detail.py
@@ -431,7 +431,6 @@ class TestCaseDetailView(AssertRedirectsMixin, ClearCachesMixin, WebTest):
             titel="",
             beschrijving="",
             registratiedatum="2021-01-12",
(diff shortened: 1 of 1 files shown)


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -467,7 +469,9 @@ def setUp(self):
                     "info_id": self.informatie_object_2["uuid"],
                 },
             ),
-            created=datetime.datetime(2021, 2, 12, 0, 0, 0),
+            created=datetime.datetime.strptime(
```
Bartvaderkin, line 472: It is safer to use native method for ISO and not mess with format strings: `datetime.fromisoformat()`

Bartvaderkin (approved):

PR #1002, opened 2024-02-01
[#2056] Add notification choices to necessary-fields form

pi-sigma:
Taiga: [#2056](https://taiga.maykinmedia.nl/project/open-inwoner/task/2056)

- Added the option to disable notifications for DigiD users via the necessary fields form upon registering
- Notifications for case status updates that indicate "action required" cannot be disabled. A disabled fake check box (without correlate in the user model) is displayed
- I avoided the `form` template tag and used the more specific `autorender_field`. This required changes to the CSS for the checkbox component


src/open_inwoner/scss/components/Form/Checkbox.scss
```diff
@@ -1,4 +1,6 @@
 .checkbox {
+  padding-top: 1em;
```
jiromaykin, line 2: Not usually my preferred distance 🙂 and we do not wish to change ALL checkboxes in the project. 
So: do not add margins to a component that needs to be re-used; so: add class 'form' to the form in order to give it display grid - which gives the components inside a 'gap'

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -230,6 +231,48 @@ def test_user_can_modify_only_email_when_digid_and_brp(self, m):
...
+
+        data = {
+            "auth_name": "533458225",
+            "auth_pass": "bar",
+        }
+
+        # post our password to the IDP
+        response = self.app.post(url, data).follow().follow()
+
+        necessary_form = response.forms[4]
```
Bartvaderkin, line 262: Using the form index is a bit risky as a modification elsewhere (eg: moving forms around) can mysteriously break this test. It is safer to use the name attribute so this test is independent from whatever is around the form we're interested in.

Bartvaderkin (approved):
Neat, just a small note for the future.

jiromaykin (changes requested):

jiromaykin (approved):

PR #1003, opened 2024-02-01
[WIP] First attempt to replace 2FA with maykin-2fa

alextreme:
All in all an easy upgrade!

A few things to note:
- Seems to clear the TOTP-devices, will every admin have to re-register their TOTP-device or is a migration path possible?
- Wasn't able to get my yubikey to work with Webauthn yet but this may be due to my Firefox upgrade recently
- Could use some additional texts/explaination (point to Google Authenticator / Microsoft Authenticator, explain what a Webauthn is or point to a resource etc)
- django-admin-index happy surprise menu appears again:


requirements/base.txt
```diff
@@ -202,9 +208,9 @@ django-ordered-model==3.4.3
     #   -r requirements/base.in
     #   django-admin-index
 django-otp==1.1.1
-    # via maykin-django-two-factor-auth
+    # via django-two-factor-auth
```
sergei-maertens, line 211: I'd also bump this one to 1.3.0 for latest and greatest :)

requirements/base.txt
```diff
@@ -381,6 +389,8 @@ packaging==23.0
     # via djangocms-text-ckeditor
 phonenumbers==8.12.33
     # via -r requirements/base.in
```
sergei-maertens, line 391: you should be able to drop this as phonenumberslite is (arbitrarily) choosen in maykin-2fa

requirements/base.txt
```diff
@@ -398,6 +408,10 @@ psycopg2==2.9.9
...
 pycparser==2.20
     # via cffi
+pydantic==1.10.14
+    # via
+    #   django-two-factor-auth
+    #   webauthn
```
sergei-maertens, line 414: if you bump to django-two-factor-auth 1.16.0 you don't pull in this dependency

pi-sigma: We're pulling `django-two-factor-auth` through `maykin-2fa`, which doesn't have a pinned version of the former. Why does this require manual upgrade?

requirements/base.txt
pi-sigma: We have custom validators using `phonenumbers`; sticking with this for now.

requirements/base.txt
sergei-maertens: `./bin/compile_dependencies.sh -P django-two-factor-auth` is the mechanism to bump versions of transitive dependencies

It is not required, but recommended since it uses a more modern version of webauthn + doesn't require `pydantic` which is a common source of dependency resolution problems.

requirements/base.txt
sergei-maertens: we may need to make this a configurable group then in maykin-2fa since that pulls phonenumberslite and they may both install their code to the `phonenumbers` module :thinking: so that could cause weird issues

do you get any warnings from pip if you install everything?

requirements/base.txt
pi-sigma: Didn't think of handling transitive deps separately.

requirements/base.txt
pi-sigma: No warnings on my end

sergei-maertens: sounds good, if you run into issues after all in the end, we'll make the necessary changes in maykin-2fa :+1:

pi-sigma:
We're running a very old version of django-admin-index (1.5.0 vs current 3.1.0), and an old version of django-ordered-model (3.4.3 vs 3.7). The upgrade is not entirely straightforward, as the styling for django-admin-index has been completely revamped. I've created a separate issue for the upgrade ([here](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2093)).

pi-sigma:
@stevenbal Just a heads-up when you're trying this out. Once you've created a TOTP device for a staff/superuser, you can no longer disable 2fa via development settings for that user (while on the ticket branch). That's not expected, but it's not (yet) determined if we should treat this as a bug or just document it and go along with it (after all, this only affects development, and you can easily create and drop superusers as you like). I've created a ticket at maykin-2fa for this ([link](https://github.com/maykinmedia/maykin-2fa/issues/14)).

stevenbal (commented):
Looks good! I ran into some issues while testing OIDC login, but this seems to be unrelated to this PR (https://taiga.maykinmedia.nl/project/open-inwoner/issue/2101)

src/open_inwoner/cms/utils/middleware.py
```diff
@@ -53,9 +53,7 @@ def __call__(self, request):
         return response
 
     def force_disable_toolbar(self, request):
-        if settings.TWO_FACTOR_FORCE_OTP_ADMIN and (
```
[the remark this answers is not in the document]
pi-sigma: The setting was from our `maykin-two-factor-auth` fork. Previously 2fa was not enabled by default. Frankly, the settings for enabling/disabling 2fa in different settings were a bit confusing; good that's being simplified.

I'm now wondering if we still need the middleware at all with the switch to `maykin-2fa`. I've tried disabling it and didn't notice the strange behavior described in the docstring  ("inline editing iframes of django-cms could be accessed..."). Perhaps @Bartvaderkin can weigh in on this.

src/open_inwoner/scss/admin/_admin_theme.scss
```diff
@@ -101,9 +100,13 @@ div.breadcrumbs {
 #content {
   /* adjusting for django-cms page */
   #changelist:not([class~='cms-pagetree-root']) {
-    display: grid;
-    grid-template-columns: 1fr 360px;
+    display: block;
```
stevenbal, line 103: I'm not sure what causes this (or if it is an issue at all, because I know very little about django-cms), but if I log in as an admin, I don't get to see the CMS toolbar if I navigate to the frontend (if I try the same on develop it does show the toolbar):

image

I can only get the toolbar to show if I go to a page in the admin and then click the view button
Screenshot 2024-02-19 at 14-47-10 Open Inwoner Deventer

stevenbal (commented):
Looks good, just have one question regarding CMS toolbar

src/open_inwoner/scss/admin/_admin_theme.scss
pi-sigma: This might be related to the changes to `DropToolbarMiddleware`. I'll look into it.

pi-sigma:
- admin is fixed
- yubikey works
- old TOTP devices work
- once enabled, 2fa cannot be circumvented for dev purposes via dev settings (you need to log in and delete the TOTP device for your user)

Bartvaderkin (approved):
I left a small observation not sure if it is something.

pi-sigma:
@Bartvaderkin Do you mean the clarification about `DropToolbarMiddleware` (I don't see anything else)?

PR #1009, opened 2024-02-08
[#2080] Add new modal with buttons

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2080


src/open_inwoner/components/templates/components/Modal/Modal.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n icon_tags button_tags %}
+
+<div class="modal" id="modal" aria-labelledby="modal__title" aria-describedby="modal__text">
```
pi-sigma, line 3: If `id` is the same as `class`, how can this help to differentiate one modal from others?

src/open_inwoner/js/components/accessibility/help_modal.js
```diff
@@ -11,16 +13,21 @@ class HelpModal {
     event.preventDefault()
     this.helpButton.classList.add('accessibility-header__modal--highlight')
     const modalId = document.getElementById('modal')
```
pi-sigma, line 15: See my previous question: if the `id` is so generic, how is this different from getting the modal via the class and `document.querySelector`?

src/open_inwoner/js/components/confirmation/index.js
```diff
@@ -11,15 +13,21 @@ class Confirmation {
     if (!this.real_submit) {
```
pi-sigma, line 13: `real_submit` is a strange name. Is there any other kind of `submit` field? I know it's not your change, but perhaps consider changing this (unless it makes sense to you and I just misunderstand).

pi-sigma (changes requested):
Minor comments/questions

src/open_inwoner/components/templates/components/Modal/Modal.html
jiromaykin: It will differentiate because this addClass will only happen on click on this very individual node.

src/open_inwoner/js/components/accessibility/help_modal.js
jiromaykin: It will differentiate because this addClass will only happen on click on a very individual node. Click the selectors for each modal and you'll see how they look different, while retaining the same functionality.

pi-sigma:
One more thing I just noticed: when you hover over `Ja, verwijdern`, the button moves a little to the top. The other button does not.

src/open_inwoner/js/components/confirmation/index.js
jiromaykin: Yeah that's not mine 🙂 but I understand why they chose this name: this name is often used when pop-up modals pop up after clicking something on the pages that just LOOKS like a Submit button, but actually isn't - and this modal warns the user about the data they are about to submit/delete - so not until they actually click the confirm button in the modal, does the "real" POST happen. So I think I'd like to keep this name.

pi-sigma: Ah, that makes sense!

src/open_inwoner/templates/pages/plans/create.html
```diff
@@ -50,13 +50,11 @@ <h1 class="h1">{% trans "Start nieuwe samenwerking" %}</h1>
...
                                     <h2 class="h2 modal__title" id="modal__title">{{ plan_template.name }}
-                                        {% button text=_("Sluiten") hide_text=True icon="close" extra_classes="modal__button modal__close-title" %}
+                                        {% button type="button" text=_("Sluiten") hide_text=True icon="close" extra_classes="modal__button modal__close-title" %}
                                     </h2>
                                     <div class="modal__text" id="modal__text">{{ plan_template.string_preview }}</div>
                                     <div class="modal__actions modal__actions--align-right" id="modal__actions">
-                                        {% spaceless %}
-                                            {% button text=_("Sluiten") extra_classes="modal__button modal__close" primary=True %}
-                                        {% endspaceless %}
+                                        <button class="button modal__button modal__close button--primary button-icon--primary" type="button" aria-label="Sluit"><span class="inner-text">Sluiten </span></button>
```
Bartvaderkin, line 59: I note this change uses plain html while the one above it uses the template tag (but I have no opinion on it)

src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -281,13 +281,12 @@ <h2 class="h2 title" id="profile-remove">{% trans "Profiel verwijderen" %}</h2><
...
                 </p>
             </div>
-            {% render_form form=form method="POST" id="delete-form" extra_classes="confirm" spaceless=True data_confirm_title=_("Weet u zeker dat u uw account wilt verwijderen?") data_confirm_text=_("Hiermee worden alleen uw persoonlijke voorkeuren verwijderd. U krijgt dan bijvoorbeeld geen e-mail meer van ons over wijzigingen van uw lopende zaken. Uw persoonsgegevens en uw lopende zaken zelf worden hiermee niet verwijderd.") data_confirm_cancel=_("Nee, annuleren") data_confirm_default=_("Ja, verwijderen") %}
+            {% render_form form=form method="POST" id="delete-form" extra_classes="confirm" spaceless=True data_confirm_title=_("Weet je zeker dat je jouw profiel wilt verwijderen?") data_confirm_text=_("Hiermee worden alleen uw persoonlijke voorkeuren verwijderd. U krijgt dan bijvoorbeeld geen e-mail meer van ons over wijzigingen van uw lopende zaken. Uw persoonsgegevens en uw lopende zaken zelf worden hiermee niet verwijderd.") data_confirm_cancel=_("Nee, annuleren") data_confirm_default=_("Ja, verwijderen") %}
```
Bartvaderkin, line 284: Nice attribute wall :see_no_evil:

src/open_inwoner/js/components/modal/index.js
```diff
@@ -53,13 +63,41 @@ export default class Modal {
     this.text.innerText = text
   }
 
+  setModalIcons(modalIcons) {
+    // Whether the modal-buttons should have icons or not
+    if (modalIcons) {
+      this.node.classList.add('show-modal-icons')
+    }
+  }
```
Bartvaderkin: I note these functions only handle the truthy case (eg: don't remove/unset).

Bartvaderkin (approved):
Just some notes for consideration.

src/open_inwoner/js/components/modal/index.js
jiromaykin: Yeah this had terrible and sad, lingering effects, so I changed that in the latest commit.

src/open_inwoner/templates/pages/plans/create.html
jiromaykin: The one in the top above it is a standard cross icon that needs no visible text so can just use the template - but the one below is where things got complicated: the original Modal was overwriting all buttons _entirely_, instead of just overwriting their inner-most text, which meant it was impossible to give Modal submit-buttons etc. an icon; so that is one of the things this wall of a PR is tackling.

jiromaykin:
> One more thing I just noticed: when you hover over `Ja, verwijdern`, the button moves a little to the top. The other button does not.

@pi-sigma I see, it's the buttons that received :focus where this was happening + I also removed the jumpy motion from the close-cross because I find it annoying + I needed to adjust for the situation where the "session" modals do not overwrite other modals properly (like when a user has not closed a modal, it doesn't get overwritten by the end-session warning). So that's what I changed in the new commit.

src/open_inwoner/scss/components/modal/_modal.scss
```diff
@@ -74,6 +74,14 @@
     @media (min-width: 768px) {
       min-width: 150px;
     }
+
+    &:focus {
+      transform: none;
+
+      &:hover {
+        transform: translateY(-1px);
+      }
```
pi-sigma, line 83: This has a strange effect: the button not only moves to the top but is also contracted (it looses one pixel on the right).

pi-sigma (changes requested):

src/open_inwoner/scss/components/modal/_modal.scss
jiromaykin: Yeah this is Firefox specific (Edge/Safari/Chrome do not have this). We also have this behaviour for ALL types of buttons that are hovered (like in the registration form etc.) so we'll promote this to a separate issue, since it's unrelated to the modals, so we can finalize the modals here :-)

jiromaykin: In case someone wants to see the new separate issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2293

pi-sigma (approved):

PR #1014, opened 2024-02-08
[#2093] Upgrade django admin index

pi-sigma:
Taiga: [#2093](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2093)

Related: recent switch to maykin-2fa ([https://github.com/maykinmedia/open-inwoner/pull/1003](https://github.com/maykinmedia/open-inwoner/pull/1003))


jiromaykin (changes requested):
👍 I like this more! No more bad responsiveness and a clearer difference in hover colors.

It does still show the same jQuery error in console sometimes where it cannot find 'Django/django is not defined' but I guess that's still part of the package, so prob. no need to do anything there.

Just one small thing that is annoying on mobile: is how `#changelist-filter` completely covers the other content (because the surrounding `#changelist` is display:grid on mobile) - you can see this on http://localhost:8000/admin/openzaak/zaaktypeconfig/ and for small screens just set it to display block.

I can't remember where but I saw some pages were checking to see if they could fin the #scroll-anchor/"style display" and shows a console error; which may be solved by adding a condition such as` if (mybutton) {...)` around the function in _src/open_inwoner/js/components/anchor-menu/scroll_to_top.js_

Screenshot 2024-02-12 at 11 59 30

pi-sigma:
@jiromaykin e2e tests are still failing but we need to move forward. Thanks for your suggestions, they're implemented!

src/open_inwoner/scss/admin/_app_overrides.scss
```diff
@@ -8,71 +8,59 @@
...
+.djai-dropdown-menu {
+  // use full width on mobile & tablet
+  width: 100%;
 
-  .dropdown-menu__item,
-  .dropdown-menu__item:link,
-  .dropdown-menu__item:visited {
-    color: $color_lightest;
+  /* Added to attach to body/breadcrumbs */
+  border-bottom: $djai-border-width solid $color_secondary;
```
jiromaykin, line 30: Not sure if this is the culprit here - but I see an odd border on the bottom here that makes the header/menu jump out strangley.
Do you also see this? I've added the inspector: if you toggle the border off, it looks normal.

adminmenu1

Clearer view on left-side only (but it's also happening on the right-hand side):

adminmenu2

jiromaykin (changes requested):
A tiny change needed, IF I am not the only one seeing this :-)

src/open_inwoner/scss/admin/_app_overrides.scss
pi-sigma: Good catch! I removed the border.

jiromaykin (approved):
I like it.

PR #1016, opened 2024-02-09
[#2069] Upgrade KvK Zoeken API to v2

pi-sigma:
Taiga: [#2069](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2069)

- [x] Add validation to API config to ensure that the API root does not contain the version number and/or endpoint
- [x] Add tests to ensure that API urls are constructed propery from the root
- [x] Update existing tests and fixtures to use v2-compliant return values
- [x] Refactor mocks to align better with OIP style
- [ ] Add data migration


src/open_inwoner/kvk/models.py
```diff
@@ -6,12 +6,18 @@
 from simple_certmanager.models import Certificate
 from solo.models import SingletonModel
 
+from .validators import validate_api_root
+
 
 class KvKConfig(SingletonModel):
-    api_root = models.URLField(
+    api_root = models.CharField(
```
stevenbal, line 13: Why was this changed to a `CharField`?

src/open_inwoner/kvk/validators.py
```diff
@@ -0,0 +1,22 @@
...
+from django.core.exceptions import ValidationError
+from django.utils.translation import gettext_lazy as _
+
+# There are only 4 possibilities, let's enumerate them
+regexes = (
+    re.compile(r"^https://api.kvk.nl/api$"),
+    re.compile(r"^https://api.kvk.nl/api/$"),
+    re.compile(r"^https://api.kvk.nl/test/api$"),
+    re.compile(r"^https://api.kvk.nl/test/api/$"),
+)
```
stevenbal, line 12: If we're going to restrict the API root to these values, I think it's better to define `choices` for the field and let people select one of these four.

I'm trying to think of a case where we'd want to use something else but can't really come up with one, is it necessary though to restrict the API root to these four values? It would be a hassle if we have to do a new release just to allow a different value here :thinking:

stevenbal (changes requested):

src/open_inwoner/kvk/validators.py
pi-sigma: I made the validation more general. I couldn't think of any reason for supporting other roots, but better safe than sorry.

requirements/base.txt
```diff
@@ -62,7 +62,9 @@ cryptography==41.0.7
     #   mozilla-django-oidc
     #   pyopenssl
 css-inline==0.13.0
-    # via -r requirements/base.in
+    # via
+    #   -r requirements/base.in
+    #   mail-editor
```
stevenbal, line 67: This is probably from another PR? Doesn't actually change anything though, so it's fine

src/open_inwoner/kvk/tests/mocks.py
```diff
@@ -24,61 +24,103 @@
...
+
+multiple_branches = {
+    "pagina": 1,
+    "resultatenPerPagina": 10,
     "totaal": 3,
     "resultaten": [
         {
             "kvkNummer": "68750110",
-            "handelsnaam": "Test BV Donald",
+            "naam": "Test BV Donald",
```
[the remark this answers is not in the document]
stevenbal: I do see that `mocks.simple` still uses `handelsnaam` https://github.com/maykinmedia/open-inwoner/blob/upgrade/2069-kvk/src/open_inwoner/kvk/tests/mocks.py#L15 (same thing for `aantal` instead of `resultatenPerPagina`)

Let's also use the v2 schema here to be consistent

stevenbal (changes requested):
Minor remark regarding `mocks.simple`. Looks good otherwise :+1: 

It might be useful for us at some point in the future to look into adding VCR.py (https://vcrpy.readthedocs.io/en/latest/), to deal with mocks and to make them more realistic instead of having to maintain them manually. Open Forms uses this as well

requirements/base.txt
pi-sigma: Yes I rebased on [https://github.com/maykinmedia/open-inwoner/pull/1014](https://github.com/maykinmedia/open-inwoner/pull/1014) because the two are closely related.

Didn't know about VCR. Looks nice and I'd be in favour of adopting it, especially seeing how I missed `mocks.simple`.

stevenbal (approved):

requirements/base.txt
pi-sigma: Double-checked: I branched off `upgrade/django-admin-index` instead of develop when starting on this. The PR has been merged though.

PR #1019, opened 2024-02-12
:art: Apply flake8 for code style

stevenbal:
Better late than never :)


alextreme:
I'm not really in favor about style enforcing with flake8, I don't think catching unused variables and ordering imports improves a codebase sufficiently that it's worthwhile to mandate it via CI. I think black is sufficient.

But I'm also not about enforcing my preferences, so if @Bartvaderkin and @pi-sigma are in favor we can merge this in

pi-sigma:
@alextreme I'm down with this. Looking at the files changed, there are quite a few superfluous variable assignments and unused imports being cleaned up, f-strings without anything interpolated being converted to regular strings etc.

I don't think this is about style. Whether lines should be 80 or 100 chars long is a matter of style and preference, but unused imports and variable assignments are just cluttering the namespace; no one would prefer to keep them if it was pointed out.

I'm assuming of course that flake8 mandates are things that everyone generally agrees on, like in the examples I've mentioned. Perhaps I'm overlooking something.

src/open_inwoner/search/resources/__init__.py
```diff
@@ -1 +1 @@
-from .resources import SynonymResource
+from .resources import SynonymResource  # noqa
```
pi-sigma, line 1: You can avoid the error from flake8 without `noqua` by explicitly declaring the class as part of the module interface:

`__all__ = [SynonymResource]`

Similarly for the othe imports in `__init__` files.

See the relevant section in [PEP 8](https://peps.python.org/pep-0008/#public-and-internal-interfaces)

pi-sigma (changes requested):

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -742,14 +742,18 @@ def test_list_cases_logs_displayed_case_ids(self, m):
         logs = list(TimelineLog.objects.all())
 
         case_log = [
-            l for l in logs if self.zaak1["identificatie"] in l.extra_data["message"]
+            log
+            for log in logs
+            if self.zaak1["identificatie"] in log.extra_data["message"]
```
Bartvaderkin, line 747: Maybe not the best example but I want to be able to use single letter variables in cases like this.

src/open_inwoner/pdc/managers.py
```diff
@@ -114,7 +114,7 @@ def filter_by_zaken(self, cases: List[Zaak]):
         pks = []
         for category in qs:
             for identificatie in category.zaaktypen:
-                if not identificatie in months_since_last_zaak_per_zaaktype:
```
Bartvaderkin, line 117: Bug?

src/open_inwoner/pdc/models/product.py
```diff
@@ -232,7 +230,7 @@ def get_absolute_url(self, category=None):
         )
 
     def has_cta_tag(self):
-        return "\[CTABUTTON\]" in self.content
+        return "\[CTABUTTON\]" in self.content  # noqa
```
Bartvaderkin, line 233: Why is this `noqa`?

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -876,7 +876,7 @@ def test_plan_list_filters_status_open(self):
 
     @freeze_time("2022-01-01")
     def test_plan_list_filters_status_closed(self):
-        open_plan = PlanFactory(
+        PlanFactory(
```
Bartvaderkin, line 879: Here information is lost by removing the unused variables.

src/open_inwoner/accounts/query.py
```diff
@@ -1,9 +1,9 @@
-from django.db.models import Case, F, Max, OuterRef, Q, Subquery, Value, When
+from django.db.models import Case, F, Max, OuterRef, Q, Subquery, When
 from django.db.models.query import QuerySet
 
 
 class MessageQuerySet(QuerySet):
-    def get_conversations_for_user(self, user: "User") -> "MessageQuerySet":
+    def get_conversations_for_user(self, user: "User") -> "MessageQuerySet":  # noqa
```
Bartvaderkin, line 6: Why is this `noqa`?

src/open_inwoner/accounts/query.py
```diff
@@ -100,7 +100,9 @@ def mark_seen(self, user, other_user) -> int:
 
 
 class InviteQuerySet(QuerySet):
-    def get_pending_invitations_for_user(self, user: "User") -> "InviteQuerySet":
+    def get_pending_invitations_for_user(
+        self, user: "User"  # noqa
```
Bartvaderkin, line 104: Another weird `noqa`?

src/open_inwoner/accounts/tests/test_message_query.py
```diff
@@ -19,7 +19,7 @@ def test_no_other_user(self):
 
     def test_1_other_user_received_last(self):
         other = UserFactory.create()
-        sent_message = MessageFactory.create(sender=self.user, receiver=other)
+        MessageFactory.create(sender=self.user, receiver=other)
```
Bartvaderkin, line 22: Information loss by removing the unused variables (also next change).

Bartvaderkin (changes requested):
This would be good to add, especially for the unused imports.

The amount of noqa's seems a bit much, and there a some that feel weird (I marked some)

Removing the unused variable names as hard rule is debatable, especially in tests they can help a lot showing what certain factories are for (I marked some examples).

Also I want to keep the option to use single letter variables, sometimes they are more readable then breaking the whole thing up in a black mess, so let's leave that up to code review instead of a dogmatic rule.

stevenbal:
@Bartvaderkin I agree, the main reason I added this was for the imports. I could add F841 (unused variables) and E741 (ambiguous/short variable names) to the ignore list.

src/open_inwoner/accounts/query.py
stevenbal: `src/open_inwoner/accounts/query.py:6:48: F821 undefined name 'User'`

I suppose I could also use `get_user_model` instead

src/open_inwoner/accounts/query.py
stevenbal: Same as the other one: src/open_inwoner/accounts/query.py:6:48: F821 undefined name 'User'

src/open_inwoner/pdc/models/product.py
stevenbal: Due to `src/open_inwoner/pdc/tests/test_product.py:271:46: W605 invalid escape sequence '\]'`

src/open_inwoner/pdc/managers.py
stevenbal: This is due to https://www.flake8rules.com/rules/E713.html. `if not var in list` could be weird to read if var is a boolean

src/open_inwoner/search/resources/__init__.py
stevenbal: I'll change this :+1:

src/open_inwoner/pdc/managers.py
Bartvaderkin: I thought the original was also not functional the same but apparently it is.

src/open_inwoner/pdc/models/product.py
Bartvaderkin: I wonder what this is supposed to be: does it even need the backslashes or was this a remnant of a regex?

src/open_inwoner/accounts/query.py
Bartvaderkin: We can just import User from .account? The pluggable user model is not really a thing in the project.

src/open_inwoner/accounts/query.py
Bartvaderkin: Ugh, even when importing in `if TYPE_CHECKING:`?

stevenbal: Using that works indeed :+1:

Bartvaderkin: Cool!

src/open_inwoner/pdc/models/product.py
stevenbal: I'm not sure, but I didn't want to change the code without knowing whether it's still necessary

Bartvaderkin:
@stevenbal Let's allow the short variables. 

I'm ambivalent about the unused variables rule but we if use it then let's convert the useful names from the tests into a comment.

stevenbal:
I added the unused variables and short variables rules to the ignorelist

Bartvaderkin (approved):
Cool, cleans up a lot.

Maybe pick a tactical time to merge as it will be a bit of a hassle for the open PR's and everybody needs to update their githooks.

alextreme (approved):
Can be merged, preferably next monday / early next week

pi-sigma (approved):

PR #1020, opened 2024-02-12
[#1808] Use long month names consistently in ssd templates

pi-sigma:
Taiga: [#1808](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1808)

- set empty title in PDF templates to prevent duplication of report name in tab
- use long month name in case no reports are found

### DIFF
--- a/src/open_inwoner/ssd/tests/test_views.py
+++ b/src/open_inwoner/ssd/tests/test_views.py
@@ -87,7 +87,7 @@ class TestMonthlyBenefitsFormView(TestCase):
         self.assertEqual(response.status_code, HTTPStatus.OK)
 
         self.assertContains(
-            response, "Geen uitkeringsspecificatie gevonden voor Dec 1985"
+            response, "Geen uitkeringsspecificatie gevonden voor december 1985"
         )
 
     @patch(
--- a/src/open_inwoner/ssd/templates/maandspecificatie.html
+++ b/src/open_inwoner/ssd/templates/maandspecificatie.html
@@ -156,7 +156,7 @@
         <div class="section extra-margin">
             <p class="monthly-report__p extra-margin">Op de 25e van de maand krijgt u het bedrag ad. &euro; {% format_currency last.bedrag.waarde_bedrag %} uitbetaald.</p>
         </div>
-        {% endif %}  
+        {% endif %}
 	</div>
 	</body>
 	<p style="page-break-after: always;">&nbsp;</p>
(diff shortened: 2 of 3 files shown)


src/open_inwoner/ssd/templates/jaaropgave.html
```diff
@@ -2,6 +2,9 @@
 
 <!DOCTYPE html>
 <html lang="nl">
+<head>
+  <title></title>
+</head>
```
stevenbal, line 7: Why was this added?

src/open_inwoner/ssd/tests/test_views.py
```diff
@@ -87,7 +87,7 @@ def test_uitkering_post_fail(self, mock_report):
         self.assertEqual(response.status_code, HTTPStatus.OK)
 
         self.assertContains(
-            response, "Geen uitkeringsspecificatie gevonden voor Dec 1985"
+            response, "Geen uitkeringsspecificatie gevonden voor december 1985"
```
stevenbal, line 90: In the taiga issue I see that the title of the `Maandspecificatie` pdf is what should be fixed, so I would expect that to be addressed too?

stevenbal (changes requested):

src/open_inwoner/ssd/tests/test_views.py
pi-sigma: This was resolved last year, see [https://github.com/maykinmedia/open-inwoner/blob/fix/1808-maandspecificatie-header/src/open_inwoner/ssd/client.py#L192](https://github.com/maykinmedia/open-inwoner/blob/fix/1808-maandspecificatie-header/src/open_inwoner/ssd/client.py#L192)

src/open_inwoner/ssd/templates/jaaropgave.html
pi-sigma: The title is duplicated in case a new tab is opened for file download (at least on Firefox). Apparently it is not (yet) standardized what browsers do with the title. The `<title><title>` effectively re-sets whatever the browser does with the title, afterwards we define our custom title. It's a hack, but I don't know a better way to deal with this. Perhaps @jiromaykin has some idea about this. I did however remove the extra `head`, which was an oversight.

pi-sigma: Discussed with Jiro, I'll revert changes and create a separate issue for this because it's an accessibility issue.

stevenbal:
@pi-sigma is this PR ready for re-review?

pi-sigma:
@ Steven Yes, this one is ready, forgot to re-request.

PR #1021, opened 2024-02-12
:wrench: [#2101] Add OIDC admin config to admin index fixture

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2101

Fixes two issues:
- If an OIDC user logs in for the first time, they are now actually authenticated (previously they had to log in twice)
- If OIDC is enabled for admins, the `is_staff` attribute is set for those users (according to the OIDC config) and they are redirected to the admin interface

### DIFF
--- a/src/open_inwoner/accounts/tests/test_oidc_views.py
+++ b/src/open_inwoner/accounts/tests/test_oidc_views.py
@@ -19,6 +19,8 @@ from digid_eherkenning_oidc_generics.views import (
     GENERIC_DIGID_ERROR_MSG,
     GENERIC_EHERKENNING_ERROR_MSG,
 )
+from open_inwoner.configurations.choices import OpenIDDisplayChoices
+from open_inwoner.configurations.models import SiteConfiguration
 from open_inwoner.kvk.branches import KVK_BRANCH_SESSION_VARIABLE
 
 from ..choices import LoginTypeChoices
@@ -34,10 +36,15 @@ class OIDCFlowTests(TestCase):
     @patch("mozilla_django_oidc_db.backends.OIDCAuthenticationBackend.get_token")
     @patch(
         "mozilla_django_oidc_db.mixins.OpenIDConnectConfig.get_solo",
-        return_value=OpenIDConnectConfig(id=1, enabled=True),
+        return_value=OpenIDConnectConfig(id=1, enabled=True, make_users_staff=True),
     )
-    def test_existing_email_updates_user(
+    @patch(
+        "open_inwoner.configurations.models.SiteConfiguration.get_solo",
+        return_value=SiteConfiguration(id=1, openid_display=OpenIDDisplayChoices.admin),
+    )
+    def test_existing_email_updates_admin_user(
         self,
+        mock_config_get_solo,
         mock_get_solo,
         mock_get_token,
         mock_verify_token,
@@ -62,11 +69,12 @@ class OIDCFlowTests(TestCase):
             callback_url, {"code": "mock", "state": "mock"}
         )
 
+        self.assertRedirects(
+            callback_response, reverse("admin:index"), fetch_redirect_response=True
+        )
+
         user.refresh_from_db()
 
-        self.assertRedirects(
-            callback_response, reverse("pages-root"), fetch_redirect_response=False
-        )
         self.assertTrue(User.objects.filter(oidc_id="some_username").exists())
         self.assertEqual(user.oidc_id, "some_username")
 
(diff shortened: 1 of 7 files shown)


src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -62,18 +69,134 @@ def test_existing_email_updates_user(
...
+        self.assertRedirects(
+            callback_response, reverse("admin:index"), fetch_redirect_response=True
+        )
+
         user.refresh_from_db()
 
+        self.assertTrue(User.objects.filter(oidc_id="some_username").exists())
+        self.assertEqual(user.oidc_id, "some_username")
+
+        db_user = User.objects.filter(oidc_id="some_username").first()
```
Bartvaderkin, line 81: In general tests are be harder if instead of .filter()/.first() (or .exists()) for single records to just use a .get() so it will also communicate and test the record must exists and cannot be multiple.

Also the test itself seems weird? I don't quite get what it tries to prove: isn't `user` the same record as `db_user`?

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -62,18 +69,134 @@ def test_existing_email_updates_user(
...
+        user.refresh_from_db()
+
+        self.assertTrue(User.objects.filter(oidc_id="some_username").exists())
+        self.assertEqual(user.oidc_id, "some_username")
+
+        db_user = User.objects.filter(oidc_id="some_username").first()
+
+        self.assertEqual(db_user.id, user.id)
+        self.assertEqual(db_user.login_type, LoginTypeChoices.oidc)
+        self.assertEqual(db_user.is_staff, False)
```
Bartvaderkin, line 141: Same as above, not quite clear what is tested here.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -62,18 +69,134 @@ def test_existing_email_updates_user(
...
+        user.refresh_from_db()
+
         self.assertTrue(User.objects.filter(oidc_id="some_username").exists())
         self.assertEqual(user.oidc_id, "some_username")
 
         db_user = User.objects.filter(oidc_id="some_username").first()
 
         self.assertEqual(db_user.id, user.id)
         self.assertEqual(db_user.login_type, LoginTypeChoices.oidc)
+        self.assertEqual(db_user.is_staff, False)
```
Bartvaderkin, line 199: Same as above.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -159,13 +295,67 @@ def test_new_user_is_created_when_new_email(
         )
 
         self.assertRedirects(
-            callback_response, reverse("pages-root"), fetch_redirect_response=False
+            callback_response, reverse("admin:index"), fetch_redirect_response=True
+        )
+
+        new_user = User.objects.filter(email="new_user@example.com").first()
```
Bartvaderkin, line 301: Here is is more clear as `new_user` and the test title work together. 

Nevertheless the filter()/first() is a bit floppy and should just be a .get() (note there is even a `assertIsNotNone(new_user)` there to work around .first()  :frowning_face:

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -62,18 +69,134 @@ def test_existing_email_updates_user(
...
+
         self.assertRedirects(
-            callback_response, reverse("pages-root"), fetch_redirect_response=False
+            callback_response, reverse("pages-root"), fetch_redirect_response=True
+        )
+
+        user.refresh_from_db()
+
+        self.assertTrue(User.objects.filter(oidc_id="some_username").exists())
+        self.assertEqual(user.oidc_id, "some_username")
```
Bartvaderkin, line 135: The filter/exists line tests the same thing as the `user.refresh_from_db()` + `self.assertEqual(user.oidc_id, "some_username")`

Bartvaderkin (changes requested):
I put some notes, although it is mostly the same feedback about filter/first being sub-optimal for tests, and showing intent.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -34,10 +36,15 @@ class OIDCFlowTests(TestCase):
     @patch("mozilla_django_oidc_db.backends.OIDCAuthenticationBackend.get_token")
     @patch(
         "mozilla_django_oidc_db.mixins.OpenIDConnectConfig.get_solo",
-        return_value=OpenIDConnectConfig(id=1, enabled=True),
+        return_value=OpenIDConnectConfig(id=1, enabled=True, make_users_staff=True),
```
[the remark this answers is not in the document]
stevenbal: It does appear to be necessary in some cases, when the user groups are updated:

```
default_groups = self.config.default_groups.all()
...
ValueError: "<OpenIDConnectConfig: OpenID Connect configuration>" needs to have a value for field "id" before this many-to-many relationship can be used.
```

src/open_inwoner/accounts/tests/test_oidc_views.py
stevenbal: You're right, the tests were a bit messy already and I was adding stuff around that :sweat_smile:.

I cleaned it up, the test is supposed to verify that an existing admin user (that is not yet OIDC), is updated if a user with the same email logs in with OIDC

src/open_inwoner/accounts/tests/test_oidc_views.py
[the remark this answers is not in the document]
Bartvaderkin: Don't worry too much, it was just an observation. The error makes sense if something in the test needs a saved instance, like we have some ForeignKey's to solo objects so we can use inline-admins.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -159,13 +285,63 @@ def test_new_user_is_created_when_new_email(
...
+            "email": "new_user@example.com",
+            "sub": "some_username",
+        }
+        UserFactory.create(email="existing_user@example.com")
+        session = self.client.session
+        session["oidc_states"] = {"mock": {"nonce": "nonce"}}
+        session.save()
+        callback_url = reverse("oidc_authentication_callback")
+
+        self.assertFalse(User.objects.filter(email="new_user@example.com").exists())
```
Bartvaderkin, line 329: I personally feel this line is not strictly necessary, I'd assume we could trust the database was cleared. Although the line clearly shows our expectations we don't really do that in general.

Bartvaderkin (approved):
Cool. I left a small note but it is not blocking.

src/open_inwoner/accounts/tests/test_oidc_views.py
stevenbal: You're right, it does indeed indicate that previously we didn't have isolated tests, so I'll remove it as we're moving away from using `setUpTestData`

PR #999, opened 2024-01-30
[#2059] Add emailnotification templates

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2059

PR #1024, opened 2024-02-13
[#939] Upgrade to Django 4.2

### LINKED ISSUE #939: Django 4.2 upgrade issues
There are incompatible versions in the resolved dependencies:
  - [x] Django>=4.2<5.0 (from -r requirements/base.in (line 16))
  - [x] django>=2.2 (from django-localflavor==3.1->-r requirements/base.in (line 28)) **localflavor 4.0 should support django 4.2**
  - [ ] django>=2.2 (from djangorestframework==3.12.4->-r requirements/base.in (line 68)) **version 3.14.x supports 4.1, no 4.2 support yet**
  - [x] django>=3.2 (from django-log-outgoing-requests==0.5.2->-r requirements/base.in (line 52)) **should be fine?**
  - [ ] django>=3.2 (from django-open-forms-client==0.2.3->-r requirements/base.in (line 47))
  - [ ] django>=3.2 (from django-privates==1.5.0->-r requirements/base.in (line 29))
  - [ ] django>=2.2 (from django-axes==5.25.0->-r requirements/base.in (line 18))
  - [ ] Django<5.0,>=2.2 (from django-cms==3.11.1->-r requirements/base.in (line 56))
  - [ ] Django>=2.2 (from drf-spectacular==0.20.1->-r requirements/base.in (line 71))
  - [ ] Django>=1.8 (from mail-editor->-r requirements/base.in (line 41))
  - [ ] Django>=2.2 (from django-cors-headers==3.10.0->-r requirements/base.in (line 30))
  - [ ] django>=2.2 (from django-timeline-logger==2.0.0->-r requirements/base.in (line 43))
  - [ ] Django>=1.8 (from django-csp==3.7->-r requirements/base.in (line 44))
  - [ ] django<5.0,>=2.2 (from django-csp-reports==1.8.1->-r requirements/base.in (line 45))
  - [ ] django>=3.2.0 (from django-digid-eherkenning==0.9.0->-r requirements/base.in (line 82))
  - [ ] Django>=2.2 (from django-extra-fields==3.0.2->-r requirements/base.in (line 69))
  - [ ] Django!=3.0.*,>=2.2 (from django-registration==3.2->-r requirements/base.in (line 36))
  - [ ] django>=3.2.0 (from zgw-consumers==0.25.0->-r requirements/base.in (line 88))
  - [ ] django<5.1,>=3.2 (from django-filer==3.1.1->-r requirements/base.in (line 19))
  - [ ] Django<4.0.0,>=2.2.0 (from maykin-django-two-factor-auth==2.0.4->-r requirements/base.in (line 26))
  - [ ] Django>=2.0 (from django-rosetta==0.9.7->-r requirements/base.in (line 23))
  - [ ] Django>=2.2 (from django-filter==21.1->-r requirements/base.in (line 70))
  - [ ] Django>=2.2 (from django-redis==5.0.0->-r requirements/base.in (line 22))
  - [ ] django>=2.2 (from django-hijack==3.0.4->-r requirements/base.in (line 20))
  - [ ] Django (from django-view-breadcrumbs==2.2.4->-r requirements/base.in (line 37))
  - [ ] Django>=3.2 (from django-htmx==1.13.0->-r requirements/base.in (line 48))
  - [ ] django>=3.1.0 (from mozilla-django-oidc-db==0.12.0->-r requirements/base.in (line 46))
  - [ ] Django>=2.2 (from django-sniplates==0.7.0->-r requirements/base.in (line 24))
  - [ ] django>=3.2.0 (from notifications-api-common==0.2.0->-r requirements/base.in (line 89))
  - [ ] Django>=2.2 (from django-import-export==2.6.1->-r requirements/base.in (line 33))
  - [ ] Django>=2.2 (from django-treebeard==4.4->-r requirements/base.in (line 25))

Issues:
- maykin-django-two-factor-auth

pi-sigma:
Closes #939

TODO:

- [x] Complete switch to maykin-2fa ([https://github.com/maykinmedia/open-inwoner/pull/1003](https://github.com/maykinmedia/open-inwoner/pull/1003))
- [x] Complete upgrade of django-admin-index ([https://github.com/maykinmedia/open-inwoner/pull/1003](https://github.com/maykinmedia/open-inwoner/pull/1003))
- [x] Upgrade mail-editor
- [x] Fix logging-related tests (`TimelineLog` objects cannot be created to to unsaved related object) 
- [x] Fix remaining tests


pi-sigma:
- I changed the logging in connection with digid user creation/change because the attempt to create a `TimelineLog` as part of the `on_bsn_change` function gave a `ValueError`: the related object (i.e. the user) is not saved. Perhaps we can revert to database logging (if this is preferred) by making the signal `pre_save` instead of `post_save`. Needs some thought/discussion.
- A few failing tests related to the contents of emails, which previously passed locally. Perhaps related to the recent email template changes ([https://github.com/maykinmedia/open-inwoner/pull/999](https://github.com/maykinmedia/open-inwoner/pull/999))? Alternatively, something not working with the mail-editor version pinned in `base.in`?

src/open_inwoner/accounts/views/inbox.py
```diff
@@ -122,12 +122,10 @@ def get_status(self, messages: MessageQuerySet) -> str:
...
         Returns the status string of the conversation.
         """
-        try:
-            return f"{_('Laatste bericht ontvangen op')} {formats.date_format(messages[-1].created_on)}"
-        except IndexError:
-            return ""
-        except AssertionError:
+        messages = list(messages)
+        if not messages:
             return ""
```
Bartvaderkin, line 127: Legacy code fun :crying_cat_face:

src/open_inwoner/configurations/tests/test_oidc.py
```diff
@@ -22,7 +24,7 @@ def setUpTestData(cls):
         openid_config.enabled = True
         openid_config.save()
 
-    def test_admin_only_enabled(self):
+    def test_admin_only_enlabled(self):
```
Bartvaderkin, line 27: *`enlabled`.. does the admin have the right label? :D

src/open_inwoner/configurations/tests/test_oidc.py
```diff
@@ -39,7 +41,9 @@ def test_admin_only_enabled(self):
 
         oidc_login_option = response.pyquery.find(".admin-login-option")
 
-        self.assertEqual(oidc_login_option.text(), _("Login with organization account"))
+        self.assertEqual(
+            oidc_login_option.text(), _("Log in met een organisatieaccount")
```
Bartvaderkin, line 45: I notice this translation changed but I don't know if that is something.

src/open_inwoner/haalcentraal/signals.py
```diff
@@ -19,8 +18,6 @@ def on_bsn_change(instance, **kwargs):
         and instance.is_prepopulated is False
         and instance.login_type == LoginTypeChoices.digid
     ):
-        system_action(
-            "Retrieving data from haal centraal based on BSN", content_object=instance
-        )
+        logger.info("Retrieving data for %s from haal centraal based on BSN", instance)
```
Bartvaderkin, line 21: I noticed the original created a TimelineLog while the change is a regular logger.

src/open_inwoner/haalcentraal/utils.py
```diff
@@ -41,4 +39,4 @@ def update_brp_data_in_db(user, initial=True):
     if initial is False:
         user.save()
 
-    system_action(_("data was retrieved from haal centraal"), content_object=user)
+    logger.info("Retrieving data for %s from haal centraal based on BSN", user)
```
Bartvaderkin, line 42: Another one logger that was a timelinelog.

src/open_inwoner/configurations/tests/test_oidc.py
pi-sigma: :) Typo from the 2fa PR this was rebased on

src/open_inwoner/haalcentraal/utils.py
pi-sigma: See previous comment

Bartvaderkin (commented):
I added some notes.

Also I find the many migrations suspicious, I suppose this is a side-effect of the strings and newer Django code?

src/open_inwoner/haalcentraal/signals.py
Bartvaderkin: I don't know but I think it could be important to have a TimelineLog entry for this functionality. 

Maybe the code can be re-arranged to work? I don't know if post-save would work because we modify the same instance. 

And at some point we should centralize/refactor the User signals (now or in the future): there are a few save signals floating around doing this kind of thing and it is hard to follow.

requirements/dev.txt
```diff
@@ -453,15 +466,20 @@ django-treebeard==4.4
     #   -r requirements/ci.txt
     #   django-cms
     #   djangocms-attributes-field
-django-view-breadcrumbs==2.2.4
+django-two-factor-auth[phonenumberslite,webauthn]==1.16.0
```
jiromaykin, line 469: For me locally I was unable to install this (all the other deps seem fine) and my terminal reports: 
`zsh: no matches found: django-two-factor-auth[phonenumberslite,webauthn]==1.16.0
`

jiromaykin: Seems like this ⬇️ solved it for me (since I do not want to disabled 'globbing' systemwide): "...if you need to pass literal square brackets as an argument to a command, you either need to escape them or quote the argument"

src/open_inwoner/haalcentraal/signals.py
pi-sigma: Turns out it works after all. The relevant tests were failing at some point during the upgrade but are passing now.

pi-sigma: @Bartvaderkin Well that was too quick. I can get the haacentraal tests to pass, but not the DigiD tests. In fact, the combination of registration/login flow and logging itself is broken in its current form. When you try to register with a new BSN, the attempt to create a timeline log results in an error. I'll revert to the previous state (replace database logs with regular logging), however feel free to take a look.

pi-sigma:
@Bartvaderkin 
> Also I find the many migrations suspicious, I suppose this is a side-effect of the strings and newer Django code?

A few of the migrations (auto-generated of cource) are cms-related changes to the `related_name` field. Not sure myself what that's about.

src/open_inwoner/haalcentraal/signals.py
Bartvaderkin: > However, I had to change to asserted log count in one of the tests

Practical learning experience :wink: 

If you want to test TimelineLog's it work better to not rely on ad-hoc queries and counts on TimelineLog model but check if the exact log(s) you expect is there.

Lot's of signals and things generate logs so the count in the tests is not reliable, because you don't know what other unexpected logs it also counts. And it obviously breaks if some other feature that logs is added and the count changes (bonus fun if a change drops one and adds another). 

This is why we have the AssertTimelineLogMixin that looks for the exact text and parameters. And yes there are old tests that do counts and it is an unreliable mess as you're now experiencing; we have no idea if they even tested correctly or not, and there are mystery counts floating around.

Anyway, if it had database logs before we'd expect them to remain after a refactor. We shouldn't just decide to not do that because it is inconvenient.

The idea of the TimelineLog is that we have a timeline in the admin and with some traceability and triage options if we have issues or complaints. It is not the same if we have to grab extra from some random log location. 

I would update the old tests and get rid of count based testing and do exact checks to verify the old tests were even correct at all, and move from there.

Bartvaderkin (changes requested):
The Django update stuff looks good but there is the signals logging, I put some input to clarify.

src/open_inwoner/accounts/views/inbox.py
```diff
@@ -105,7 +105,7 @@ def get_other_user(self, conversations: dict) -> Optional[User]:
 
         return get_object_or_404(User, uuid=other_user_uuid)
 
-    def get_messages(self, other_user: User) -> MessageQuerySet:
```
stevenbal, line 108: I'm not sure why `List` was added here, it just returns a queryset right?

stevenbal (changes requested):
Just some minor remarks/questions

src/open_inwoner/accounts/views/inbox.py
pi-sigma: Slicing a queryset results in another queryset, but slicing with step (`queryset[1:10:2]`) results in a list. Who would have thought this might lead to confusion and bugs :)

src/open_inwoner/haalcentraal/signals.py
pi-sigma: It makes sense to clean up the logging tests that do rely on order and counts.

However, the test for haalcentraal signals does not fail because it relies on "ad-hoc queries and counts on TimelineLog model"; there are simply no logs created. I don't understand why/how the test could previously pass, since the function is triggered by a `pre-save` signal, but the log object to be created is related to the user object, which hasn't been saved at the time of logging.

jiromaykin (commented):
I'm not quite sure how to review this - at least a fun thing to know is that the admin dashboard looks better in dark-mode now 👍
We do need to change the position of the top of the page in the mail-editor (which doesn't have the main nav menu) but that is outside of the scope of this PR.

src/open_inwoner/accounts/views/inbox.py
stevenbal: But it would then be a list of messages, not a list of querysets right?

stevenbal (approved):
Looks good to me, just the one comment about the typing, but that's not crucial.

It does need to be rebased due to merge conflicts in requirements unfortunately

Bartvaderkin:
@pi-sigma I added a PR #1062 on this PR with a fix for the Haalcentraal timelinelog issue.

Can you check if that works? And if there is anything else we to cover so we can merge this soon.

Bartvaderkin (approved):
:+1: lets go! :rocket:

PR #1028, opened 2024-02-15
Debugging playwright e2e tests

Bartvaderkin:
Trying to find out why random e2e tests are failing on locator timeouts.

I've seen it happen on webkit, chromium and firefox, and either the cookie_accept test or the multifile upload.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_action_views.py
+++ b/src/open_inwoner/accounts/tests/test_action_views.py
@@ -344,7 +344,7 @@ class ActionsPlaywrightTests(PlaywrightSyncLiveServerTestCase):
 
         mock_solo.return_value.cookiebanner_enabled = False
 
-        context = self.browser.new_context(storage_state=self.user_login_state)
+        context = self.new_browser_context(storage_state=self.user_login_state)
 
         page = context.new_page()
         page.goto(self.live_reverse("profile:action_list"))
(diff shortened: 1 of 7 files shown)


Bartvaderkin:
Ugh no.. still `playwright._impl._errors.TimeoutError: Timeout 10000ms exceeded.` but this time in Firefox & the multifile upload

Bartvaderkin:
I tried downgrading playwright to playwright==1.39.0 since we updated it to 1.41 for the Django update but same results.

Bartvaderkin:
I think I smushed the cookie accept test as I haven't seen it in a while, but the multiple file upload test is still super flaky.

I added some waits and assertions but it even fails on the most basic visibility check.

pi-sigma:
@Bartvaderkin Could we test the hypothesis (that we're running up against some infrastructure limit) by disabling most of the e2e tests and keep around the cookie-accept and the multiple-file-upload test (these are the ones I've seen failing myself)?

Bartvaderkin:
@pi-sigma Seems reasonable, we can slap some `@skip` and see what happens.

stevenbal:
I see these tests being flakey when I run the tests locally too, so I don't think it's an infrastructure thing?

Bartvaderkin:
I don't know, I never had them fail locally and tried it a lot.

Bartvaderkin:
Closing this as Steven is fixing it in his own branch and we don't need these extra timeouts etc right now.

PR #1029, opened 2024-02-15
:lipstick: [#2115] Forced default bold for headings 

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2115
Also: a small fix for the dyslexia drop-down menu

### DIFF
--- a/src/open_inwoner/scss/views/App.scss
+++ b/src/open_inwoner/scss/views/App.scss
@@ -255,6 +255,8 @@
 
 @font-face {
   font-family: 'Heading';
+  font-style: normal;
+  font-weight: bold;
   src: url('/media/custom_fonts/heading_font.ttf') format('truetype'),
     url('/static/fonts/ubuntu/Ubuntu-M.ttf') format('truetype');
 }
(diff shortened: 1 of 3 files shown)


src/open_inwoner/scss/views/App.scss
```diff
@@ -255,6 +255,8 @@
 
 @font-face {
   font-family: 'Heading';
+  font-style: normal;
+  font-weight: bold;
   src: url('/media/custom_fonts/heading_font.ttf') format('truetype'),
```
pi-sigma, line 260: Are we going to need a separate font for bold headings, similarly to the body fonts?

pi-sigma (approved):

src/open_inwoner/scss/views/App.scss
jiromaykin: > Are we going to need a separate font for bold headings, similarly to the body fonts?

No, because by default Headings are rendered as 'bold' explicitly (just like when you'd use 'bold' for other content) so 'bold' is implied here.

stevenbal: Just for my understanding: does this work because there previously was some default (numeric) value for heading `font-weight`? At first glance it doesn't make a lot of sense to see the addition of `font-weight: bold` make things less bold :sweat_smile:.

stevenbal: check, thanks for the explanation :+1:

stevenbal (approved):

PR #1030, opened 2024-02-15
:sparkles: [#2088] Implement Zaken search for eHerkenning users

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2088

### DIFF
--- a/src/open_inwoner/openzaak/clients.py
+++ b/src/open_inwoner/openzaak/clients.py
@@ -41,12 +41,17 @@ logger = logging.getLogger(__name__)
 class ZakenClient(APIClient):
     def fetch_cases(
         self,
-        user_bsn=None,
-        user_kvk_or_rsin=None,
-        max_requests=4,
-        identificatie=None,
-        vestigingsnummer=None,
+        user_bsn: Optional[str] = None,
+        user_kvk_or_rsin: Optional[str] = None,
+        max_requests: int = 4,
+        identificatie: Optional[str] = None,
+        vestigingsnummer: Optional[str] = None,
     ):
+        if user_bsn and (user_kvk_or_rsin or vestigingsnummer):
+            raise ValueError(
+                "either `user_bsn` or `user_kvk_or_rsin`/`vestigingsnummer` should be supplied, not both"
+            )
+
         if user_bsn:
             return self.fetch_cases_by_bsn(
                 user_bsn, max_requests=max_requests, identificatie=identificatie
--- a/src/open_inwoner/openzaak/clients.py
+++ b/src/open_inwoner/openzaak/clients.py
@@ -39,11 +39,32 @@ logger = logging.getLogger(__name__)
 
 
 class ZakenClient(APIClient):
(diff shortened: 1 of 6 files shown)


src/open_inwoner/openzaak/clients.py
```diff
@@ -40,11 +40,32 @@
...
 class ZakenClient(APIClient):
+    def fetch_cases(
+        self,
+        user_bsn=None,
+        user_kvk_or_rsin=None,
+        max_requests=4,
+        identificatie=None,
+        vestigingsnummer=None,
+    ):
+        if user_bsn:
```
Bartvaderkin, line 51: For correctness we could check if not both `user_bsn` and `user_kvk_or_rsin` are supplied as they're mutually exclusive. Or `vestigingsnummer` and `user_bsn` etc.

Also some typing in the signature would be good.

src/open_inwoner/openzaak/utils.py
```diff
@@ -143,3 +144,25 @@ def translate_single_status(status_text: str) -> str:
         )
     except StatusTranslation.DoesNotExist:
         return ""
+
+
+def get_fetch_parameters(request) -> dict:
```
Bartvaderkin, line 149: Good idea to clean this up. Maybe use a name like `get_user_fetch_parameters()` so we keep it for user-related params only?

src/open_inwoner/search/views.py
```diff
@@ -62,10 +64,9 @@ def search(self, form):
             self.log_user_action(user, _("search query: {query}").format(query=query))
 
         # Check if the query exactly matches with a case that belongs to the user
-        # TODO should be implemented for KVK as well
-        if hasattr(self.request.user, "bsn"):
+        if search_params := get_fetch_parameters(self.request):
```
Bartvaderkin, line 67: Here you see it's unclear: this makes get_fetch_parameters() look like magic. We're getting the fetch params for the user, not a generic mystery search.

Bartvaderkin (changes requested):
Some small notes.

Bartvaderkin (changes requested):

src/open_inwoner/openzaak/clients.py
```diff
@@ -42,12 +42,16 @@
...
-        vestigingsnummer=None,
+        user_bsn: Optional[str] = None,
+        user_kvk_or_rsin: Optional[str] = None,
+        max_requests: Optional[int] = 4,
+        identificatie: Optional[str] = None,
+        vestigingsnummer: Optional[str] = None,
     ):
+        assert not (
+            user_bsn and (user_kvk_or_rsin or vestigingsnummer)
+        ), "either `user_bsn` or `user_kvk_or_rsin`/`vestigingsnummer` should be supplied, not both"
```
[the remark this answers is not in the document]
stevenbal: Oh TIL, I always assumed `Optional` was for optional arguments, but you're right

Bartvaderkin (approved):

PR #1035, opened 2024-02-16
💚 Add font files in playwright tests

stevenbal:
these tests seem to fail since https://github.com/maykinmedia/open-inwoner/pull/963 was merged, possibly related to the App.scss font-face changes

### DIFF
--- a/src/open_inwoner/utils/tests/playwright.py
+++ b/src/open_inwoner/utils/tests/playwright.py
@@ -2,12 +2,16 @@ import os
 from typing import Callable
 
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
+from django.core.files.base import ContentFile
 from django.urls import reverse
 
 from furl import furl
 from playwright.sync_api import Browser, Playwright, sync_playwright
 
 from open_inwoner.accounts.models import User
+from open_inwoner.configurations.choices import CustomFontName
+from open_inwoner.utils.files import OverwriteStorage
+from open_inwoner.utils.test import temp_media_root
 
 BROWSER_DRIVERS = {
     # keys for the E2E_DRIVER environment variable (likely from test matrix)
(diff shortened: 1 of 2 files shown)


src/open_inwoner/utils/tests/playwright.py
```diff
@@ -94,6 +96,19 @@ def setUpClass(cls):
...
+            "heading_font",
+            "body_font_regular",
+            "body_font_bold",
+            "body_font_italic",
+            "body_font_bold_italic",
+        ]
+        for font in fonts:
+            default_storage.save(
+                f"custom_fonts/{font}.ttf", ContentFile(b"", name=f"{font}.ttf")
+            )
```
Bartvaderkin, line 110: There could be an issue that `default_storage.save()` generates a new name if the file already exists. IIRC @pi-sigma did something for it in the implementation of the feature.

Also maybe @pi-sigma knows if this list of names is defined somewhere in code so this list self-maintains if we change them.

Bartvaderkin (commented):
I'm fine with this, but left a comment. 

If you want it today @pi-sigma could look at my comment and press the appropriate buttons.

src/open_inwoner/utils/tests/playwright.py
stevenbal: I'll reuse `CustomFontName` to get this list of names and I'll look into the duplicate file issue

Bartvaderkin: There might also be a fun interaction with the `@temp_media_root()`decorator (although IMHO that should always be active anyway so we could add that).

pi-sigma: I defined a custom storage (in `open_inwoner.utils.files`) to remove duplicates. That's necessary because otherwise the fille is added with some random string attached to the file name. E.g. if `heading_font` already exists, then adding another yields `heading_font_s40dfg` or something like it, making it impossible to refer to the file.

stevenbal: I reused the custom storage and also added the `temp_media_root` decorator

pi-sigma (approved):

PR #1038, opened 2024-02-19
:white_check_mark: [#1916] Fix flakey password reset throttling test

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1916

### DIFF
--- a/src/open_inwoner/accounts/tests/test_password_reset_view.py
+++ b/src/open_inwoner/accounts/tests/test_password_reset_view.py
@@ -1,11 +1,26 @@
 from django.urls import reverse
 
 from django_webtest import WebTest
+from freezegun import freeze_time
 
 from open_inwoner.utils.test import ClearCachesMixin
 
 
 class PasswordResetViewTests(ClearCachesMixin, WebTest):
+    """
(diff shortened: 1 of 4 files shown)


src/open_inwoner/accounts/tests/test_password_reset_view.py
```diff
@@ -4,8 +4,12 @@
 
 from open_inwoner.utils.test import ClearCachesMixin
 
+from freezegun import freeze_time
+
 
 class PasswordResetViewTests(ClearCachesMixin, WebTest):
+    # Freeze the time to ensure a consistent cache key for throttling
+    @freeze_time("2024-01-01T12:00:00")
```
stevenbal, line 12: Looking at the implementation it makes sense why this was failing:

https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/utils/mixins.py#L54

If `time()` for instance is about to reach a value where `mod 60` equals zero, the cache key changes, meaning that the throttling essentially resets for that IP.

```python
>>> 1708335959 - (1708335959 % 60)
1708335900
>>> 1708335960 - (1708335960 % 60)
1708335960
```

PR #1045, opened 2024-02-21
[#2099] Re-design mijn vragen

pi-sigma:
Taiga: [#2099](https://taiga.maykinmedia.nl/project/open-inwoner/task/2099)

Related: PR https://github.com/maykinmedia/open-inwoner/pull/1059

### DIFF
--- a/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
+++ b/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
@@ -1,6 +1,52 @@
 .contactmomenten {
   margin-top: var(--spacing-giant);
 
+  &__scrolldown {
+    padding-top: 3em;
+
+    .form-container {
+      width: 50%;
+    }
+
+    .button {
+      margin-top: var(--spacing-giant);
+    }
+  }
+
+  &__contact-form {
+    text-align: center;
+
+    .form-container {
+      margin: auto;
+      width: 50%;
+    }
+
+    .form-heading {
+      padding-bottom: var(--spacing-giant);
+      text-align: left;
+    }
+
+    form {
+      textarea {
+        width: 100%;
+      }
+
+      .form__control {
+        label {
+          padding-bottom: var(--spacing-extra-large);
+        }
+
+        .input {
+          max-width: 100% !important;
+        }
+      }
+
+      .button {
+        float: left;
+      }
+    }
+  }
+
   & .grid {
     grid-row-gap: var(--spacing-medium);
     grid-column-gap: var(--spacing-extra-large);
@@ -72,4 +118,16 @@
       }
     }
   }
+}
+
+// overrides of elements outside of components
+section {
+  display: block;
+  padding-bottom: 2em;
+}
+
+.pagination {
+  display: table !important;
+  margin: 0 auto !important;
+  padding-top: 1em !important;
 }
--- a/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
+++ b/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
@@ -40,10 +40,6 @@
           max-width: 100% !important;
         }
       }
-
-      .button {
-        float: left;
-      }
     }
   }
 
--- a/src/open_inwoner/templates/pages/cases/contact_form.html
+++ b/src/open_inwoner/templates/pages/cases/contact_form.html
@@ -9,7 +9,7 @@
     {% endblocktrans %}
 </p>
 
-{% render_form form=contact_form method="POST" id="contact-form" hxpost=hxpost_contact_action hxtarget="#cases-contact-form" extra_classes="case-contact-form" %}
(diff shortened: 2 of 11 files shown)


src/open_inwoner/templates/pages/contactmoment/list.html
```diff
@@ -39,4 +63,31 @@ <h1 class="utrecht-heading-1" id="contactmomenten">{{ page_title }} ({{ contactm
...
+                </p>
+            </div>
+            <div>
+                <form action="#" method="POST" id="question-form">
+                    {% csrf_token %}
+                    {% for field in question_form.fields %}
+                        {% autorender_field question_form field %}
+                    {% endfor %}
+
+                    <button class="button button--textless button--icon button--icon-after button--primary" type="submit">
```
jiromaykin, line 85: We do not like to float :-) and it's missing an icon - Here it's easier to use the HTML that already exists in '_src/open_inwoner/templates/pages/cases/contact_form.html_' and replace this button with this (containing the icon too):
 
```HTML
<div class="form__submit">
  {% button_row %}
   {% button text=_("Vraag versturen") icon="east" icon_position="after" icon_outlined=True id="submit_contact" type="submit" disabled=True primary=True name="contact_form_submit" %}
  {% endbutton_row %}
</div>
```

Also it's missing the disabled button behaviour - for example look at '_src/open_inwoner/js/components/cases/case_contact_form.js_' so you can add a new JS file for this Vragen Form submit button,  - or make 1 file that looks for a new form-selector that can be enabled or disabled depending on whether the user is actually typing something, in such a way the JS can be used by both forms so it becomes a more general JS styling thing.

src/open_inwoner/templates/pages/contactmoment/list.html
```diff
@@ -1,10 +1,33 @@
...
+            <div class="form-heading">
+                <h2 class="h2">{% trans "Stel een vraag" %}</h2>
+                <p class="p">
+                {% blocktrans trimmed %}
+                    Heeft u een vraag? Dan kunt u deze hier stellen.
+                    Een van onze medewerkers beantwoord uw vraag z.s.m.
+                {% endblocktrans %}
+                </p>
+            </div>
+            {% button href="#question-form" text=_("Stel een vraag") title=_("Stel een vraag") primary=True icon="arrow_downward" icon_position="after" %}
```
jiromaykin, line 19: The design isn't done yet but I would still like to see this wrapped with a Div with class `button-row` to give it correct padding,
see also: '_src/open_inwoner/components/templates/components/Button/ButtonRow.html_'

jiromaykin (changes requested):

Bartvaderkin (approved):

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -1,6 +1,59 @@
 .contactmomenten {
   margin-top: var(--spacing-giant);
 
+  .contact-form {
```
jiromaykin, line 4: This needs to be taken outside of .contactmomenten or else the disabled button won't work for the other forms.

src/open_inwoner/templates/pages/cases/contact_form.html
```diff
@@ -9,7 +9,7 @@ <h2 class="h2" id="contact">{% trans 'Stel een vraag' %}</h2>
     {% endblocktrans %}
 </p>
 
-{% render_form form=contact_form method="POST" id="contact-form" hxpost=hxpost_contact_action hxtarget="#cases-contact-form" extra_classes="case-contact-form" %}
+{% render_form form=contact_form method="POST" id="contact-form" hxpost=hxpost_contact_action hxtarget="#cases-contact-form" extra_classes="contact-form" %}
```
jiromaykin, line 12: You may like to move the. disabled styling to the .form SCSS instead of into .Contactform SCSS (?) - since here it will not work on this class as a general style.

src/open_inwoner/templates/pages/contactmoment/list.html
```diff
@@ -39,4 +65,33 @@ <h1 class="utrecht-heading-1" id="contactmomenten">{{ page_title }} ({{ contactm
...
+                    Heeft u een vraag? Dan kunt u deze hier stellen.
+                    Een van onze medewerkers beantwoord uw vraag z.s.m.
+                {% endblocktrans %}
+                </p>
+            </div>
+            <div>
+                <form action="{% url 'cases:contactmoment_list' %}" method="POST" id="question-form" class="contact-form">
+                    {% csrf_token %}
+
+                    {% for field in form.fields %}
```
jiromaykin, line 84: The "expand_more" icon is missing from the select-input here, so you may need to expand the form here into its separate fields. (Examples of the icon can be found in Samenwerken etc.)
 
select-caret

jiromaykin (changes requested):

jiromaykin (approved):

PR #1046, opened 2024-02-22
🐛 [#2143] Show all categories to staff users

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2143

### DIFF
--- a/src/open_inwoner/pdc/managers.py
+++ b/src/open_inwoner/pdc/managers.py
@@ -35,7 +35,10 @@ class CategoryPublishedQueryset(MP_NodeQuerySet):
 
     def visible_for_user(self, user: User):
         if user.is_authenticated:
-            if user.bsn:
+            # Show all categories to staff users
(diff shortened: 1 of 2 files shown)


Bartvaderkin (changes requested):
Readability would improve in cases like this if we don't use random numbered variables but give the categories useful names, like `category_anonymous`, `category_all`, `category_citizens`, `category_companies` so the test itself is meaningful and you don't have to go back and forth to the setup to read the tests.

And optional note: this test file looks pretty old, can we update it? There are unusual user setups in the setUpTestData(), with passwords and custom email and manual `self.client.login(email=self.user.email, password="12345")`. 

That is all irrelevant to the test, and passwords are actually already set in the factory  :frowning: The more recent tests just create an appropriate user right there in the test and do a force_login.

pi-sigma (approved):

Bartvaderkin (approved):

PR #1048, opened 2024-02-22
[#2042] Fix margins for Userfeed and Home

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2042 

This is a first start to begin transforming the main grid in such a way that 'components no longer have outside margins around themselves - the parent grid should determine the spacing between components, not the components themselves - see also [the temporary margin-top fix here](https://github.com/maykinmedia/open-inwoner/pull/969#pullrequestreview-1831949190) that was the original reason for this PR.

This PR also enables spacing between plugins on the Home page by wrapping them with a Section - so you can move around any of the plug-ins in any order without having to constantly update any combination of margins between cards & heading, between cards and questionnaires, between userfeed and video, etc, etc.


src/open_inwoner/templates/pages/faq.html
```diff
@@ -2,7 +2,7 @@
 {% load i18n faq_tags %}
 
 {% block content %}
-    <h1 class="h1">
+    <h1 class="utrecht-heading-1">
```
Bartvaderkin, line 5: Gemeente specific class :D

Bartvaderkin (changes requested):

pi-sigma (approved):
Looking good. I do wonder about the styling convention. `h1` elements are now styled as `utrecht-heading-1` but `h2` elementss continue to receive `class="h2"`? Is that an oversight or intentional? Just so I know which convention to follow when I write this myself...

src/open_inwoner/templates/pages/faq.html
jiromaykin: @Bartvaderkin Actually it's not quite :-D this is one of the NL Design-System (design token) classes we will be using more and more often - the first municipality that sets the standard, also gives the code their name... Since open-inwoner is not a municipality we will not be suggesting names for classes etc. In future you will see more Utrecht en Den Haag names even :-O

jiromaykin:
> Looking good. I do wonder about the styling convention. `h1` elements are now styled as `utrecht-heading-1` but `h2` elementss continue to receive `class="h2"`? Is that an oversight or intentional? Just so I know which convention to follow when I write this myself...

Currently this was intentional - I need another design-token sprint in order to create more values from NL Design System classes that we can use. Also: in Open-Inwoner we have a bit of a complicated H2 element that uses flexbox in order to get 'indicator' icons and sometimes the H2 wraps itself around a button so H2 in particular is more complicated to completely rework; the other Headings will be much more straightforward.

src/open_inwoner/templates/pages/faq.html
Bartvaderkin: Ugh, well ok then. Let's hope they thought about how this is going to scale in the future :shrug:

Bartvaderkin (approved):

src/open_inwoner/templates/pages/faq.html
jiromaykin: Expectation is that it will take at least 5 years for all of these names to become something like "NLDS-heading-bla-bla" once there will be definite community standards 🙈

PR #1049, opened 2024-02-23
:zap: Fetch case information objects in parallel

stevenbal:


stevenbal:
seems to be more work then expected to fix the tests and the performance gains seems marginal, so closing this

PR #1051, opened 2024-02-23
:wastebasket: [#2062] Remove dashboard templatetags

stevenbal:


src/open_inwoner/components/templates/components/Faq/Faq.html
```diff
@@ -1,7 +1,10 @@
 {% load i18n link_tags render_tags %}
 
-<aside class="faq" aria-label="{% trans "Veelgestelde vragen" %}">
-    {% if title %}<h2 id="faq" class="h2">{{ title }}</h2>{% endif %}
+{% trans "Veelgestelde vragen" as default_title %}
+<aside class="faq" aria-label="{{ default_title }}">
```
pi-sigma, line 4: Is there a point for the `aria-label`? Either the `h2` is visible, then the content is accessible to screen readers, or the `h2` is hidden, but then the `h1` element from `pages/faq.html` is visible and contains `{% trans "Veelgestelde vragen" %}`, so again the content is accessible.

pi-sigma (approved):
Small point about the use of `aria-label` in one place. Not sure if this needs to be addressed, so feel free to adopt or ignore as you see fit.

src/open_inwoner/components/templates/components/Faq/Faq.html
stevenbal: Hmm I'm not so sure,  @jiromaykin can we omit the `aria-label` here?

Bartvaderkin (approved):

PR #1052, opened 2024-02-23
🐛 [#2146] Avoid KvK branch select redirects for media files

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2146

### DIFF
--- a/src/open_inwoner/kvk/middleware.py
+++ b/src/open_inwoner/kvk/middleware.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
 from django.http import HttpResponseRedirect
 from django.urls import NoReverseMatch, reverse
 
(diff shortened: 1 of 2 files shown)


pi-sigma (approved):
This is fine by me, though I wonder if it would be worthwhile to address this kvk-redirect issue in a more principled way. The middleware approach I adopted seems to be causing more problems than it solves. Perhaps a mixin, applied to select views, would be a better approach. Let's discuss next week.

Bartvaderkin (commented):
Do we also need this for the static and private-media?

stevenbal:
@Bartvaderkin static files if served via Django don't seem to trigger the middleware, and on staging/production I assume we serve staticfiles via nginx (https://bitbucket.org/maykinmedia/maykin-deployment/src/f5a08f425136c452553d0f30cd1b33cbb684ef47/projects/open-inwoner/nginx-production.conf#lines-66)

I just saw that we serve media files via nginx as well, but at least this fix makes this branch selection screen less slow locally.

I'm not sure if private media could pose an issue here, but I suppose it can't hurt to add a bypass for it in the middleware

Bartvaderkin (approved):

PR #1056, opened 2024-02-26
[#2137] Show disabled pseudo-checkbox for actions

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2137 
To explicitly show required-actions notifications cannot be turned off (on http://localhost:8000/mijn-profiel/notificaties/).

➕ also making the pseudo checkbox accessible (⚠️ ordinary **disabled** inputs are **_not_** detected by screenreaders and are not read + cannot be focussed, so making a _visual_ fake checkbox here that actually gets detected).
A disabled input that cannot be enabled by the user in any way is a violation of accessibility.

### DIFF
--- a/src/open_inwoner/templates/pages/profile/notifications.html
+++ b/src/open_inwoner/templates/pages/profile/notifications.html
@@ -20,8 +20,8 @@
         {# pseudo checkbox for notifications that cannot be disabled #}
         <div class="checkbox">
             <span role="checkbox" aria-disabled="true" aria-checked="true" class="checkbox__input checkbox__pseudo checkbox__input--disabled" aria-labelledby="id_cases_notifications_action_required" tabindex="0"></span>
-            <span class="checkbox__label checkbox__pseudo-label" id="id_cases_notifications_action_required">{% trans "Zaaknotificaties - action required" %}</span>
-            <p class="p">{% trans "E-mail notifications for case updates that require action (cannot be disabled)" %}</p>
+            <span class="checkbox__label checkbox__pseudo-label" id="id_cases_notifications_action_required">{% trans "Zaaknotificaties - actie is nodig" %}</span>
+            <p class="p">{% trans "E-mailnotificaties wanneer er actie nodig is voor een zaak (kan niet uitgeschakeld worden)" %}</p>
         </div>
     </div>
 
--- a/src/open_inwoner/templates/pages/profile/notifications.html
+++ b/src/open_inwoner/templates/pages/profile/notifications.html
@@ -9,6 +9,23 @@
     {% trans "Kies voor welk onderwerp je meldingen wilt ontvangen" %}
 </p>
 
-{% form form_object=form method="POST" id="change-notifications" submit_text=_("Opslaan") secondary_href='profile:detail' secondary_text=_('Terug') secondary_icon='arrow_backward' secondary_icon_position="before" %}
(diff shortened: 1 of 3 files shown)


src/open_inwoner/templates/pages/profile/notifications.html
```diff
@@ -9,6 +9,23 @@ <h1 class="h1" id="title">
...
 
-{% form form_object=form method="POST" id="change-notifications" submit_text=_("Opslaan") secondary_href='profile:detail' secondary_text=_('Terug') secondary_icon='arrow_backward' secondary_icon_position="before" %}
+<form method="POST" id="change-notifications" action="{% url 'profile:notifications' %}" class="form" novalidate>
+    {% csrf_token %}
+
+    {% for field in form.fields %}
+        {% autorender_field form field %}
+    {% endfor %}
+
+    <div class="form__control">
```
stevenbal, line 19: Maybe we could move this disabled checkbox to a separate template that can be included, to avoid duplication.

Also, is there a reason why this is a `<span role="checkbox">` and not an `<input type="checkbox" disabled>`?

jiromaykin: Let's not make a reusable template for bad patterns such as actual disabled inputs that cannot be enabled by the user, as I explained in the description of this PR above - also: [new designs are coming soon](https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=1373%3A5943&mode=design&t=na0I2hZEz4Y4U2c0-1) and this PR is an intermediate change.
Also: yes, this should _not_ be an actual input tag, because accessibility devices completely ignore disabled inputs and do not give any textual/label information.

stevenbal: check, didn't know that :+1:

stevenbal (approved):

src/open_inwoner/templates/pages/profile/notifications.html
```diff
@@ -9,6 +9,23 @@ <h1 class="h1" id="title">
...
+        {# pseudo checkbox for notifications that cannot be disabled #}
+        <div class="checkbox">
+            <span role="checkbox" aria-disabled="true" aria-checked="true" class="checkbox__input checkbox__pseudo checkbox__input--disabled" aria-labelledby="id_cases_notifications_action_required" tabindex="0"></span>
+            <span class="checkbox__label checkbox__pseudo-label" id="id_cases_notifications_action_required">{% trans "Zaaknotificaties - action required" %}</span>
```
pi-sigma, line 23: `{% trans "Zaaknotificaties - action required" %}` is a mix of Dutch and English (should have been cleaned up as part of my PR).

pi-sigma (changes requested):

src/open_inwoner/templates/pages/profile/notifications.html
jiromaykin: Guess I should make it all in Dutch (including the paragraph)? 
(We have new translations for this one now I can use)

pi-sigma: Yes, client-facing text should be Dutch, preferably. It will be translated anyways, but this will make it easier and more consistent.

pi-sigma (approved):

PR #1057, opened 2024-02-26
✨ [#2112] Retrieve open tasks on login and store in userfeed 

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2112


src/open_inwoner/userfeed/tests/test_external_tasks.py
```diff
@@ -0,0 +1,115 @@
...
+    @requests_mock.Mocker()
+    def test_fetch_tasks_after_login(self, m):
+        request = RequestFactory().get("/dummy")
+        request.user = self.user
+
+        ESuiteData().install_mocks(m)
+
+        user_logged_in.send(User, user=self.user, request=request)
+
+        qs_after_first_login = FeedItemData.objects.all()
```
Bartvaderkin, line 36: I think you need a list() here to save the actual results because otherwise the queryset is lazily evaluated later and it gets a bit unclear.

src/open_inwoner/userfeed/tests/test_external_tasks.py
```diff
@@ -0,0 +1,115 @@
...
+from open_inwoner.accounts.tests.factories import DigidUserFactory
+from open_inwoner.openzaak.models import OpenZaakConfig
+from open_inwoner.openzaak.tests.mocks import ESuiteData
+from open_inwoner.openzaak.tests.shared import FORMS_ROOT
+from open_inwoner.userfeed.choices import FeedItemType
+from open_inwoner.userfeed.models import FeedItemData
+from open_inwoner.userfeed.tests.factories import FeedItemDataFactory
+
+
+class UserFeedExternalTasksTestCase(TestCase):
```
Bartvaderkin, line 17: This case is a good example of my earlier comment about splitting the trigger/decision with the actual functionality. 

Here we have a mix of signal/login stuff, and actual feed update things. Eg: to test the feed update we need to simulate a login.. 

So I'd say do a `test_login_triggers_update()`, with a mocked `update_user_task_feed(user)` that is checked with just an `assert_called_once_with()`. Then we could easily refactor to a `test_render_triggers_update()` if we decide to move it to the plugin render, without having to mess with the feed update testing part that is now intertwined.

Then add one or more `test_update_user_task_feed()` tests to test the update function, expiration etc etc without caring where it was used or mixing in signal or plugin render code.

src/open_inwoner/userfeed/tests/test_external_tasks.py
```diff
@@ -0,0 +1,115 @@
...
+            self.assertEqual(item2.type, FeedItemType.external_task)
+            self.assertEqual(
+                item2.type_data,
+                {
+                    "url": "https://maykinmedia.nl",
+                    "naam": "Aanvullende informatie gewenst",
+                    "uuid": "d74f6a5c-297d-43a3-a923-1774164d852d",
+                    "startdatum": "2023-10-11",
+                    "identificatie": "4321-2023",
+                    "formulier_link": "https://maykinmedia.nl",
```
Bartvaderkin, line 72: Here we see half the data is stuff we don't really care for.

src/open_inwoner/apimock/apis/esuite-read/formulieren-api/openstaande-taken.json
```diff
@@ -0,0 +1,18 @@
...
+    },
+    {
+        "url": "https://maykinmedia.nl",
+        "uuid": "d74f6a5c-297d-43a3-a923-1774164d852d",
+        "identificatie": "4321-2023",
+        "naam": "Aanvullende informatie gewenst",
+        "startdatum": "2023-10-11",
+        "formulierLink": "https://maykinmedia.nl"
+    }
+]
```
Bartvaderkin, line 18: Github diff complains there is no ending newline here.

src/open_inwoner/openzaak/tests/mocks.py
```diff
@@ -28,9 +28,30 @@ def __init__(self):
             ],
         }
 
+        self.task1 = {
```
Bartvaderkin, line 31: Would it be a good idea to split this mock data class? These tasks are not related to the open submissions. We could have a class ESuiteSubmissionData and a ESuiteTaskData:

Bartvaderkin (changes requested):
I added some notes, but also used this as a good demo case for practical testing with mock/patch.

pi-sigma (approved):

src/open_inwoner/userfeed/hooks/external_task.py
```diff
@@ -0,0 +1,69 @@
...
+    for task in openstaande_taken:
+        if task.uuid in existing_uuids:
+            continue
+
+        data = {
+            "user": user,
+            "type": FeedItemType.external_task,
+            "ref_uuid": task.uuid,
+            "action_required": True,
+            "type_data": dataclasses.asdict(task),
```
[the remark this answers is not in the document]
stevenbal: I modified it to only contain the data we use and renamed them. I also added an update for existing items, if their `type_data` changed

src/open_inwoner/userfeed/hooks/external_task.py
```diff
@@ -0,0 +1,69 @@
...
+        type=FeedItemType.external_task, user=user, ref_uuid__in=completed_uuids
+    ).mark_completed()
+
+
+@receiver(user_logged_in)
+def fetch_open_tasks(sender, user, request, *args, **kwargs):
+    if user.login_type == LoginTypeChoices.digid:
+        if client := build_client("form"):
+            tasks = client.fetch_open_tasks(user.bsn)
+            create_external_task_items(user, tasks)
```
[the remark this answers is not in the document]
stevenbal: Changed it (though because we no longer rely on a signal now, this became a separate function anyway)

src/open_inwoner/userfeed/tests/test_external_tasks.py
stevenbal: I split it up and changed the login test to a CMS plugin render test. I did also add a check for the display of the FeedItem in the UserFeed.

stevenbal:
@Bartvaderkin I made `has_change_permission` and `has_add_permission` conditional based on `settings.DEBUG`, to make local development a bit easier

src/open_inwoner/userfeed/hooks/external_task.py
```diff
@@ -0,0 +1,69 @@
...
+            "type": FeedItemType.external_task,
+            "ref_uuid": task.uuid,
+            "action_required": True,
+            "type_data": dataclasses.asdict(task),
+        }
+        create_data.append(FeedItemData(**data))
+    FeedItemData.objects.bulk_create(create_data)
+
+    # Mark all tasks with UUIDs not occurring in the fetched results as completed
+    completed_uuids = existing_uuids - set(task.uuid for task in openstaande_taken)
```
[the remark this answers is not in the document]
stevenbal: I moved the fetch to the plugin render and removed the signal

src/open_inwoner/userfeed/hooks/external_task.py
```diff
@@ -0,0 +1,88 @@
...
+    existing_uuids = set(existing_uuid_mapping.keys())
+
+    update_data = []
+    create_data = []
+    for task in openstaande_taken:
+        type_data = dict(
+            action_url=task.formulier_link,
+            task_name=task.naam,
+            task_identificatie=task.identificatie,
+        )
```
Bartvaderkin, line 51: I would make this a dict literal instead of a dict constructor so it is clear at a glance these are arbitrary keys. We want them highly visible because they are important.

src/open_inwoner/userfeed/hooks/external_task.py
```diff
@@ -0,0 +1,88 @@
...
+            "type": FeedItemType.external_task,
+            "ref_uuid": task.uuid,
+            "action_required": True,
+            "type_data": dict(
+                action_url=task.formulier_link,
+                task_name=task.naam,
+                task_identificatie=task.identificatie,
+            ),
+        }
+        create_data.append(FeedItemData(**data))
```
Bartvaderkin, line 69: .. but this should be a FeedItemData() constructor so the IDE can type check. Using an intermediate dict literal and **kwargs to put them in the constructor is unnecessary.  (this is opposite from earlier comment about the type_data)

Bartvaderkin (changes requested):
Cool, much clearer. 

I put a style note about dicts because the distinction helps a lot in reading existing code.

src/open_inwoner/userfeed/hooks/external_task.py
stevenbal: oops I did it the wrong way around, the type_data is now a regular dict (was also duplicated before, so I removed the redeclaration) and i use the constructor with keyword args for FeedItemData

Bartvaderkin (approved):

src/open_inwoner/userfeed/hooks/external_task.py
Bartvaderkin: Cool, thanks. Much clearer.

PR #1059, opened 2024-02-27
[#2098] Show case-related questions in case detail view

pi-sigma:
Taiga: [#2098](https://taiga.maykinmedia.nl/project/open-inwoner/task/2098)

Related: PR https://github.com/maykinmedia/open-inwoner/pull/1045

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -3,11 +3,12 @@ from datetime import datetime
 from typing import Optional, TypedDict
 
 from django.contrib.auth.mixins import AccessMixin
-from django.http import Http404
+from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
+from django.views import View
 from django.views.generic import TemplateView
 
 from view_breadcrumbs import BaseBreadcrumbMixin
@@ -211,3 +212,31 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             },
         ]
         return ctx
+
+
+class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
+    """
+    Redirect to `KlantContactMomentDetailView` on the basis of contactmoment uuid
+    """
+
+    def get(self, request, *args, **kwargs):
+        klanten_client = build_client("klanten")
+        contactmoment_client = build_client("contactmomenten")
+
+        klant = klanten_client.retrieve_klant(**get_fetch_parameters(self.request))
+        kcms = contactmoment_client.retrieve_klantcontactmomenten_for_klant(klant)
+
+        if not kcms:
+            raise Http404
+
+        contactmoment_uuid = kwargs["uuid"]
+        kcm = next(
+            (kcm for kcm in kcms if str(kcm.contactmoment.uuid) == contactmoment_uuid)
+        )
+
+        if not kcm:
+            raise Http404
+
+        return HttpResponseRedirect(
+            reverse("cases:contactmoment_detail", kwargs={"kcm_uuid": kcm.uuid})
+        )
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -195,6 +195,7 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             ctx["zaak"] = getattr(ocm, "object", None)
 
         contactmoment: KCMDict = self.get_kcm_data(kcm)
+
         ctx["contactmoment"] = contactmoment
         ctx["metrics"] = [
             {
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -195,7 +195,6 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             ctx["zaak"] = getattr(ocm, "object", None)
 
         contactmoment: KCMDict = self.get_kcm_data(kcm)
-
         ctx["contactmoment"] = contactmoment
         ctx["metrics"] = [
             {
--- a/src/open_inwoner/cms/cases/tests/test_contactform.py
+++ b/src/open_inwoner/cms/cases/tests/test_contactform.py
@@ -19,6 +19,7 @@ from open_inwoner.accounts.tests.factories import (
     DigidUserFactory,
     eHerkenningUserFactory,
 )
+from open_inwoner.openklant.clients import ContactmomentenClient
 from open_inwoner.openklant.constants import Status
 from open_inwoner.openklant.models import OpenKlantConfig
 from open_inwoner.openklant.tests.data import CONTACTMOMENTEN_ROOT, KLANTEN_ROOT
@@ -41,6 +42,12 @@ PATCHED_MIDDLEWARE = [
 
 
 @requests_mock.Mocker()
+@patch.object(
+    ContactmomentenClient,
+    "retrieve_objectcontactmomenten_for_zaak",
+    autospec=True,
+    return_value=[],
+)
 @override_settings(
     ROOT_URLCONF="open_inwoner.cms.tests.urls", MIDDLEWARE=PATCHED_MIDDLEWARE
 )
@@ -326,14 +333,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
             json=self.contactmoment,
         )
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_is_shown_if_open_klant_api_configured(self, m, ocm_mock):
+    def test_form_is_shown_if_open_klant_api_configured(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         self.assertTrue(self.ok_config.has_api_configuration())
 
@@ -343,14 +345,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.assertTrue(response.context["case"]["contact_form_enabled"])
         self.assertTrue(contact_form)
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_is_shown_if_open_klant_email_configured(self, m, ocm_mock):
+    def test_form_is_shown_if_open_klant_email_configured(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         self.ok_config.register_email = "example@example.com"
         self.ok_config.register_contact_moment = False
@@ -365,14 +362,11 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.assertTrue(response.context["case"]["contact_form_enabled"])
         self.assertTrue(contact_form)
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_is_shown_if_open_klant_email_and_api_configured(self, m, ocm_mock):
+    def test_form_is_shown_if_open_klant_email_and_api_configured(
+        self, m, contactmoment_mock
+    ):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         self.ok_config.register_email = "example@example.com"
         self.ok_config.save()
@@ -386,7 +380,7 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.assertTrue(response.context["case"]["contact_form_enabled"])
         self.assertTrue(contact_form)
 
-    def test_no_form_shown_if_open_klant_not_configured(self, m):
+    def test_no_form_shown_if_open_klant_not_configured(self, m, contactmoment_mock):
         self._setUpMocks(m)
 
         # reset
@@ -406,14 +400,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.assertFalse(response.context["case"]["contact_form_enabled"])
         self.assertFalse(contact_form)
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_no_form_shown_if_contact_form_disabled(self, m, ocm_mock):
+    def test_no_form_shown_if_contact_form_disabled(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         CatalogusConfig.objects.all().delete()
         self.zaak_type_config.delete()
@@ -429,14 +418,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.assertFalse(response.context["case"]["contact_form_enabled"])
         self.assertFalse(contact_form)
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_success_with_api(self, m, ocm_mock):
+    def test_form_success_with_api(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         response = self.app.get(self.case_detail_url, user=self.user)
         form = response.forms["contact-form"]
@@ -470,14 +454,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
             },
         )
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_success_with_api_eherkenning_user(self, m, ocm_mock):
+    def test_form_success_with_api_eherkenning_user(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         for use_rsin_for_innNnpId_query_parameter in [True, False]:
             with self.subTest(
@@ -540,14 +519,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
                     },
                 )
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_success_with_email(self, m, ocm_mock):
+    def test_form_success_with_email(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         self.ok_config.register_email = "example@example.com"
         self.ok_config.register_contact_moment = False
@@ -578,14 +552,9 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
             _("Contact formulier inzending vanaf Open Inwoner Platform"),
         )
 
-    @patch(
-        "open_inwoner.cms.cases.views.status.InnerCaseDetailView.get_objectcontactmomenten"
-    )
-    def test_form_success_with_bearth_email_and_api(self, m, ocm_mock):
+    def test_form_success_with_both_email_and_api(self, m, contactmoment_mock):
         self._setUpMocks(m)
         self._setUpExtraMocks(m)
-
-        ocm_mock.return_value = []
 
         self.ok_config.register_email = "example@example.com"
         self.ok_config.save()
--- a/src/open_inwoner/cms/cases/tests/test_contactform.py
+++ b/src/open_inwoner/cms/cases/tests/test_contactform.py
@@ -56,6 +56,7 @@ class CasesContactFormTestCase(AssertMockMatchersMixin, ClearCachesMixin, WebTes
         self.document_service = ServiceFactory(
             api_root=DOCUMENTEN_ROOT, api_type=APITypes.drc
         )
+
         # openzaak config
         self.oz_config = OpenZaakConfig.get_solo()
         self.oz_config.zaak_service = self.zaak_service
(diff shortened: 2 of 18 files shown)


src/open_inwoner/cms/cases/tests/test_contactform.py
```diff
@@ -296,6 +296,15 @@ def _setUpExtraMocks(self, m):
...
         )
+        self.objectcontactmoment = generate_oas_component_cached(
+            "cmc",
+            "schemas/ObjectContactMoment",
+            uuid="77880671-b88a-44ed-ba24-dc2ae688c2ec",
+            url=f"{CONTACTMOMENTEN_ROOT}objectcontactmomenten/77880671-b88a-44ed-ba24-dc2ae688c2ec",
+            object=self.zaak["url"],
+            object_type="zaak",
+            contactmoment=self.contactmoment["url"],
+        )
```
stevenbal, line 307: This functionality is not actually tested in this file, right? If that is the case, I think I'd be in favor of mocking (like you did for `test_multiple_file_upload` as well)

src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -276,6 +277,15 @@ def setUp(self) -> None:
...
         )
+        self.objectcontactmoment = generate_oas_component_cached(
+            "cmc",
+            "schemas/ObjectContactMoment",
+            uuid="77880671-b88a-44ed-ba24-dc2ae688c2ec",
+            url=f"{CONTACTMOMENTEN_ROOT}objectcontactmomenten/77880671-b88a-44ed-ba24-dc2ae688c2ec",
+            object=self.zaak["url"],
+            object_type="zaak",
+            contactmoment=self.contactmoment["url"],
+        )
```
stevenbal, line 288: Same comment about mocking this

src/open_inwoner/openklant/tests/factories.py
```diff
@@ -1,9 +1,37 @@
...
 
 class ContactFormSubjectFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = "openklant.ContactFormSubject"
 
     subject = factory.Faker("sentence")
     subject_code = factory.Faker("word")
+
+
+def make_contactmoment(contact_moment_data: dict):
```
stevenbal, line 14: I think instead of this, we could use from `zgw_consumers.api_models.base.factory`:

```python
from zgw_consumers.api_models.base import factory

factory(ContactMoment, data)
```

stevenbal (changes requested):

jiromaykin:
> [...] we need to paginate inside a detail view + the query param for the page is not easily accessible because of the inner/outer view split for cases). Let me know if this functionality is not needed, then I'll get rid of it.

So, I think we agree the pagination is not needed for the detailview - the detailview needs a collapse/readmore button. 
The other task (2148) about the pagination was just a pure styling task (make pagination centered on all views, so make them look the same through the entire project, but that task is for me).

jiromaykin (changes requested):
I need more 'Vragen' data in order to properly review this.

src/open_inwoner/templates/pages/cases/status_outer.html
```diff
@@ -9,7 +9,7 @@
     <div class="cases__detail" id="cases-detail-content"><div>
 
     <div class="loader-container" id="spinner-container">
-        <div class="spinner" id="spinner-id" hx-get="{{ hxget }}" hx-trigger="load" hx-target="#cases-detail-content">
+        <div class="spinner" id="spinner-id" hx-get="{{ hxget }}" hx-trigger="load" hx-target="#cases-detail-content" hx-include="[name='page']">
```
jiromaykin, line 12: Out of curiosity: is this AJAX specific?

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -58,7 +58,7 @@ <h2 class="h2" id="documents-upload">{% trans "Document toevoegen" %}</h2>
                 {% endbutton_row %}
             {% endif %}
 
-            {#  Documents. #}
+            {#  Documents #}
```
jiromaykin, line 61: Nice :-)

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -73,6 +73,49 @@ <h2 class="h2 {% if case.new_docs %}indicator{% endif %}" id="documents">{% tran
                 </section>
             {% endif %}
 
+            {# Questions/contactmomenten #}
```
jiromaykin, line 76: I think it would be a bit nicer to add an include here? So Something like `{% include 'pages/cases/questions.html' %}` This way it will be easier to swap the order of the contact-moment form with this new part, because in the design you can see they need to be reversed 
(questions first, then the contact-form all the way at the bottom of the page):

Also the contact-form should now be 100% the width of the container/column.

Screenshot 2024-03-04 at 16 43 38

jiromaykin (changes requested):

src/open_inwoner/templates/pages/cases/status_outer.html
pi-sigma: No longer needed if we don't use pagination

src/open_inwoner/templates/pages/cases/status_inner.html
pi-sigma: There is no contact form on this page, perhaps you're thinking of the vragen list view? Anyways, the suggestion sounds reasonable, if only to prevent the case detail template from growing too big.

stevenbal:
@pi-sigma is this PR ready for re-review?

pi-sigma:
@ Steven Yes, I just pushed latest changes. Forgot to put it in draft while I was working on it.

src/open_inwoner/js/components/card/ToggleHideSelection.js
```diff
@@ -0,0 +1,46 @@
...
+    this.node = node
+    this.button = document.querySelector('#toggle-hide-elements')
+    this.button?.addEventListener('click', this.toggleHide.bind(this))
+    this.icon = document.querySelector('.readmore__icon')
+    this.node.showMoreText = document.querySelector(
+      '.showmore__button-text'
+    ).textContent
+
+    // hide all but the first 3 cards by default
+    var allCards = document.querySelectorAll('.card')
```
stevenbal, line 14: While this is not an issue right now (probably), but this will select all cards on the page, which might cause issues if we reuse this somewhere else

src/open_inwoner/scss/components/Card/Card.scss
```diff
@@ -11,7 +15,7 @@
   border: var(--card-size-border) solid var(--card-color-border);
   border-radius: var(--border-radius);
   box-sizing: border-box;
-  display: flex;
+  // display: flex;
```
stevenbal, line 18: Not sure if we can just remove this without it causing problems elsewhere? @jiromaykin

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -1,4 +1,4 @@
-{% load i18n utils anchor_menu_tags card_tags file_tags grid_tags table_tags solo_tags link_tags button_tags icon_tags notification_tags %}
+{% load i18n utils anchor_menu_tags card_tags file_tags grid_tags list_tags table_tags solo_tags link_tags button_tags icon_tags notification_tags pagination_tags %}
```
stevenbal, line 1: `list_tags` and `pagination_tags` are unused

src/open_inwoner/templates/pages/questions/questions.html
```diff
@@ -0,0 +1,37 @@
+{% load i18n icon_tags list_tags %}
+
+<div class="contactmomenten">
+<h2 class="h3 contactmomenten__title">{% trans "Vragen" %}</h2>
```
stevenbal, line 4: `<h2>` that has `h3` as its class, not sure if that is intentional?

stevenbal (changes requested):
I'm not sure what @jiromaykin and @Bartvaderkin think of this, but generally I'd prefer it if PR feedback is added in a new commit instead of amending the original commits (unless it's small trivial stuff). That makes it easier to go over the processed feedback and see what changed

src/open_inwoner/templates/pages/questions/questions.html
jiromaykin: If this is a reusable component, it would be better to only use its own unique classes, like contactmomenten__title, and check if it has the correct heading-font and font-size like in the design.

src/open_inwoner/js/components/card/ToggleHideSelection.js
```diff
@@ -0,0 +1,46 @@
+export class ToggleHideSelection {
```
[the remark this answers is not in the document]
pi-sigma: The `hide-me` attribute was toggled for the wrong target. Strange it worked for me previously (though not after coming back to it).

I also noticed that the questions are in the wrong order (they should be reversed). Do you think it's safe to simply order by `registratiedatum` in reverse? The `registratiedatum` of a `Contactmoment` is apparently optional, but I haven't seen any without date.

src/open_inwoner/js/components/card/ToggleHideSelection.js
```diff
@@ -0,0 +1,46 @@
...
+
+    // toggle button text + icon
+    for (var i = 0; i < cardsTail.length; i++) {
+      if (cardsTail[i].classList.contains('hide-me')) {
+        this.icon.textContent = 'expand_more'
+        this.button.innerHTML = this.icon.outerHTML + this.node.showMoreText
+        break
+      } else {
+        this.icon.textContent = 'expand_less'
+        this.button.innerHTML = this.icon.outerHTML + 'Minder toon'
```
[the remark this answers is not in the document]
jiromaykin: We are using very, very little translation in JS currently, and yes, you would need dataset here.
We do use translation in the accessibility header, but that uses innerText which is often not a safe thing to use.
We're also using JS translation for FileInput by way of `<ul class="file-list__list" aria-live="polite" data-label-delete="{% trans 'Verwijderen' %}"></ul>` en then in JS like this: `const labelDelete = this.getFilesList().dataset.labelDelete || 'Delete'`
But I made a much easier solution when I created the expand for  the ReadMore component, where I would either show/hide the already translated span: https://github.com/maykinmedia/open-inwoner/blob/c677c1df82aedd6807a3edb1a3e540a8d000f37d/src/open_inwoner/templates/pages/product/detail.html#L43

src/open_inwoner/templates/pages/questions/questions.html
pi-sigma: I was following the pattern of `status` heading on the same page, which is also an `h2` with class attribute `h3`. I realize this is strange and I commented on this myself before. However, we ought to be consistent not just with the design but also with our own templates. I thought @jiromaykin was going to take care of `h2` elements as part of a later sprint (our `h1` are already styled with `utrecht-heading`), that's why I wrote it as is.

Bartvaderkin (changes requested):
I added some style notes. Also don't forget about `autospec=True` if you mock things where the signature could change (everywhere? don't know)

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -2081,3 +2230,68 @@ def test_request_error_in_connecting_doc_with_zaak_shows_proper_message(
                 f"Een fout is opgetreden bij het uploaden van {self.uploaded_informatie_object['bestandsnaam']}"
             ),
         )
+
+    def test_kcm_redirect(self, m):
+        """Check redirect from question embedded in case detail to klant_contactmoment detail"""
+
+        self._setUpMocks(m)
```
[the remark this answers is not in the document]
pi-sigma: Yes, the test needs the `contactmoment` mock, which is also used elsewhere in the suite. I'm defining additional mocks inside the test only if they aren't needed anywhere else. At first I simply added these new mocks to `setUp`, which would install them for every test, but I got complaints about missing mock addresses.

I thought it would make sense to install these extra mocks only where they're needed, but if it's poor style I can move them back into the general `setUp` function.

stevenbal (changes requested):
Some minor remarks still, looks good otherwise

src/open_inwoner/cms/cases/tests/test_htmx.py
```diff
@@ -43,6 +45,9 @@
 
 @tag("e2e")
 @requests_mock.Mocker()
+@patch.object(
+    ContactmomentenClient, "retrieve_objectcontactmomenten_for_zaak", autospec=True
+)
```
Bartvaderkin, line 50: See my other note. Why is this here and not a request mock?

src/open_inwoner/cms/cases/tests/test_contactform.py
```diff
@@ -39,6 +42,9 @@
 
 
 @requests_mock.Mocker()
+@patch.object(
+    ContactmomentenClient, "retrieve_objectcontactmomenten_for_zaak", autospec=True
+)
```
Bartvaderkin, line 47: This looks odd; in most cases we mock API responses through the request mocker instead of mocking the API client. 

Also why is it here at class level? Does every test function need this? It adds a lot of noise to the signatures but in many of the test functions I don't see the mock object being used or asserted.

If it is needed to get an empty list response then please add an explicit `return_value=[]` to the patch so it is more clear, and decide if this really has to be at class level.

jiromaykin (changes requested):

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -211,3 +212,31 @@ def get_context_data(self, **kwargs):
             },
         ]
         return ctx
+
+
+class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
+    """
+    Redirect to `KlantContactMomentDetailView` on the basis of contactmoment uuid
+    """
```
Bartvaderkin, line 220: Is this view really needed? I'd expect any place we have a contactmoment we'd also have the klantcontactmoment.

It feels weird we need to have a view just for this lookup, especially if we're going to to the same requests at the place it redirects to.

Bartvaderkin (changes requested):

src/open_inwoner/cms/cases/tests/test_contactform.py
stevenbal: I asked if it could be moved to the class level, because it seemed like none of these tests actually test this behavior (they test the contactform instead) and we just want to mock it out for all tests in this class

src/open_inwoner/cms/cases/tests/test_htmx.py
pi-sigma: I had a mock earlier but it wasn't actually used (only returned an empty list), so I switched to `patch` per Steven's request.

Bartvaderkin: Can you add a `return_value=[]` so it is explicit we want this to return an empty list?

src/open_inwoner/cms/cases/tests/test_contactform.py
Bartvaderkin: Ok fine, but it is kinda noisy to have to boilerplate this on every test function, while the empty mock would be part of the mocks, especially if it isn't even important to the tests.

src/open_inwoner/accounts/views/contactmoments.py
pi-sigma: Case detail view is getting crowded, so it was suggested to move this into a separate view.

Bartvaderkin: Do you have more context then "it was suggested"?

Anyway, the question was why we need a redirect view to do a lookup for information we have both before and after the redirect? I'd expect you could make URLs with the klantcontactmoment uuid.

stevenbal: The issue is that if we do this in the case detail view, we would have to do a bunch of fetches to get the associated `KlantContactMoment`en for each `ObjectContactMoment.contactmoment`, because we need the `KlantContactMoment.uuid` to build the URL for the Mijn vraag detail view. Because the case detail view is already quite massive, we figured it would be a better idea to have an intermediate view to resolve this

Bartvaderkin: Ah ok. I assumed we only show ContactMomenten for a Zaak for this user's Klant so we'd have the KlantContactMoment, but we show all ContactMomenten for a Zaak so then this makes sense.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -141,6 +141,59 @@ def get_context_data(self, **kwargs):
...
+                    "voorkeurstaal": "Avo",
+                    "vorig_contactmoment": None,
+                    "volgend_contactmoment": None,
+                    "onderwerp_links": ["https://www.kramer.biz/"],
+                    "initiatiefnemer": "klant",
+                    "medewerker": "https://davis.com/",
+                }
+                questions.append(factory(ContactMoment, data))
+            questions.sort(key=lambda q: q.registratiedatum, reverse=True)
+
```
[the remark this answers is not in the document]
pi-sigma: Hideous, I know, but atm it's difficult for Jiro (or anyone else) to actually test the toggle/hide behavior, since our mocks don't provide enough contactmomenten.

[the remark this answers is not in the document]
Bartvaderkin: If this is the apimocks in the repo then we could add more. I think it'd work to just duplicate the items in the list resources (without even changing the uuids etc)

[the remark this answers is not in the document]
pi-sigma: The uuids need to be checked as well. I've created a separate ticket, though it's not urgent ([https://taiga.maykinmedia.nl/project/open-inwoner/issue/2196](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2196)).

src/open_inwoner/js/components/card/ToggleHideSelection.js
```diff
@@ -0,0 +1,60 @@
+export class ToggleHideSelection {
```
jiromaykin, line 1: Currently the logic here makes it so that ALL cards get the hide-me class when expanded, but i am not sure why - the scope is very tangled.
Also we do not actually seem to need a toggle for the expand-button (so there would be no need for toggling data-attributes).

I can make all cards show with this script and remove the button like this, but it is not pretty:

```export class ToggleHideSelection {
  static selector = '.card--toggle-hide'

  constructor() {
    this.icon = document.querySelector('.expand-icon')
    this.button = document.getElementById('toggle-hide-elements')
    this.button.addEventListener('click', this.toggleHide.bind(this))
    this.button.innerHTML =
      this.icon.outerHTML + this.button.dataset.labelReveal

    // Hide all but the first three cards by default
    const allCards = document.querySelectorAll(ToggleHideSelection.selector)
    const cardsTail = Array.from(allCards).slice(3)

    if (cardsTail.length === 0) {
      this.button.classList.add('hide-me')
    } else {
      cardsTail.forEach((element) => {
        element.classList.add('hide-me')
      })
    }
  }

  toggleHide(event) {
    event.preventDefault()

    // Toggle the expand/collapse state of the button
    const isExpanded = this.button.getAttribute('aria-expanded') === 'false'
    this.button.setAttribute('aria-expanded', isExpanded ? 'true' : 'false')
    this.icon.textContent = isExpanded ? 'expand_more' : 'expand_less'
    this.button.innerHTML =
      this.icon.outerHTML +
      (isExpanded
        ? this.button.dataset.labelReveal
        : this.button.dataset.labelCollapse)

    if (isExpanded) {
      console.log(`Cards are expanded`)
      this.button.remove()
    }

    // Toggle 'hide-me' class on cards beyond the first three
    const allCards = document.querySelectorAll(ToggleHideSelection.selector)
    const cardsTail = Array.from(allCards).slice(3)

    cardsTail.forEach((element) => {
      element.classList.toggle('hide-me', !isExpanded)
    })
  }
}

document
  .querySelectorAll(ToggleHideSelection.selector)
  .forEach((toggleHide) => new ToggleHideSelection(toggleHide))
```

jiromaykin (dismissed):
I am not sure how to fix the tangled javascript, but if we change the 'dropdown' class into an 'expand' class than this may work as an intermediate solution.
Also some small requests to change the Card SCSS and some html classes for the primary buttons.

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -58,19 +58,29 @@ <h2 class="h2" id="documents-upload">{% trans "Document toevoegen" %}</h2>
...
                 <h2 class="h2 {% if case.new_docs %}indicator{% endif %}" id="documents">{% trans 'Eerder toegevoegde documenten' %}</h2>
                 {% case_document_list case.documents %}
             {% endif %}
 
-            {#  Contact moment form #}
-            {% if case.contact_form_enabled %}
+            {# Questions/contactmomenten #}
+            {% if case.questions %}
+                {% include "pages/questions/questions.html" with questions=case.questions only %}
+                <div class="dropdown dropdown--spaced">
```
jiromaykin, line 70: This should be <div class="expand"> so you do not need to change the dropdown CSS, and move .hide-me to .card.

src/open_inwoner/js/components/card/ToggleHideSelection.js
```diff
@@ -0,0 +1,54 @@
...
+      console.log(`Cards are expanded`)
+      this.button.remove()
+    }
+
+    // Toggle 'hide-me' class on cards beyond the first three
+    const allCards = document.querySelectorAll(ToggleHideSelection.selector)
+    const cardsTail = Array.from(allCards).slice(3)
+
+    cardsTail.forEach((element) => {
+      element.classList.toggle('hide-me', !isExpanded)
```
stevenbal, line 47: I'm not sure if it's just me, but adding the `force` flag here actually causes the expand button to do nothing, why do we need it?

pi-sigma: This worked for me just a few seconds ago (and seems to have worked for Jiro). I modified the code to make the toggle work (without it the spacing/padding is messed up when the button is hidden) and show the total number of questions, as in the design. Give it another try and let me know if this works now.

requirements/dev-local.txt
```diff
@@ -0,0 +1,3 @@
+ipython
+pdbr
+pytype
```
Bartvaderkin, line 3: I don't think this belongs in the shared dependencies as it is personal tooling.

pi-sigma: Right, missing from my global .gitignore

src/open_inwoner/js/components/card/ToggleHideSelection.js
stevenbal: Checked it, works for me now as well :+1:

stevenbal (approved):

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -456,8 +533,8 @@ def setUp(self):
                     "info_id": self.informatie_object["uuid"],
                 },
             ),
-            created=datetime.datetime.fromisoformat(
-                self.zaak_informatie_object["registratiedatum"]
+            created=dateutil.parser.parse(
+                self.zaak_informatie_object_old["registratiedatum"]
```
Bartvaderkin, line 537: Note for other reviewers: we discussed this and `dateutil.parser.parse` is acceptable here because the zgw_consumers `factory()` also uses that and it produces different timezones then `datetime.fromisoformat` so the equality check failed. 

There might be a way to convert it but let's declare that out of scope for this PR because it is a hassle.

Bartvaderkin (approved):
:+1: Let's go! :rocket:

PR #1061, opened 2024-02-29
:bug: Avoid necessary fields redirect for media files

stevenbal:
Same fix as https://github.com/maykinmedia/open-inwoner/pull/1052, forgot to apply it to `NecessaryFieldsMiddleware`

### DIFF
--- a/src/open_inwoner/accounts/middleware.py
+++ b/src/open_inwoner/accounts/middleware.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.conf import settings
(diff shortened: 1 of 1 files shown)


stevenbal:
@alextreme I figured that until we change this middleware setup, it would be good to make sure this doesn't happen in the current setup

alextreme:
I guess this only applies for dev-environments (as static files are served by Nginx). Can be checked by Bart and merged imho

Bartvaderkin (approved):
Smooth development is worth it.

PR #1062, opened 2024-02-29
Fixed user save-signal issue with Haalcentraal BRP retrieval logging

Bartvaderkin:
This was kind of gnarly:

1. We can't use `pre_save` signal because we need a saved user
2. Calling save() from either a `pre_save` or `post_save` handler creates a signal loop

Solution: use `post_save` so we have a saved user, and break the loop.

Downside is this will trigger any other `post_save` handler multiple times, which is a vote to centralize signal based user update stuff. Most of it is currently `user_logged_in` so not an issue.

Note this PR targets the open Django 4.x branch #1024

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -44,7 +44,7 @@ def log_user_login(sender, user, request, *args, **kwargs):
 
     if user.login_type == LoginTypeChoices.digid:
         if brp_config.service:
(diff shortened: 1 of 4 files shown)


pi-sigma (approved):
Nice idea with checking/setting the flag in `on_bsn_change`! Strange how this could pass with Django 3.2 and only fail with the upgrade, but this seems like the right behavior.

PR #1063, opened 2024-02-29
[#2058] Added `status_text` attribute to FeedItem, fixed issue in case_status item

Bartvaderkin:
Followup on #1058

### DIFF
--- a/src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
+++ b/src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
@@ -16,14 +16,21 @@
     </div>
     <div class="card-container card-container--columns-2">
         {% for item in userfeed.items %}
-        <a href="{{ item.action_url }}" class="card {% if item.action_required %}card--status{% endif %} {% if item.is_completed %}card--completed{% endif %}">
+        <a href="{{ item.action_url }}" class="card card--status card--status-{{ item.status_indicator }} {% if item.is_completed %}card--completed{% endif %}">
+        {% if item.action_required %}
+            <div class="userfeed-marker userfeed-marker--warning">
+                {% icon icon="warning_amber" icon_position="after" extra_classes="icon--warning" outlined=True %}
+                <span class="userfeed__status_text">{{ item.status_text }}</span>
+            </div>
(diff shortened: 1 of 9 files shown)


src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -16,14 +16,21 @@ <h2 class="h2 {% if userfeed.action_required %}indicator{% endif %}">
...
+            <div class="userfeed-marker userfeed-marker--warning">
+                {% icon icon="warning_amber" icon_position="after" extra_classes="icon--warning" outlined=True %}
+                <span class="userfeed__status_text">{{ item.status_text }}</span>
+            </div>
+        {% else %}
+            <div class="userfeed-marker userfeed-marker--info">
+                {% icon icon="info" icon_position="after" extra_classes="icon--default" outlined=True %}
+                <span class="userfeed__status_text">{{ item.status_text }}</span>
+            </div>
+        {% endif %}
```
Bartvaderkin, line 30: @jiromaykin 

The status in the classnames should be dynamic, eg: `userfeed-marker--{{ item.status_indicator }}` or  `icon-{{ item.status_indicator }}`.  

So not an if/else switched on action_required, just use the value from the item.

So you have `warning`, `info`, `error` and `success` versions of the classes. The default is `info`. These are the values from `open_inwoner.openzaak.constants.StatusIndicators`

jiromaykin: This means I still needed to add conditions to retrieve the correct icons in my most recent commit, 
as I don't seem to be able to do something like this here: 
`{% include "components/StatusIndicator/StatusIndicator.html" with status_indicator=item.status_indicator .....`

jiromaykin (approved):
Needs to be tested with real-life data.

stevenbal (approved):

src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
Bartvaderkin: Feels weird that doesn't work.

PR #1065, opened 2024-02-29
[#2245] Registration notifications redesign

jiromaykin:
Design: https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=2217%3A9149&mode=design&t=ObhqwB8PWC3lvpzO-1

Note: instead of the Accent color, we decided to use the secondary color for all radio/checkbox inputs.

userstory : https://taiga.maykinmedia.nl/project/open-inwoner/us/2210
new issue number: https://taiga.maykinmedia.nl/project/open-inwoner/task/2245

### DIFF
--- a/src/open_inwoner/js/components/form/ChoiceListMultiple.js
+++ b/src/open_inwoner/js/components/form/ChoiceListMultiple.js
@@ -1,0 +1,34 @@
+export class ChoiceListMultiple {
+  static selector = '.choice-list-multiple .choice-list-multiple__item'
+
+  constructor(listitem) {
+    this.listitem = listitem
+    this.checkbox = listitem.querySelector(
+      '.choice-list-multiple input[type="checkbox"]'
+    )
+    this.label = listitem.querySelector(
+      '.choice-list-multiple .choice-list-multiple__item .checkbox__label'
+    )
+
+    if (this.checkbox.checked) {
+      this.listitem.classList.add('selected')
+    }
+
+    this.listitem.addEventListener(
+      'click',
+      this.toggleChoiceItemBorder.bind(this)
+    )
+  }
+
+  toggleChoiceItemBorder() {
+    if (this.checkbox.checked) {
+      this.listitem.classList.add('selected')
+    } else {
+      this.listitem.classList.remove('selected')
+    }
+  }
+}
+
+document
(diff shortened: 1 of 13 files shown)


src/open_inwoner/js/components/form/ChoiceListMultiple.js
```diff
@@ -0,0 +1,35 @@
...
+      '.choice-list-multiple input[type="checkbox"]'
+    )
+    this.label = listitem.querySelector(
+      '.choice-list-multiple .choice-list-multiple__item .checkbox__label'
+    )
+    // Add 'selected' parentclass if checkbox is already checked onLoad
+    if (this.checkbox.checked) {
+      this.listitem.classList.add('selected')
+    }
+    // If selector exists
```
pi-sigma, line 16: Update/remove comment

pi-sigma (changes requested):
When I deselect one of the options, the text becomes underlined. It stays that way when I re-select the option and only disappears when click outside the card.

Screenshot from 2024-03-25 12-35-01

Is this intended?

Also, wouldn't it make sense to use the same design for changing notification prefs in the profile?

pi-sigma (approved):

alextreme:
> Attentie @alextreme : This one has been approved, but... here i assumed that checkboxes have to wrap both the Label and the info text - and we may not want that (?) because of the future checkbox designs.

If you want to revisit this feel free to do so, I've removed Paul's approval

jiromaykin:
@Bartvaderkin @pi-sigma I've removed the `...p {{ help_extra_cls }}` class from the Checkbox template since I think it's superfluous.

jiromaykin:
@pi-sigma This one is now fully ready for review.

pi-sigma (changes requested):
Why is the siteimprove commit part of this PR? Otherwise looking good.

jiromaykin:
> Why is the siteimprove commit part of this PR? Otherwise looking good.

Just a little explanation for future questions: The Siteimprove tracker has to track _extremely_ specific elements in the site. Like, for example, the second checkbox inside a list of 3. So if the structure/class of the tracked element changes, I will also have to update the tracker code... It's going to be a 'challenge' to not forget to do this with future changes.

PR #1066, opened 2024-03-01
[#2116] Added 'contact_phonenumber' field to SiteConfiguration, mail context & mail templates

Bartvaderkin:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2116


src/open_inwoner/configurations/models.py
```diff
@@ -19,7 +19,7 @@
 from ..utils.css import clean_stylesheet
 from ..utils.fields import CSSField
 from ..utils.files import OverwriteStorage
-from ..utils.validators import FilerExactImageSizeValidator
+from ..utils.validators import DutchPhoneNumberValidator, FilerExactImageSizeValidator
```
jiromaykin, line 22: Hopefully this validator also accepts the short phonenumbers municipalities now use? Like Groningen only uses "14 050" without an area code etc.

src/open_inwoner/conf/parts/maileditor.py
```diff
@@ -233,7 +233,7 @@ def _readfile(name):
                 "description": _("Name of the site"),
             },
             {
-                "name": "contactpage",
+                "name": "contact_page",
```
jiromaykin, line 236: This would need 'contact_phonenumber' as well, but I can add that later myself, since some parts of the templates will be changing a bit.

jiromaykin (approved):
We communicated and apparently shortened phonenumbers are valid as well, so approving now, so we can use this.

PR #1067, opened 2024-03-01
♻️ [#2150] Throttle user emails based on template name

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2150


src/open_inwoner/openzaak/models.py
```diff
@@ -563,11 +564,11 @@ class Meta:
...
         ]
 
-    def has_received_similar_notes_within(self, period: timedelta) -> bool:
+    def has_received_similar_notes_within(
+        self, period: timedelta, template_name: str
+    ) -> bool:
         return UserCaseStatusNotification.objects.has_received_similar_notes_within(
-            self.user, self.case_uuid, period, not_record_id=self.id
-        ) or UserCaseInfoObjectNotification.objects.has_received_similar_notes_within(
-            self.user, self.case_uuid, period
```
stevenbal, line 570: I wasn't too sure about removing this, but in practice we never send `UserCaseStatusNotification`s that have the same `template_name` as `UserCaseInfoObjectNotification`s

src/open_inwoner/openzaak/tests/test_notification_zaak_status.py
```diff
@@ -569,24 +569,34 @@ def test_handle_status_update(self, mock_send: Mock, mock_feed_hook: Mock):
         case = factory(Zaak, data.zaak)
         case.zaaktype = factory(ZaakType, data.zaak_type)
 
-        status = factory(Status, data.status_final)
-        status.statustype = factory(StatusType, data.status_type_final)
+        status1 = factory(Status, data.status_initial)
+        status1.statustype = factory(StatusType, data.status_type_initial)
+
+        status2 = factory(Status, data.status_final)
+        status2.statustype = factory(StatusType, data.status_type_final)
```
Bartvaderkin, line 576: A general feedback: let's stop arbitrary numbering our test data and use meaningful postfixes so things stay readable. 

In this case you don't even have to invent a scheme but can just copy form the data class, eg: `status1` is `status_initial` and `status2` is `status_final` (as well as the related objects).

Bartvaderkin (changes requested):
It feels a little inaccurate and kludgy to reuse/co-opt `template_name` for this and not expand to a more general `notification_type`. 

I don't know if we want to block on that but it feels like one of those things where a convenient quickfix shortcut now gives headaches in the future.

src/open_inwoner/openzaak/models.py
Bartvaderkin: "in practice.. (at this point in time)", famous last words. It's is probably fine but an assumption and not a hard enforced rule. Let's see what happens.

Bartvaderkin:
Something like that, the difference between like "throttle notifications per message type" instead of a "throttle notifications based on what template they happen to use because that's the only difference right now", but I don't know if we want to go for that.

Bartvaderkin:
Proposal to rename the field on the model to something like 'collision_key' ofzo 'dedupe_key' or something like that (and still use the template name but could be anything).

stevenbal:
@Bartvaderkin renamed `template_name` to `collision_key`

Bartvaderkin (approved):

PR #1074, opened 2024-03-04
:card_file_box: [#2041] Added hard database constraints on User's identifier/login_type fields

stevenbal:


src/open_inwoner/accounts/models.py
```diff
@@ -278,11 +278,11 @@ class Meta:
                 | Q(login_type=LoginTypeChoices.eherkenning),
                 name="check_kvk_or_rsin_only_set_when_login_eherkenning",
             ),
-            CheckConstraint(
-                check=((Q(kvk="") & ~Q(rsin="")) | (~Q(kvk="") & Q(rsin="")))
-                | ~Q(login_type=LoginTypeChoices.eherkenning),
-                name="check_kvk_or_rsin_exclusive_when_login_eherkenning",
-            ),
+            # CheckConstraint(
```
stevenbal, line 281: I left this one out for now, because our current implementation uses both KVK and RSIN (OpenZaak requires RSIN and eSuite KVK)

src/open_inwoner/accounts/models.py
```diff
@@ -246,30 +244,44 @@ class Meta:
...
+            ),
             # CheckConstraint(
-            #     # maybe this is not correct?
-            #     check=(Q(bsn="") | Q(bsn__isnull=True))
-            #     | ~Q(login_type=LoginTypeChoices.digid),
-            #     name="check_digid_bsn_required_when_digid",
+            #     check=((Q(kvk="") & ~Q(rsin="")) | (~Q(kvk="") & Q(rsin="")))
+            #     | ~Q(login_type=LoginTypeChoices.eherkenning),
+            #     name="check_kvk_or_rsin_exclusive_when_login_eherkenning",
             # ),
```
stevenbal, line 285: I left this one out, because our current implementation uses both KVK and RSIN (OpenZaak requires RSIN and eSuite KVK)

Bartvaderkin: This rule enforces that when a user is using eherkenning you can't have both kvk and rsin set on the same instance.

This was an assumption for review at the time I made this, I'm not 100% into eherkenning so maybe it is legal to have both at same time?

src/open_inwoner/openklant/tests/data.py
```diff
@@ -38,10 +39,13 @@ def setUpServices(cls):
 
 class MockAPIReadPatchData(MockAPIData):
     def __init__(self):
+        User.objects.filter(email="old@example.com").delete()
```
Bartvaderkin, line 42: These delete()'s feel a bit kludgey. Are we sure this is needed or if we can fix the tests that want this?

Bartvaderkin (changes requested):

src/open_inwoner/accounts/models.py
stevenbal: It is legal to have both: https://www.kvk.nl/over-het-handelsregister/rsin-nummer/

Apparently all companies (except eenmanszaken) have both KVK nummer and RSIN. I'll create a separate issue to check how we handle the eenmanszaak situation if we connect to ZGW apis right now, because I'm fairly sure that won't work with our current implementation

Bartvaderkin: Cool, thanks for checking. Then we could remove the commented-out constraint and replace it with a text comment that reminds the next reader we explicitly didn't constrained this situation because it is legal to have both.

stevenbal: On second thought, this eenmanszaak edge case might mean that we cannot add the constraint that users can only have a BSN when they log in using DigiD, because we'll (probably?) have to use the BSN of the owner of the eenmanszaak to fetch Zaken etc.

src/open_inwoner/accounts/tests/test_user_constraints.py
```diff
@@ -0,0 +1,121 @@
...
+
+class UserIdentifierConstraintTests(TestCase):
+    def test_unique_email_when_login_default(self):
+        UserFactory(email="foo@example.com", login_type=LoginTypeChoices.default)
+        UserFactory(email="bar@example.com", login_type=LoginTypeChoices.default)
+
+        with self.assertRaises(IntegrityError):
+            UserFactory(email="foo@example.com", login_type=LoginTypeChoices.default)
+
+    def test_not_unique_email_when_login_digid_or_eherkenning(self):
```
pi-sigma, line 19: We currently have a test that asserts that DigiD users cannot register with a duplicate email that belongs to a non-DigiD user (https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/tests/test_auth.py#L1151). Is that still relevant (and more generally, the entire `DuplicateEmailRegistrationTest` class)?

src/open_inwoner/cms/cases/tests/test_contactform.py
```diff
@@ -442,6 +443,9 @@ def test_form_success_with_api_eherkenning_user(self, m):
             with self.subTest(
                 use_rsin_for_innNnpId_query_parameter=use_rsin_for_innNnpId_query_parameter
             ):
+                # FIXME for some reason creating the user outside of the loop
+                # makes the second iteration fail?
+                User.objects.filter(kvk="12345678").delete()
```
pi-sigma, line 448: Same/similar issue as in https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/tests/test_profile_views.py#L547, where the mock has to be recreated because it is overriden somehow? Something's up with our mock machinery that needs to be fixed, though not sure if this can be done as part of your ticket; perhaps create a new one for this.

pi-sigma (changes requested):
Not sure if something needs to be done about `DuplicateEmailRegistrationTest`; take a quick look to confirm.

src/open_inwoner/cms/cases/tests/test_contactform.py
Bartvaderkin: Agree these delete()'s are a bit smelly.

src/open_inwoner/accounts/tests/test_user_constraints.py
stevenbal: It does make me wonder why we relaxed the constraint for `email` though :thinking:

Bartvaderkin: In practice some Digid users are sharing the same email, like couples or caregivers helping someone else with their digital issues.

src/open_inwoner/openklant/tests/data.py
stevenbal: I removed them and fixed the issues for the tests that needed it

pi-sigma (approved):

src/open_inwoner/accounts/tests/test_user_constraints.py
```diff
@@ -0,0 +1,121 @@
...
+
+        with self.assertRaises(IntegrityError):
+            UserFactory(rsin="12345678", login_type=LoginTypeChoices.eherkenning)
+
+    @skip("Not yet implemented")
+    def test_not_both_kvk_and_rsin_when_login_eherkenning(self):
+        with self.assertRaises(IntegrityError):
+            UserFactory(
+                kvk="12345678", rsin="12345678", login_type=LoginTypeChoices.eherkenning
+            )
```
Bartvaderkin, line 58: Should this test be changed and enabled to test this is allowed?

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -548,9 +551,14 @@ def test_eherkenning_user_updates_klant_api(self, m):
                 # `m` is overridden somewhere, which causes issues when `MockAPIReadPatchData.install_mocks`
                 # is run for the second time
                 with requests_mock.Mocker() as m:
-                    data = MockAPIReadPatchData().install_mocks_eherkenning(
+                    # Reset the email, otherwise we get errors that a user already exists
+                    # with the updated email
+                    data.eherkenning_user.email = original_email
```
Bartvaderkin, line 556: This could have been/become two tests with less hassle but it's your choice.

Bartvaderkin (changes requested):
Some small notes, and I wonder if we covered the eenmanszaken situation from an earlier comment (if we always have kvk and rsin, or if eenmanszaken can have a bsn but not kvk or rsin).

src/open_inwoner/accounts/tests/test_profile_views.py
stevenbal: Yeah it's becoming quite messy, I split it into two

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -540,52 +541,78 @@ def test_modify_phone_and_email_updates_klant_api(self, m):
...
+
+        config = OpenKlantConfig.get_solo()
+        config.use_rsin_for_innNnpId_query_parameter = False
+        config.save()
+
+        response = self.app.get(self.url, user=data.eherkenning_user)
+
+        # reset noise from signals
+        m.reset_mock()
+        self.clearTimelineLogs()
```
Bartvaderkin, line 595: Same as above

Bartvaderkin (changes requested):

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -541,52 +541,78 @@ def test_modify_phone_and_email_updates_klant_api(self, m):
...
+
+        config = OpenKlantConfig.get_solo()
+        config.use_rsin_for_innNnpId_query_parameter = True
+        config.save()
+
+        response = self.app.get(self.url, user=data.eherkenning_user)
+
+        # reset noise from signals
+        m.reset_mock()
+        self.clearTimelineLogs()
```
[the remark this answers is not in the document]
stevenbal: I removed the `reset_mock`, but `clearTimelineLogs` is actually still needed (it's also used in other tests in this file that don't have subtests)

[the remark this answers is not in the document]
Bartvaderkin: Existing code may or may not be perfect.

Could be cool to see why it happens but if you tested it fails if we remove it I'll believe it (could be because the code earlier in the tests generates similar logs)

Bartvaderkin:
@stevenbal Are we sure about the `check_bsn_only_set_when_login_digid` rule? In relation to the eherkenning & eenmanszaken from earlier?

alextreme:
The suggestion (OpenZaak + RSIN + eHerkenning -> dont allow eenmanszaken to log in via eHerkenning) is sufficient for now and is already in the project. For eenmanszaken BSN == RSIN, so logging in via DigiD would give them the same information.

stevenbal:
@Bartvaderkin I rebased this on master, can you re-review/approve this? :)

alextreme:
@swrichards could you take a look at this PR somewhere in the following weeks? No rush but now it is languishing

PR #1075, opened 2024-03-04
✨ [#2149] Display new answer header for mijn vragen list

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2149

This can be tested locally by connecting to the mock Klanten/Contactmomenten endpoints:
- http://localhost:8000/apimock/openklant-read/klanten/
- http://localhost:8000/apimock/openklant-read/contactmomenten/

Logging in with DigiD and navigating to `Mijn vragen`, this should should the new answer header for the vraag, which should disappear after visiting the detail page


src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -39,7 +52,8 @@ def setUpTestData(cls):
             config=OpenKlantConfig.get_solo(),
         )
 
-    def test_list_for_bsn(self, m):
+    # @patch("open_inwoner.accounts.views.contactmoments.get_local_kcm_mapping", autospec=True)
```
Bartvaderkin, line 55: Lingering comment here

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -273,11 +357,15 @@ def test_show_detail_for_bsn_with_zaak_reformat_esuite_id(self, m):
...
         )
 
-    def test_show_detail_for_kvk_or_rsin(self, m):
+    def test_show_detail_for_kvk_or_rsin(self, m, mock_get_local_kcm_mapping):
         for use_rsin_for_innNnpId_query_parameter in [True, False]:
             with self.subTest(
                 use_rsin_for_innNnpId_query_parameter=use_rsin_for_innNnpId_query_parameter
             ):
+                # Avoid having a `KlantContactMomentLocal` with the same URL for
+                # different users
```
Bartvaderkin, line 366: I think comments are somewhat exempt from line length rule. Like putting the last two words on the next line is arguably worse then making the line slightly longer.

src/open_inwoner/openklant/wrap.py
```diff
@@ -95,3 +96,27 @@ def get_fetch_parameters(request, use_vestigingsnummer: bool = False) -> dict:
...
         return parameters
     return {}
+
+
+def get_local_kcm_mapping(
+    kcms: list[KlantContactMoment], user: User
+) -> dict[str, KlantContactMomentLocal]:
+    existing_kcms = KlantContactMomentLocal.objects.filter(user=user).values_list(
+        "kcm_url", flat=True
+    )
```
Bartvaderkin, line 106: I think this should to be wrapped in a `set()`? Otherwise you're doing `in` into a queryset.

src/open_inwoner/templates/pages/contactmoment/list.html
```diff
@@ -34,6 +34,10 @@ <h2>{% trans "U heeft op dit moment nog geen vragen." %}</h2>
             {% for contactmoment in page_obj.object_list %}
                 {% render_column start=forloop.counter_0|multiply:4 span=4 %}
                 <div class="card card--compact card--stretch">
+                    {% if contactmoment.new_answer_available %}
+                        {% translate "Nieuw antwoord beschikbaar" as new_answer_text %}
+                        {% include "components/StatusIndicator/StatusIndicator.html" with status_indicator="info" status_indicator_text=new_answer_text %}
```
Bartvaderkin, line 39: I note there is no `only`

Bartvaderkin (changes requested):
Preliminary review as I'm only here in the morning. 

I note you removed some of the apimocks. I don't mind and I assume there is a good reason and Paul and Jiro don't need them (I heard they actually would like more for proper list development but that could be for another time)

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -69,12 +72,27 @@ class KCMDict(TypedDict):
...
+    def get_kcm_data(
+        self,
+        kcm: KlantContactMoment,
+        local_kcm_mapping: Optional[dict[str, KlantContactMomentLocal]] = None,
+    ) -> KCMDict:
+        _is_new = is_new(
+            kcm.contactmoment,
+            "registratiedatum",
+            timedelta(days=settings.CONTACTMOMENT_NEW_DAYS),
+        )
```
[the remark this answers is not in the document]
stevenbal: Renamed it to `instance_is_new`, wasn't sure how to make it more descriptive without the name getting out of hand :see_no_evil:

stevenbal:
> I note you removed some of the apimocks.

The problem was that because the endpoints just return what is in the files, the objectcontactmomenten can only belong to one of the `Contactmomenten`, meaning that the detailview for the second contactmoment raises assertionerrors

Bartvaderkin (approved):

pi-sigma:
I stumbled over the API mocks myself and created an issue for it: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2196

We can discuss on Monday if we want to further modify/expand them so that they're more useful.

pi-sigma (changes requested):
I don't know if the `AttributeError` is due to misconfiguration on my part or the missing mock.

A few remarks about the placement of `KlantcontactMomenten` in the admin the naming of classes/functions. Doesn't have to be changed, but double-check.

src/open_inwoner/openklant/models.py
```diff
@@ -127,3 +127,28 @@ class Meta(OrderedModel.Meta):
 
     def __str__(self):
         return self.subject
+
+
+class KlantContactMomentLocal(models.Model):
```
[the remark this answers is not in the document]
stevenbal: You're right, I changed it

stevenbal:
> I don't know if the AttributeError is due to misconfiguration on my part or the missing mock.

do you have the Klanten API configured in your local Open Klant configuration?

pi-sigma:
I do have the API configured but I receive a `401 Client Error` for the call to our `openklant-read` test API. Strange as it worked before and now doesn't, but the problem seems to be on my end.

pi-sigma (approved):

PR #1078, opened 2024-03-05
[#2175] Fix rounded corners for mail-templates

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2175

- check rounded corners for Gmail on Android

This fix does not use fallbacks, instead the issue was that the Table with rounded area got a white background, which cut through the rounded corners and made it seem not-rounded.
PS: the border-radiuses in the design are only 3px so not dramatically visible.

### DIFF
--- a/src/open_inwoner/templates/mail/_base.html
+++ b/src/open_inwoner/templates/mail/_base.html
@@ -102,8 +102,8 @@
             display: block;
         }
 
-        .td__alignright {
-            text-align: left;
+        .td__alignright, .td__rightcell {
+            text-align: left !important;
         }
     }
 </style>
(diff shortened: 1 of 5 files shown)


src/open_inwoner/templates/mail/_base.html
```diff
@@ -137,7 +137,7 @@
 
     .button--primary {
         background-color: {{ theming.primary_color }};
-        border: 1px solid {{ theming.primary_color }};
```
stevenbal, line 140: Was this space intentionally removed?

jiromaykin: Ew, it was sadness from cherrypicking... Will revert.

stevenbal (approved):

Bartvaderkin (approved):

PR #1080, opened 2024-03-07
:lipstick: [#2173] Make userfeed indicators/icons conditional

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2173

Should work when proper data is available,
I made the decision to still show a color banner in case there is no status_text set at all (also easier for the admin to spot if things need to be configured):

acties

### DIFF
--- a/src/open_inwoner/scss/components/UserFeed/UserFeed.scss
+++ b/src/open_inwoner/scss/components/UserFeed/UserFeed.scss
@@ -84,17 +84,12 @@
 
   .card--status {
     // Default color
-    padding: 0;
+    padding: var(--spacing-medium) 0 0 0;
     background-color: var(--color-info-lighter);
-  }
 
-  &-marker {
-    display: flex;
-    align-items: center;
-    border-top-left-radius: var(--border-radius);
-    border-top-right-radius: var(--border-radius);
-    color: var(--color-info-darker);
-    font-size: var(--font-size-body);
+    .card__header {
+      display: flex;
+    }
 
     &--warning {
       background-color: var(--color-danger-lightest);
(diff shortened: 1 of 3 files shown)


src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -16,35 +16,21 @@ <h2 class="h2 {% if userfeed.action_required %}indicator{% endif %}">
...
-                    {% icon icon="check_circle" outlined=True %}
-                {% elif item.status_indicator == "failure" %}
-                    {% icon icon="cancel" outlined=True %}
-                {% elif item.status_indicator == "warning" %}
-                    {% icon icon="warning_amber" outlined=True %}
-                {% else %}
-                    {% icon icon="info" outlined=True %}
+            <a href="{{ item.action_url }}" class="card card--status card--status--{{ item.status_indicator }} {% if item.is_completed %}card--completed{% endif %}">
+                {% if item.status_text %}
+                    {% include "components/StatusIndicator/StatusIndicator.html" with status_indicator=item.status_indicator status_indicator_text=item.status_indicator_text %}
```
stevenbal, line 21: This needs to be `status_text` to make sure it uses the right property: https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/userfeed/hooks/case_status.py#L106

```suggestion
                    {% include "components/StatusIndicator/StatusIndicator.html" with status_indicator=item.status_indicator status_indicator_text=item.status_text %}
```

src/open_inwoner/scss/components/UserFeed/UserFeed.scss
```diff
@@ -84,17 +84,12 @@
```
stevenbal, line 84: I assume the empty status indicator is supposed to scale with the height of the indicator with text on the same row, to keep the cards aligned. So this is probably intentional, right? Just wanted to double check
image

stevenbal (changes requested):

src/open_inwoner/scss/components/UserFeed/UserFeed.scss
jiromaykin: Oh dear, the Warning one on the right does not look good... Let's see if I can fix that for empty statuses.

jiromaykin: I think these below are the most common edge/weird cases for no-content/too-long content, but in reality there will likely always be a one-line status_text - if this 'marker' is not what we want for empty statuses, I will have to ask the designer next week:

Screenshot 2024-03-07 at 15 52 20

PR #1086, opened 2024-03-11
[#2106] New contact-form mail-template

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2106

We might miss a couple of data things for this design + Some sentences in the design are not correct so I changed them for now, may be changed later again.

This is the mail-template for clients, not for municipality-staff.

### DIFF
--- a/src/open_inwoner/conf/parts/contactform_registration.html
+++ b/src/open_inwoner/conf/parts/contactform_registration.html
@@ -1,29 +1,89 @@
-<p>Beste</p>
+<table role="presentation" border="0" cellspacing="0" cellpadding="0" width="100%" class="table-mail">
+    <tbody>
+    <tr>
+        <td class="td-mail td__padding-all" colspan="6">
+            <p>Beste inwoner,</p>
+            <p>Wij hebben uw vraag ontvangen.</p>
+        </td>
+    </tr>
+    <tr>
+        <td class="spacing__extra" colspan="6">
+            &nbsp;
+        </td>
+    </tr>
+    <tr class="tr--responsive">
+        <td class="td--responsive td-mail td-mail__bg-success td__padding-all td__leftcell" colspan="3">
+            <p><span class="text-color__small-gray-900">Vraag: </span></p>
+            <h2>{{ subject }} / {{ question }}</h2>
+        </td>
+        <td class="td--responsive td-mail td-mail__bg-success td__padding-all td__rightcell td__alignright" colspan="3">
+            <p><!-- <span class="text-color__small-gray-900">Vraagnummer: </span><br/>{{ identification }} --></p>
+        </td>
+    </tr>
 
-<table>
+    <!-- Table with different column-widths -->
     <tr>
-        <th>Onderwerp:</th>
-        <td>{{ subject }}</td>
+        <td class="td-mail td-mail__bg-success td__padding-all" colspan="6">
+            <div class="td-mail__bg-white">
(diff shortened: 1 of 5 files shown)


src/open_inwoner/conf/parts/maileditor.py
```diff
@@ -410,6 +411,28 @@ def _readfile(name):
                 "name": "question",
                 "description": _("Vraag"),
             },
+            {
+                "name": "site_name",
+                "description": _("Name of the site"),
+            },
+            {
+                "name": "login_page",
```
jiromaykin, line 419: @Bartvaderkin This should probably be `question_link` instead of the login_page

src/open_inwoner/conf/parts/maileditor.py
```diff
@@ -410,6 +411,28 @@ def _readfile(name):
                 "name": "question",
```
jiromaykin, line 411: @Bartvaderkin We do not want the question itself to actually become visible in the email, but we do need the subject and perhaps the question number.

src/open_inwoner/conf/parts/contactform_registration.html
```diff
@@ -1,29 +1,89 @@
...
+            &nbsp;
+        </td>
+    </tr>
+    <tr class="tr--responsive">
+        <td class="td--responsive td-mail td-mail__bg-success td__padding-all td__leftcell" colspan="3">
+            <p><span class="text-color__small-gray-900">Vraag: </span></p>
+            <h2>{{ subject }}</h2>
+        </td>
+        <td class="td--responsive td-mail td-mail__bg-success td__padding-all td__rightcell td__alignright" colspan="3">
+            <p><!-- <span class="text-color__small-gray-900">Vraagnummer: </span><br/>{{ identification }} --></p>
```
jiromaykin, line 20: @Bartvaderkin I commented out the stuff for the question-number/identification for now.

src/open_inwoner/conf/parts/contactform_registration.html
```diff
@@ -1,29 +1,89 @@
...
+    </tr>
+    <tr>
+        <td class="spacing__extra" colspan="6">
+            &nbsp;
+        </td>
+    </tr>
+    <tr class="tr--responsive">
+        <td class="td--responsive td-mail td-mail__bg-success td__padding-all td__leftcell" colspan="3">
+            <p><span class="text-color__small-gray-900">Vraag: </span></p>
+            <h2>{{ subject }} / {{ question }}</h2>
```
jiromaykin, line 17: @Bartvaderkin Here is something we may not actually have...? A subject? We only want to show what kind of question this is - not the entire question itself; the design needs changing?

Bartvaderkin (changes requested):
As discussed this is not the correct template: "contactform_registration" goes to the municipality (as alternative to the klanen/contacten API). 

So if we need this as a new"contactform_confirmation"

See the Taiga ticket for more.

src/open_inwoner/conf/parts/contactform_registration.html
Bartvaderkin: Resolving this as it was based on a confusion.

src/open_inwoner/conf/parts/contactform_registration.html
Bartvaderkin: Resolving this as it was based on a confusion.

Bartvaderkin (approved):
Looks good. I take the taiga ticket for the backend part but we could already merge this so we don't get tangled.

PR #1091, opened 2024-03-14
✨ [#2192] Display newsletter form errors to user

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2192

~~Based on #1082, so that should be merged first~~

### DIFF
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -330,6 +330,10 @@ class NewsletterSubscribeView(
     def form_valid(self, form):
         form.save(self.request)
 
+        # Display errors raised by Laposta API
+        if form.errors:
+            return self.form_invalid(form)
+
         messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
         self.log_user_action(
             self.request.user, _("users newsletter subscriptions were modified")
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -332,6 +332,9 @@ class NewsletterSubscribeView(
 
         # Display errors raised by Laposta API
         if form.errors:
+            self.log_user_action(
+                self.request.user, _("failed to modify user newsletter subscription")
+            )
             return self.form_invalid(form)
 
         messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
--- a/src/open_inwoner/laposta/client.py
+++ b/src/open_inwoner/laposta/client.py
@@ -63,6 +63,7 @@ class LapostaClient(APIClient):
             # Handle scenario where a subscription does not exists in the API,
             # but it does exist locally
             if error.get("code") == 203 and error.get("parameter") == "member_id":
+                logger.info("Subscription does not exist for user")
                 return None
 
         data = get_json_response(response)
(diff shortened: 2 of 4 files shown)


src/open_inwoner/accounts/views/profile.py
```diff
@@ -330,6 +330,10 @@ def get_form_kwargs(self) -> dict[str, Any]:
     def form_valid(self, form):
         form.save(self.request)
 
+        # Display errors raised by Laposta API
+        if form.errors:
+            return self.form_invalid(form)
```
pi-sigma, line 335: Should be logged before returning?

src/open_inwoner/laposta/client.py
```diff
@@ -33,39 +33,39 @@ def get_lists(self) -> list[LapostaList]:
...
-            return None
+        response = self.delete(f"member/{member_id}", params={"list_id": list_id})
+
+        if response.status_code == 400:
+            data = response.json()
+            error = data.get("error", {})
+            # Handle scenario where a subscription does not exists in the API,
+            # but it does exist locally
+            if error.get("code") == 203 and error.get("parameter") == "member_id":
+                return None
```
pi-sigma, line 66: In `create_subscription` there is a logging call in the same spot, here it's missing

src/open_inwoner/laposta/forms.py
```diff
@@ -38,26 +45,45 @@ def __init__(self, user=None, *args, **kwargs):
...
+        to_create = []
+        for list_id in newsletters:
+            if list_id in existing_subscriptions:
+                continue
+
+            try:
+                member = client.create_subscription(list_id, user_data)
+            except (RequestException, ClientError):
+                logger.exception(
+                    "Something went wrong while trying to create subscription"
```
pi-sigma, line 75: Perhaps catch the errros as `exc` or something and add then add `exc_info=exc` to the call

src/open_inwoner/laposta/forms.py
```diff
@@ -67,17 +93,33 @@ def save(self, request, *args, **kwargs):
...
+            user=request.user, list_id__in=unsubscribe_from_ids
+        )
+
+        to_delete_ids = []
+        for subscription in unsubscribe_from:
+            try:
+                client.remove_subscription(subscription.list_id, subscription.member_id)
+            except (RequestException, ClientError):
+                logger.exception(
+                    "Something went wrong while trying to create subscription"
```
pi-sigma, line 110: See previous

pi-sigma (changes requested):

src/open_inwoner/laposta/forms.py
stevenbal: `logging.exception` adds the traceback by default: https://docs.python.org/3/library/logging.html#logging.Logger.exception

src/open_inwoner/accounts/views/profile.py
stevenbal: Added a `log_user_action`

pi-sigma (approved):

src/open_inwoner/laposta/forms.py
Bartvaderkin, line 110: This log line mentions create (instead of remove/unsubscribe)

Bartvaderkin (changes requested):
So close: found one log string.

Bartvaderkin (approved):

PR #1093, opened 2024-03-18
Fixed autoload of django-admin-index

pi-sigma:
Error handling when django-admin-index could not be loaded was missing exception info and return.


Bartvaderkin (approved):
Yes please

PR #1094, opened 2024-03-18
:label: [#1718] Update typehints to use PEP585 generics

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1718

### DIFF
--- a/src/open_inwoner/openzaak/api_models.py
+++ b/src/open_inwoner/openzaak/api_models.py
@@ -2,7 +2,7 @@ import logging
 import re
 from dataclasses import dataclass, field
 from datetime import date, datetime
-from typing import Dict, Optional, Union
+from typing import Optional, Union
 
 from dateutil.relativedelta import relativedelta
 from zgw_consumers.api_models.base import Model, ZGWModel
@@ -272,7 +272,7 @@ class Notification(Model):
     hoofd_object: str
     actie: str
     aanmaakdatum: datetime
-    kenmerken: Dict = field(default_factory=dict)
+    kenmerken: dict = field(default_factory=dict)
 
 
 @dataclass
--- a/src/open_inwoner/ssd/service/jaaropgave/jaar_opgave_client.py
+++ b/src/open_inwoner/ssd/service/jaaropgave/jaar_opgave_client.py
@@ -1,5 +1,4 @@
 from dataclasses import dataclass, field
-from typing import Optional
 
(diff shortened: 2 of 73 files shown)


src/open_inwoner/openzaak/api_models.py
```diff
@@ -107,29 +106,29 @@ class ZaakType(ZGWModel):
...
+    versiedatum: date | None = None
+    concept: bool | None = None
 
 
 @dataclass
 class ZaakInformatieObject(ZGWModel):
     url: str
-    informatieobject: Union[str, "InformatieObject"]
-    zaak: Union[str, Zaak]
+    informatieobject: "str | InformatieObject"
```
Bartvaderkin, line 126: Typo

src/open_inwoner/ssd/service/jaaropgave/jaar_opgave_client.py
```diff
@@ -17,16 +16,14 @@ class Meta:
         name = "Envelope"
         namespace = "http://schemas.xmlsoap.org/soap/envelope/"
 
-    header: Optional[
-        "JaarOpgaveClientPortTypeSendJaarOpgaveClientInput.Header"
-    ] = field(
+    header: "JaarOpgaveClientPortTypeSendJaarOpgaveClientInput.Header | None" = field(
```
Bartvaderkin, line 19: Typo

Bartvaderkin (changes requested):
73 files changed is nice. But it does look like your replace missed the string references in some Optional's and Union's and wrapped the whole replacement in a string.

So I'd search for ` | None"` en `"str | ` en ` | str"` (etc?)

src/open_inwoner/openzaak/api_models.py
stevenbal: The union syntax doesn't work with string operands (yet?): https://github.com/python/cpython/issues/90015#issuecomment-1093938214.

I'll revert the changes in this file though, because replacing `Union` and `Optional` seems to cause issues with `zgw_consumers.factory`

pi-sigma (approved):

src/open_inwoner/openzaak/api_models.py
Bartvaderkin: The Union/forward reference thing is pretty disappointing. I expected the operator version to be just syntax sugar on Union but apparently not.

Bartvaderkin (approved):

PR #1095, opened 2024-03-18
:clown_face: [#2196] Inject query parameter values in Mock API responses

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2196

### DIFF
--- a/src/open_inwoner/apimock/apis/openklant-read/contactmomenten/objectcontactmomenten.json
+++ b/src/open_inwoner/apimock/apis/openklant-read/contactmomenten/objectcontactmomenten.json
@@ -1,8 +1,23 @@
 {
-    "count": 1,
+    "count": 4,
     "previous": null,
     "next": null,
     "results": [
+        {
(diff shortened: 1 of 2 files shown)


Bartvaderkin (approved):
Cool hack :rescue_worker_helmet:

stevenbal:
> Perhaps the hardcoded strings in the reponse objects should be removed or replaced with a dummy string. Currently they create the impression that a specific object/zaak (for instance) will be returned, when in fact the return value is now dynamically determined.

I was thinking about this too, but this will actually make it impossible to test the link from Mijn vraag to Mijn aanvraag, because in that case we do an API call that filters just by ContactMoment (and not by object, meaning the dummy value will stay in the response).

> The mocks could be extended to include multiple questions per case, so that we can test the expand/hide feature, although this could also be done in a separate ticket.

It's good that you said this, because I added three extra objectcontactmomenten, and it seems that the expand feature currently doesn't work if the total number of questions is even :sweat_smile:

PR #1098, opened 2024-03-18
[#2217] Use 'statustekst' as default case status label

pi-sigma:
Taiga: [#2217](https://taiga.maykinmedia.nl/project/open-inwoner/task/2217)

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -369,8 +369,11 @@ class InnerCaseDetailView(
         end_statustype_data = None
         if not status_types_mapping.get(end_statustype.url):
             end_statustype_data = {
-                "label": status_translate(
-                    end_statustype.omschrijving, default=_("No data available")
+                "label": (
+                    end_statustype.statustekst
+                    or status_translate(
+                        end_statustype.omschrijving, default=_("No data available")
+                    )
                 ),
                 "status_indicator": getattr(
                     self.statustype_config_mapping.get(end_statustype.url),
@@ -541,8 +544,11 @@ class InnerCaseDetailView(
         return [
             {
                 "date": s.datum_status_gezet,
-                "label": lookup.from_glom(
-                    s, "statustype.omschrijving", default=_("No data available")
+                "label": (
+                    s.statustype.statustekst
+                    or lookup.from_glom(
+                        s, "statustype.omschrijving", default=_("No data available")
+                    )
                 ),
                 "status_indicator": getattr(
                     statustype_config_mapping.get(s.statustype.url),
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -214,7 +214,6 @@ class InnerCaseDetailView(
                 "statuses": self.get_statuses_data(
                     statuses, status_translate, self.statustype_config_mapping
                 ),
-                # "end_statustype_data": end_statustype_data,
                 "end_statustype_data": end_statustype_data,
                 "second_status_preview": second_status_preview,
                 "documents": documents,
@@ -369,11 +368,9 @@ class InnerCaseDetailView(
         end_statustype_data = None
         if not status_types_mapping.get(end_statustype.url):
             end_statustype_data = {
-                "label": (
-                    end_statustype.statustekst
-                    or status_translate(
-                        end_statustype.omschrijving, default=_("No data available")
-                    )
+                "label": status_translate(
+                    (end_statustype.statustekst or end_statustype.omschrijving),
+                    default=_("No data available"),
                 ),
                 "status_indicator": getattr(
                     self.statustype_config_mapping.get(end_statustype.url),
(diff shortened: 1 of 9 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -369,8 +369,11 @@ def handle_end_statustype_data(
         end_statustype_data = None
         if not status_types_mapping.get(end_statustype.url):
             end_statustype_data = {
-                "label": status_translate(
-                    end_statustype.omschrijving, default=_("No data available")
+                "label": (
+                    end_statustype.statustekst
+                    or status_translate(
+                        end_statustype.omschrijving, default=_("No data available")
+                    )
```
Bartvaderkin, line 376: This doesn't match the Taiga task:

>  statustekst is to be used if filled in, otherwise fallback to the omschrijving. The resulting text can then be mapped to a different one via the status mapping tabel

So I'd expect something like this here:

    status_translate((end_statustype.statustekst or end_statustype.omschrijving), default=_("No data available")))

This applies to all changes that involve the translation.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -541,8 +544,11 @@ def get_statuses_data(
         return [
             {
                 "date": s.datum_status_gezet,
-                "label": lookup.from_glom(
-                    s, "statustype.omschrijving", default=_("No data available")
+                "label": (
+                    s.statustype.statustekst
+                    or lookup.from_glom(
+                        s, "statustype.omschrijving", default=_("No data available")
+                    )
```
Bartvaderkin, line 551: It feels inconsistent to use glom for one attribute but then use attribute access for the other.

For compactness you can also modify the `from_glom()` helper (or add a specific method) to also accept a tuple/list of multiple gloms (and one default)

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -191,21 +191,23 @@ def setUp(self):
             vertrouwelijkheidaanduiding=VertrouwelijkheidsAanduidingen.openbaar,
             indicatieInternOfExtern="intern",
         )
-        self.status_type1 = generate_oas_component_cached(
+        self.status_type_initial = generate_oas_component_cached(
```
Bartvaderkin, line 194: :clinking_glasses: Nice

Bartvaderkin (changes requested):

src/open_inwoner/openzaak/api_models.py
```diff
@@ -79,8 +80,11 @@ def process_data(self) -> dict:
             "start_date": self.startdatum,
             "end_date": getattr(self, "einddatum", None),
             "description": self.zaaktype.omschrijving,
-            "current_status": status_translate.from_glom(
-                self, "status.statustype.omschrijving", default=""
+            "current_status": (
+                glom(self, "status.statustype.statustekst")
+                or status_translate.from_glom(
+                    self, "status.statustype.omschrijving", default=""
+                )
```
Bartvaderkin, line 87: I forgot to add a note that that this relates to earlier about modifying/expanding the `from_glom()` method

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -1384,9 +1384,8 @@ def test_page_translates_statuses(self, m):
         response = self.app.get(
             self.case_detail_url, user=self.user, headers={"HX-Request": "true"}
         )
-        self.assertNotContains(response, st1.status)
+        self.assertNotContains(response, st1.translation)
         self.assertNotContains(response, st2.status)
-        self.assertContains(response, st1.translation)
         self.assertContains(response, st2.translation)
```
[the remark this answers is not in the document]
pi-sigma: The asserts had to be changed to reflect the handling of translating `statustekst` and `omschrijving`. I changed the variable names and re-arranged the asserts, which should make things clearer.

src/open_inwoner/openzaak/notifications.py
```diff
@@ -591,10 +591,9 @@ def send_case_update_email(
         "case_link": case_detail_url,
     }
     if status:
-        context["status_description"] = translate_single_status(
-            status.statustype.omschrijving
-        )
-
+        status_type = status.statustype
+        status_text = status_type.statustekst or status_type.omschrijving
+        context["status_description"] = translate_single_status(status_text)
```
Bartvaderkin, line 596: I note in the code above the `or` condition didn't need a temporary variable but was fine directly as argument.

Also we probably should add the default _("No data available") everywhere it makes sense (this is a legacy thing we can fix).

src/open_inwoner/userfeed/hooks/case_status.py
```diff
@@ -23,7 +23,8 @@ def case_status_notification_received(user: User, case: Zaak, status: Status):
         "case_uuid": case.uuid,
         "case_identificatie": case.identificatie,
         "case_omschrijving": case.omschrijving,
-        "status_omschrijving": status.statustype.omschrijving,
+        "status_omschrijving": status.statustype.statustekst
+        or status.statustype.omschrijving,
```
Bartvaderkin, line 27: For the future: these 'or' assignments statements in a dictionary will format a nudge better if you wrap with a parenthesis (visually more connected and less looking for subtle comma's).

Bartvaderkin (changes requested):

src/open_inwoner/userfeed/hooks/case_status.py
pi-sigma: The parentheses are removed by black

alextreme:
- Doublecheck that the emailnotification for a case statuschange has the statustekst instead of omschrijving
- Check the userfeed

Bartvaderkin (approved):

PR #1099, opened 2024-03-19
:sparkles: [#2179] Display list of Qmatic appointments in profiel

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2179

Can be tested by logging in with an account with email `foo@bar.org` and navigating to Mijn profiel

I adapted the client code from Open Forms and replaced `TypedDict` with pydantic's `BaseModel` to deserialize the data. Although we don't actually use the same client methods as OF (because OF implements creating appointments and we just need to implement the list of appointments + modify/cancel), this should probably be merged into a library at some point?


src/open_inwoner/templates/pages/profile/appointments.html
```diff
@@ -0,0 +1,49 @@
...
+                    <div class="card__body">
+                        {% render_list %}
+                            <h2 class="card__heading-2">{{ appointment.title }}</h2>
+                            {% timezone appointment.branch.timeZone %}
+                                {% list_item text=appointment.start|date:"l j F Y H:i" compact=True strong=False %}
+                            {% endtimezone %}
+                            {% list_item text=appointment.notes compact=True strong=False %}
+                            {% list_item text=appointment.branch.name caption=_("Locatie") compact=True strong=False %}
+                            {% list_item text=appointment.branch.addressCity compact=True strong=False %}
+                            {% list_item text=appointment.branch.addressLine2 compact=True strong=False %}
```
stevenbal, line 24: I'm not sure if these are the right fields to display, because the documentation isn't too clear about what is stored in these fields

src/open_inwoner/accounts/views/profile.py
```diff
@@ -343,3 +348,30 @@ def form_valid(self, form):
...
+        context = super().get_context_data(**kwargs)
+        # TODO email should be verified
+        try:
+            client = QmaticClient()
+        except NoServiceConfigured:
+            logger.exception("Error occurred while creating Qmatic client")
+            context["appointments"] = []
+        else:
+            context["appointments"] = client.list_appointments_for_customer(
+                quote(self.request.user.email)
```
stevenbal, line 368: This assumes that the municipality uses a citizen's email as `external_id` in Qmatic (still awaiting confirmation from Hoorn if this is correct). If we want to support `bsn` or something else, we'll probably have to add a feature flag or choicefield on the config to enable that.

src/open_inwoner/qmatic/client.py
```diff
@@ -0,0 +1,141 @@
...
+
+from open_inwoner.utils.api import JSONEncoderMixin
+
+from .exceptions import QmaticException
+from .models import QmaticConfig
+
+# API DATA DEFINITIONS
+
+
+class ServiceDict(BaseModel):
```
Bartvaderkin, line 16: If these aren't TypedDict's anymore the name doesn't need the Dict postfix :wink:

src/open_inwoner/qmatic/client.py
```diff
@@ -0,0 +1,141 @@
...
+    assert isinstance(service, Service)
+    return build_client(service, client_factory=Client)
+
+
+def startswith_version(url: str) -> bool:
+    if url.startswith("v1/"):
+        return True
+    if url.startswith("v2/"):
+        return True
+    return False
```
Bartvaderkin, line 104: This could be like this:

````python
    return url.startswith(("v1/", "v2/",))  # it accepts a tuple
````

File: src/open_inwoner/templates/pages/profile/appointments.html (hunk 1 above)
Comment (Bartvaderkin) [reply src/open_inwoner/templates/pages/profile/appointments.html:24]: Feels like a testing and acceptance problem :wink:

File: src/open_inwoner/qmatic/client.py [hunk 5]
```diff
@@ -0,0 +1,141 @@
...
+            raise QmaticException(
+                f"Server error (HTTP {response.status_code}): {error_msg}"
+            )
+
+        return response
+
+    def list_appointments_for_customer(
+        self, customer_externalid: str
+    ) -> list[Appointment]:
+        endpoint = f"customers/externalId/{customer_externalid}/appointments"
```
Bartvaderkin, line 136: Are we confident we can build URL's with just f-string instead of something that applies URL-encoding?

src/open_inwoner/qmatic/client.py
```diff
@@ -0,0 +1,141 @@
...
+        response = self.get(endpoint)
+        if response.status_code == 404:
+            return []
+        response.raise_for_status()
+        return [Appointment(**entry) for entry in response.json()["appointmentList"]]
```
Bartvaderkin, line 141: Is this the appropriate pattern to parse Pydantic from json? It feels a little simple to just dump the data as kwargs, but I haven't used Pydantic so I don't know.

src/open_inwoner/accounts/views/profile.py
Bartvaderkin: This `quote()` doesn't belong here but should be in the client method. It is not the responsibility of the using code to encode the values (as conceptually it doesn't even know how the client works internally).

This relates to my earlier comment I put in the client.

src/open_inwoner/qmatic/client.py
```diff
@@ -0,0 +1,141 @@
...
+def QmaticClient() -> "Client":
+    """
+    Create a Qmatic client instance from the database configuration.
+    """
+    config = QmaticConfig.get_solo()
+    assert isinstance(config, QmaticConfig)
+    if (service := config.service) is None:
+        raise NoServiceConfigured("No Qmatic service defined, aborting!")
+    assert isinstance(service, Service)
+    return build_client(service, client_factory=Client)
```
Bartvaderkin, line 96: if (service := config.service) is None:

I'm having doubts if this is a good pattern or possibly too clever walrus abuse. Usually you do the walrus so the assignment is there in the if-body, but this flips it around.

Bartvaderkin (changes requested):
Pretty clean, with some subtle feedback.

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -1139,3 +1143,119 @@ def test_render_form_limit_newsletters_to_admin_selection(self, m):
...
+
+        response = self.app.get(self.appointments_url, user=self.user)
+
+        self.assertIn(_("Een overzicht van uw afspraken"), response.text)
+
+        cards = response.pyquery(".appointment-info")
+
+        self.assertEqual(len(cards), 2)
+
+        passport_appointment = PQ(cards[0]).find("ul").children()
```
pi-sigma, line 1241: Is it necessary to use PyQuery directly, since you're already using `response.pyquery`, or just more convenient?

src/open_inwoner/qmatic/client.py
```diff
@@ -0,0 +1,141 @@
...
+
+def QmaticClient() -> "Client":
+    """
+    Create a Qmatic client instance from the database configuration.
+    """
+    config = QmaticConfig.get_solo()
+    assert isinstance(config, QmaticConfig)
+    if (service := config.service) is None:
+        raise NoServiceConfigured("No Qmatic service defined, aborting!")
+    assert isinstance(service, Service)
```
pi-sigma: Are the asserts necessary (is there a reason to think that they could fail)?

src/open_inwoner/qmatic/client.py
pi-sigma: Instead of making this a boolean check, you could make it return or create an appropriate url, like:
```
def get_or_create_url_with_version(url):
    if url.startswith("v1/"):
        return url
    if url.startswith("v2/"):
        return url
    return f"v1/{url}"
```
Then below in `Client`: ```url = get_or_create_url_with_version(url)```

src/open_inwoner/qmatic/models.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Global configuration and defaults
+    """
+
+    service = models.OneToOneField(
+        "zgw_consumers.Service",
+        verbose_name=_("Calendar API"),
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.orc},
+        related_name="+",
```
pi-sigma, line 23: Not very descriptive name. Intentional?

src/open_inwoner/qmatic/client.py
stevenbal: I agree, I've been a bit too eager copy pasting :sweat_smile:. I'll switch it around such that the assignment variable is used inside the if statement.

src/open_inwoner/qmatic/client.py
stevenbal: I wasn't too sure about this either initially, but from in the examples from the docs this same pattern is used: https://docs.pydantic.dev/latest/#pydantic-examples

pi-sigma (approved):
Couple minor questions + suggestions

src/open_inwoner/qmatic/client.py
Bartvaderkin: Maybe we should add the try/except ValidationError and a log when parsing external data, so it doesn't blow up if weird data comes in.

src/open_inwoner/qmatic/models.py
stevenbal: Setting the related_name to `"+"` makes sure that `Service` doesn't have a backwards relation to `QmaticConfig` (which I think is right in this case, because trying to get a `SingletonModel` from a `Service` instance isn't right): https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.ForeignKey.related_name

src/open_inwoner/qmatic/client.py
stevenbal: Good point, I'll add this and add a test for it

src/open_inwoner/accounts/tests/test_profile_views.py
stevenbal: It actually is necessary, it's a bit weird but for some reason I can't use `.find` and other methods on the element that I get from `cards[0]` directly

src/open_inwoner/qmatic/client.py
stevenbal: I was a bit too eager copying stuff from Open Forms I think, I don't really see how these asserts serve much of a purpose either, so I'll remove them

src/open_inwoner/accounts/tests/test_profile_views.py
Bartvaderkin: This looks familiar: iirc pyquery doesn't wrap recursively like some other api's do.

src/open_inwoner/qmatic/client.py
stevenbal: Hmm I think I'll keep it as it is now, that way the function is a bit more clear in what it does imo

stevenbal:
@Bartvaderkin can you re-review this? :)

Bartvaderkin (approved):

PR #1101, opened 2024-03-19
:lipstick: [#1811] Fix spacing warning-header text

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1811

### DIFF
--- a/src/open_inwoner/components/templates/components/Header/WarningHeader.html
+++ b/src/open_inwoner/components/templates/components/Header/WarningHeader.html
@@ -9,7 +9,7 @@
 
 <div class="warning-header">
     <div class="warning-header__container">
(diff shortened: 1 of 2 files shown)


pi-sigma (changes requested):
Doesn't seem to work for me (I have `npm run watch`).

Screenshot from 2024-03-19 15-28-36

jiromaykin:
> Doesn't seem to work for me

@pi-sigma Your screenshot looks good to me. 
What doesn't work? 
We still need a little bit of a top-margin for the text, or else it would look really weird if there is only one line:

oneline

pi-sigma (approved):
I thought the top of the icon should align perfectly with the top of the text, but the explanation makes sense; it looks weird with only one line.

PR #1103, opened 2024-03-19
✨ [#2180] Implement reschedule and delete for Qmatic appointments

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2180

based on #1099, so that should be merged first


src/open_inwoner/templates/pages/profile/appointments.html
```diff
@@ -24,18 +24,12 @@ <h2 class="card__heading-2">{{ appointment.title }}</h2>
                             {% list_item text=appointment.branch.addressLine2 compact=True strong=False %}
                         {% endrender_list %}
 
-                        <span class="link link--icon link--secondary"
-                            aria-label="{% trans "Wijzig afspraak" %}"
-                            title="{% trans "Wijzig afspraak" %}">
-                            <span class="link__text">{% trans "Wijzig afspraak" %}</span>
+                        <a href="{{ booking_base_url }}{{ appointment.publicId }}" class="link link--icon link--secondary"
```
Bartvaderkin, line 27: It is a bit dirty to build URL's in the template if it is not necessary or legacy patterns.

src/open_inwoner/templates/pages/profile/appointments.html
```diff
@@ -24,18 +24,12 @@ <h2 class="card__heading-2">{{ appointment.title }}</h2>
...
+                            aria-label="{% trans "Wijzig of annuleer afspraak" %}"
+                            title="{% trans "Wijzig of annuleer afspraak" %}">
```
Bartvaderkin, line 29: A bit late, but maybe we don't need both title and aria-label? Aria should pickup standard attributes.

Bartvaderkin (changes requested):

pi-sigma (approved):

src/open_inwoner/qmatic/client.py
```diff
@@ -135,10 +136,16 @@ def list_appointments_for_customer(
...
         response.raise_for_status()
+        config = QmaticConfig.get_solo()
         try:
-            return [
+            appointments = [
                 Appointment(**entry) for entry in response.json()["appointmentList"]
             ]
+            for appointment in appointments:
+                appointment.url = (
+                    f"{config.booking_base_url}{quote(appointment.publicId)}"
```
stevenbal, line 146: Turns out I can't actually use `furl` here, because the URL Hoorn uses has `/#/` in its path, which `furl` parses as the fragment and moves it to the end of the URL

Bartvaderkin: Ugh.. seems fishy but if it works :man_shrugging:

Bartvaderkin (approved):

PR #1106, opened 2024-03-21
[#2106, #2203] Implemented 'contactform_confirmation' email

Bartvaderkin:
The registration code is pretty spicy. I tried refactoring but it does need many conditionals so except some small changes I kept it this way.


stevenbal (approved):

pi-sigma (approved):

src/open_inwoner/openklant/migrations/0011_openklantconfig_api_sends_email_confirmation.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ("openklant", "0010_openklantconfig_register_channel"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="openklantconfig",
+            name="api_sends_email_confirmation",
+            field=models.BooleanField(
+                default=False,
```
alextreme, line 17: Lets turn this around to make the usage via the admin simpler: "Stuur contactformulier e-mailbevestiging" / "send_email_confirmation" which is off by default.

alextreme (changes requested):
Rename admin setting

alextreme (approved):

PR #1108, opened 2024-03-21
🥅 [#2197] Block eHerkenning login for ZZP if RSIN is required

stevenbal:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2197

### DIFF
--- a/src/eherkenning/tests/test_mock_views.py
+++ b/src/eherkenning/tests/test_mock_views.py
@@ -108,8 +108,15 @@ class PasswordLoginViewTests(eHerkenningMockTestCase):
         self.assertContains(response, reverse("login"))
         self.assertNoEHerkenningURLS(response)
 
-    @patch("open_inwoner.kvk.client.KvKClient.get_all_company_branches")
-    def test_post_redirects_and_authenticates(self, mock_kvk):
+    @patch(
+        "open_inwoner.kvk.signals.KvKClient.retrieve_rsin_with_kvk",
+        return_value="123456789",
+        autospec=True,
+    )
+    @patch("open_inwoner.kvk.client.KvKClient.get_all_company_branches", autospec=True)
+    def test_post_redirects_and_authenticates(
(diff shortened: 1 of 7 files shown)


src/eherkenning/tests/test_mock_views.py
```diff
@@ -108,8 +108,14 @@ def test_get_returns_valid_response(self):
         self.assertContains(response, reverse("login"))
         self.assertNoEHerkenningURLS(response)
 
+    @patch(
+        "open_inwoner.kvk.signals.KvKClient.retrieve_rsin_with_kvk",
+        return_value="123456789",
+    )
```
Bartvaderkin, line 114: Let's make a habit of putting `auto_spec=True` in all the patches.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -133,8 +140,30 @@ def get_success_url(self):
...
 
+class BlockEenmanszaakLoginMixin:
+    def get(self, request):
+        response = super().get(request)
+
+        openzaak_config = OpenZaakConfig.get_solo()
+        openklant_config = OpenKlantConfig.get_solo()
+        if (
+            hasattr(request.user, "rsin")
+            and not request.user.rsin
```
Bartvaderkin, line 151: Looks like `getattr(request.user, "rsin", None)`

Bartvaderkin (approved):
Just optional notes

src/open_inwoner/accounts/views/auth.py
stevenbal: `request.user` can be AnonymousUser (in case OIDC gives errors) and in that case I want the OIDCCallbackView to handle those errors

src/eherkenning/tests/test_mock_views.py
stevenbal: good point, added them

PR #1111, opened 2024-03-22
[#2187] WIP Untangling registration views, redirects and middleware

Bartvaderkin:
This started in the email-verification task but is spiraling a little bit because of how entangled the user-upfix redirects are with various views and login flows.


Bartvaderkin:
Closing because it is borked.

PR #1112, opened 2024-03-23
Make vragen detail conform to design

pi-sigma:
Design: [Figma](https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=1080-6442&mode=design)


src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,42 @@
...
                 {% include "components/Dashboard/Dashboard.html" with metrics=metrics only %}
             {% endrender_column %}
-            {% render_column start=4 span=6 %}
+        {% endrender_grid %}
 
-                <div class="contactmoment__details">
-                {% contactmoment_table contactmoment zaak=zaak %}
+        <div class="container--no-margin">
+            <div class="contactmoment">
+                <h3>{% trans "Uw vraag" %}</h3>
```
jiromaykin, line 15: Do not jump from H1 to H3 without using a H2 Tag...! (bad for the hierarchy in accessibility software).
Also looking at the design you can see all H3's in your template need to actually be H2's (but... with a font-size of 20px) - you could use the "h2" class here as a basis, and/or create a special H2 heading within Contactmoment SCSS. Something like ".contactmoment__heading-2" but then that one would need to be a Flexbox because of the date field.

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,42 @@
...
+                <p>{{ contactmoment.text }}</p>
+            </div>
+            <div class="contactmoment">
+                <h3>{% trans "Antwoord" %}</h3>
```
jiromaykin, line 19: The datum/date is missing here - you can use a H2 element that's wrapping an anchor just like we are using on the Homepage.
Look at the design: 
https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=1625-11082&mode=design&t=irQankQdfNj4iA6o-0

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,42 @@
...
+                {% if contactmoment.antwoord %}
+                    <p>{{ contactmoment.antwoord }}</p>
+                {% else %}
+                    <p>{% trans "Nog geen antwoord beschijkbaar" %}</p>
```
jiromaykin, line 23: All your paragraphs still need the proper line-height and `color: var(--font-color-body);` you can do this by either giving them the 'p class or create a contactmoment-p class where you can control their spacing as well.
(Our default body font color is not black: giving it a bit of a gray color is better for accessibility).

jiromaykin (changes requested):

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -351,9 +353,9 @@ def test_show_detail_for_bsn_with_zaak_reformat_esuite_id(
             },
         )
 
-        zaak_link = response.pyquery(".case-detail__link")
+        zaak_link = response.pyquery("#origin_link")
 
-        self.assertEqual(zaak_link.text(), _("Ga naar zaak"))
+        self.assertIn(_("Terug naar aanvrag"), zaak_link.text())
```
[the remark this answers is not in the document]
pi-sigma: Discussed with @alextreme, we're not going to show the name of the aanvraag; this would complicate things unnecessarily.

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -362,6 +364,13 @@ def test_show_detail_for_bsn_with_zaak_reformat_esuite_id(
             ),
         )
 
+        contactmoment_link = response.pyquery("#destination_link")
+        self.assertIn(_("Bekijk alle vragen"), contactmoment_link.text())
```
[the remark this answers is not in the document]
pi-sigma: There are two possibilities:
1. user comes from contactmomenten list
2. user comes from aanvraag detail

You're describing the first. In the second case, the primary CTA is "Terug naar aanvraag" and the secondary link is "Bekijk alle vragen".

The design has a little note in the top right corner explaining this.

src/open_inwoner/templates/pages/contactmoment/detail.html
pi-sigma: We don't have the data available. This is the second place where we're departing from the design.

src/open_inwoner/templates/pages/contactmoment/detail.html
pi-sigma: Fortunately, we're not displaying the response date, so this complication doesn't arise.

jiromaykin: Alright, but you will still need to use an H2 _tag_ here instead of an H3 _tag_.

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -74,46 +74,21 @@
...
-      word-break: break-word;
-      white-space: normal;
-
-      &.table__item--status {
-        .table__item--wrapper {
-          display: flex;
-          justify-content: space-between;
-        }
-      }
+  p {
```
jiromaykin, line 77: Do not style tags without classes - we only do that if we do not have control over the components (like with CKeditor) - so your contactmoment__text class is perfectly fine , and just add the color in there.

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -74,46 +74,21 @@
...
-        }
-      }
-    }
+  &__heading {
+    font-size: var(--font-size-body-large);
+  }
+
+  &__text {
+    padding-bottom: var(--spacing-large);
+    line-height: var(--font-line-height-body-medium);
```
jiromaykin, line 91: No new line-height variable needed here because we already have that, it's in "--font-line-height-body: 24px" so you can use that variable.

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
+                    {% endif %}
+                </div>
+                <div class="contactmoment">
+                    <a href="{{ origin.url }}" id="origin_link" class="button button--secondary">
+                        {% icon "arrow_backward" %}{{ origin.label }}
+                    </a>
+                </div>
+                {% if destination %}
+                    <div class="contactmoment">
+                        <a href="{{ destination.url }}" id="destination_link" class="button button--transparent">
```
jiromaykin, line 34: Both your button--transparent, as well as your primary button need to be Bold - This style looks a lot like what we already use for Forms with a fullwidth actions area. You should wrap these two buttons with a Div with class `contactmoment__actions` so you can give them both the bold font-weight.

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
                 {% include "components/Dashboard/Dashboard.html" with metrics=metrics only %}
             {% endrender_column %}
-            {% render_column start=4 span=6 %}
+        {% endrender_grid %}
 
-                <div class="contactmoment__details">
-                {% contactmoment_table contactmoment zaak=zaak %}
+        {% render_grid %}
+            {% render_column start=4 span=6 %}
+                <div class="contactmoment">
```
jiromaykin, line 15: Here you should just add 1 wrapping Div (or Section) with class contactmoment - it is not necessary to add the same class over and over again to all the divs below if you already have it in a wrapper.

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
+                    <h2 class="contactmoment__heading">{% trans "Uw vraag" %}</h2>
+                    <p class="contactmoment__text">{{ contactmoment.text }}</p>
                 </div>
+                <div class="contactmoment">
```
jiromaykin, line 19: Class not needed if you wrap this in a div/section with class contactmoment.

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
+                    <h2 class="contactmoment__heading">{% trans "Antwoord" %}</h2>
+                    {% if contactmoment.antwoord %}
+                        <p class="contactmoment__text">{{ contactmoment.antwoord }}</p>
+                    {% else %}
+                        <p class="contactmoment__text">{% trans "Nog geen antwoord beschijkbaar" %}</p>
+                    {% endif %}
+                </div>
+                <div class="contactmoment">
```
jiromaykin, line 27: Class not needed if you wrap this in a div/section 🙂

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
+                    <a href="{{ origin.url }}" id="origin_link" class="button button--secondary">
+                        {% icon "arrow_backward" %}{{ origin.label }}
+                    </a>
+                </div>
+                {% if destination %}
+                    <div class="contactmoment">
```
jiromaykin, line 33: DRY

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -1,21 +1,44 @@
...
+                <div class="contactmoment">
+                    <h2 class="contactmoment__heading">{% trans "Antwoord" %}</h2>
+                    {% if contactmoment.antwoord %}
+                        <p class="contactmoment__text">{{ contactmoment.antwoord }}</p>
+                    {% else %}
+                        <p class="contactmoment__text">{% trans "Nog geen antwoord beschijkbaar" %}</p>
+                    {% endif %}
+                </div>
+                <div class="contactmoment">
+                    <a href="{{ origin.url }}" id="origin_link" class="button button--secondary">
```
jiromaykin, line 28: Needs to be button--primary.

jiromaykin (changes requested):

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -74,46 +74,22 @@
...
-      white-space: normal;
-
-      &.table__item--status {
-        .table__item--wrapper {
-          display: flex;
-          justify-content: space-between;
-        }
-      }
+  &__heading {
+    font-size: var(--font-size-body-large);
```
jiromaykin, line 78: This is not the correct CSS variable, and you have not applied the correct Font-family for Headings and the other needed values. 
We already have these Heading sizes, but you - you would need --font-size-heading-3; here (which is absolutely fine to use for a Level 2 heading).
You can just inherit everything from Heading 3:

```color: var(--font-color-heading-3);
  font-family: var(--font-family-heading);
  font-size: var(--font-size-heading-3);
  font-weight: var(--font-weight-heading-3);
  line-height: var(--font-line-height-heading-3);
  margin: 0;
```

jiromaykin (changes requested):

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -74,46 +74,22 @@
...
+    font-size: var(--font-size-heading-3);
+    font-weight: var(--font-weight-heading-3);
+    line-height: var(--font-line-height-heading-3);
+    margin: 0;
+  }
 
-      .case-detail__link {
-        color: var(--color-primary);
-        text-decoration: none;
+  &__text {
```
jiromaykin, line 85: Since you're not using the "CSS Reset" styling you will need to style **_everything_** for paragraphs here. A "CSS reset" means you're setting paddings and margins to zero, because many HTML elements have a default padding/margin you'll need to 'reset/null' in order to give it the proper values from the actual design.

TODO: 
- remove the padding from these '&__text' paragraphs
- add margin: 0 0 var(--row-height) 0;

Note: --row-height=40px.

It would also be slightly more proper to add the font-family=body here, though it's already set on the Body tag.

Look at all the stuff that the `.p` class is 'resetting'; these are the values you will need every time you want to give a paragraph the proper styling if you do not wish to use the `.p` class (and yeah: we are slowly trying to get rid of that one) so this is how it is 'resetting' default HTML values:

```
.p {
  color: var(--font-color-body);
  font-family: var(--font-family-body);
  font-size: var(--font-size-body);
  line-height: var(--font-line-height-body);
  font-weight: normal;
  margin: 0;
}
```

src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -74,46 +74,22 @@
...
-          display: flex;
-          justify-content: space-between;
-        }
-      }
+  &__heading {
+    font-family: var(--font-family-heading);
+    font-size: var(--font-size-heading-3);
+    font-weight: var(--font-weight-heading-3);
+    line-height: var(--font-line-height-heading-3);
+    margin: 0;
```
jiromaykin, line 82: Nice margin reset :-) but... in the design you can see you will still need a bottom margin for Headings (var(--spacing-large)).

jiromaykin (changes requested):
If you can't see the proper values in Figma, look at the design in `Dev mode`. If you are unable to measure the distance between elements in Figma, you can toggle Dev mode off and CTRL-select an element and use the ALT key to see the distance to another element (by hovering).

These are the proper margins in Figma (although they may look a bit too big in a browser 🤔):

figma-proper-devmode

jiromaykin (approved):
The two navigation buttons at the bottom look a bit strange but that is because they will contain less text than they do in the design, so that's okay for now. 
Also: this branch will need rebasing.

PR #1115, opened 2024-03-25
[#2277] Use first OIP subject for questions as fallback

pi-sigma:
If no mapping exists for e-suite for a particular e-suite subject code (or if duplicate codes exist), we use the first OIP configured subject as fallback

Taiga: [#2277](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2277)

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -140,14 +140,6 @@ class KlantContactMomentBaseView(
             )
             return e_suite_subject_code
 
-        if not subject.subject:
-            logger.warning(
-                "Could not determine OIP subject for contactmoment %s; "
-                "falling back on e-suite subject code ('onderwerp')",
-                kcm.contactmoment.url,
-            )
-            return e_suite_subject_code
-
         return subject.subject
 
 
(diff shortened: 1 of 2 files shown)


src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -131,7 +130,7 @@ def get_kcm_subject(
                 kcm.contactmoment.url,
                 exc_info=exc,
             )
-            return e_suite_subject_code
+            return ContactFormSubject.objects.first()
```
alextreme, line 133: Discussed and request is to split this up, situation with multipleobjectsreturned should show the first of the subjects with the subject_code matching the esuite subject code

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -114,23 +114,38 @@ def get_kcm_subject(
...
+                exc_info=exc,
+            )
+            return e_suite_subject_code
+
+        if not subject.subject:
+            logger.warning(
+                "Could not determine OIP subject for contactmoment %s; "
+                "falling back on e-suite subject code ('onderwerp')",
+                kcm.contactmoment.url,
+            )
```
stevenbal, line 148: This can probably be removed if we do this in the try ... except block: `subject = ContactFormSubject.objects.get(~Q(subject=""), subject_code=e_suite_subject_code)`. Though on second thought, do we need this at all? `ContactFormSubject.subject` is a required field: https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openklant/models.py#L111-L114

pi-sigma: Good catch. I've removed this part and modified the test.

stevenbal (approved):

alextreme (approved):

PR #1116, opened 2024-03-25
:sparkles: [#2262] Add setup configuration steps for ZGW and KIC

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2262

Once django-setup-configuration supports loading YAML (see 2nd iteration https://github.com/maykinmedia/django-setup-configuration/issues/1), we might want to start using that to configure the services


src/open_inwoner/conf/base.py
```diff
@@ -928,6 +929,89 @@
...
+    FORMULIEREN_API_ROOT  # this is still required by the form, but not actually used
+)
+FORMULIEREN_API_CLIENT_ID = config("FORMULIEREN_API_CLIENT_ID", "")
+FORMULIEREN_API_SECRET = config("FORMULIEREN_API_SECRET", "")
+# ZGW config options
+ZAAK_MAX_CONFIDENTIALITY = config("ZAAK_MAX_CONFIDENTIALITY", None)
+DOCUMENT_MAX_CONFIDENTIALITY = config("DOCUMENT_MAX_CONFIDENTIALITY", None)
+ACTION_REQUIRED_DEADLINE_DAYS = config("ACTION_REQUIRED_DEADLINE_DAYS", None)
+ALLOWED_FILE_EXTENSIONS = config("ALLOWED_FILE_EXTENSIONS", None)
+MIJN_AANVRAGEN_TITLE_TEXT = config("MIJN_AANVRAGEN_TITLE_TEXT", None)
```
stevenbal, line 980: I wasn't sure about prefixing all these variables, because that would make the names rather long. I tried to keep them the same as the model field name they are used for, but some (like `title_text`) were a bit too generic in my opinion

src/open_inwoner/configurations/bootstrap/kic.py
```diff
@@ -0,0 +1,176 @@
...
+        "CONTACTMOMENTEN_API_SECRET",
+    ]
+    enable_setting = "KIC_API_CONFIG_ENABLE"
+
+    def is_configured(self) -> bool:
+        return bool(Service.objects.filter(api_root=settings.CONTACTMOMENTEN_API_ROOT))
+
+    def configure(self):
+        organization = settings.OIP_ORGANIZATION or settings.ENVIRONMENT
+        org_label = f"Open Inwoner {organization}".strip()
```
pi-sigma, line 80: I'm not sure where this is used, but do we want to hardcode "Open Inwoner" into the organization label? This doesn't seem to be configurable, so might not be desirable.

src/open_inwoner/configurations/bootstrap/kic.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+
+        config.save()
+
+    def test_configuration(self):
+        """
+        make requests to the APIs and verify that a connection can be made
+        """
+        klanten_client = build_client("klanten")
+        contactmomenten_client = build_client("contactmomenten")
```
pi-sigma, line 156: I've noticed that we somtimes use `contactmomenten_client` (plural) and sometimes `contactmoment_client` (singular). It would be good to make this consistent so that it is easier to grep on the variable.

src/open_inwoner/conf/base.py
```diff
@@ -928,6 +929,95 @@
...
+FORMULIEREN_API_ROOT = config("FORMULIEREN_API_ROOT", "")
+if FORMULIEREN_API_ROOT and not FORMULIEREN_API_ROOT.endswith("/"):
+    FORMULIEREN_API_ROOT = f"{FORMULIEREN_API_ROOT.strip()}/"
+FORMULIEREN_OAS_URL = (
+    FORMULIEREN_API_ROOT  # this is still required by the form, but not actually used
+)
+FORMULIEREN_API_CLIENT_ID = config("FORMULIEREN_API_CLIENT_ID", "")
+FORMULIEREN_API_SECRET = config("FORMULIEREN_API_SECRET", "")
+# ZGW config options
+ZAAK_MAX_CONFIDENTIALITY = config("ZAAK_MAX_CONFIDENTIALITY", None)
```
pi-sigma, line 982: Here you define the default for `ZAAK_MAX_CONFIDENTIALITY` as `None`. Later (in `bootstrap/zgw.py`) you set `config.zaak_max_confidentiality = settings.ZAAK_MAX_CONFIDENTIALITY`, but only if the latter is not `None`.

However, the default for  `config.zaak_max_confidentiality` (defined in `openzaak/models.py`) is `VertrouwelijkheidsAanduidingen.openbaar`. Why not use that default value here? Perhaps I don't fully understand the intended use of `django-setup-configuration`. Are we configuring these services in two steps?

src/open_inwoner/configurations/bootstrap/zgw.py
```diff
@@ -0,0 +1,309 @@
...
+        config.document_service = Service.objects.get(
+            api_root=settings.DOCUMENTEN_API_ROOT
+        )
+        config.form_service = Service.objects.get(
+            api_root=settings.FORMULIEREN_API_ROOT
+        )
+
+        # General config options
+        if settings.ZAAK_MAX_CONFIDENTIALITY:
+            config.zaak_max_confidentiality = settings.ZAAK_MAX_CONFIDENTIALITY
```
pi-sigma, line 234: Related to my comment above: here we set `config.zaak_max_confidentiality` to the value of `ZAAK_MAX_CONFIDENTIALITY`. But if the latter is `None`, we use the default defined by `OpenZaakConfig`. Why not use that default in our `settings/base.py`?

pi-sigma (changes requested):

src/open_inwoner/conf/base.py
stevenbal: I did it like this, because otherwise I'd have to duplicate all the defaults set in `models.py` in this settings file. Although the likelihood of the defaults changing isn't that high, I'd rather just keep the defaults in one place (`models.py`)

src/open_inwoner/conf/base.py
stevenbal: You're right, I think I'll prefix them to make it clear that they're used to setup the configuration

src/open_inwoner/configurations/bootstrap/kic.py
stevenbal: It's the human readable user identification shown in the audit trail, so when Open Inwoner makes a request using an application to i.e. Zaken API, that request is logged by Zaken API and this label is associated with the "user" (application) than made the request.

I could make the whole label configurable, but I'm not sure what else a municipality would want to show here besides the name of the application (Open Inwoner) and their organization name.

pi-sigma (approved):

src/open_inwoner/configurations/bootstrap/kic.py
```diff
@@ -0,0 +1,190 @@
...
+                "user_representation": org_label,
+            },
+        )
+        if not created:
+            klanten_service.oas = settings.KIC_CONFIG_KLANTEN_API_ROOT
+            klanten_service.client_id = settings.KIC_CONFIG_KLANTEN_API_CLIENT_ID
+            klanten_service.secret = settings.KIC_CONFIG_KLANTEN_API_SECRET
+            klanten_service.user_id = settings.KIC_CONFIG_KLANTEN_API_CLIENT_ID
+            klanten_service.user_representation = org_label
+            klanten_service.save()
```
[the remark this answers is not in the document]
stevenbal: Ah you're right, I thought that `defaults` were only used if the instance is created, but now I see that there is `create_defaults` for that https://docs.djangoproject.com/en/5.0/ref/models/querysets/#update-or-create. I'll remove these if blocks.

Bartvaderkin (approved):
Will be an interesting script that defines all these variables, and more fun debugging typos, but it is what was requested.

src/open_inwoner/configurations/bootstrap/kic.py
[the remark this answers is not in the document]
stevenbal: Oh whoops, in that case we do actually need the if block, because without it existing instances won't be updated in case we run the command with `--overwrite`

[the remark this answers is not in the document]
Bartvaderkin: Are you sure? The kwargs select the instance (eg: api_root): if it exists it gets updated with `defaults`, if it doesn't it gets created with the values from both kwargs and `defaults`.

I don't see anything here that would do something different when `--overwrite` or whatever would be be specified. It is a plain `update_or_create()` (not a `get_or_create()` with an condition on `created` and `--overwrite`)

Bartvaderkin (changes requested):

src/open_inwoner/configurations/bootstrap/kic.py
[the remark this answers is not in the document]
stevenbal: Ah I was confusing it with `get_or_create` :facepalm:

Bartvaderkin (approved):
Cool

PR #1121, opened 2024-03-26
:sparkles: [#2242] Add CMS plugin for mijn afspraken

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2242


src/open_inwoner/scss/components/Appointments/Appointments.scss
```diff
@@ -0,0 +1,7 @@
+.appointments {
+  .card__body {
+    span.button:last-child {
+      bottom: var(--spacing-large) !important;
```
jiromaykin, line 4: We like to avoid "!important" - what if you try to make it more specific by adding the parents with  ">" selectors? Have not tested this but maybe this is specific enough: 
`.plugin-card > .card__body > .button.button--icon-before.button--transparent`

src/open_inwoner/scss/components/PluginCard/PluginCard.scss
```diff
@@ -0,0 +1,54 @@
+.plugin-card {
```
jiromaykin, line 1: This seems specific to Userfeed? We do have multiple plugin-cards, but apparently those do not need this styling...?

jiromaykin (changes requested):
More questions than actual change-requests 🙃

Bartvaderkin (approved):
Pretty neat, only gripe I have is the name. I'd propose to call the whole thing `UserAppointments` instead of `MyAppointments`, similar to `UserFeed`. Because 'me' and 'my' names are a bit django tutorial and technically it can display any user.

Bartvaderkin (changes requested):
Pretty neat, only gripe I have is the name. I'd propose to call the whole thing UserAppointments instead of MyAppointments, similar to UserFeed. Because 'me' and 'my' names are a bit django tutorial and technically it can display any user.

src/open_inwoner/scss/components/PluginCard/PluginCard.scss
stevenbal: I thought I'd make this generic, because the appointments plugin cards have (almost) the same styling and I didn't want to duplicate it:

image

jiromaykin: OK - seems like the appointments don't need a notification color (yet) so approving this now.

jiromaykin (approved):

Bartvaderkin (approved):

PR #1123, opened 2024-03-28
[#2221] Fix klant creation with e-suite

pi-sigma:
- Make sure Klant is created via e-suite on user login
- Replace API specs with e-suite compatible versions
- Change default on OpenZaak and OpenKlant config to not use RSIN
- Refactor tests + mocks

Taiga: [#2221](https://taiga.maykinmedia.nl/project/open-inwoner/task/2221)


src/open_inwoner/openklant/clients.py
```diff
@@ -24,18 +24,68 @@
...
-            logger.exception("exception while making request", exc_info=e)
+        except (RequestException, ClientError):
+            logger.exception("exception while making request")
             return
 
         klant = factory(Klant, data)
 
         return klant
 
+    def _create_klant_for_bsn(self, user_bsn: str) -> Klant:
```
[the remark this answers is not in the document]
stevenbal: The same occurs for eHerkenning, is this an issue on the side of eSuite?

src/open_inwoner/openklant/services.py
```diff
@@ -15,12 +15,16 @@ def update_user_from_klant(request):
...
         return
 
-    klant = client.retrieve_klant(**get_fetch_parameters(request))
+    fetch_params = get_fetch_parameters(request)
 
-    if not klant:
+    if klant := client.retrieve_klant(**fetch_params):
+        msg = "retrieved klant for user"
+    elif klant := client.create_klant(**fetch_params):
+        msg = "created klant (%s) for user" % klant.klantnummer
```
stevenbal, line 23: could be an f-string, but not really an issue

src/open_inwoner/openklant/tests/test_signal.py
```diff
@@ -141,3 +145,55 @@ def test_update_user_after_login_skips_existing_email(self, m):
...
+            bsn="123456789",
+        )
+
+        request = RequestFactory().get("/dummy")
+        request.user = user
+        user_logged_in.send(User, user=user, request=request)
+
+        self.assertTimelineLog(
+            f"created klant ({self.klant_bsn['klantnummer']}) for user"
+        )
```
stevenbal, line 167: If this log is present, it does imply that things worked, but I would also add assertions to check that `m.request_history` contains the expected requests (one for the GET and one for the POST, and no others)

src/open_inwoner/openklant/tests/test_signal.py
```diff
@@ -141,3 +145,55 @@ def test_update_user_after_login_skips_existing_email(self, m):
...
+            kvk="87654321",
+        )
+
+        request = RequestFactory().get("/dummy")
+        request.user = user
+        user_logged_in.send(User, user=user, request=request)
+
+        self.assertTimelineLog(
+            f"created klant ({klant_eherkenning['klantnummer']}) for user"
+        )
```
stevenbal, line 199: same comment as above

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -57,7 +58,7 @@ def test_list_for_bsn(self, m, mock_get_kcm_answer_mapping):
 
         detail_url = reverse(
             "cases:contactmoment_detail",
-            kwargs={"kcm_uuid": data.klant_contactmoment["uuid"]},
+            kwargs={"kcm_uuid": uuid_from_url(data.klant_contactmoment["uuid"])},
```
stevenbal, line 61: I think you forgot to remove the `uuid` from the `klant_contactmoment` (applies to other cases in this test file as well)

stevenbal (changes requested):

src/open_inwoner/openklant/clients.py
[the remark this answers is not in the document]
pi-sigma: Same here. The mock for information retrieval that's needed for testing user creation on the side of the E-suite is disabled.

pi-sigma:
By default, `requests_mock` forces query params to be lower case  (see [https://requests-mock.readthedocs.io/en/latest/knownissues.html](https://requests-mock.readthedocs.io/en/latest/knownissues.html)). I thought it would be good to change that, since whether or not query params are considered case sensitive is not settled by RFC; it seems to depend on the application. We have other tests (`CaseListViewTests`) that could also be updated.

src/open_inwoner/openklant/clients.py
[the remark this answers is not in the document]
pi-sigma: With the mock enabled, things work as expected. Calls to the Klanten and Contactmomenten APIs return `200`/`201`, the klant shows up in the E-suite, question is created and linked to the new user.

post-success

src/open_inwoner/utils/test.py
```diff
@@ -84,3 +84,7 @@ class SessionMiddleware(DefaultSessionMiddleWare):
 
     def __init__(self, *args, **kwargs):
         super().__init__(get_response=lambda x: "dummy")
+
+
+def uuid_from_url(url: str) -> str:
+    return url.split("/")[-1]
```
Bartvaderkin, line 90: This could ideally do some checks to see if the bit looks like a UUID.

Bartvaderkin (approved):

stevenbal (approved):

stevenbal:
Tested it just now, works both for BSN and KVK :+1:

PR #1128, opened 2024-04-02
✨ [#2216] Add KCM survey link

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2216

Design for mobile devices will be implemented in a separate issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2294


src/open_inwoner/templates/master.html
```diff
@@ -136,6 +136,13 @@
             </main>
         {% endblock main_outer %}
 
+        {% get_solo 'configurations.SiteConfiguration' as config %}
+        {% if config.kcm_survey_link_url and config.kcm_survey_link_text %}
+            <div class="kcm-survey">
+                {% link bold=True href=config.kcm_survey_link_url text=config.kcm_survey_link_text blank=True hide_external_icon=True secondary=True %}
+            </div>
+        {% endif %}
```
stevenbal, line 144: I'm not entirely sure where to place to not mess with our tab order, I figured placing it before the footer made sense

src/open_inwoner/scss/components/KCMSurvey/KCMSurvey.scss
```diff
@@ -0,0 +1,13 @@
+.kcm-survey {
+  position: fixed;
+  top: 40%;
```
pi-sigma, line 3: Vertical positioning appears to be tricky since the thing is rotated. I note that the element is displayed slightly below the center of the screen.

src/open_inwoner/templates/master.html
pi-sigma: Alternatively: below the footer, since it is the only element that doesn't contain information relevant to the user; instead, it asks for the user's opinion about the entire site, so it would also make sense to put it at the end. However, your approach seems fine as well.

jiromaykin: I see on https://gemeente.groningen.nl/ they've indeed put it below the footer, right before the Body closes.
Currently we have the slightly similar floating element `<div class="anchor__scroll" id="scroll-anchor">` in Master above the Main section (because this used to be visible on Desktop) but I guess I'll have to move that one down in future also; so I would group these two together below the Footer (outside of the scope of this PR).

src/open_inwoner/scss/components/KCMSurvey/KCMSurvey.scss
jiromaykin: Yeah this is quite tricky to make perfect, so I feel like this is pretty close? If perfection is needed, I would remove the style from the 'kcm-survey' class and add the positioning to the link-styles (again positioning only) from the https://gemeente.groningen.nl/ "Uw mening" anchor link ;-) and then correct for our own padding and colors etc.

jiromaykin (changes requested):

src/open_inwoner/scss/components/KCMSurvey/KCMSurvey.scss
```diff
@@ -0,0 +1,13 @@
...
+  right: 0;
+  z-index: 1000;
+  transform: rotate(-90deg);
+  transform-origin: right bottom;
+  padding: var(--spacing-medium);
+  color: var(--color-info-darker);
```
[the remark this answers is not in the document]
stevenbal: Ah you're right, I'll add it to `.kcm-survey.link` specifically and remove it from `.kcm-survey`

src/open_inwoner/scss/components/KCMSurvey/KCMSurvey.scss
stevenbal: I'll take the styling from Groningen :+1:

jiromaykin (approved):

PR #1132, opened 2024-04-03
[#2288] Handle None questions in case detail view

pi-sigma:
Taiga: [#2288](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2288)

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_case_detail.py
+++ b/src/open_inwoner/openzaak/tests/test_case_detail.py
@@ -27,6 +27,7 @@ from zgw_consumers.constants import APITypes
 from open_inwoner.accounts.choices import LoginTypeChoices
 from open_inwoner.accounts.tests.factories import UserFactory, eHerkenningUserFactory
 from open_inwoner.cms.cases.views.status import InnerCaseDetailView, SimpleFile
+from open_inwoner.openklant.api_models import ObjectContactMoment
 from open_inwoner.openklant.clients import ContactmomentenClient
 from open_inwoner.openklant.constants import Status as ContactMomentStatus
 from open_inwoner.openklant.models import OpenKlantConfig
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -2327,3 +2327,17 @@ def test_kcm_redirect(self, m):
             status_code=302,
             target_status_code=200,
         )
+
+    @patch.object(
+        ContactmomentenClient,
+        "retrieve_objectcontactmomenten_for_zaak",
+        autospec=True,
+        return_value=[None],
```
Bartvaderkin, line 2335: This mock is not accurate and doesn't seem to test the modification. 

It makes `retrieve_objectcontactmomenten_for_zaak()` return a list with a single None value.

While the code change doesn't handle that. It loops over the `objectcontactmomenten` as if they can't be null and then does a null check on an attribute of an object in that list.

Bartvaderkin (changes requested):

src/open_inwoner/openzaak/tests/test_case_detail.py
pi-sigma: The modification ensures sure that `object.contactmoment` is not `None`, so that `contactmoment.registratiedatum` further down doesn't raise `AttributeError`. `getattr(None, "contactmoment", None)` returns `None`, in exactly the same way that `getattr(objectcontactmoment, "contactmoment", None)` would return `None` for an `objectcontactmoment` that happens to lack a `contactmoment`. The test will therefore pass with the modification and fail without it. I didn't see a reason to create a fake `objectcontactmoment`; the mechnism is exactly the same (`getattr -> AttributeError -> None`).

Also, the code does not assume AFAIK that `objectcontactmomenten` can't be null; it creates an empty list, calls `retrieve_objectcontactmomenten_for_zaak` to populate it, then loops over the list.

Bartvaderkin: You made a modification that handles if an external sub resource is None, so the test should test with that same subresource being None. And not rely on some random other None value somewhere that happens to make the test pass for some unrelated reason. 

With this test we still don't know if you modification works or makes sense, because the situation it handles is not present in the test.

Bartvaderkin (changes requested):

src/open_inwoner/openzaak/tests/test_case_detail.py
Bartvaderkin: You made a modification that handles if an external sub resource is None, so the test should test with that same subresource being None. And not rely on some random other None value somewhere that happens to make the test pass for some unrelated reason.

With this test we still don't know if you modification works or makes sense, because the situation it handles is not present in the test.

Bartvaderkin (approved):

PR #1134, opened 2024-04-04
:lipstick: [#2272] Position indicator dot headings

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2272

The nature of 'shrinking' a breaking text [where the wrap-width is unknown](https://stackoverflow.com/questions/37406353/make-container-shrink-to-fit-child-elements-as-they-wrap) turned out quite difficult; therefor I am using a  rather unique grid in order to make the indicator appear as close to the breaking text as possible with media-queries.

In this PR I am also fixing the ellipsis for long file-names 
➕ minor improvements of the case-detail page.

To test dot for file-list, increase default value of `DOCUMENT_RECENT_DAYS = config("DOCUMENT_RECENT_DAYS", default=1)`

### DIFF
--- a/src/open_inwoner/scss/components/File/File.scss
+++ b/src/open_inwoner/scss/components/File/File.scss
@@ -31,7 +31,6 @@
   &__data {
     display: flex;
     flex-direction: column;
-    adding: var(--spacing-large);
     font-family: var(--font-family-body);
     line-height: var(--spacing-extra-large);
     padding: var(--spacing-large);
--- a/src/open_inwoner/scss/components/File/File.scss
+++ b/src/open_inwoner/scss/components/File/File.scss
@@ -31,8 +31,18 @@
   &__data {
     display: flex;
     flex-direction: column;
+    adding: var(--spacing-large);
+    font-family: var(--font-family-body);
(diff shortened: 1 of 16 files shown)


src/open_inwoner/scss/components/File/File.scss
```diff
@@ -31,8 +31,18 @@
   &__data {
     display: flex;
     flex-direction: column;
+    adding: var(--spacing-large);
```
stevenbal, line 34: Typo probably? I see padding below, so I think this can be removed

stevenbal (approved):

src/open_inwoner/scss/components/File/File.scss
jiromaykin: Oops... this should not be approved :-D since there is another padding statement a couple of lines lower. I'll fix this.
There were multiple statements for this very same class in multiple places so I wanted to group them all together here.

jiromaykin: I have fixed this :-D

pi-sigma (approved):

PR #1137, opened 2024-04-05
:sparkles: [#2263] Setup configuration steps for authentication

stevenbal:
task: https://taiga.maykinmedia.nl/project/open-inwoner/task/2263

Implements setup configuration for:
- DigiD & eHerkenning via OpenID Connect
- Admin login via OpenID connect
- DigiD & eHerkenning via SAML

I also removed the old DigiD SAML settings, because they have been deprecated: https://github.com/maykinmedia/django-digid-eherkenning/blob/master/CHANGELOG.rst#050-2022-10-31


src/open_inwoner/configurations/bootstrap/auth.py
```diff
@@ -0,0 +1,482 @@
...
+        form_data = {
+            field.name: getattr(config, field.name)
+            for field in OpenIDConnectDigiDConfig._meta.fields
+        }
+
+        # Only override field values with settings if they are defined
+        for setting in self.all_settings:
+            value = getattr(settings, setting, None)
+            if value is not None:
+                model_field_name = setting.split("DIGID_OIDC_")[1].lower()
```
stevenbal, line 77: I rely on the name being prefixed by `DIGID_OIDC_` and the rest of the variable name being identical to the name of the model field. Some fields on the config start with `oidc_`, that's why some settings contain `OIDC_` twice (e.g. `DIGID_OIDC_OIDC_RP_CLIENT_ID`).

src/open_inwoner/configurations/bootstrap/auth.py
```diff
@@ -0,0 +1,490 @@
...
+        form_data["enabled"] = True
+
+        # Use the admin for to apply validation and fetch URLs from the discovery endpoint
+        form = OpenIDConnectConfigForm(data=form_data)
+        if not form.is_valid():
+            raise ConfigurationRunFailed(
+                f"Something went wrong while saving configuration: {form.errors}"
+            )
+
+        form.save()
```
pi-sigma, line 247: In other places, `form.save()` is wrapped in `try/except`, here it it is not.

pi-sigma (changes requested):

Bartvaderkin (changes requested):
It looks like this mixes regular settings with one-time configuration settings? Can we somehow split that in the code? Or are these always applied?

Also I'd really recommend to look into a solution that would add the description (and other info to generate the documentation) in the same place because it's going to be awful to manage multiple massive lists like this accurately.

src/open_inwoner/configurations/bootstrap/auth.py
stevenbal: The SAML models can throw `ValidationError`s from methods called in `.save()`, but the OIDC form/models don't do this (validation happens during `form.is_valid` instead)

stevenbal:
> It looks like this mixes regular settings with one-time configuration settings? Can we somehow split that in the code? Or are these always applied?

You mean `DIGID_ENABLED` right? I think I will split this one up to have a setting to enable the setup config for DigiD (SAML) and keep the old setting to enable DigiD in general (show the login button, etc)

src/open_inwoner/configurations/bootstrap/auth.py
```diff
@@ -0,0 +1,490 @@
...
+class DigiDOIDCConfigurationStep(BaseConfigurationStep):
+    """
+    Configure DigiD authentication via OpenID Connect
+    """
+
+    verbose_name = "Configuration for DigiD via OpenID Connect"
+    required_settings = [
+        "DIGID_OIDC_OIDC_RP_CLIENT_ID",
+        "DIGID_OIDC_OIDC_RP_CLIENT_SECRET",
+        # NOTE these are part of the model, but not actually part of the admin form
```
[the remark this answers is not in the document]
stevenbal: I added them to the form and removed the comments

Bartvaderkin (approved):
I approve it for now if we need it for this release, noting this overlaps with the other configuration tickets.

PR #1139, opened 2024-04-05
[#2207] Avoid showing the current status of the case twice in the situation where the zaak.status URL isn't in the status history

alextreme:
…


alextreme:
@pi-sigma from my analysis (see the Taiga issue) the duplication occurs regardless of zaaktype when our eSuite-workaround is necessary

alextreme:
By the way, Andy is checking if he can fix in the eSuite that this workaround is no longer necessary

pi-sigma:
That would be nice!

pi-sigma:
Not sure what can be done about the failing test. If `fetch_status_history` returns an empty list, it still seems necessary to add the status to `statuses` (not just the mapping). I haven't seen the strange behavior with the UUIDs Steven reports in the Taiga discussion.

However, the second test concerning missing statuses and the hidden upload button ([link](https://github.com/maykinmedia/open-inwoner/blob/c8ff97b56676dc6f26c3a18d082421e1360c0f99/src/open_inwoner/openzaak/tests/test_case_detail.py#L1225)) is broken; it passes independently of whether `fetch_status_history` returns an empty list or not. Should be marked with `@skip()` and reviewed (or deleted).

alextreme:
Discuss the failing tests at the office next week with Steven or Bart. If fetch_status_history returns an empty list then having statuses be the empty list makes sense, this seems to be different than the situation we see so imho this test is to be rewritten or if it's no longer applicable removed.

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -1175,53 +1175,6 @@ def test_document_ordering_by_name(self, m):
         self.assertEqual(documents[1].name, self.informatie_object["titel"])
         self.assertEqual(documents[2].name, self.informatie_object_no_date["titel"])
 
-    @patch(
-        "open_inwoner.openzaak.clients.ZakenClient.fetch_status_history", autospec=True
-    )
-    def test_e_suite_missing_current_status_fetch_status(
```
stevenbal, line 1181: @alextreme I'm now unsure what is still and issue and what is not, but from what I understood from the taiga issue, this regression test is no longer needed?

alextreme: Indeed it can be removed

pi-sigma (changes requested):
`test_e_suite_missing_current_status_upload_button_displayed` should probably also be removed

https://github.com/maykinmedia/open-inwoner/blob/013e1e23064eacdcd03dac9ff3c45614a96150ec/src/open_inwoner/openzaak/tests/test_case_detail.py#L1181

pi-sigma (approved):

PR #1140, opened 2024-04-05
[#2266] Add setup for siteconfiguration

pi-sigma:
Taiga: [#2266](https://taiga.maykinmedia.nl/project/open-inwoner/task/2266)


src/open_inwoner/configurations/bootstrap/siteconfig.py
```diff
@@ -0,0 +1,105 @@
...
+
+    verbose_name = "Site configuration"
+    required_settings = [
+        "SITE_NAME",
+        "SITE_SECONDARY_COLOR",
+        "SITE_ACCENT_COLOR",
+    ]
+
+    def is_configured(self):
+        return all((getattr(settings, required) for required in self.required_settings))
```
stevenbal, line 21: This should check if the required fields are set on the `SiteConfiguration` singleton, otherwise `is_configured` will always be true (because the settings have to be defined), causing the `configure` method to be skipped

src/open_inwoner/configurations/bootstrap/siteconfig.py
```diff
@@ -0,0 +1,105 @@
...
+class SiteConfigurationStep(BaseConfigurationStep):
+    """
+    Set up general configuration ("Algemene configuratie")
+    """
+
+    verbose_name = "Site configuration"
+    required_settings = [
+        "SITE_NAME",
+        "SITE_SECONDARY_COLOR",
+        "SITE_ACCENT_COLOR",
```
stevenbal, line 17: Is there a reason why the primary_color isn't required?

src/open_inwoner/configurations/bootstrap/siteconfig.py
```diff
@@ -0,0 +1,105 @@
...
+        for key, value in setting_to_config.items():
+            setting = getattr(settings, key)
+            if setting is not None:
+                setattr(config, value, setting)
+        config.save()
+
+    def test_configuration(self):
+        """
+        actual testing is done in final step
+        """
```
stevenbal, line 105: This comment can be replaced with a `...`, since we don't really have anything to test for this configuration

stevenbal (changes requested):

src/open_inwoner/configurations/bootstrap/siteconfig.py
pi-sigma: The `SiteConfiguration` model has a default for the primary color, so I thought it's not necessary to make this required?

stevenbal: Hmm that's true, though I do feel like if a municipality is automating this, the primary color of the website (and other colors in general) is something they should definitely always set explicitly

pi-sigma: Makes sense, I'll add it.

src/open_inwoner/configurations/bootstrap/siteconfig.py
```diff
@@ -12,13 +12,18 @@ class SiteConfigurationStep(BaseConfigurationStep):
 
     verbose_name = "Site configuration"
     required_settings = [
-        "SITE_NAME",
-        "SITE_SECONDARY_COLOR",
-        "SITE_ACCENT_COLOR",
+        "name",
+        "secondary_color",
+        "accent_color",
     ]
```
stevenbal, line 18: The `required_settings` list should contain the names of the settings in `conf.settings`, otherwise `validate_requirements` wont work: https://github.com/maykinmedia/django-setup-configuration/blob/main/django_setup_configuration/configuration.py#L16-L31

You could maybe define the `setting_to_config` mapping on the class level and then in `is_configured` you could loop over the `required_settings` and grab the config field name from the mapping

stevenbal (approved):

PR #1142, opened 2024-04-08
[#2260] Removed StatusTranslation model

Bartvaderkin:


src/open_inwoner/openzaak/api_models.py
```diff
@@ -69,17 +71,14 @@ def process_data(self) -> dict:
...
             "identification": self.identification,
             "uuid": str(self.uuid),
             "start_date": self.startdatum,
             "end_date": getattr(self, "einddatum", None),
             "description": self.zaaktype.omschrijving,
-            "current_status": status_translate.from_glom_multiple(
+            "current_status": glom_multiple(
                 self,
                 ("status.statustype.statustekst", "status.statustype.omschrijving"),
                 default="",
```
pi-sigma, line 84: For the `label` in `get_statuses_data` we have `default=_("No data available")`. Should we use this too for the `current_status`, or should the default be an empty string?

src/open_inwoner/openzaak/tests/test_notification_utils.py
```diff
@@ -65,7 +60,7 @@ def test_send_case_update_email(self):
         body_html = email.alternatives[0][0]
         self.assertIn(case.identificatie, body_html)
         self.assertIn(case.zaaktype.omschrijving, body_html)
-        self.assertIn("My Translated Status", body_html)
+        self.assertIn(status.statustype.omschrijving, body_html)
```
pi-sigma, line 63: Are we showing the `omschrijving` or the `statustekst`? They could be the same for some statustypes, different for others.

src/open_inwoner/userfeed/tests/hooks/test_case_status.py
```diff
@@ -99,7 +92,7 @@ def test_status_update(self, mock_get_active_app_names: Mock):
             strip_tags(item.message),
             escape(
                 _("Case status has been changed to '{status}'").format(
-                    status="translated status"
+                    status=status2.statustype.omschrijving
```
pi-sigma, line 95: See previous comment about `omschrijving` vs `statustekst`

pi-sigma (changes requested):
Nice cleanup! Couple minor questions.

alextreme:
Discussed with Groningen, and to ease the transition I'm delaying this PR until 1.17

src/open_inwoner/openzaak/api_models.py
Bartvaderkin: Nah, it should be the same so I added it.

src/open_inwoner/openzaak/tests/test_notification_utils.py
Bartvaderkin: Yes you are right, statustekst has priority, I updated the test

pi-sigma (approved):

PR #1145, opened 2024-04-08
[#2274] Pass data in Haalcenntraal POST as JSON

pi-sigma:
Taiga: [#2274](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2274)

### DIFF
--- a/src/open_inwoner/haalcentraal/tests/test_client.py
+++ b/src/open_inwoner/haalcentraal/tests/test_client.py
@@ -1,0 +1,21 @@
+from django.test import TestCase
+
+import requests_mock
+
+from ..api import BRP_2_1
+from .mixins import HaalCentraalMixin
+
+
+@requests_mock.Mocker()
+class BRPClientTest(HaalCentraalMixin, TestCase):
+    def test_request_content_type(self, m):
+        self._setUpMocks_v_2(m)
+        self._setUpService()
+
+        api_client = BRP_2_1()
+        response = api_client.make_request("999993847")
+
+        self.assertEqual(
+            response.request.body,
+            b'{"fields": ["naam.geslachtsnaam", "naam.voorletters", "naam.voornamen", "naam.voorvoegsel", "geslacht.omschrijving", "geboorte.plaats.omschrijving", "geboorte.datum.datum", "verblijfplaats.verblijfadres.officieleStraatnaam", "verblijfplaats.verblijfadres.huisnummer", "verblijfplaats.verblijfadres.huisletter", "verblijfplaats.verblijfadres.huisnummertoevoeging", "verblijfplaats.verblijfadres.postcode", "verblijfplaats.verblijfadres.woonplaats"], "type": "RaadpleegMetBurgerservicenummer", "burgerservicenummer": ["999993847"]}',
+        )
--- a/src/open_inwoner/haalcentraal/tests/test_client.py
+++ b/src/open_inwoner/haalcentraal/tests/test_client.py
@@ -1,3 +1,5 @@
+import json
+
 from django.test import TestCase
 
 import requests_mock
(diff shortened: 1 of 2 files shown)


src/open_inwoner/haalcentraal/tests/test_client.py
```diff
@@ -0,0 +1,21 @@
...
+
+import requests_mock
+
+from ..api import BRP_2_1
+from .mixins import HaalCentraalMixin
+
+
+@requests_mock.Mocker()
+class BRPClientTest(HaalCentraalMixin, TestCase):
+    def test_request_content_type(self, m):
```
stevenbal, line 11: I think this test should also check if the request in `m.request_history` has the `Content: application/json` request header, because now we're just testing if the mock returns the right response.

src/open_inwoner/haalcentraal/tests/test_client.py
```diff
@@ -0,0 +1,21 @@
...
+        self._setUpMocks_v_2(m)
+        self._setUpService()
+
+        api_client = BRP_2_1()
+        response = api_client.make_request("999993847")
+
+        self.assertEqual(
+            response.request.body,
+            b'{"fields": ["naam.geslachtsnaam", "naam.voorletters", "naam.voornamen", "naam.voorvoegsel", "geslacht.omschrijving", "geboorte.plaats.omschrijving", "geboorte.datum.datum", "verblijfplaats.verblijfadres.officieleStraatnaam", "verblijfplaats.verblijfadres.huisnummer", "verblijfplaats.verblijfadres.huisletter", "verblijfplaats.verblijfadres.huisnummertoevoeging", "verblijfplaats.verblijfadres.postcode", "verblijfplaats.verblijfadres.woonplaats"], "type": "RaadpleegMetBurgerservicenummer", "burgerservicenummer": ["999993847"]}',
```
stevenbal, line 20: I would define this as a dictionary outside the assert and then do `json.dumps(data).encode("utf-8")`, that way it can be formatted with black and we don't have this big one-liner

stevenbal (changes requested):

src/open_inwoner/haalcentraal/tests/test_client.py
```diff
@@ -0,0 +1,44 @@
...
+                "verblijfplaats.verblijfadres.huisletter",
+                "verblijfplaats.verblijfadres.huisnummertoevoeging",
+                "verblijfplaats.verblijfadres.postcode",
+                "verblijfplaats.verblijfadres.woonplaats",
+            ],
+            "type": "RaadpleegMetBurgerservicenummer",
+            "burgerservicenummer": ["999993847"],
+        }
+
+        self.assertEqual(response.request.body, json.dumps(data).encode("utf-8"))
```
Bartvaderkin, line 44: While this works for now I'd like to note I would've gone the other way: JSON parse the request body and compare it to the expected dict. 

JSON serialization is technically not 100% stable (there are multiple valid JSON strings possible for the same data, because of dict ordering and pretty pretting), and in the assertion you'd get a structural diff instead of a text diff.

pi-sigma: Good point!

src/open_inwoner/haalcentraal/tests/test_client.py
```diff
@@ -16,6 +18,27 @@ def test_request_content_type(self, m):
...
+                "verblijfplaats.verblijfadres.huisletter",
+                "verblijfplaats.verblijfadres.huisnummertoevoeging",
+                "verblijfplaats.verblijfadres.postcode",
+                "verblijfplaats.verblijfadres.woonplaats",
+            ],
+            "type": "RaadpleegMetBurgerservicenummer",
+            "burgerservicenummer": ["999993847"],
+        }
+
+        self.assertEqual(json.loads(response.request.body), data)
```
stevenbal, line 44: this could have used `response.json`, but I'll approve to get this merged for the release

stevenbal (approved):

src/open_inwoner/haalcentraal/tests/test_client.py
alextreme: Agreed, feels unnecessary

PR #1150, opened 2024-04-09
[#2297] Add documentation for automating configurations

pi-sigma:
Documentation for configuration setup via environment variables:

- [x] General configuration
- [x] Klanten
- [x] ZGW
- [x] Authentication (DigiD, eHerkenning, Azure)
- [ ] CMS
- [x] Document non-supported features
- [x] Document usage for deployment/CI
- [x] Automate creation of docs

Docs are generated by running `python src/manage.py generate_config_docs`. Pass one of the supported options ("siteconfig", "kic", "zgw", "digid_oidc", "eherkenning_oidc") to generate docs for a specific config, or run the command without arguments to generate all.

The docs are made available under `docs/configuration/".

Taiga: [#2297](https://taiga.maykinmedia.nl/project/open-inwoner/task/2297)

Related: 
- #1116 
- #1140
- #1137


src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Meta class for populating configuration classes with descriptors
+    """
+
+    def __init__(cls, name, bases, dct):
+        model_fields = (
+            field
+            for field in vars(cls.model)
+            if not field.startswith("__") and field not in cls.excluded_fields
+        )
```
Bartvaderkin, line 45: This should filter on the Field base class (the way the `_iter_descriptors` does) (there could be more members without an underscore then just fields)

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,124 @@
...
+    def __get__(self, instance, objtype=None):
+        return config(self.get_setting_name(), default=None)
+
+    def __set__(self, instance, value):
+        pass
+
+    def get_setting_name(self) -> str:
+        if setting_name := getattr(self, "setting_name", None):
+            return setting_name
+        return f"{self.namespace}_" + self.name.upper()
```
Bartvaderkin, line 32: This is odd: the `def __set_name__(self, owner, name):` does similar and `self.setting_name` should always be there.

src/open_inwoner/configurations/bootstrap/siteconfig.py
```diff
@@ -4,88 +4,17 @@
...
-        "SITE_OPENID_DISPLAY": "openid_display",
-        "SITE_REDIRECT_TO": "redirect_to",
-        "SITE_ALLOW_MESSAGES_FILE_SHARING": "allow_messages_file_sharing",
-        "SITE_HIDE_CATEGORIES_FROM_ANONYMOUS_USERS": "hide_categories_from_anonymous_users",
-        "SITE_HIDE_SEARCH_FROM_ANONYMOUS_USERS": "hide_search_from_anonymous_users",
-        "SITE_DISPLAY_SOCIAL": "display_social",
-        "SITE_EHERKENNING_ENABLED": "eherkenning_enabled",
-    }
+    required_settings = SiteConfigurationSettings.get_required_settings()
+    setting_to_config = SiteConfigurationSettings.get_config_mapping()
```
Bartvaderkin, line 17: This looks like an artifact from earlier implementation. 

Class attributes run at the module level (eg: import time). it is convenient to declare static data at class level but we don't really need the data at that time. Instead you can just call these methods when the data is actually needed.

src/open_inwoner/configurations/bootstrap/constants.py
```diff
@@ -0,0 +1,229 @@
...
+
+from .utils import generate_api_fields_from_template
+
+CONFIDENTIALITY_CHOICES = [
+    choice[0] for choice in VertrouwelijkheidsAanduidingen.choices
+]
+FILE_EXTENSION_CHOICES = [ext for ext in generate_default_file_extensions()]
+USER_INFO_CLAIM_SOURCES_CHOICES = [
+    choice[0] for choice in UserInformationClaimsSources.choices
+]
```
Bartvaderkin, line 14: 'Choices' as naming convention in Django means a value and a label. If it is just a set of single values then call it something else.

docs/install/configurations.rst
```diff
@@ -0,0 +1,341 @@
...
+    SITE_PRIMARY_FONT_COLOR="#111111"
+    SITE_SECONDARY_FONT_COLOR="#222222"
+    SITE_ACCENT_FONT_COLOR="#333333"
+    SITE_WARNING_BANNER_ENABLED=<True | False>
+    SITE_WARNING_BANNER_TEXT="warning banner text"
+    SITE_WARNING_BANNER_BACKGROUND_COLOR="#444444"
+    SITE_WARNING_BANNER_FONT_COLOR="#555555"
+    SITE_LOGIN_SHOW=False
+    SITE_LOGIN_ALLOW_REGISTRATION=<True | False>
+    SITE_LOGIN_2FA_SMS=<True | False>
```
Bartvaderkin: These lists of values should be sorted so it more readable and related values are grouped together. Also I miss the meta information like type, description, default etc.

src/open_inwoner/configurations/management/commands/create_docs.py
```diff
@@ -0,0 +1,90 @@
...
+TEMPLATE_DIR = (
+    Path(os.path.abspath(os.path.dirname(__file__))).parent.parent
+    / "bootstrap"
+    / "templates"
+)
+TARGET_DIR = (
+    Path(os.path.abspath(os.path.dirname(__file__))).parent.parent.parent.parent.parent
+    / "docs"
+    / "configuration"
+)
```
Bartvaderkin, line 27: I'm really curious about the thought process here. Surely we can do better then `.parent.parent.parent.parent.parent`? Where does that even go?

Start from `settings.BASE_DIR` if the target is not in immediate vicinity of the local file.

src/open_inwoner/configurations/bootstrap/constants.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+siteconfig_fields = {
+    "name": {"values": "string"},
```
Bartvaderkin, line 18: I'm slightly mystified why this is here: went for pulling the fields and the meta information from the config model, instead of manually declaring every field using the class+descriptors because you didn't want to make declarations for every field. 

But this still declares info for every field.. and as loose dictionaries instead of something more controlled like with the class+descriptors. :confused:

Bartvaderkin (changes requested):

docs/install/configurations.rst
pi-sigma: I added more information as part of the refactoring. Not sure how to approach the sorting. I agree the order could be improved, I don't see an easy way to automate this.

src/open_inwoner/configurations/management/commands/create_docs.py
```diff
@@ -0,0 +1,87 @@
...
+        config_option = kwargs["config_option"]
+        if not config_option or config_option not in SUPPORTED_OPTIONS:
+            return
+
+        config = self.get_config(config_option)
+
+        required_settings = [
+            config.get_setting_name(field) for field in config.required_fields
+        ]
+        all_settings = [config.get_setting_name(field.name) for field in config.fields]
```
Bartvaderkin, line 76: You could also sort here where we're iterating over fields.

docs/install/configurations.rst
Bartvaderkin: I'd assume if you can print a list of things you could sort them before the printing? I put some notes of possible locations.

src/open_inwoner/configurations/management/commands/generate_config_docs.py
```diff
@@ -0,0 +1,115 @@
...
+        self,
+        template_path: os.PathLike,
+        template_variables: dict[str, list],
+        output_path: os.PathLike,
+    ):
+        with open(template_path, "r") as template:
+            template_str = template.read()
+            template = Environment(
+                loader=FileSystemLoader(TEMPLATE_DIR), autoescape=select_autoescape()
+            ).from_string(template_str)
```
Bartvaderkin, line 77: We should stick to one template engine, in this case Django templates, and follow the regular template finder/loader pattern, instead of introducing custom Jinja that is not used elsewhere and afaics not specifically necessary here. 

So move the templates to a normal Django app template directory, with a subdirectory named 'configurations' (the app name) and a normal .rst extension (without .template). They are very specific to this app so you can keep the templates in the app directory (and not the shared templates in the project root).

src/open_inwoner/configurations/bootstrap/constants.py
```diff
@@ -0,0 +1,154 @@
...
+    "all": [],
+    "required": [
+        "oidc_rp_client_id",
+        "oidc_rp_client_secret",
+    ],
+}
+
+populate_fields(
+    eherkenning_oidc_fields, require=None, exclude=exclude, all_fields=all_fields
+)
```
Bartvaderkin, line 154: I'd expect these blocks to be part of each ConfigSettingsBase subclass. 

Is there a reason these live and run on module level? I guess maybe to put them as class attributes on the actual bases?

Class attributes on class-based-design are a convenience for easy overrides, not a requirement. So these blocks could just be a `get_fields()` method on each ConfigSettingsBase subclass. (this would also make it more dynamic)

src/open_inwoner/configurations/bootstrap/utils.py
```diff
@@ -0,0 +1,82 @@
...
+            name=field.name,
+            verbose_name=field.verbose_name,
+            description=field.description,
+            field_type=field.get_internal_type(),
+            default_value=get_default_value(field),
+            values=get_example_value(field),
+        )
+        fields["all"].append(config_field)
+        if require and config_field.name in require:
+            fields["required"].append(config_field)
```
Bartvaderkin, line 34: In general let's try to move away from ad-hoc dicts with important keys and use dataclasses.

src/open_inwoner/configurations/management/commands/generate_config_docs.py
```diff
@@ -0,0 +1,115 @@
...
+        return mapping[config_option]
+
+    def get_detailed_info(self, config: ConfigSetting) -> list[str]:
+        ret = []
+        for field in config.fields["all"]:
+            model_field = config.model._meta.get_field(field.name)
+            part = []
+            part.append(f"{'Variable':<20}{config.get_setting_name(field)}")
+            part.append(f"{'Setting':<20}{str(model_field.verbose_name)}")
+            part.append(f"{'Description':<20}{str(model_field.help_text)}")
```
Bartvaderkin, line 60: At this point you shouldn't need the model field anymore, the `populate_fields()` earlier was already responsible to collect the meta data (eg: `verbose_name` and `help_text`) and put them in the `ConfigField` instances that would them be the abstraction and source of truth for everything else that follows.

Bartvaderkin (changes requested):

src/open_inwoner/configurations/management/commands/generate_config_docs.py
pi-sigma: I couldn't find a way to get rid of spacing between the settings (they are all separted by a newline when generated). `{% spaceless %}` doesn't seem to work, and Jinja has a convenient method to trim blocks of content (e.g. `{%- content %}` will eliminate newlines before `content`). Is there a way to achieve this with Django templates?

src/open_inwoner/configurations/bootstrap/templates/digid_oidc.rst.template
```diff
@@ -0,0 +1,11 @@
...
+
+{% block link -%}
+.. _digid_oidc:
+{% endblock -%}
+
+{% block title -%}
+========================
+DigiD OIDC configuration
+========================
+{% endblock %}
```
stevenbal, line 11: I feel like having a separate template for each config is not completely necessary? The only differing elements are the rst link and the name, which we could probably configure somewhere. This is not super crucial right now though

docs/configuration/digid_oidc.rst
```diff
@@ -0,0 +1,195 @@
...
+
+Detailed Information
+====================
+
+::
+
+    Variable            DIGID_OIDC_OIDC_RP_CLIENT_ID
+    Setting             OpenID Connect client ID
+    Description         OpenID Connect client ID provided by the OIDC Provider
+    Model field type    CharField
```
stevenbal, line 58: I know we need this to figure out the possible values, but we probably don't need to display it in the actual docs

src/open_inwoner/configurations/bootstrap/utils.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def get_default_value(field: Any) -> str:
+    default = field.default
+
+    if default is NOT_PROVIDED:
+        return "No default"
+    if not isinstance(default, (str, bool, int)):
+        return "No information"
```
stevenbal, line 43: In which cases does this happen?

src/open_inwoner/configurations/management/commands/generate_config_docs.py
```diff
@@ -0,0 +1,115 @@
...
+ConfigSetting: TypeAlias = (
+    DigiDOIDCConfigurationSettings
+    | KICConfigurationSettings
+    | SiteConfigurationSettings
+    | ZGWConfigurationSettings
+    | eHerkenningDOIDCConfigurationSettings
+)
+
+
+SUPPORTED_OPTIONS = ["siteconfig", "kic", "zgw", "digid_oidc", "eherkenning_oidc"]
```
stevenbal, line 27: I'm seeing the same list quite a few times here, is there a way to make this simpler such that we don't have to add a new configuration in multiple places?

stevenbal (changes requested):
Some minor comments and some comments that we should probably create separate issues for. It might also be an idea to move this documentation functionality to the django-setup-configuration library at some point

src/open_inwoner/configurations/bootstrap/utils.py
pi-sigma: Sometimes the default is a function. For example, for `DIGID_OIDC_OIDC_RP_SCOPES_LIST`, the default will be `&lt;function get_default_scopes_bsn at 0x7f0c994dcb80&gt;`

stevenbal: I think in that case we could do a check using `callable` (https://docs.python.org/3/library/functions.html#callable) and in case it is callable we can call it to get the defaults

pi-sigma: Good idea, thanks!

src/open_inwoner/configurations/management/commands/generate_config_docs.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get_config(self, config_option: str) -> dict[str, ConfigSetting]:
+        mapping = {
+            "siteconfig": SiteConfigurationSettings,
+            "kic": KICConfigurationSettings,
+            "zgw": ZGWConfigurationSettings,
+            "digid_oidc": DigiDOIDCConfigurationSettings,
+            "eherkenning_oidc": eHerkenningDOIDCConfigurationSettings,
+        }
+        return mapping[config_option]
```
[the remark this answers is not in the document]
pi-sigma: Not quite there yet, but I've improved this

docs/configuration/kic.rst
```diff
@@ -0,0 +1,97 @@
...
+"""""""""
+
+::
+
+    KIC_CONFIG_CONTACTMOMENTEN_API_CLIENT_ID
+    KIC_CONFIG_CONTACTMOMENTEN_API_CLIENT_SECRET
+    KIC_CONFIG_CONTACTMOMENTEN_API_ROOT
+    KIC_CONFIG_KLANTEN_API_CLIENT_ID
+    KIC_CONFIG_KLANTEN_API_CLIENT_SECRET
+    KIC_CONFIG_KLANTEN_API_ROOT
```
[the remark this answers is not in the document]
pi-sigma: This will need some tinkering

pi-sigma:
I've disabled autoescape because it messes up the output. It did work previously with `autoescape=select_autoescape()`, but stopped working after the refactor. If we want to keep it on, it will need some investigation, along with the template engine issue brought up by @Bartvaderkin

pi-sigma:
@stevenbal @Bartvaderkin 

You'll notice that I've changed some of the setting names (e.g. `klanten_api_client_id` -> `klanten_service_client_id`), because the name of a setting variable is currently determined on the basis of the name of the model field (`klanten_service_client_id` -> `KIC_CONFIG_KLANTEN_SERVICE_CLIENT_ID`). Previously I more or less hand rolled the data for `ForeignKey`s like API services. If you have alternative suggestions for moving from model fields to settings variables, let me know.

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,296 @@
...
+        "zaak_max_confidentiality",
+    )
+    excluded_fields = ()
+
+
+class DigiDOIDCConfigurationSettings(ConfigSettingsBase):
+    model = OpenIDConnectDigiDConfig
+    display_name = "DigiD OIDC Configuration"
+    namespace = "DIGID_OIDC"
+    api_fields = tuple()
```
Bartvaderkin, line 244: These `api_fields` seems lost? I don't see them used.

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,296 @@
...
+        "document_max_confidentiality",
+        "enable_categories_filtering_with_zaken",
+        "fetch_eherkenning_zaken_with_rsin",
+        "max_upload_size",
+        "reformat_esuite_zaak_identificatie",
+        "skip_notification_statustype_informeren",
+        "title_text",
+        "zaak_max_confidentiality",
+    )
+    excluded_fields = ()
```
Bartvaderkin, line 237: Here the empty value is `()`, but elsewhere it is `tuple()` (but see earlier note that you should set them on the base class and not per subclass).

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,296 @@
...
+class ConfigSettingsBase:
+    model: ConfigModel
+    display_name: str
+    namespace: str
+    required_fields: tuple[str, ...]
+    included_fields: tuple[str, ...]
+    excluded_fields: tuple[str, ...]
+
+    def __init__(self):
+        self.config_fields = Fields(all=set(), required=set())
```
Bartvaderkin, line 40: See earlier not about default_factory.

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,296 @@
...
+        return f"{cls.namespace}_" + field.name.upper()
+
+    @staticmethod
+    def get_default_value(field: Any) -> str:
+        default = field.default
+
+        if default is NOT_PROVIDED:
+            return "No default"
+        if callable(default):
+            default = default.__call__()
```
Bartvaderkin, line 60: Sometimes I wonder how you come up with stuff.. like I have never seen a callable being called with `__call__()` :confused: 

Let's just call it:

````python
    if callable(default):
        default = default()
````

Anyway, as a base rule it is a big smell if you need dunder methods for general tasks, so if you feel you need to then reevaluate if you're on the right track.

**Review by Bartvaderkin — CHANGES_REQUESTED** [review]

**Review by stevenbal — CHANGES_REQUESTED** [review]
Mostly minor comments, looks good overall. I created a separate issue (taiga 2419) to use the docstring of the `ConfigurationSteps` in the generated documentation as well

File: src/open_inwoner/configurations/bootstrap/typing.py [hunk 23]
```diff
@@ -0,0 +1,17 @@
+from typing import TypeAlias
```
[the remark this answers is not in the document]
pi-sigma: I wanted this too, but creates a circular import becase `typing.py` imports the settings classes from `models`

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,286 @@
...
+                )
+
+                if config_field.name in self.required_fields:
+                    self.config_fields.required.add(config_field)
+
+                # use combination of whitelist/blacklist for all fields
+                if self.included_fields:
+                    if (
+                        config_field.name in self.included_fields
+                        and config_field not in self.excluded_fields
```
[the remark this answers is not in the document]
pi-sigma: I'm getting inconsistent results if I try to do without this (admittedly complicated) test: too many fields get included in `all_fields` (I've renamed `included_fields` per your suggestion). Will look into this at a later point, perhaps when moving this functionality to the lib.

src/open_inwoner/configurations/bootstrap/models.py
```diff
@@ -0,0 +1,286 @@
...
+    def get_required_settings(self) -> tuple[str, ...]:
+        return tuple(
+            self.get_setting_name(field) for field in self.config_fields.required
+        )
+
+    def get_config_mapping(self) -> dict[str, ConfigField]:
+        return {self.get_setting_name(field): field for field in self.config_fields.all}
+
+
+class SiteConfigurationSettings(ConfigSettingsBase):
```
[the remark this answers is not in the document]
pi-sigma: Makes sense, included in the latest push

stevenbal (changes requested):
Just the final comment regarding the descriptions for JSONFields and PrivateMediaFileFields, if that is addressed I'll approve :)

pi-sigma:
@stevenbal 
Descriptions are added, though I had to add a description for `FileField` instead of `PrivateMediaField` because the former is returned as the type at runtime. We probably have to deal with the different kinds of `PrivateMediaField` separately, so I added a description for `ImageField` as well (even though we're not currently using it).

I simplified the filtering of config fields. Think we can double-check this when we migrate to the library.

The migration user story now has an issue for checking the handling relational fields: https://taiga.maykinmedia.nl/project/open-inwoner/task/2437

stevenbal (approved):

alextreme:
@Bartvaderkin please re-review this PR, imho we can merge this in without the CMS envvars

src/open_inwoner/configurations/bootstrap/base.py
```diff
@@ -0,0 +1,166 @@
...
+            values = [choice[0] for choice in choices]
+            return ", ".join(values)
+
+        # other fields
+        field_type = field.get_internal_type()
+        match field_type:
+            case item if item in BasicFieldDescription.names:
+                return getattr(BasicFieldDescription, field_type)
+            case _:
+                return "No information available"
```
Bartvaderkin, line 93: This could be a normal if/else statement that would be readable at a glance.

Bartvaderkin (approved):

PR #1160, opened 2024-04-16
[#2259] Show case result in list view

Bartvaderkin:
https://taiga.maykinmedia.nl/project/open-inwoner/task/2259

### DIFF
--- a/src/open_inwoner/openzaak/cases.py
+++ b/src/open_inwoner/openzaak/cases.py
@@ -13,7 +13,7 @@ from .utils import is_zaak_visible
 logger = logging.getLogger(__name__)
 
 
-def resolve_zaak_type(case: Zaak, client: CatalogiClient = None) -> None:
+def resolve_zaak_type(case: Zaak, client: CatalogiClient | None = None) -> None:
     """
     Resolve `case.zaaktype` (`str`) to a `ZaakType(ZGWModel)` object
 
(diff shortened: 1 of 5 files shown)


src/open_inwoner/openzaak/cases.py
```diff
@@ -6,14 +6,14 @@
...
-from .clients import build_client
+from .clients import CatalogiClient, ZakenClient, build_client
 from .models import ZaakTypeConfig, ZaakTypeStatusTypeConfig
 from .utils import is_zaak_visible
 
 logger = logging.getLogger(__name__)
 
 
-def resolve_zaak_type(case: Zaak, client=None) -> None:
+def resolve_zaak_type(case: Zaak, client: CatalogiClient = None) -> None:
```
pi-sigma, line 16: `client: CatalogiClient | None = None`

Similarly for the other annotations

src/open_inwoner/utils/glom.py
```diff
@@ -0,0 +1,13 @@
+from glom import glom
+
+
+def glom_multiple(obj, paths: tuple[str, ...], *, default=None):
+    if not len(paths) > 1:
+        raise ValueError("no paths provided")
```
pi-sigma, line 6: A bit misleading, as someone might pass a single path, thinking that `glom_multiple` works for both cases. Perhaps `ValueError("glom_multiple requires a tuple of at least two paths")`.

pi-sigma (approved):

pi-sigma:
I double-checked by changing statuses/results on the E-suite and now I'm not sure it's working as intended. On the list view, the result is shown instead of the status, but the detail view still displays the status

E-suite:
[
esuite
](url)


List:

list



Detail:

detail

Bartvaderkin:
The ticket was to change the list to replace the status with the same value as the result on the detail page. 

I didn't change anything on the detail page.

pi-sigma:
Let's discuss next week and create a separate ticket for the detail view, if necessary, then we can merge this. Unless @alextreme thinks otherwise.

alextreme:
@pi-sigma good comment, I've updated the user story 2208. The change is as requested so we'll leave it at this due to the lack of remaining time in this sprint

PR #1161, opened 2024-04-16
:lipstick: [#2194] Correct use of Sections and other gaps

jiromaykin:
This PR solves wrong section styles and some other faulty gaps/margins/paddings here and there.
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2194 

the main issue is a bit of a ['blame' because some styles were put inside 'contactmomenten' that did not belong there](https://github.com/maykinmedia/open-inwoner/pull/1045/files#diff-715ac754308c4741d154fa5bad16a14a6a78c544ac23aaff2571b33a8cd0e2a2R120)  + all Section tags in the entire project were styled wrong because of the CSS that was targeting all Section Tags instead of a `.[componentname]-section` class (we need to target components instead of everything everywhere all at once 🙂)

inside scope of this issue:

- [x] remove margins from section tags, check to see if section styling doesn't break everywhere

outside scope of this issue so did is as an extra

- [x] delete pagination styling from wrong file - move to .pagination component and remove table display
- [x] improve card spacing for Userfeed and other Cards
- [x] remove unnecessary margin on Mijn Aanvragen page

### DIFF
--- a/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
+++ b/src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
@@ -44,8 +44,7 @@
   }
 
   & .grid {
-    grid-row-gap: var(--spacing-medium);
-    grid-column-gap: var(--spacing-extra-large);
+    grid-gap: var(--spacing-extra-large);
 
     // Tablets
     @media (min-width: 500px) and (max-width: 767px) {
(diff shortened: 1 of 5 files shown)


src/open_inwoner/scss/components/Contactmomenten/Contactmomenten.scss
```diff
@@ -99,15 +98,3 @@
     padding: var(--spacing-medium) 0 0 0;
   }
 }
```
jiromaykin, line 100: Removing this code presented the original issue of wrongly styled sections.

pi-sigma (approved):

PR #1163, opened 2024-04-18
:heavy_plus_sign: Re-add beautifulsoup4 to deps

stevenbal:
this was previously removed, but it was still in use

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,6 +20,8 @@ attrs==21.2.0
     # via
     #   glom
     #   jsonschema
+beautifulsoup4==4.12.3
+    # via -r requirements/base.in
 billiard==3.6.4.0
     # via celery
 boltons==21.0.0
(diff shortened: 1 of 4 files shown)


requirements/base.txt
```diff
@@ -481,7 +482,13 @@ requests==2.31.0
     #   maykin-python3-saml
     #   messagebird
     #   mozilla-django-oidc
+    #   requests-mock
+    #   responses
     #   zgw-consumers
+requests-mock==1.12.1
+    # via maykin-python3-saml
+responses==0.25.0
+    # via maykin-python3-saml
```
stevenbal, line 491: I'm not sure why these are added now?

Bartvaderkin:
We looked at the mystery changes but they appear now because Alex didn't recompile after a manual saml/eherkenning update, and have some test dependencies because it looks like saml includes those.

Bartvaderkin (approved):

requirements/base.txt
alextreme: Sorry, my bad

PR #1165, opened 2024-04-18
:white_check_mark: [#2324] Add test for setup_configuration command

stevenbal:


Bartvaderkin (changes requested):
Could this be a little dryer? Every step exists 4 times here (one as patch, three as new instances), and there are quite a lot.

Bartvaderkin (approved):
Cool, so dynamic.

PR #1170, opened 2024-04-22
:sparkles: [#2303] Improvements/fixes for Laposta

stevenbal:
Fixes for laposta integration after testing done in #2303

Changes:
* update LapostaConfig helptext to explain about caching
* Display Laposta form errors and keep selectboxes checked
* inject `toestemming` field


src/open_inwoner/laposta/forms.py
```diff
@@ -58,13 +63,14 @@ def save(self, request, *args, **kwargs):
...
         newsletters = self.cleaned_data["newsletters"]
+        has_errors = False
 
         list_name_mapping = dict(self.fields["newsletters"].choices)
         user_data = UserData(
             ip=get_client_ip(request)[0],
             email=user.verified_email,
             source_url=None,
-            custom_fields=None,
+            custom_fields={"toestemming": "Ja, ik wil de nieuwsbrief ontvangen"},
```
stevenbal, line 73: this is actually the expected value, instead of `True`

Bartvaderkin (approved):

src/open_inwoner/laposta/forms.py
alextreme: @stevenbal ugh. Yes, please make this configurable on our end. Then we can use the same text for asking permission.

Are you sure this is the same value for all newsletters? This feels like something which is newsletter-specific...

stevenbal: From what I saw, it seems to be the same for all forms (I would have to double check), but it is configurable per form so it's technically possible that they're different

pi-sigma (approved):

src/open_inwoner/laposta/forms.py
alextreme: Split off my request into https://taiga.maykinmedia.nl/project/open-inwoner/issue/2413 , it's out of scope

PR #1172, opened 2024-04-22
:wheelchair: [#2346] Remove spans, correct list-items

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2346 

Turns out there were faulty Spans around LI's inside UL's in some Cards pages + also wrong use of paragraph tags in case of empty list-items. Also: Headings should not be a direct child inside UL's.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_profile_views.py
+++ b/src/open_inwoner/accounts/tests/test_profile_views.py
@@ -1257,12 +1257,11 @@ class UserAppointmentsTests(ClearCachesMixin, WebTest):
 
         passport_appointment = PQ(cards[0]).find("ul").children()
 
-        self.assertEqual(passport_appointment[0].text, "Aanvraag paspoort")
-        self.assertEqual(PQ(passport_appointment[1]).text(), "Datum\n1 januari 2020")
-        self.assertEqual(PQ(passport_appointment[2]).text(), "Tijd\n13:00 uur")
-        self.assertEqual(PQ(passport_appointment[3]).text(), "Locatie\nHoofdkantoor")
-        self.assertEqual(PQ(passport_appointment[4]).text(), "Dam 1")
-        self.assertEqual(PQ(passport_appointment[5]).text(), "1234 ZZ Amsterdam")
+        self.assertEqual(PQ(passport_appointment[0]).text(), "Datum\n1 januari 2020")
+        self.assertEqual(PQ(passport_appointment[1]).text(), "Tijd\n13:00 uur")
+        self.assertEqual(PQ(passport_appointment[2]).text(), "Locatie\nHoofdkantoor")
+        self.assertEqual(PQ(passport_appointment[3]).text(), "Dam 1")
+        self.assertEqual(PQ(passport_appointment[4]).text(), "1234 ZZ Amsterdam")
         self.assertEqual(
             PQ(cards[0]).find("a").attr("href"),
             f"{self.data.config.booking_base_url}{self.data.appointment_passport.publicId}",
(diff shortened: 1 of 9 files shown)


jiromaykin:
Adding both @stevenbal and @pi-sigma as reviewers since you've worked with these pages and can check to see if the stying is still the same after these changes, in for example "Mijn afspraken" etc.

src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -186,41 +184,38 @@ <h4 class="card__heading-4"><span class="link link__text">{% trans "Mijn Interes
...
+                                        <li class="card__text--small">{% trans "U heeft geen gemeentelijke begeleider." %}</li>
                                     {% endfor %}
-                                </span>
-                            {% endrender_list %}
+                                {% endrender_list %}
 
-                            {% if inbox_page_is_published %}
-                                <span class="link link--icon link--secondary" aria-label="{% trans "Stuur een bericht" %}" title="{% trans "Stuur een bericht" %}">
+                                {% if inbox_page_is_published %}
+                                    <span class="link link--icon link--secondary" aria-label="{% trans "Stuur een bericht" %}" title="{% trans "Stuur een bericht" %}">
```
pi-sigma, line 198: Do we need the `aria-label` here, since we have "Stuur een bericht" immediately below?

src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -259,57 +251,60 @@ <h4 class="card__heading-4"><span class="link link__text">{% trans "Acties" %}</
...
+                            <h4 class="card__heading-4"><span class="link link__text">{% trans "Mijn uitkeringen" %}</span></h4>
+                            {% render_list %}
                                 {% list_item text="Jaaropgaven" compact=True strong=False active=False %}
                                 {% list_item text="Maandspecificaties" compact=True strong=False %}
+                            {% endrender_list %}
+                            <span class="link link--icon link--secondary"
+                                  aria-label="{% trans "Bekijk uitkeringen" %}"
+                                  title="{% trans "Bekijk uitkeringen" %}">
+                                <span class="link__text">{% trans "Bekijk uitkeringen" %}</span>
+                                {% icon icon="arrow_forward" icon_position="after" primary=True outlined=True %}
```
pi-sigma, line 264: See previous comment about `aria-label`.

pi-sigma (changes requested):
Looks good. Minor questions about the use of `aria-label`.

Also, I noticed that the link for the `Mijn begeleider` card in the profile page behaves differently from the link for `Mijn contacten` in case the user has no contacts. The latter leads to  `/mijn-profiel/contacts/` (even if there aren't any contacts), while the latter leads to `mijn-profiel/#` (instead of `/mijn-profiel/contacts?type=begeleider` ). I think this should be made consistent.

I would prefer if both lead to the contacts page. It makes sense to direct to the contacts page even if the user doesn't have any existing contacts or mentors, because they can create new ones.

It's an easy fix, but could also be addressed in a separate PR.

src/open_inwoner/templates/pages/profile/me.html
jiromaykin: That is actually an issue that will come up with the other thirty accessibility tasks still to come, but I will correct it on the profiel page here anyway.
Important to note: aria-label does not work on static elements such as div or span; it is meant for interactive elements that need a label, like buttons.
(The OIP project is quite cluttered with them...).

jiromaykin:
> ...noticed that the link for the `Mijn begeleider` card in the profile page behaves differently from the link for `Mijn contacten` in case the user has no contacts. ...
> I would prefer if both lead to the contacts page. It makes sense to direct to the contacts page even if the user doesn't have any existing contacts or mentors, because they can create new ones.

@pi-sigma  That issue is outside of the scope of accessibility so I put it in Taiga here: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2408

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -1257,25 +1257,23 @@ def test_render_list_if_appointments_are_found(self, m):
...
+        self.assertEqual(PQ(passport_appointment[3]).text(), "Dam 1")
+        self.assertEqual(PQ(passport_appointment[4]).text(), "1234 ZZ Amsterdam")
         self.assertEqual(
             PQ(cards[0]).find("a").attr("href"),
             f"{self.data.config.booking_base_url}{self.data.appointment_passport.publicId}",
         )
 
         id_card_appointment = PQ(cards[1]).find("ul").children()
 
-        self.assertEqual(id_card_appointment[0].text, "Aanvraag ID kaart")
```
stevenbal, line 1273: Same comment as above

stevenbal (changes requested):
Just two comments about removed assertions, other than that it looks good :+1:

src/open_inwoner/accounts/tests/test_profile_views.py
stevenbal:
```python
self.assertEqual(PQ(cards[1]).find(".card__heading-2").text(), "Aanvraag ID kaart")
```

pi-sigma (approved):

stevenbal (approved):

PR #1174, opened 2024-04-23
[#2240] Avoid open redirects through contactform

pi-sigma:
Taiga: [#2240](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2240)

As far as I can see, this was the only place that didn't either check for open redirects or redirect to some hardcoded url, an object's canonical url etc. (it was also the only source reported by GitHub).


Bartvaderkin (approved):

alextreme:
This is missing a regressiontest, but okay

PR #1176, opened 2024-04-24
[#2236] Get choices from Open Forms on form creation

pi-sigma:
The open forms client loads choices for the form field of `Product` on application startup. This moves the retrieval to the constructor of the form that needs the choices.

Depends on https://github.com/open-formulieren/open-forms-client-django/pull/11, which needs to be merged first.

Taiga: [#2236](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2236)


src/open_inwoner/pdc/admin/product.py
```diff
@@ -62,6 +62,10 @@ def __init__(self, *args, **kwargs):
                     "categories"
                 ].initial = self.instance.categories.intersection(user_categories)
 
+        self.fields["form"].choices = Product.form.field.get_choices(
+            include_blank=Product.form.field.blank
+        )
```
stevenbal, line 67: Is there no other way to do this? It would be nice if the library itself dealt with this

pi-sigma: Yeah, this is a bit odd. The library currently handles the field choices on the model level, hence the problem and this hackish workaround. I'll put the PR in draft for the time being and look into improving this in the library.

pi-sigma:
No longer needed, the library handles this by itself.

PR #1177, opened 2024-04-24
Use Ubuntu 20.04 in CI

pi-sigma:


pi-sigma:
Worth a try, but doesn't work

PR #1178, opened 2024-04-24
[#2062] Remove table tags + associated template

pi-sigma:
Taiga: [#2062](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2062)


Bartvaderkin (approved):
Wasn't even used anymore :smile: 

Makes you wonder what other old stuff is floating around in those `{% load %}` tags.

pi-sigma:
> Wasn't even used anymore 😄
> 
> Makes you wonder what other old stuff is floating around in those `{% load %}` tags.

Yes, I thought we were running out of easy cleanups, so this came as a surprise.

PR #1180, opened 2024-04-25
:lipstick: [#2064] Fix userfeed styling

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2064 

based on design: https://www.figma.com/file/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?type=design&node-id=1460%3A5026&mode=dev

Fixing heading(=description) styling + improved the pseudo-element spacing a bit ;
not yet adding the arrow link, will be new issue if actually needed in future.

Note: I have also changed the order and HTML semantics, as this relates to this issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2352

### DIFF
--- a/src/open_inwoner/cms/plugins/tests/test_userfeed.py
+++ b/src/open_inwoner/cms/plugins/tests/test_userfeed.py
@@ -44,10 +44,10 @@ class TestUserFeedPlugin(TestCase):
         items = pyquery.find(".card-container .card")
         self.assertEqual(len(items), 1)
 
-        title = items.find(".p.tabled__value").text()
+        title = items.find(".userfeed__title").text()
         self.assertEqual(title, "Test message")
 
-        message = items.find(".userfeed-card__heading").text()
+        message = items.find(".userfeed-card__description").text()
         self.assertEqual(message, "Hello")
 
         action_url = items[0].attrib["href"]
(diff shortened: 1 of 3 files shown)


pi-sigma (commented):
Looks good to me. I noticed two departures from the design: 

(1) in the design the the text is bold (ours is regular)
(2) in the design the arrow is always centered in the box (on our end it's aligned with the bottom line of the text)

Double-check that this is as intended, then we're good to go.

jiromaykin:
@pi-sigma Let's look at this together again, because 
1) I don't see a bold/normal difference with the design?
2) in Figma there is actually no design for a situation where the text exceeds 1 line, and I would say it is important to keep the arrow on the same height or else it would be floating at a different heights. But if we keep the arrow floating at the bottom it looks more peaceful to the eye.:

userfeed-arrow-bottom

jiromaykin:
@pi-sigma I checked again to see if it looks weird, but it doesn't. For the larger cards, like Vragen we also have the arrows 'sticking' to the bottoms of the cards, not the center.
And: The only situation where the 'bold' text would be floating 'above' the arrow, is when it is part of a grid where the card next to it is higher.
It's better to let the card border grow in order to give some peace to the eye in those cases, so it could look like this in a grid where cards that are next to each other will be given the same height:

userfeed-heights

pi-sigma (approved):

PR #1181, opened 2024-04-25
:wheelchair: [#2373] Make aria-labels same as (hidden) texts

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2373

 
A small change with big effects.
aria-labels only work on interactive elements and need to at least 'contain' keywords that are inside the visible or invisible text of the component.
 

Side note: title attributes are not meant for accessibility, and should **actually** contain different, more meaningful text than the readable text of an interactive element. But this has not been addressed as an issue anywhere in the report - so we're only changing the aria-label here, as per request.

TL;DR: I changed the order of `{% firstof title text %}`

### DIFF
--- a/src/open_inwoner/components/templates/components/Button/Button.html
+++ b/src/open_inwoner/components/templates/components/Button/Button.html
@@ -3,9 +3,9 @@
 {% if href %}
     <a
         class="{{ classes }}"
-        href="{{href}}"
+        href="{{ href }}"
         title="{% firstof title text %}"
-        aria-label="{% firstof title text %}"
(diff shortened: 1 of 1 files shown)


Bartvaderkin (approved):
Cool. 

We'll have to see how accurate these title vs text arguments have been set throughout the project but that would be that other task.

Maybe we should also make this choices more of a thing because I'm not so up-2-date on those rules, and it is likely we don't pass enough info into some contexts.

pi-sigma (approved):

PR #1182, opened 2024-04-26
[#2404] Avoid redirect cycle for DigiD login with email required

pi-sigma:
A redirect cycle occurs for DigiD login when email verification is required

Taiga: [#2404](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2404)

### DIFF
--- a/src/open_inwoner/utils/middleware.py
+++ b/src/open_inwoner/utils/middleware.py
@@ -74,6 +74,8 @@ class BaseConditionalUserRedirectMiddleware(abc.ABC):
             for e in self.extra_pass_prefixes:
                 try:
                     prefixes.append(resolve_url(e))
+                    prefixes.append(reverse("profile:registration_necessary"))
+                    prefixes.append(reverse("profile:email_verification_user"))
                 except NoReverseMatch:
                     # support testing without cms app mounted
                     pass
--- a/src/open_inwoner/utils/middleware.py
+++ b/src/open_inwoner/utils/middleware.py
@@ -31,6 +31,27 @@ def get_always_pass_prefixes() -> tuple[str, ...]:
     )
 
 
+def get_app_pass_prefixes() -> tuple[str, ...]:
+    """
+    Urls for apps that may not be active
+    """
+    return (
+        "profile:registration_necessary",
+        "profile:email_verification_user",
+    )
+
+
+def resolve_urls(urls) -> tuple[str, ...]:
+    ret = list()
+    for url in urls:
+        try:
+            ret.append(resolve_url(url))
+        except NoReverseMatch:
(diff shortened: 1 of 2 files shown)


pi-sigma:
For some reason, the test `request.session.REDIRECT_SESSION_VARIABLE` always evaluates to `False`, whereas `request.session.REDIRECT_SESSION_VARIABLE.__dict__` evaluates to `True` after the variable has been set.

src/open_inwoner/utils/middleware.py
```diff
@@ -98,6 +98,9 @@ def __call__(self, request):
         if request.path.startswith(self.get_pass_prefixes(request)):
             return self.get_response(request)
 
+        if "REDIRECT_SESSION_VARIABLE" in request.session.__dict__:
```
stevenbal, line 101: The `request.session` can be indexed like a dict, so you shouldn't use `.__dict__` here
```suggestion
        if request.session.get("REDIRECT_SESSION_VARIABLE"):
```

Also I'm not too sure about the name of `REDIRECT_SESSION_VARIABLE`, in what situation is it expected to be true? And I never see this session variable being removed or set to false, so after becoming true, it will stay that way for that session

src/open_inwoner/utils/middleware.py
```diff
@@ -106,5 +109,7 @@ def __call__(self, request):
             url.args.update(request.GET)
             return redirect(str(url))
 
-        # fallthrough
+        # fallthrough (avoid redirect cycle by modifying session)
+        request.session.REDIRECT_SESSION_VARIABLE = True
```
stevenbal, line 113:
```suggestion
        request.session["REDIRECT_SESSION_VARIABLE"] = True
```

stevenbal (dismissed):

src/open_inwoner/utils/middleware.py
pi-sigma: I ended up with a different (and simpler) solution

Bartvaderkin (changes requested):
Please just add these to the list in `get_always_pass_prefixes()` above instead of hard coding them in the base class.

pi-sigma:
> Please just add these to the list in `get_always_pass_prefixes()` above instead of hard coding them in the base class.

Doesn't work because the cms profile app might not be active, which results in failing tests (no match for `reverse(profile:...)`).

pi-sigma:
The test fails because the redirect for KVK login is skipped after adding the urls to `get_pass_prefixes`. It seems to work fine when testing manually, so I'm not sure if the test is to blame or if sth is wrong with the middleware as it is.

Bartvaderkin (approved):

PR #1185, opened 2024-04-26
Attempt to use in-memory sqlite3 instead as DB to see if this speeds CI up

alextreme:


alextreme:
Nevermind, postgis...

PR #1191, opened 2024-04-30
:bug: [#2381] Make zakelijk tab hidden when eHerkenning is off

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2381 

Branched off from #1189


alextreme:
FYI, the issue for this PR has been marked as blocking the 1.16 release for Groningen

pi-sigma (changes requested):
I think it would be preferable to hide the `particulier` tab as well if `eHerkenning` is disabled. There is no point in having a single tab.

pi-sigma (approved):

PR #1192, opened 2024-05-01
[#2424] Automatically check that config docs have been created/updated

pi-sigma:
- [x] Create management command to check that config docs have been created/updated
 - [x]  Update CI with command

Depends on https://github.com/maykinmedia/open-inwoner/pull/1150, which needs to be merged first

Taiga: [#2242](https://taiga.maykinmedia.nl/project/open-inwoner/task/2424)


pi-sigma:
Implemented via https://github.com/maykinmedia/django-setup-configuration/pull/9, so this is no longer needed.

PR #1195, opened 2024-05-02
[#2338] Add newsletters to sidebar navigation

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2338

### DIFF
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -103,6 +103,7 @@ class MyProfileView(
         context["anchors"] = [
             ("#personal-info", _("Persoonlijke gegevens")),
             ("#notifications", _("Voorkeuren voor meldingen")),
+            ("#newsletters", _("Nieuwsbrieven")),
             ("#overview", _("Overzicht")),
             ("#profile-remove", _("Profiel verwijderen")),
         ]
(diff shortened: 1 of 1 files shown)


src/open_inwoner/accounts/views/profile.py
```diff
@@ -103,6 +103,7 @@ def get_context_data(self, **kwargs):
         context["anchors"] = [
             ("#personal-info", _("Persoonlijke gegevens")),
             ("#notifications", _("Voorkeuren voor meldingen")),
+            ("#newsletters", _("Nieuwsbrieven")),
```
stevenbal, line 106: This should only be added if Laposta is configured and if the user has a verified email. You can put the import at the top of the file and the conditional below the instantiation of `context["anchors"]`

```python
from open_inwoner.laposta.models import LapostaConfig

...

if LapostaConfig.get_solo().api_root and user.has_verified_email():
    context["anchors"].insert(2, ("#newsletters", _("Nieuwsbrieven")))
```

stevenbal (changes requested):

src/open_inwoner/accounts/views/profile.py
jiromaykin: Ah yes, makes sense.
At least now you know where these scrollanchors are as well :)

stevenbal (approved):

PR #1200, opened 2024-05-08
Upgrade to Debian bookworm in Dockerfile

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/dimpact-enschede-ssc-ict-1/issue/81


alextreme (approved):
can be merged after double-checking that PDF exports are working without the pango-error

pi-sigma:
Confirmed, works.

PR #1201, opened 2024-05-08
Add init volume to docker-compose

pi-sigma:
The PR:
- adds an init volume to docker-compose so we can initialize containers with SQL dumps
- changes the port for Elasticsearch to `9201` so the Docker container can run alongside the system installation of ES without blocking the default port

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .vscode/
 .sass-cache/
 *.db
+docker-init-db.sql/*
 *.pyc
 __pycache__
 *~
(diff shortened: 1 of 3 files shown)


alextreme (approved):

pi-sigma:
I'll add some documentation to make clear how the database can be initialized

pi-sigma:
@alextreme I'll add @Bartvaderkin as reviewer to double-check the documentation for correctness and clarity.

Bartvaderkin (approved):

PR #1202, opened 2024-05-08
[#2454] Add IPv6Address to JSONEncoderMixin

pi-sigma:
Support for encoding IPv6 addresses was missing from the `JSONEncoderMixin` used for creating subscriptions

Taiga: [#2454](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2454)


Bartvaderkin (approved):
Code seems fine but is is unclear if they support IPv6, let's assume they do.

alextreme:
We'll know soon enough...

PR #1204, opened 2024-05-11
[#2460] Refactor KVK branch view

pi-sigma:
- Save company_branches on form instance instead of retrieving it through get_context_data to avoid multiple IO calls
- Move custom validation of branch number out of the POST of the form view and into the form itself

Taiga: [#2460](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2460)


Bartvaderkin (approved):
It'd be cool to have more functionality into the form .save() but this is already a big improvement.

PR #1205, opened 2024-05-13
Disabled the 'Run compile_dependencies on MacOS' CI task

Bartvaderkin:
It is borked and I don't know how to fix it.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -38,24 +38,26 @@ jobs:
       - name: Run bin/compile_dependencies.sh
         run: ./bin/compile_dependencies.sh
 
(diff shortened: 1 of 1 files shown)


pi-sigma (changes requested):
If `pg_config` is missing (the cause of the failing check?), that could be because postgres is missing from the machine or the file is in the wrong path. Did you try installing postgres via brew before compiling deps?

```
    - name: Run bin/compile_dependencies.sh
      run: |
          brew install postgresql
          ./bin/compile_dependencies.sh
```
What's the error for that?

pi-sigma:
The dependency check for the celery ticket is fixed.

Bartvaderkin:
Fixed in another PR

PR #1208, opened 2024-05-14
[#2461] Add setup configuration steps for CMS

pi-sigma:
Taiga: [#2461](https://taiga.maykinmedia.nl/project/open-inwoner/task/2461)

Depends on https://github.com/maykinmedia/django-setup-configuration/pull/9, which needs to be merged first.


src/open_inwoner/configurations/bootstrap/cms.py
```diff
@@ -0,0 +1,171 @@
...
+    extension_settings = {
+        "CMS_SSD_REQUIRES_AUTH": "requires_auth",
+        "CMS_SSD_REQUIRES_AUTH_BSN_OR_KVK": "requires_auth_bsn_or_kvk",
+        "CMS_SSD_MENU_INDICATOR": "menu_indicator",
+        "CMS_SSD_MENU_ICON": "menu_icon",
+    }
+
+    def __init__(self):
+        self.app_name = "ssd"
+        self.app_hook = SSDApphook
```
Bartvaderkin, line 78: You can set these as class level attributes like the other values (eg: not per instance in the constructor).

Bartvaderkin (approved):

PR #1210, opened 2024-05-15
[#2327+#2355] Send verification e-mail immedately

alextreme:


alextreme:
@pi-sigma please adress the 2 failing tests in open_inwoner.mail.tests.test_verification.TestMailVerificationMiddlewareFlow.

The way how I did it may not be a good idea and require discussion with @Bartvaderkin. If that's the case wait until after the current sprint to tackle this

pi-sigma:
Tests are fixed. I added @Bartvaderkin as a reviewer to check if we want to roll with this.

Bartvaderkin (approved):

pi-sigma (approved):

PR #1215, opened 2024-05-21
[#2480] Add NLDS paragraph components

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2480

paragraph modifiers that Utrecht doesn't have are: `--compact`, `--small`, `--muted`, `--centered`
This means we will have to make "`--oip`" modifiers with our own designtokens, so that municipalities that do not wish to use our design-tokens can use either their own or add custom-styling themselves.

HTML example of the proper way of working: 
This is what it looked/looks like in the past/legacy code:
`<p class="p p--small">…..` 

But now this should become:
`<p class="utrecht-paragraph  utrecht-paragraph--oip  utrecht-paragraph--oip-small">.....`

We will not yet be implementing this for **_ALL_** `<p>` tags because we have them in our own components that do not exist in NLDS at all. (Like the paragraphs inside the '**Mijn profiel**' tabled sections for example).

Also: in future I would love to completely remove all "legacy styling" that's meant to keep things backwards compatible for now. Plus it seems we have unused code, like `location-card-list` which might be safe to delete but not sure. So 'clean-up' will need to become a separate issue.

This PR relies on changes in our [NLDS token package PR nr.9](https://github.com/maykinmedia/open-inwoner-design-tokens/pull/9) so to view and check to see if all styles look the same, do an NPM ci --legacy-peer-deps install and rebuild.

Note to self: the two unique classes `utrecht-paragraph--oip-title-text` and `utrecht-paragraph--oip-summary` are not in design tokens becasue they should probably be defined by their parent component (like "readmore" component.

### DIFF
--- a/src/open_inwoner/components/templates/components/Messages/Messages.html
+++ b/src/open_inwoner/components/templates/components/Messages/Messages.html
@@ -2,8 +2,8 @@
 
 <section class="messages">
     <header class="messages__header">
-        <h4 class="utrecht-heading-4">{{ subject }}</h4>
-        <p class="p p--muted">{{ status }}</p>
+        <h3 class="utrecht-heading-4">{{ subject }}</h3>
+        <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-muted">{{ status }}</p>
     </header>
 
 
@@ -17,7 +17,7 @@
             {% for day in days %}
                 <li class="messages__day">
                     <header class="messages__day-header">
-                        <h4 class="p p--muted">{{ day.text }}</h4>
+                        <p class="utrecht-paragraph utrecht-paragraph--oip-paragraph-muted">{{ day.text }}</p>
                     </header>
 
                     <ol class="messages__list">
@@ -68,7 +68,7 @@
                 {% form_actions primary_text=_("Verzenden") primary_icon='arrow_forward' %}
             {% endrender_form %}
         {% else %}
-            <p class="p p--muted">{% trans "De gebruiker is inactief. Het is niet mogelijk een bericht te sturen." %}</p>
+            <p class="utrecht-paragraph utrecht-paragraph--oip-paragraph-muted">{% trans "De gebruiker is inactief. Het is niet mogelijk een bericht te sturen." %}</p>
         {% endif %}
     </div>
 </section>
--- a/src/open_inwoner/components/templates/components/Messages/Messages.html
+++ b/src/open_inwoner/components/templates/components/Messages/Messages.html
@@ -2,7 +2,7 @@
 
 <section class="messages">
     <header class="messages__header">
-        <h3 class="utrecht-heading-4">{{ subject }}</h3>
+        <h2 class="utrecht-heading-4">{{ subject }}</h2>
         <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-muted">{{ status }}</p>
     </header>
 
@@ -17,7 +17,7 @@
             {% for day in days %}
                 <li class="messages__day">
                     <header class="messages__day-header">
-                        <p class="utrecht-paragraph utrecht-paragraph--oip-paragraph-muted">{{ day.text }}</p>
+                        <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-paragraph-muted">{{ day.text }}</p>
                     </header>
 
                     <ol class="messages__list">
@@ -68,7 +68,7 @@
                 {% form_actions primary_text=_("Verzenden") primary_icon='arrow_forward' %}
             {% endrender_form %}
         {% else %}
-            <p class="utrecht-paragraph utrecht-paragraph--oip-paragraph-muted">{% trans "De gebruiker is inactief. Het is niet mogelijk een bericht te sturen." %}</p>
+            <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-paragraph-muted">{% trans "De gebruiker is inactief. Het is niet mogelijk een bericht te sturen." %}</p>
         {% endif %}
     </div>
 </section>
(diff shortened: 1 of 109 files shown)


src/open_inwoner/pdc/tests/test_views.py
```diff
@@ -242,11 +242,11 @@ def test_category_detail_description_rendered(self):
         response = self.client.get(url)
 
         self.assertIn(
-            '<p class="p">A <em>descriptive</em> description</p>',
+            "A <em>descriptive</em> description",
```
jiromaykin, line 245: @Bartvaderkin This test `test_category_detail_description_rendered`  we are not sure about. I can't replace the `p` class with the NLDesignSystem class here (since then this test doesn't not pass). But if I remove the class entirely it does pass. @pi-sigma wasn't sure about this assertion.

Bartvaderkin: @jiromaykin I'm not quite sure. I think I would just remove the opening `<p .. >` (and leave the closing one).

````py
self.assertIn("A <em>descriptive</em> description</p>", response.rendered_content)
````

And the second assertion (the `self.assertNotIn(` seems a bit odd and probably could be removed (not quite sure why it is there, maybe @pi-sigma remembers).

**Comment by jiromaykin** [comment]
@Bartvaderkin and @pi-sigma : So here is another fun big changes NLDS thing. This time I'm using the proper way to style 'modifiers' like `--muted` etc.. And I'm really hoping we can completely remove the 'legacy' styling - that's why I separated those to the bottom of the SCSS files.

File: src/open_inwoner/pdc/tests/test_views.py (hunk 1 above)
Comment (jiromaykin) [reply src/open_inwoner/pdc/tests/test_views.py]: I've adjusted/removed them, so this is ready for review.

File: src/open_inwoner/components/templates/components/Card/RenderCard.html [hunk 2]
```diff
@@ -13,13 +13,13 @@
     <div class="card__body{% if compact %}--compact{% endif %} {% if direction %} card__body--direction-{{ direction }}{% endif %}{% if grid %} card__body--grid{% endif %}">
         {% if title %}
             {% if compact %}
-            <h3 class="utrecht-heading-3 card__heading-3">
+            <p class="utrecht-heading-3 card__heading-3">
```
pi-sigma, line 16: Why are we switching to `p` for the card title? It would seem that some `h` element (like the `h3` that we currently) is appropriate (especially since we keep the corresponding style)?

src/open_inwoner/components/templates/components/Messages/Messages.html
```diff
@@ -2,8 +2,8 @@
 
 <section class="messages">
     <header class="messages__header">
-        <h4 class="utrecht-heading-4">{{ subject }}</h4>
-        <p class="p p--muted">{{ status }}</p>
+        <h3 class="utrecht-heading-4">{{ subject }}</h3>
```
pi-sigma, line 5: Do we need to switch to `h3` if we keep styling corresponding to `h4` elements?

src/open_inwoner/scss/components/Typography/P.scss
```diff
@@ -1,63 +1,64 @@
...
-  }
+  &--oip {
+    // Takes style values from our own design-tokens NPM package
+    margin: 0;
 
-  &--muted {
-    color: var(--color-mute);
-  }
+    &-small {
+      font-size: var(--oip-utrecht-paragraph-small-font-size);
```
pi-sigma, line 11: I would prefer variable names where the size indicator is at the end, e.g. `oip-utrecht-paragraph-font-size-small`, `oip-utrecht-paragraph-font-size-medium` etc., as this would be easier to discern.

pi-sigma (changes requested):

src/open_inwoner/components/templates/components/Messages/Messages.html
jiromaykin: Actually we do. And it should even be an H2 tag instead, but styled like an h4 tag. So I will change that. 
This messages template really needs a rework. Too many things that do not function like a heading were given a heading tag.
Remember: the HTML hierarchy of the **tags** need to be correct. For accessibility you will need to often make headings style like as if they are headings of different higher/lower levels, as long as their HTML hierarchy remains correct, in order to allow screenreaders to read-out a correct main headings-menu in the correct order. Screenreaders respond to semantic HTML, not to CSS.

src/open_inwoner/scss/components/Typography/P.scss
jiromaykin: This is not about 'preference', this is how NLDS SCSS variables are built. In this case I am making a variant of the already existing "**paragraph-small**" so that is actually its entire name. 
Indeed note that the SCSS variables/values do not quite follow the same order as HTML classes.
So they look like they a different hierarchy than the usual BEM HTML classes, but they actually don't. It looks confusing because the HTML classnames look a lot like their CSS variables, but not quite...

Here in these screenshots you can see examples of Open-Formulieren and Amsterdam and how it will make themeswitching easier in the future if we keep the NLDS order intact. I know this looks confusing... And Open-Forms actually creates local variables for their own custom components, which is something we could also do, but I decided to not do that and just create these as coming from our own design-tokens.

NLDS-amsterdam-order

NLDS-OF-order

pi-sigma: Whoever decided to build the variables this way made the wrong choice IMO, but oh well....

pi-sigma (changes requested):
Rebasing on `develop` will create a merge conflict due to `pdc/tests/test_views.py`. I would fix it myself, but rebasing also marked a bunch of other files as having changed, so we should wait until Monday. Will take this up with Jiro first thing in the morning.

pi-sigma (approved):

PR #1216, opened 2024-05-27
[#2465] Postprocess HTML for NLDS styling

pi-sigma:
Taiga: [#2465](https://taiga.maykinmedia.nl/project/open-inwoner/task/2465)

### DIFF
--- a/src/open_inwoner/pdc/tests/test_views.py
+++ b/src/open_inwoner/pdc/tests/test_views.py
@@ -242,11 +242,11 @@ class CategoryDetailViewTest(TestCase):
         response = self.client.get(url)
 
         self.assertIn(
-            '<p class="p">A <em>descriptive</em> description</p>',
+            "A <em>descriptive</em> description</p>",
             response.rendered_content,
         )
         self.assertNotIn(
-            '[<p class="p">A <em>descriptive</em> description</p>, <em>descriptive</em>]',
+            "[A <em>descriptive</em> description</p>, <em>descriptive</em>]",
             response.rendered_content,
         )
 
--- a/src/open_inwoner/utils/ckeditor.py
+++ b/src/open_inwoner/utils/ckeditor.py
@@ -5,15 +5,15 @@ import markdown
 from bs4 import BeautifulSoup
 
 CLASS_ADDERS = [
-    ("h1", "h1"),
-    ("h2", "h2"),
-    ("h3", "h3"),
-    ("h4", "h4"),
-    ("h5", "h5"),
-    ("h6", "h6"),
+    ("h1", "utrecht-heading-1"),
+    ("h2", "utrecht-heading-2"),
+    ("h3", "utrecht-heading-3"),
+    ("h4", "utrecht-heading-4"),
+    ("h5", "utrecht-heading-5"),
+    ("h6", "utrecht-heading-6"),
     ("img", "image"),
     ("li", "li"),
-    ("p", "p"),
+    ("p", "utrecht-paragraph"),
     ("a", "link link--secondary"),
     ("table", "table table--content"),
     ("th", "table__header"),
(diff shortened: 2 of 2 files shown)


pi-sigma:
@Bartvaderkin The security warning is technically valid, since the content is user-provided. However, this user is a staff user who creates content via the Ckeditor in the admin. So can we treat the content as trusted?

src/open_inwoner/utils/ckeditor.py
```diff
@@ -105,3 +118,18 @@ def get_product_rendered_content(product):
                 element.append(icon)
 
     return soup
+
+
+def convert_styling_to_nlds(response: requests.Response) -> None:
+    """
+    Post-process rendered HTML content, replace legacy styling with NLDS classes
+    """
```
Bartvaderkin, line 126: Superfluous comment feels dogmatic: it just duplicates information already present in the function signature.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -105,3 +118,18 @@ def get_product_rendered_content(product):
...
+    raw_html = response.content
+    html_tree = html.fromstring(raw_html)
+
+    for tag, style_class in NLDS_MAP.items():
+        elems = html_tree.cssselect(tag)
```
Bartvaderkin, line 131: Let's use `.iterfind(tag)` to find node names: it is an iterator and doesn't involve a CSS engine to find a tag name.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -105,3 +118,18 @@ def get_product_rendered_content(product):
...
+        for item in elems:
+            item.set("class", style_class)
```
Bartvaderkin, line 133: This clobbers the current class(es). Instead we want to add our class to whatever else might be on there. 

(this is especially a problem when it unintentionally processes the whole page)

src/open_inwoner/pdc/views.py
```diff
@@ -225,6 +226,16 @@ def get_context_data(self, **kwargs):
         context["display_social"] = config.display_social
         return context
 
+    def render_to_response(self, context, **response_kwargs):
+        """
+        Override to post-process rendered HTML content
+        """
+        response = super().render_to_response(context, **response_kwargs)
+
+        response.add_post_render_callback(convert_styling_to_nlds)
```
Bartvaderkin, line 235: This modifies the whole document and not just the content from the field, so I don't think this is what we want.

Bartvaderkin (changes requested):

Bartvaderkin:
> @Bartvaderkin The security warning is technically valid, since the content is user-provided. However, this user is a staff user who creates content via the Ckeditor in the admin. So can we treat the content as trusted?

@pi-sigma This code parse the whole `response.content`, which is a bit more then just the field's content (eg: before it gets added to the page_.

src/open_inwoner/pdc/tests/test_views.py
```diff
@@ -242,11 +242,11 @@ def test_category_detail_description_rendered(self):
         response = self.client.get(url)
 
         self.assertIn(
-            '<p class="p">A <em>descriptive</em> description</p>',
+            "A <em>descriptive</em> description</p>",
             response.rendered_content,
         )
         self.assertNotIn(
-            '[<p class="p">A <em>descriptive</em> description</p>, <em>descriptive</em>]',
+            "[A <em>descriptive</em> description</p>, <em>descriptive</em>]",
```
jiromaykin, line 249: We are going to have merge-conflict here since I've also changed these lines, but differently, in #1215 :-)

pi-sigma: We had a bug a while ago where CKeditor duplicated some of the content, so there's a reason whey the second assertion is in there

src/open_inwoner/utils/ckeditor.py
```diff
@@ -78,7 +78,10 @@ def get_product_rendered_content(product):
                 element.string = ""
                 element.attrs.update(
                     {
-                        "class": "button button--textless button--icon button--icon-before button--primary cta-button",
+                        "class": (
+                            "utrecht-button button--textless button--icon "
```
jiromaykin, line 82: Currently we do not have NLDS button components yet so you can remove this particular button stuff here and revert it to what it was, or else the styling won't work.
I will be changing this one dramatically when I continue work on the Buttons (the NLDS buttons are no longer part of a sprint currently).

jiromaykin (changes requested):

pi-sigma:
@Bartvaderkin Got rid of most of the code for this change

Bartvaderkin (approved):

PR #1219, opened 2024-05-27
[#2509] Create celery task to rebuild search index

pi-sigma:
Taiga: [#2509](https://taiga.maykinmedia.nl/project/open-inwoner/task/2509)


alextreme:
schedule to be added

PR #1220, opened 2024-05-27
[#2507] Fixed and added test for 'django-admin-index' fixture

Bartvaderkin:
I think the mechanism wasn't broken, it just errored on bad data (there was a content-type in there that was removed).

Note it listens to the post-migrate signal (eg: not application startup). If we runt this on application start it would run for every test and management command as well (and it deletes and re-creates all the records)

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/2482

### DIFF
--- a/src/open_inwoner/accounts/apps.py
+++ b/src/open_inwoner/accounts/apps.py
@@ -7,10 +7,10 @@ from django.core.management import call_command
 from django.db.models.signals import post_migrate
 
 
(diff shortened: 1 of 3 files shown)


Bartvaderkin:
@pi-sigma In hindsight I don't think the issue was the signal, just the bad data, and running this on application startup is not actually desirable.

pi-sigma (approved):
Seems reasonable

PR #1221, opened 2024-05-27
[#2477] Add site name to dynamic mail context

pi-sigma:
Taiga: [#2477](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2477)


Bartvaderkin (changes requested):
I would add it to the tests here as well: https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/mail/tests/test_context.py

Bartvaderkin (approved):

PR #1222, opened 2024-05-28
[#2510] Updated version number

Bartvaderkin:
I looked into pulling this from git but some installations like docker don't have a checkout so Alex agreed to keep using this method.

### DIFF
--- a/src/open_inwoner/__init__.py
+++ b/src/open_inwoner/__init__.py
@@ -1,6 +1,6 @@
 from .celery import app as celery_app  # noqa
 
-__version__ = "1.16"
+__version__ = "1.16.5"
 __author__ = "Maykin Media"
 __homepage__ = "https://github.com/maykinmedia/open-inwoner"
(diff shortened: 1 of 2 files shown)


src/open_inwoner/__init__.py
```diff
@@ -1,6 +1,6 @@
 from .celery import app as celery_app  # noqa
 
-__version__ = "1.16"
+__version__ = "1.16.5"
```
pi-sigma, line 3: You could get the value from `RELEASE` in `open_inwoner.conf.base` (the other way around is not possible because of circular import).

pi-sigma (approved):
One of the hard-coded strings could be eliminated, but it's not a big deal

src/open_inwoner/__init__.py
Bartvaderkin: I found the circular import but didn't try the other way and that worked so I updated the PR

PR #1223, opened 2024-05-28
[#2503] Add dyslexia modus with NL-DesignSystem components

jiromaykin:
Also added font-enlarger
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2503

I am aware this looks very unattractive visually 😕 - specifically when both enlarging AND toggling the Dyslexia font...

### DIFF
--- a/src/open_inwoner/js/components/accessibility/enlarge_font.js
+++ b/src/open_inwoner/js/components/accessibility/enlarge_font.js
@@ -1,29 +1,102 @@
-class EnlageFont {
+export class EnlargeFont {
+  static selector = '.accessibility--enlarge-font'
   constructor(node) {
     this.node = node
     this.text = node.querySelector('.link__text')
     this.icon = node.querySelector('.material-icons')
     this.node.addEventListener('click', this.enlarge.bind(this))
+
+    // Set initial values for default styles
+    let root = document.documentElement
+    const bodyFontSizeDefault = '--utrecht-document-font-size'
+    const paragraphFontSizeDefault = '--utrecht-paragraph-font-size'
+    const paragraphFontSizeSmall = '--utrecht-paragraph-small-font-size'
+    const headingThreeFontSize = '--utrecht-heading-3-font-size'
+    const headingFourFontSize = '--utrecht-heading-4-font-size'
+    const oipBodyFontSizeDefault = '--font-size-body'
+    const oipBodyFontSizeSmall = '--font-size-body-small'
+    const oipHeadingThreeFontSize = '--font-size-heading-3'
+    const oipHeadingFourFontSize = '--font-size-heading-4'
+    const oipCardHeadingFontSize = '--font-size-heading-card'
+    const baseSizeDefault = '16px'
+    const baseSizeSmall = '14px'
+    const baseHeadingThreeFontSize = '20px'
+    const baseHeadingFourFontSize = '16px'
+    const baseCardHeadingFontSize = '18px'
+
+    root.style.setProperty(bodyFontSizeDefault, baseSizeDefault)
+    root.style.setProperty(paragraphFontSizeDefault, baseSizeDefault)
+    root.style.setProperty(paragraphFontSizeSmall, baseSizeSmall)
+    root.style.setProperty(headingThreeFontSize, baseHeadingThreeFontSize)
+    root.style.setProperty(headingFourFontSize, baseHeadingFourFontSize)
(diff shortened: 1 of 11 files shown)


src/open_inwoner/js/components/accessibility/enlarge_font.js
```diff
@@ -1,29 +1,95 @@
...
+      headingThree: '20px',
+      headingFour: '16px',
+      cardHeading: '18px',
+    }
+
+    // Enlarge the different types of body-font as well as paragraphs
+    // and set all lower types of headings (lower than H2) to get the same larger font-size
+    this.enlargeSizes = {
+      default: '20px',
+      small: '17px',
```
pi-sigma, line 41: The ratio between default and small should be constant (unless there's a reason to depart; it may well be desirable to have a slightliy different ration in the case of larger fonts): in `this.baseSizes` it is `16 - 14 = 2`, here we have `20 - 17 = 3`.

src/open_inwoner/scss/views/App.scss
```diff
@@ -195,49 +195,49 @@
...
+  // Font-family = 'Body' coming from both NLDS and/or the font-upload functionality
+  --font-family-body: var(--utrecht-document-font-family);
   --font-color-body: var(--color-gray-dark);
-  --font-size-body: 16px;
+  --font-size-body: var(--utrecht-document-font-size);
   --font-line-height-body: 24px;
 
   --font-size-body-extra-small: 12px;
-  --font-size-body-small: 14px;
+  --font-size-body-small: var(--utrecht-paragraph-small-font-size);
```
pi-sigma, line 235: `--font-size-body` is defined in terms of `--utrecht-document` but `font-size-body-small` is defined in terms of `--utrecht-paragraph`. Is there no `--utrecht-document-small-font-size`?

src/open_inwoner/scss/views/App.scss
```diff
@@ -195,49 +195,49 @@
...
   --font-size-body-large: 20px;
```
pi-sigma, line 236: Why is this hard-coded (it's the only case of font size not defined in terms of utrecht classes)?

pi-sigma (changes requested):

src/open_inwoner/scss/views/App.scss
pi-sigma: Good for now. If we start using this size elsewhere, we should introduce the modifier as you indicated

src/open_inwoner/js/components/accessibility/enlarge_font.js
jiromaykin: If you'd set Small to 18px when enlarged, I personally think it looks a bit ugly, and I do not really see a clear difference with the default font-size anymore. Where I feel like you can really visually see a difference between small and default in the original sizes. 
Visually it doesn't seem to be an 'absolute number' thing but more of an 'optical' thing, but I can change this if it would really be an improvement, even though the font-size actually falls outside of the scope of the issue - but currently the toggling in the accessibility-header doesn't to work at all so I wanted to add an extra improvement.

src/open_inwoner/scss/views/App.scss
jiromaykin: Do you mean the '--font-size-body-large'? I did not change anything in this PR there. 
Also: no this value does not exist in NLDS for the Document level.
The NLDS Document only has a couple of attributes and values; all other modifiers I guess should belong to actual components. 

If you look at Utrecht's Storybook here https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-document--docs you can inspect their Document/paragraph component, which will point to a long file with all of their CSS variables. It does not contain many `-document-...` variables.

Utrecht does have a variable '--utrecht-typography-scale-lg-font-size' but that is not part of the NLDS standard, and is a 'scaling' variable font-size so not the same as our set pixel size.

Additionally: this '--font-size-body-large' variable is only used in OIP for: just a few icons, 2 asterisks and the mobile menu... But currently the accessibility header is not visible on mobile so that is outside of the scope here. 
(That is a future task that needs a new design).
➕ Forgive me for not adding an enlarger to a font-size that is already 'large' 😉 which is why I also omitted H1 and H2 since those are already quite large.

src/open_inwoner/js/components/accessibility/enlarge_font.js
pi-sigma: No, this is enough explanation. Thought it might be an accident.

pi-sigma (approved):

PR #1224, opened 2024-05-28
:wheelchair: [#2345] A11y replace disallowed heading tags in Cards

jiromaykin:
issue: if there is no content, then there should not be any headings.
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2345 

Cards that only have an image and no text content at all, should use a paragraph that may visually look like a Heading.
A real Heading is an HTML tag that should summarize/pin-point the section below it, but in case of an image it does not make sense in accessibility/meaningfulness.

Also screen-readers will read out headings as a list with hierarchy, so there better be something meaningful there.

Note : it is **outside of the scope** of this issue but some of the other cards are incorrect as well - in case a card has only a title, it's again not really a heading. Example: the sub-category card here: https://open-inwoner-test.maykin.nl/onderwerpen/vervoer/ 
And on the Profile page it is better to just consistently use paragraphs, since many of the 'Overzicht' cards do not have 'meaningful' content when read by a screenreader as a hierarchical list.

Perhaps none of the cards need a true Heading tag, but at least our 'Aanvragen/Vragen/Samenwerkingen' cards have a lot more content that may benefit from having a heading.

Also note: NL Design System has its own Card component, which we are not using yet in OIP.

### DIFF
--- a/src/open_inwoner/components/templates/components/Card/CategoryCard.html
+++ b/src/open_inwoner/components/templates/components/Card/CategoryCard.html
@@ -22,12 +22,13 @@
 </div>
 
 {% else %}
+{# If this card has no content section #}
     <a href="{{ category.slug }}" class="card">
         <div class="card__body">
             {% if category %}
-                <h3 class="utrecht-heading-3 card__heading-3">
+                <p class="utrecht-heading-3 card__heading-3">
                     <span class="link link__text">{{ category }}</span>
-                </h3>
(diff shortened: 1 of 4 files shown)


jiromaykin:
@alextreme You are also allowed to review this one, since it's quite simple.

alextreme (approved):

PR #1227, opened 2024-05-29
[#2448] Link to manual from docs

pi-sigma:
Taiga: [#2448](https://taiga.maykinmedia.nl/project/open-inwoner/task/2448)


alextreme (changes requested):
Wrong PDF, it's the manual on https://taiga.maykinmedia.nl/project/open-inwoner/wiki/beheerhandleiding-openinwonerplatform

Note that it is an older version, we should get an update in a week or two. 

Lets point to the version on Github from the Taiga wiki once this is moved over

pi-sigma:
@alextreme The task description only specifies a link to the manual. Wouldn't it make sense to provide a link to the Taiga page containing all the docs instead, since the client will also be interested in the document explaining the primary/secondary colors?

alextreme:
@pi-sigma well, the context of the issue is that originally we wanted to have all content of the admin-manual on readthedocs, publically accessible. This is quite a step content-wise as the content needs to be converted to rst/markdown, so publishing the PDF and linking to it was the first quick fix

pi-sigma:
@alextreme Got it. I thought since the manual says "vertrouwelijk" on the title page it should remain behind the Taiga login. I'll put the PDF in the docs folder and update the link.

alextreme:
Noted, this is indeed against our security policy. You have approval to publish it, I'll ensure this is noted as a public document for the next iteration

alextreme (approved):

PR #1228, opened 2024-05-29
Swapping around migrations due to InconsistentMigrationHistory on acc…

alextreme:
Problem that occurred on acceptance environments with v1.17.0 which I switched around on v1.17.1. This was most likely due to the fact that we delayed removing the statustranslation table to 1.17

### DIFF
--- a/src/open_inwoner/openzaak/migrations/0047_delete_statustranslation.py
+++ b/src/open_inwoner/openzaak/migrations/0047_delete_statustranslation.py
@@ -6,7 +6,7 @@ from django.db import migrations
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("openzaak", "0046_alter_openzaakconfig_fetch_eherkenning_zaken_with_rsin"),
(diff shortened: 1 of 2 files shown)


alextreme:
This would be sufficient if all we have are SaaS environments (production has the same issue where 0047 is missing).

However we also have Dimpact-installed environments that need a clean migration path between releases, these are done with our Docker image but we don't have much of a say how and when a new version of OIP is installed. And we also have separate installs managed by Sjoerd using our Docker image separate from our SaaS installs. The main thing we could do is mention the steps you say in the release notes, but it wouldn't be easy for them.

So this PR is to avoid future upgrade problems and also a reminder that leaving a PR + migration out of a release will cause headaches during future releases. Of course, doing a squash would avoid confusion for new installs and allow us to clean up this history in due time.

image

pi-sigma (approved):
@alextreme Thanks for the explanation, that makes sense! Let's discuss the option of squashing migrations on Monday.

PR #1090, opened 2024-03-14
[#2199] Updated settings for django-yubin

Bartvaderkin:

PR #1231, opened 2024-05-30
Replace django-mailer with django-yubin

pi-sigma:
Reverting the previous change in [https://github.com/maykinmedia/open-inwoner/pull/1090](https://github.com/maykinmedia/open-inwoner/pull/1090)

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -46,11 +46,11 @@ mozilla-django-oidc-db
 django-open-forms-client
 django-htmx
 playwright
-django-mailer
 django-image-cropping
 django-log-outgoing-requests
 xsdata
 django-setup-configuration
+django-yubin
 
 # django cms
 django-cms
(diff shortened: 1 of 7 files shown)


alextreme (changes requested):
This seems okay but isn't fully complete. Note that only the send_mail task seems to be added automatically (this will need to be verified after merging and deploying to the testenvironment).

The management commands are gone with Yubin 2.x so afaik we need to add all yubin tasks we need to CELERY_BEAT_SCHEDULE (we don't want to manually add celery beat tasks individually per install)

https://django-yubin.readthedocs.io/en/latest/queue.html#tasks

For retry_emails (1x per hour) and delete_old_emails (1x per day) we also have (less frequent) cronjobs

> You don’t usually need to create a send_email task, Yubin email backend does it automatically. For retry_emails and delete_old_emails, you can use [Celery Beat](https://django-celery-beat.readthedocs.io/en/latest/) to schedule periodic task.

alextreme (approved):

alextreme:
Can be merged in develop, however PR is still in draft

PR #1235, opened 2024-06-03
Add support for resolving instance attributes in the cache decorator

swrichards:
Taiga issue [#2524](https://taiga.maykinmedia.nl/project/open-inwoner/task/2524).

This will be followed by a PR to actually use this in the ZGW clients (and will probably also remove a few tests that were really just testing the underlying caching logic, which I've included here in a more isolated form by testing the cache behavior directly).

### DIFF
--- a/src/open_inwoner/utils/decorators.py
+++ b/src/open_inwoner/utils/decorators.py
@@ -1,5 +1,7 @@
 import inspect
 import logging
+import re
+import uuid
 from collections.abc import Callable
 from functools import wraps
 from typing import TypeVar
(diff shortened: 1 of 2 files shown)


src/open_inwoner/utils/decorators.py
```diff
@@ -1,26 +1,61 @@
...
+from django.core.cache import BaseCache, caches
 
 logger = logging.getLogger(__name__)
 
 
 RT = TypeVar("RT")
 
 
-def cache(key: str, alias: str = "default", **set_options):
+def _map_cache_key_instance_attrs_to_placeholders(key):
```
pi-sigma, line 17: Could use a brief docstring explaining the need for this mapping due to supporting multiple backends with different endpoints.

src/open_inwoner/utils/decorators.py
```diff
@@ -50,19 +81,39 @@ def wrapped(*args, **kwargs) -> RT:
                 }
                 key_kwargs[argspec.varkw] = var_kwargs
 
-            cache_key = key.format(**key_kwargs)
-
-            _cache = caches[alias]
+            (
+                cache_key_with_attr_placeholders,
+                attr_mapping,
+            ) = _map_cache_key_instance_attrs_to_placeholders(key)
```
pi-sigma, line 87: Subjective, but I find this easier to parse:
```
cache_key_with_attr_placeholders, attr_mapping = (
     _map_cache_key_instance_attrs_to_placeholders(key)
)
```

src/open_inwoner/utils/tests/test_utils.py
```diff
@@ -0,0 +1,205 @@
...
+
+from open_inwoner.utils.decorators import cache
+
+MockCache = mock.create_autospec(DummyCache)
+
+
+@override_settings(
+    CACHES={"default": {"BACKEND": "open_inwoner.utils.tests.test_utils.MockCache"}}
+)
+class DynamicCacheKeyTests(DjangoTestCase):
```
pi-sigma, line 18: class names should be singular (`class DynamicCacheKeyTest`)

src/open_inwoner/utils/tests/test_utils.py
```diff
@@ -0,0 +1,205 @@
...
+                mock.call("alpha:charlie:bravo:42"),
+                mock.call("alpha:bar:bravo:baz"),
+                mock.call("alpha:bar:bravo:5"),
+                mock.call("alpha:bar:bravo:baz:charlie:charlie:42"),
+                mock.call("static"),
+            ]
+        )
+
+    def test_non_existent_attr_key_raises(self):
+        class TestInstance:
```
pi-sigma, line 67: I would call it `class TestClass` for consistency with the previous case

src/open_inwoner/utils/decorators.py
```diff
@@ -50,19 +81,39 @@ def wrapped(*args, **kwargs) -> RT:
...
+                    )
+
+                bound_instance = args[0]
+                for mapped_attr, original_attr in attr_mapping.items():
+                    try:
+                        key_kwargs[mapped_attr] = getattr(bound_instance, original_attr)
+                    except AttributeError:
+                        raise AttributeError(
+                            f"Attribute `{original_attr}` does not exist on bound instance"
+                        )
```
pi-sigma, line 101: You can also add a note to Python's `AttributeError` instead of raising a new one:
```
except AttributeError as exc:
    exc.add_note(f"Attribute `{original_attr}` does not exist on bound instance")
    raise
```

pi-sigma (changes requested):
Couple minor remarks. I see no tests for the error cases (`ValueError`, `AttributeError`). Would be nice to have those, but it's not critical IMO and I don't want to hold you up, so it's optional.

pi-sigma (approved):

PR #1236, opened 2024-06-03
[#2523] Prevent display of unpublished pages

pi-sigma:
Taiga: [#2523](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2523)

### DIFF
--- a/src/open_inwoner/pdc/tests/test_product.py
+++ b/src/open_inwoner/pdc/tests/test_product.py
@@ -152,6 +152,17 @@ class TestPublishedProducts(WebTest):
         )
         self.assertEqual(list(response.context["products"]), [product1])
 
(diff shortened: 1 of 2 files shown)


pi-sigma:
The search needs to be filtered by `product.published` (a `bool`). Couldn't get this to work in a timely fashion for the `FacetedSearch` class. Documentation: https://elasticsearch-dsl.readthedocs.io/en/latest/faceted_search.html

alextreme:
The suggested patch for the detail view seems sufficient for now, we can then investigate the search-results later (the search results send the user to the detail view anyway). PR should have contained a regressiontest in my eyes but it was still in-draft.

@swrichards for you to decide if we can include this in 1.17.2 or not

alextreme (approved):
Can be merged after tests pass

PR #1237, opened 2024-06-03
[#2416] Scope ZGW clients cache keys by the underlying service endpoint

swrichards:
Taiga [#2516](https://taiga.maykinmedia.nl/project/open-inwoner/task/2516).


alextreme (approved):
I was first a bit in doubt if I'd prefer 'key:url' or 'url:key' but as the proposed order goes from general-to-specific this feels like the way to go

PR #1240, opened 2024-06-05
[#2526] Allow for additional ZGW service backends to be configured

swrichards:
This is a first step towards allowing multiple ZGW backends throughout the platform. To maintain backwards compatibility with the current datamodel, we proxy the current service config fields to the new ZGWApiGroupConfig model as we work our way towards making all ZGW client invocations multi-backend aware.


alextreme:
Seems okay, also for @pi-sigma to review first after his holiday

alextreme:
@stevenbal review vanuit jouw kant zou wenselijk zijn hiervoor, dit hadden we volgens mij in april samen met Bart al voorbesproken maar lag even stil

src/open_inwoner/openzaak/admin.py
```diff
@@ -18,23 +18,20 @@
     ZaakTypeInformatieObjectTypeConfig,
     ZaakTypeResultaatTypeConfig,
     ZaakTypeStatusTypeConfig,
+    ZGWApiGroupConfig,
 )
 
 
+class ZGWApiGroupConfig(admin.TabularInline):
```
stevenbal, line 25: Let's use `StackedInline` here to avoid overflow to the right:

image

src/open_inwoner/openzaak/models.py
```diff
@@ -46,45 +47,160 @@ def generate_default_file_extensions():
...
-class OpenZaakConfig(SingletonModel):
-    """
-    Global configuration and defaults for zaken and catalogi services.
-    """
+class ZGWApiGroupConfig(models.Model):
+    """A set of of ZGW service configurations."""
+
+    open_zaak_config = models.ForeignKey(
+        "openzaak.OpenZaakConfig", on_delete=models.PROTECT, related_name="api_groups"
+    )
```
stevenbal, line 55: `OpenZaakConfig` is a `SingletonModel` and there will only be one instance of it, so I don't think we need a `ForeignKey` to it on this model and can just use `.get_solo` if we need it?

src/open_inwoner/openzaak/migrations/0050_migrate_zgw_root_fields_to_multi_backend.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+def migrate_zgw_service_config_to_default_group_config(apps, schema_editor):
+    ZGWApiGroupConfig = apps.get_model("openzaak", "ZGWApiGroupConfig")
+    OpenZaakConfig = apps.get_model("openzaak", "OpenZaakConfig")
+
+    for config in OpenZaakConfig.objects.all():
```
stevenbal, line 12: I don't think looping here is necessary, because `OpenZaakConfig` is a `SingletonModel`, but it's not a big issue

src/open_inwoner/openzaak/migrations/0050_migrate_zgw_root_fields_to_multi_backend.py
```diff
@@ -0,0 +1,53 @@
...
+            ztc_service=config.catalogi_service,
+            form_service=config.form_service,
+            open_zaak_config=config,
+        )
+
+
+def reverse_migrate_zgw_service_config_to_default_group_config(apps, schema_editor):
+    OpenZaakConfig = apps.get_model("openzaak", "OpenZaakConfig")
+
+    for config in OpenZaakConfig.objects.all():
```
stevenbal, line 26: Same comment with regards to the for-loop

src/open_inwoner/openzaak/tests/test_migrations.py
```diff
@@ -0,0 +1,68 @@
...
+            catalogi_service=Service.objects.get(id=self.catalogi_service.id),
+            document_service=Service.objects.get(id=self.documenten_service.id),
+            form_service=Service.objects.get(id=self.forms_service.id),
+        )
+
+    def test_migration_0048_to_0051_multi_zgw_backend(self):
+        ZGWApiGroupConfig = self.apps.get_model("openzaak", "ZGWApiGroupConfig")
+        OpenZaakConfig = self.apps.get_model("openzaak", "OpenZaakConfig")
+
+        (config,) = OpenZaakConfig.objects.all()  # should be 1
```
stevenbal, line 35:
```suggestion
        config = OpenZaakConfig.objects.get()
```

stevenbal (changes requested):
Some minor remarks, looks good overall :+1:

src/open_inwoner/openzaak/models.py
stevenbal: Hmm yeah that's true, I don't feel too strongly about either way, so it's fine

pi-sigma (approved):

PR #1246, opened 2024-06-10
:wheelchair: [#2376] Add aria-expanded for desktop navigation menus

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2376


pi-sigma (commented):
Do we only need this for the "Onderwerpen" button? The "Welkom" button currently doesn't have an `aria-expanded` attribute

jiromaykin:
> Do we only need this for the "Onderwerpen" button? The "Welkom" button currently doesn't have an `aria-expanded` attribute

? But it does. And yes, they both need it.

pi-sigma (approved):
Sorry, was looking in the wrong place for the "Welkom" button. Works fine (though only tested on FireFox and Chrome).

PR #1252, opened 2024-06-11
♿  [#2366] Add link purpose to document download links

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2366

### DIFF
--- a/src/open_inwoner/components/templates/components/File/File.html
+++ b/src/open_inwoner/components/templates/components/File/File.html
@@ -18,11 +18,11 @@
                 <span class="file__name">
                     {{ name }}
                     {% if extension and size %}
-                        ({{extension}}, {{size|readable_size}})
+                        ({{ extension }}, {{ size|readable_size }})
                     {% elif extension %}
-                        ({{extension}})
+                        ({{ extension }})
                     {% elif size %}
-                        ({{size|readable_size}})
+                        ({{ size|readable_size }})
                     {% endif %}
                 </span>
                 <span class="file__date">{{ created|date:'j F Y' }}</span>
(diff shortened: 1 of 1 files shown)


src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -49,8 +49,8 @@
                     </div>
                 {% enddropdown %}
             {% elif show_download %}
-                {% trans "Download" as download %}
-                {% link href=url text=download primary=True download=True hide_text=True extra_classes="file__download" icon="download" icon_outlined=True icon_position="before" %}
+                {% trans "Download " as download %}
```
pi-sigma, line 52:
```suggestion
                {% trans "Download" as download %}
```

src/open_inwoner/components/templates/components/File/File.html
```diff
@@ -49,8 +49,8 @@
...
+                {% link href=url text=download|add:name primary=True download=True hide_text=True extra_classes="file__download" icon="download" icon_outlined=True icon_position="before" %}
```
pi-sigma, line 53:
```suggestion
                {% link href=url text=download|add:" "|add:name primary=True download=True hide_text=True extra_classes="file__download" icon="download" icon_outlined=True icon_position="before" %}
```

src/open_inwoner/components/templates/components/File/File.html
pi-sigma: I would avoid mixing translations with formatting and instead call the filter twice.

pi-sigma (changes requested):

pi-sigma (approved):

PR #1253, opened 2024-06-12
[#2523] Avoid indexing unpublished products into elasticsearch

alextreme:


src/open_inwoner/search/tests/test_search_products.py
```diff
@@ -83,6 +83,20 @@ def test_search_product_with_typo(self):
         self.assertEqual(len(results), 1)
         self.assertEqual(int(results[0].meta.id), self.product1.id)
 
+    def test_search_product_hide_unpublished(self):
+        product3 = ProductFactory.create(
```
pi-sigma, line 87: Generally we try to avoid indexed variables, as they are meaningless: `product_unpublished` > `product3`. Not a big thing here, as it's the only variable.

pi-sigma (approved):

src/open_inwoner/search/tests/test_search_products.py
alextreme: Makes sense, updated

pi-sigma (approved):

PR #1257, opened 2024-06-13
♿ [#2352] Remove faulty aria-label and add external link

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2343

This approach ensures that screen readers can read both the visible link text (inside the a-href anchor) _**and**_ the additional information about the link opening a new/different website, which is information coming from the meaningful icon.

### DIFF
--- a/src/open_inwoner/utils/ckeditor.py
+++ b/src/open_inwoner/utils/ckeditor.py
@@ -104,16 +104,16 @@ def get_product_rendered_content(product):
                     }
                 )
                 icon.append("open_in_new")
-                # Create the screen reader only text span element
-                sr_only_text = soup.new_tag("span")
-                sr_only_text.attrs.update(
+
+                screen_reader_only_text = soup.new_tag("span")
+                screen_reader_only_text.attrs.update(
                     {
                         "class": "sr-only",
                     }
                 )
-                sr_only_text.append(_("Opens external website"))
-                # Append both elements to the anchor
+                screen_reader_only_text.append(_("Opens external website"))
+
                 element.append(icon)
-                element.append(sr_only_text)
+                element.append(screen_reader_only_text)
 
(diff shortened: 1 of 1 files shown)


alextreme:
Assigning to Paul for review next week, requires a double-check

src/open_inwoner/utils/ckeditor.py
```diff
@@ -100,11 +100,20 @@ def get_product_rendered_content(product):
...
                 icon.append("open_in_new")
+                # Create the screen reader only text span element
+                sr_only_text = soup.new_tag("span")
+                sr_only_text.attrs.update(
+                    {
+                        "class": "sr-only",
+                    }
+                )
+                sr_only_text.append(_("Opens external website"))
+                # Append both elements to the anchor
```
pi-sigma, line 115: Comment is unnecessary, the code is self-explanatory.

src/open_inwoner/utils/ckeditor.py
```diff
@@ -100,11 +100,20 @@ def get_product_rendered_content(product):
                 icon.attrs.update(
                     {
                         "aria-hidden": "true",
-                        "aria-label": _("Opens in new window"),
                         "class": "material-icons",
                     }
                 )
                 icon.append("open_in_new")
+                # Create the screen reader only text span element
+                sr_only_text = soup.new_tag("span")
```
pi-sigma, line 108: Comment is unnecessary. Add a newline between the `icon` code and this part, and make the code self-explanatory by using a more explicit variable name (e.g. `screen_reader_only_text`. It's more verbose but we only use it here so it's ok)

pi-sigma (changes requested):

pi-sigma (approved):

PR #1259, opened 2024-06-17
[#2560] Treat catalogus as a required field on zaaktype response

swrichards:
Previously we allowed for two edge-cases in resolving zaaktypes:

1. We allowed for catalogus to be absent on a ZaakType to accommodate eSuite, in which the field was absent.
2. We allowed for importing a ZaakType if it included a URL pointing to a Catalogus that had not previously been created.

This commit no longer allows for these exceptions. With regards to 1, eSuite has since implemented changes to include the catalogus field. As for 2: it was already an edge-case we felt should raise rather than be suppressed, but this is especially true now as we're moving to support multiple ZGW backends. In that scenario, we'll clearly want to only include ZaakTypes from a known Catalogus API, to avoid cross-backend ambiguity.


src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -107,20 +107,16 @@ def import_zaaktype_configs() -> list[ZaakTypeConfig]:
...
-            if zaak_type.catalogus:
-                catalog = catalog_lookup.get(zaak_type.catalogus)
-                if not catalog:
-                    # weird edge-case: if the zaak_type has a catalogus-url but we don't have the object
-                    # TODO this is bad, log/raise something
-                    pass
+            try:
+                catalog = catalog_lookup[zaak_type.catalogus]
+            except KeyError:
+                raise ValueError(
```
pi-sigma, line 113: Is there a reason for raising `ValueError` instead of adding the msg to `KeyError`? According to the docs, `ValueError` is "Raised when an operation or function receives an argument that has the right type but an inappropriate value". Doesn't seem to be the case here.

pi-sigma (changes requested):

src/open_inwoner/openzaak/zgw_imports.py
pi-sigma: `RuntimeError` seems fine as well, though I would then `except KeyError as exc` and `raise RunTimeError from exc` so the stack trace is preserved. I wonder though if it's worth the effot to apply this distinction of perspectives to our case, since *we* are the caller (the calling code is even in the same package). I find this more intuitive in the case of a library.

pi-sigma (approved):

alextreme (approved):

PR #1260, opened 2024-06-17
♿ [#2340] Header/footer logo use alt from image configuration

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2340

Both the header logo and the footer logo need to use the same ALT that can be set in the Admin in the '**mappen**' configuration: http://localhost:8000/admin/filer/folder/

Each image in there (navigate to any in the mappen/directories) has an 'ALT' field that can be set - but when empty the image should just receive value from the Site Name.

Screenshot 2024-06-17 at 17 41 58

### DIFF
--- a/src/open_inwoner/templates/master.html
+++ b/src/open_inwoner/templates/master.html
@@ -149,7 +149,8 @@
                 {% endif %}
                 {% if footer.logo %}
                     <div class="footer__logo">
-                        {% include "components/Logo/Logo.html" with src=footer.logo.url alt=footer.logo_alt href=footer.logo_url svg_height=60 only %}
+                    {% firstof config.footer_logo.default_alt_text config.name as logo_alt_text %}
+                        {% include "components/Logo/Logo.html" with src=footer.logo.url alt=logo_alt_text href=footer.logo_url svg_height=60 only %}
                     </div>
                 {% endif %}
             </div>


src/open_inwoner/utils/context_processors.py
```diff
@@ -55,7 +55,7 @@ def settings(request):
         },
         "footer": {
             "logo": config.footer_logo,
-            "logo_alt": config.name,
+            "logo_alt": config.logo.default_alt_text,
```
jiromaykin, line 58: @pi-sigma Not sure I am using the correct approach here... Not quite sure why this value is set differently for the footer anyway 
(would this just be for emailtemplates?)

src/open_inwoner/templates/master.html
```diff
@@ -149,7 +149,8 @@
                 {% endif %}
                 {% if footer.logo %}
                     <div class="footer__logo">
-                        {% include "components/Logo/Logo.html" with src=footer.logo.url alt=footer.logo_alt href=footer.logo_url svg_height=60 only %}
+                    {% firstof config.logo.default_alt_text config.name as logo_alt_text %}
```
pi-sigma, line 152:
```suggestion
                    {% firstof config.footer_logo.default_alt_text config.name as logo_alt_text %}
```

pi-sigma (changes requested):

pi-sigma (approved):

PR #1261, opened 2024-06-18
:arrow_up: [#2562] Bump lxml and xmlsec to their latest versions

sergei-maertens:
xmlsec has done efforts to support binary wheels together with lxml to avoid compile/installation/runtime issues, which should also speed up installation since no source compilation is required.

Commit message from eeac088fe42ce943dbd904437066c3f9e7c9f830:

Several teammembers have faced sporadic issues installing xmlsec due to missing binaries and the resulting difficulties in building the packages with the right library versions. Requiring at least 1.3.14 allows us to use the pre-built binaries.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -379,7 +379,7 @@ jsonschema==4.1.0
     # via drf-spectacular
 kombu==5.3.7
     # via celery
-lxml==4.9.1
+lxml==5.2.2
     # via
     #   django-digid-eherkenning
     #   mail-editor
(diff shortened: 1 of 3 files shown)


alextreme:
The lxml upgrade is causing problems with our uwsgi-installs ( e.g. https://stackoverflow.com/questions/78582042/how-to-solve-the-xmlsec-error-100-lxml-xmlsec-libxml2-library-version-mism ). Requesting @swrichards to revert this PR for the time being as in my eyes the upgrade is currently causing more problems that it solves, I'd prefer to wait with upgrading lxml until the installation hassles are ironed out

sergei-maertens:
I managed to pinpoint it to the presence of the libxml2-dev package in Open Formulieren: https://github.com/open-formulieren/open-forms/commit/8e06944b72dced6de41eed1c9a5233bea64203da

I don't know what in uwsgi causes this to go wrong, but this setup works for us with binary wheel installations. Other workarounds are indeed reverting back to source builds for both xmlsec and lxml, which should only remove the benefit of slower dependency installations.

PR #1262, opened 2024-06-18
[#2564] Mark catalogus as a required field on ZaakTypeconfig

swrichards:


src/open_inwoner/openzaak/tests/test_migrations.py
```diff
@@ -66,3 +70,47 @@ def test_migration_0048_to_0051_multi_zgw_backend(self):
             expected,
             msg="Service config should have been moved to a new ZGWApiGroupConfig",
         )
+
+
+class TestMakeZaakTypeConfigCatalogusRequired(TestCase):
```
pi-sigma, line 75: We have a utility class for migration tests that sets + resets the constraint checks (among other things). You don't have to use it, but take a look: https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/utils/tests/test_migrations.py

pi-sigma (changes requested):
There are two places that rely on the condition that `catalogus` is `null` and should be changed:

- https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openzaak/models.py#L389
- https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openzaak/managers.py#L101

pi-sigma (approved):

PR #1263, opened 2024-06-18
[#2566] Persist the originating ZGW service on the client

swrichards:
[#2566] Persist originating service on zgw clients
    
To support multiple zgw backends, we frequently need to determine
which backend to query for which resource. As a first step to
keeping better track of where ZGW objects come from, in this
commit we persist the zgw service used to configure a client
on the client object, so that we can later fetch and persist
this service to facilitate easier target backend resolution.

### DIFF
--- a/src/open_inwoner/openzaak/clients.py
+++ b/src/open_inwoner/openzaak/clients.py
@@ -1,6 +1,7 @@
 import base64
 import logging
 from datetime import date
+from typing import Literal, Mapping, Type, TypeAlias
 
 from django.conf import settings
 from django.core.files.uploadedfile import InMemoryUploadedFile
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/clients.py
```diff
@@ -680,9 +693,13 @@ def fetch_open_tasks(self, bsn: str) -> list[OpenTask]:
...
 
-def _build_zgw_client(type_) -> APIClient | None:
+ZgwClientType = Literal["zaak", "catalogi", "document", "form"]
+ZgwClient: TypeAlias = ZakenClient | CatalogiClient | DocumentenClient | FormClient
+
+
+def _build_zgw_client(type_: ZgwClientType) -> ZgwClient | None:
     config = OpenZaakConfig.get_solo()
-    services_to_client_mapping = {
+    services_to_client_mapping: Mapping[ZgwClientType, ZgwClient] = {
```
pi-sigma, line 702: There is a mismatch between annotation and assignment: the values in the dictionary are of type `Type`, but what's expected according to annotation is a `Union` of clients.

src/open_inwoner/openzaak/clients.py
```diff
@@ -38,7 +40,18 @@
 logger = logging.getLogger(__name__)
 
 
-class ZakenClient(APIClient):
+class ZgwAPIClient(APIClient):
+    """A client for interacting with ZGW services."""
+
+    configured_from: Service
+    """The Service instance used to configure this client."""
```
pi-sigma, line 47: A second docstring? Seems unusual. Also, the attribute seems to me sufficiently clear, no comment necessary.

pi-sigma (changes requested):

pi-sigma (approved):

alextreme (approved):

PR #1267, opened 2024-06-19
Support both  successful and failing migrations in the migration test utility

swrichards:
I've had to write a few non-trivial migrations for the multi-ZGW backend work, and I suspect there will be a few more. To keep things consistent and readable, I've expanded the migration testing utility to support both successful and failing migrations.

### DIFF
--- a/src/open_inwoner/utils/tests/test_migrations.py
+++ b/src/open_inwoner/utils/tests/test_migrations.py
@@ -1,21 +1,34 @@
+import contextlib
+
 from django.db import connection
 from django.db.migrations.executor import MigrationExecutor
 from django.test import TestCase
 
 
-class TestMigrations(TestCase):
+class TestMigrationsBase(TestCase):
+    app = None
+    migrate_from = None
+    migrate_to = None
+    extra_migrate_from: list[tuple[str, str]] = None
     """
     Test the effect of applying a migration
     Adapted from https://github.com/open-formulieren/open-forms/blob/e64c9368264d3f662542866e2d7d5ba15e0f265c/src/openforms/utils/tests/test_migrations.py
     """
 
-    app = None
-    migrate_from = None
-    migrate_to = None
+    @contextlib.contextmanager
+    def _immediate_constraints(self):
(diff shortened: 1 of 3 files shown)


src/open_inwoner/utils/tests/test_migrations.py
```diff
@@ -37,18 +35,71 @@ def setUp(self):
...
+    ```
+    MyModel = self.apps.get_model(self.app, "MyModel")
+    ```
+    """
+
+    def setUp(self):
+        self.revert_to_migrate_from()
+        self.setUpBeforeMigration(self.old_apps)
+        self.apply_migration_to()
+        self.restore_constraint_checks()
```
pi-sigma, line 71: The level of abstraction strikes me as uneven across the class. Restoration of constraint checks is abstracted in the method `restore_constraint_checks`, but the original change (`"SET CONSTRAINTS ALL IMMEDIATE"`) is implicit in `setUpBeforeMigration`. Loooking just at this code, I can see immediately that constraints are being restored, but I don't see why that's happening because the original change is hidden. If possible (perhaps there are complications), I'd prefer having both of these implicit or both explicit in a separate method.

src/open_inwoner/utils/tests/test_migrations.py
```diff
@@ -37,18 +35,71 @@ def setUp(self):
...
+    def test_migration_should_fail(self):
+        # Setup failure conditions
+        self.assertRaises(SomeError):
+            self.attempt_migration()
+    ```
+    """
+
+    def setUp(self):
+        self.revert_to_migrate_from()
+        self.setUpBeforeMigration(self.old_apps)
```
pi-sigma, line 97: `setUpBeforeMigration` has `cursor.execute("SET CONSTRAINTS ALL IMMEDIATE")`, which is executed independently of whether the migration succeeds. Do we not need to restore the constraints?

pi-sigma (changes requested):

pi-sigma (approved):

PR #1273, opened 2024-06-24
♿ [#2435] Add aria-expanded to mobile anchors navigation

jiromaykin:
issue https://taiga.maykinmedia.nl/project/open-inwoner/task/2435 

note: 

- `aria-expanded` only works for the elements that indicate if a **control** is expanded or collapsed.
- `aria-controls` is not a hard requirement, but 'recommended'

### DIFF
--- a/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
+++ b/src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
@@ -3,8 +3,8 @@
 <aside class="anchor-menu {% if desktop %}anchor-menu--desktop{% else %}anchor-menu--mobile{% endif %}" aria-label="{% trans "Secundaire paginanavigatie" %}">
     {% if not desktop %}
     <div class="anchor-menu--container">
-        {% button href="#" icon="expand_more" icon_position="after" extra_classes="anchor-menu--mobile__title anchor-menu__toggle" bordered=False text=_("Op deze pagina") %}
-        <ul class="anchor-menu__list">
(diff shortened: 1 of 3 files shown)


src/open_inwoner/components/templates/components/AnchorMenu/AnchorMenu.html
```diff
@@ -3,8 +3,8 @@
 <aside class="anchor-menu {% if desktop %}anchor-menu--desktop{% else %}anchor-menu--mobile{% endif %}" aria-label="{% trans "Secundaire paginanavigatie" %}">
     {% if not desktop %}
     <div class="anchor-menu--container">
-        {% button href="#" icon="expand_more" icon_position="after" extra_classes="anchor-menu--mobile__title anchor-menu__toggle" bordered=False text=_("Op deze pagina") %}
-        <ul class="anchor-menu__list">
+        {% button icon="expand_more" icon_position="after" extra_classes="anchor-menu--mobile__title anchor-menu__toggle" bordered=False text=_("Op deze pagina") ariaExpanded="false" ariaControls="anchor-menu-list-mobile" %}
+        <ul class="anchor-menu__list" id="anchor-menu-list-mobile">
```
pi-sigma, line 7: I don't see where the `id` is used. Is this necessary?

jiromaykin: Yes. That ID is needed for the WAI_ARIA attribute "aria-controls".
Read more here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

pi-sigma (approved):

PR #1274, opened 2024-06-24
[#2577] Use CKEditorWidget for SSD help-text fields

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2577

### DIFF
--- a/src/open_inwoner/templates/pages/ssd/monthly_reports_list.html
+++ b/src/open_inwoner/templates/pages/ssd/monthly_reports_list.html
@@ -1,54 +1,41 @@
-{% extends 'master.html' %}
-{% load i18n grid_tags dropdown_tags form_tags button_tags ssd_tags %}
+{% extends 'pages/ssd/reports_base.html' %}
+{% load i18n grid_tags dropdown_tags form_tags button_tags render_tags ssd_tags %}
 
-{% block content %}
-    {% render_grid %}
-        {% render_column span=9 %}
-            <h1 class="utrecht-heading-1" id="title">
-                {% trans "Mijn uitkeringen" %}
-            </h1>
-            <p>
-                {% blocktrans with mijn_uitkeringen_text=client.config.mijn_uitkeringen_text %}
-                    {{ mijn_uitkeringen_text }}
-                {% endblocktrans %}
-            </p>
-            <div class="tab--container">
-                <div class="tabs">
-                    <ul class="list tabs__headers">
-                        <li class="list-item tab__header--item"><span id="monthly" class="link tab__header active">{% trans "Maandspecificaties" %}</span></li>
-                        <li class="list-item tab__header--item"><a href="{% url 'ssd:yearly_benefits_index' %}" id="yearly" class="link tab__header">{% trans "Jaaropgaven" %}</a></li>
-                    </ul>
-                    {% if client.config.maandspecificatie_enabled is True %}
-                        <div class="tabs__body">
-                        {# Note: Mijn uitkeringen tab-content styles need to be independent from the URL, unlike the Tabs of the Login page #}
-                        {# URL-dependent styles would be coming from src/open_inwoner/js/components/tab-panels #}
-                        {# Setting class to 'tab__contents' instead of 'tab__content' is a quick-fix #}
-                            <div id="maandspecificaties" class="tab__contents active">
-                                <p>
-                                    {% blocktrans with display_text=client.config.maandspecificatie_display_text %}
-                                        {{ display_text }}
-                                    {% endblocktrans %}
-                                </p>
-                                {% render_form form=form method="POST" id="monthlyreports-form" show_required=False submit_text=_('Download') %}
-                                    {% csrf_token %}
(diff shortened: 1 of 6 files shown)


pi-sigma:
We also need to decide which formatting options we want to support, hence the update to `CKEDITOR_CONFIGS` in the settings. I made a stab at what seems like a reasonalbe default to me; let me know if this is too much or not enough.

src/open_inwoner/templates/pages/ssd/monthly_reports_list.html
```diff
@@ -1,54 +1,41 @@
...
+{% block report %}
+    <div class="tab--container">
+        <div class="tabs">
+            <ul class="list tabs__headers">
+                <li class="list-item tab__header--item"><span id="monthly" class="link tab__header active">{% trans "Maandspecificaties" %}</span></li>
+                <li class="list-item tab__header--item"><a href="{% url 'ssd:yearly_benefits_index' %}" id="yearly" class="link tab__header">{% trans "Jaaropgaven" %}</a></li>
+            </ul>
+            {% if client.config.maandspecificatie_enabled is True %}
+                <div class="tabs__body">
+                {# Note: Mijn uitkeringen tab-content styles need to be independent from the URL, unlike the Tabs of the Login page #}
```
jiromaykin, line 13: Something with rebasing happening here?

src/open_inwoner/conf/base.py
```diff
@@ -947,7 +947,41 @@
...
+                ],
+            },
+            {"name": "links", "items": ["Link", "Unlink"]},
+            {
+                "name": "insert",
+                "items": ["Image", "Table", "HorizontalRule", "Smiley", "SpecialChar"],
+            },
+            "/",
+            {"name": "styles", "items": ["Styles", "Format", "Font", "FontSize"]},
+            {"name": "colors", "items": ["TextColor", "BGColor"]},
```
jiromaykin, line 983: NO.
😉

src/open_inwoner/conf/base.py
```diff
@@ -947,7 +947,41 @@
 
 CKEDITOR_CONFIGS = {
     "default": {
+        "height": 200,
+        "width": "80%",
```
jiromaykin, line 951: General: Percentages are (usually) bad and won't account for needed breaks on smaller screens. 
Also: width and height should not be static like this but you could use values like min-height/max-height/max-width in these types of situations.

More specific: we do not need new additions in base? Since we already have this CSS here: `src/open_inwoner/scss/admin/_ck_editor.scss`
For an example of the way we use this wysiwyg editor, you can also peek at any of the cases in http://localhost:8000/admin/openzaak/zaaktypeconfig/ and look at how we treated the 'ZAAKTYPE STATUSTYPE CONFIGURATIES' of each casetype.

And most important: we should only give a limited number of options to the editors.
So this would be the maximum of stuff we should make available really (do not give them the option to add colors or unnecessary textstyles like "typewriter"): 👇

editor-options-casetype

jiromaykin (changes requested):

src/open_inwoner/templates/pages/ssd/reports_base.html
```diff
@@ -0,0 +1,20 @@
+{% extends 'master.html' %}
+{% load i18n grid_tags dropdown_tags form_tags button_tags ssd_tags render_tags %}
+
+{% block content %}
+    {% render_grid %}
+        {% render_column span=9 %}
+            <h1 class="utrecht-heading-1" id="title">
+                {% trans "Mijn uitkeringen" %}
+            </h1>
+            <div class="wysiwyg">
```
jiromaykin, line 10: What is this template for? Is this for the PDF? 
(if so: in that case we cannot use the wysiwyg class here; we will need the styling from the PDF CSS file).

src/open_inwoner/templates/pages/ssd/monthly_reports_list.html
```diff
@@ -1,54 +1,41 @@
...
+                {# URL-dependent styles would be coming from src/open_inwoner/js/components/tab-panels #}
+                {# Setting class to 'tab__contents' instead of 'tab__content' is a quick-fix #}
+                    <div id="maandspecificaties" class="tab__contents active">
+                        <div class="wysiwyg">
```
jiromaykin, line 17: Very nice :-)

But now that we removed styling for the paragraph, we still need some spacing for the tab-panel, so I'd suggest to go to `src/open_inwoner/scss/components/TabPanel/TabPanel.scss` and then add the 'quickfix' tab__contents classes to the existing styles (these will change in another PR that changes the styling for Uitkeringen only and will make it different, but still look like the Login page Tabs anyway later on):

```
  .tab__content,
  .tab__content.active,
  .tab__contents,
  .tab__contents.active {
    margin: 1em;
    scroll-margin-top: 150px;
  }
```

TLDR: add the classes that have an extra 's' in the end of their name.

jiromaykin (changes requested):
Minor style addition needed + I am quite happy users can still add headings... They cannot add an H1 anyway currently, so no problem for accessibility.

jiromaykin (approved):

PR #1278, opened 2024-06-25
[#2587] Require all services except forms in ZGWApiGroupConfig

swrichards:
Note that the changes in the test files were necessary because the `OpenZaakConfig` legacy getter/setter methods relied on most services being nullable, enabling it to dynamically create each missing service as it was set. That is obviously no longer possible, so now you'll see a lot of:

```python
ZGWApiGroupConfigFactory(
  ztc_service__api_root=CATALOGI_ROOT,
  zrc_service__api_root=ZAKEN_ROOT,
  form_service=None,
)
```

in place of

```python
self.config = OpenZaakConfig.get_solo()
self.config.zaak_service = ServiceFactory(api_root=ZAKEN_ROOT, api_type=APITypes.zrc)
self.config.catalogi_service = ServiceFactory(api_root=CATALOGI_ROOT, api_type=APITypes.ztc)
self.config.save()       
```

This also functions as a general cleanup to pave the way for the coming changes: the test setups would've had to evolve eventually anyway, and this will make the logic around multiple backends a bit easier.

Taiga: [#2587](https://taiga.maykinmedia.nl/project/open-inwoner/task/2587)


src/open_inwoner/openzaak/models.py
```diff
@@ -65,23 +65,26 @@ class ZGWApiGroupConfig(models.Model):
         on_delete=models.PROTECT,
```
pi-sigma, line 65: Do we still need this if the service is required? Same for the other services below?

src/open_inwoner/openzaak/models.py
```diff
@@ -65,23 +65,26 @@ class ZGWApiGroupConfig(models.Model):
...
     )
     drc_service = models.ForeignKey(
         "zgw_consumers.Service",
         verbose_name=_("Documenten API"),
         on_delete=models.PROTECT,
         limit_choices_to={"api_type": APITypes.drc},
         related_name="zgwset_drc_config",
-        null=True,
+        null=False,
+        blank=False,
```
pi-sigma, line 78: Is this not the default behavior?

src/open_inwoner/openzaak/tests/test_migrations.py
```diff
@@ -172,3 +172,32 @@ def test_migration_0051_to_0052_raises_with_descriptive_exception_message(self):
             " This field is now required: please manually update all the affected rows or re-sync"
             " your ZGW objects to ensure the field is included.",
         )
+
+
+class TestAllServicesExceptFormsRequiredForZGWApiGroup(TestFailingMigrations):
```
pi-sigma, line 177: Since this is testing the migration, perhaps `ZGWAPIGroupRequirementsFailingMigrationTest` or something like it?

Also, could use a test for a migration that succeeds.

pi-sigma (changes requested):

pi-sigma (approved):

PR #1280, opened 2024-06-25
♿ [#2379] Search feedback notification for screenreaders

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2379
the issue from the audit is not quite correct, when it says "...without adding focus...". 
We only have dynamic content in our HTMX components (like the Spinner in Mijn Aanvragen) but we do not have dynamic content for our Notifications. Nor for our form Errors. These only becoem visible on PageLoad.
So we need to add focus on them or else most screen readers will ignore them.

`role=alert` and `aria-live` usually only work for dynamically loaded content, so hardly work for Divs that are loaded on new pageload.

Adding the negative` tabindex= -1` will make the text focusable for javascript, without interfering with the natural Tab order.

Forcing focus feels 'hacky' but this is because there is no 'on page load' solution that will work the same for all screen readers.


jiromaykin:
@pi-sigma This one may look simple but it's quite the juggling of what not-not-not to do. So 'd be happy to explain what I am trying to do here in live conversation.

TLDR:
- add focus for javascript only
- keep the focused element OUT of Tab control
- removing visible focus style from the notification-content (my own decision, not sure if this is good/bad)
- [terribly changing of the context ](https://www.w3.org/WAI/WCAG21/Understanding/status-messages.html#dfn-changes-of-context)but this seems the only way for elements that are not 'dynamic' and can never make use of 'role=status' or 'role=alert' so 'aria-live' will never work.

jiromaykin:
@pi-sigma I forgot: You can test this actual issue by searching and going to the Search page (activate Elasticsearch) and submit feedback ("_Heeft u gevonden wat u zocht?"/"Feedback versturen_") which will generate a Notifications section that contains 1 Notification.

Still my PR needs more tweaking because I based it on this answer, amongst others, and this means I will need to add `role="region"` at least:
https://stackoverflow.com/a/41508795/21179736

but I am still not sure what to do if there are multiple Notifications, since you can only add focus on 1 element and I am not sure if I could just add focus to the surrounding `NotificationS.html` template, instead of adding focus to just the child 'list item' in `notification.html`. Styling is possibly applied in a 'cascading' order.

Plus: I did add overkill when also adding the focus to the `Error.html` template... Which is outside of the scope of this issue actually. 
But those Error notifications are a bit problematic for screen-readers because in Forms we now set Focus on the first input element, and not on any of the errors (guess that should be a new issue).

src/open_inwoner/js/components/notifications/NotificationsList.js
```diff
@@ -0,0 +1,39 @@
...
+export class NotificationsList {
+  static selector = '.notifications'
+  constructor(notificationContents) {
+    this.notificationContents = notificationContents
+  }
+
+  scrollToFirstNotification() {
+    if (this.notificationContents.length > 0) {
+      // Scroll to the first notification, since there could be multiple
+      this.notificationContents[0].scrollIntoView({
```
pi-sigma, line 14: `notificationContents` behaves like a stack, so this will srcoll into the *last* notification (because it was the last one added). If you want to start with the *first* notification, you need to loop in reverse (so -1 instead of 0).

I don't know what's the intended (or best) behavior, though. If you want to focus on the last notification, you can ignore this.

src/open_inwoner/js/components/notifications/NotificationsList.js
```diff
@@ -0,0 +1,39 @@
...
+        block: 'center',
+        behavior: 'smooth',
+      })
+
+      // Add a delay before setting focus for screen readers
+      setTimeout(() => {
+        // Set focus on the first notification content
+        this.notificationContents[0].focus()
```
pi-sigma, line 22: See previous comment.

pi-sigma (changes requested):

src/open_inwoner/js/components/notifications/NotificationsList.js
jiromaykin: @pi-sigma Actually no. If you would manually add multiple templates for notifications by hand (_after doing Git pull and rebuild NPM_), you can see the reverse is not needed.

The code correctly targets the first notification content by using `this.notificationContents[0]`. 
There is no need to loop in reverse or use a different index. The code already achieves the intended behavior of scrolling to and focusing on the first notification, because I am "re-indexing the Nodelist" here.

If you want to visually see which 1 element truly has focus after PageReload, you can adjust/remove the `_Notification.SCSS` in this PR where I removed the browsers default focus-visible style.

You can test this manually this way: paste in multiple types of notifications in the wrong order, after the last line of the `Notification.html` template, so you can see the re-ordering and focus is happening correctly and generate these notifications by using the "Feedback versturen" button after doing a Search query:

```
<div class="notification notification--success ">
   <div class="notification__icon">
      <span aria-hidden="true" class="material-icons-outlined ">check_circle</span>
   </div>
   <div class="notification__content" role="status" tabindex="-1">
      <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-compact">Success success</p>
   </div>
   <button class="button notification__close button--textless button--icon button--transparent" type="" name="" value="" title="Sluit deze melding" aria-label="Sluit deze melding">
   <span aria-hidden="true" class="material-icons ">close</span>
   </button>
</div>
<div class="notification notification--info ">
   <div class="notification__icon">
      <span aria-hidden="true" class="material-icons-outlined ">check_circle</span>
   </div>
   <div class="notification__content" role="status" tabindex="-1">
      <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-compact">Info blue info</p>
   </div>
   <button class="button notification__close button--textless button--icon button--transparent" type="" name="" value="" title="Sluit deze melding" aria-label="Sluit deze melding">
   <span aria-hidden="true" class="material-icons ">close</span>
   </button>
</div>
<div class="notification notification--error ">
   <div class="notification__icon">
      <span aria-hidden="true" class="material-icons-outlined ">check_circle</span>
   </div>
   <div class="notification__content" role="status" tabindex="-1">
      <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-compact">Error error</p>
   </div>
   <button class="button notification__close button--textless button--icon button--transparent" type="" name="" value="" title="Sluit deze melding" aria-label="Sluit deze melding">
   <span aria-hidden="true" class="material-icons ">close</span>
   </button>
</div>
<div class="notification notification--warning ">
   <div class="notification__icon">
      <span aria-hidden="true" class="material-icons-outlined ">check_circle</span>
   </div>
   <div class="notification__content" role="status" tabindex="-1">
      <p class="utrecht-paragraph utrecht-paragraph--oip utrecht-paragraph--oip-compact">Warning warning</p>
   </div>
   <button class="button notification__close button--textless button--icon button--transparent" type="" name="" value="" title="Sluit deze melding" aria-label="Sluit deze melding">
   <span aria-hidden="true" class="material-icons ">close</span>
   </button>
</div>
```

src/open_inwoner/js/components/notifications/index.js
```diff
@@ -25,16 +53,25 @@ export class Notification {
...
    */
   scrollToNotification() {
-    const notificationContent = document.querySelector('.notification__content')
+    const notificationContents = Array.from(
+      this.node.querySelectorAll('.notification__content')
+    )
+
+    if (notificationContents) {
+      notificationContents.forEach((content) => {
+        // If errors are present, scroll and trigger the opened state
```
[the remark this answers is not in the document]
jiromaykin: I was hoping to separate the surrounding 'NotificationS' **list** template logic with those from the individual 'notification' **items**. But I have to target both the parent, as well as the children at the same time so I think I can just remove all this code here.

pi-sigma (approved):

PR #948, opened 2024-01-11
[#1916] Fix test suite

pi-sigma:
Taiga: [#1916](https://taiga.maykinmedia.nl/project/open-inwoner/issue/1916)
- Refactor flaky tests to use hard-coded values
- Temporarily disable parallel tests due to a Django bug: [#32144](https://code.djangoproject.com/ticket/32114) ([PR 17650](https://github.com/django/django/pull/17650))
- Update failing openklant tests + mocks (PR #945 broke develop because it was not rebased)

We keep an eye on the Django ticket and when/if the PR gets merged, we can try switching back to parallel tests

PR #1282, opened 2024-06-26
[#2589] Run CI tests in parallel

swrichards:
We have made several previous attempts at running the CI tests in
parallel, which were ultimately reverted because we ran into the
issue described at https://github.com/django/django/pull/17650 and
https://code.djangoproject.com/ticket/32114 (in this repo, see PRs
#1050, #948 and #690.

The Django ticket has been fixed, but will will only be included
in 5.1 onwards, and won't be backported to 4.2.x. However, it seems
the current testsuite no longer fails with the issue reported in the
Django ticket (pickle related errors). It is possible the potentially
failing tests are flaky, but so far we've not run into it during a
handful of parallel runs. This commit will re-enable parallel tests,
and we'll have to see if we run into the same issues again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,9 @@ jobs:
   tests:
     name: Run the Django test suite
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        test-type: [main, elastic]
 
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


alextreme (approved):
I'm okay with trying this out again for the time being, worst case is that we run into the flakey tests issue again and have to revert this CI change

PR #1283, opened 2024-06-26
[#2558] Multi backend support for the catalogi and zaaktypes ZGW importer

swrichards:


src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -48,32 +52,35 @@ def import_catalog_configs() -> list[CatalogusConfig]:
...
-        )
-        return []
+    proxy = MultiZgwClientProxy(build_catalogi_clients())
+    result = proxy.fetch_catalogs_no_cache()
+
+    if result.has_errors:
+        for response in result.failing_responses:
+            logger.exception(
+                "Client %s encountered an exception. Import will continue, for any other configured clients",
+                exc_info=response.exception,
```
pi-sigma, line 62: I don't see anything being assigned to `%s`. Perhaps;

```
logger.exception(
                "Client %s encountered an exception. Import will continue, for any other configured clients" % "error_msg",
                exc_info=response.exception,
)
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -87,17 +94,16 @@ def import_zaaktype_configs() -> list[ZaakTypeConfig]:
...
+    proxy = MultiZgwClientProxy(build_catalogi_clients())
+    result = proxy.fetch_zaaktypes_no_cache()
 
-    zaak_types = get_configurable_zaaktypes(client)
-    if not zaak_types:
-        return []
+    if result.has_errors:
+        for response in result.failing_responses:
+            logger.exception(
+                "Client %s encountered an exception", exc_info=response.exception
```
pi-sigma, line 103: - See previous comment about log formatting.
- If the import continues (the function doesn't return or raise anything), log message should be the same as above

pi-sigma (changes requested):

alextreme (approved):

PR #1284, opened 2024-06-27
[#2570] Add multi-backend support to ZGW iotypes importer

swrichards:


src/open_inwoner/openzaak/tests/test_zgw_imports_command.py
```diff
@@ -32,55 +31,77 @@ class ZGWImportTest(ClearCachesMixin, TestCase):
...
+        aaaaa - 123456789 [catalogi.nl]
+        bbbbb - 123456789 [andere-catalogi.nl]
+        bbbbb - 123456789 [catalogi.nl]
 
-        imported 2 new zaaktype configs
+        imported 4 new zaaktype configs
+        AAA - zaaktype-aaa
         AAA - zaaktype-aaa
         BBB - zaaktype-bbb
+        BBB - zaaktype-bbb
```
pi-sigma, line 74: Can we distinguish between these configs as in the case of the catalogi messages above (e.g. by displaying the netloc of the service api root or something else)?. It's quite useful in the catalogus case.

pi-sigma (changes requested):
Looking good. One small question/suggestion about the messages, though not sure if it's feasible.

pi-sigma (approved):

PR #1285, opened 2024-06-27
♿ [#2374] Add more verbose error-text for input fields

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2374
This one is more for cognitive disabilities, where error notifications need to be more clear in the actions they expect from the user.
Would be good to have this in as many forms as possible and not just the 1 emailfield in the profiel-edit page.


jiromaykin:
The audit report did not mention the text for the error-notification (the ones that always load on top of page) so that's probably not needed here, only the warning error for each input field.

Screenshot 2024-06-27 at 17 38 54

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -674,6 +676,39 @@ def test_modify_email_updates_klant_api_but_skip_unchanged_phone(self, m):
         )
 
 
+class TestForm(ErrorMessageMixin, forms.Form):
+    name = forms.CharField(required=True, label="Naam")
+    email = forms.EmailField(required=True, label="E-mailadres")
+
+
+class ErrorMessageMixinTests(TestCase):
```
pi-sigma, line 684: Let's stick to singular nouns for classes. There are many tests in `ErrorMessageMixinTest`, but there is only one class.  I know our code is not consistent, but we should try.

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -674,6 +676,39 @@ def test_modify_email_updates_klant_api_but_skip_unchanged_phone(self, m):
...
+    def test_default_error_messages(self):
+        form = TestForm(data={})
+        self.assertFalse(form.is_valid())
```
pi-sigma, line 687: Not needed.

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -674,6 +676,39 @@ def test_modify_email_updates_klant_api_but_skip_unchanged_phone(self, m):
...
+            ],
+        )
+
+    def test_custom_error_messages(self):
+        custom_messages = {
+            "name": {"required": _("Naam is verplicht.")},
+            "email": {"required": _("E-mail is verplicht.")},
+        }
+        form = TestForm(data={}, custom_error_messages=custom_messages)
+        self.assertFalse(form.is_valid())
```
pi-sigma, line 707: Not needed

src/open_inwoner/accounts/views/profile.py
```diff
@@ -40,6 +41,24 @@
 logger = logging.getLogger(__name__)
 
 
+class ErrorMessageMixin(FormMixin):
```
pi-sigma, line 44: I agree with your comment that displaying the error message twice is overkill, so this class can be removed.

src/open_inwoner/utils/forms.py
```diff
@@ -6,6 +6,44 @@
 from django.utils.translation import gettext_lazy as _
 
 
+class ErrorMessageMixin:
+    custom_error_messages = {
```
pi-sigma, line 10: The name is confusing, because these are *not* the custom error messages supplied by the caller. In your tests these are the `default_error_messages`.

src/open_inwoner/utils/forms.py
```diff
@@ -6,6 +6,44 @@
...
+        "required": _(
+            "Het verplichte veld %s is niet (goed) ingevuld. Vul het veld in."
+        )
+    }
+
+    def __init__(self, *args, **kwargs):
+        user_added_messages = {}
```
pi-sigma, line 17: This should be simply `custom_error_messages`, it contrasts with the class-level `default_error_messages`. This way the code is consistent with the test you wrote.

src/open_inwoner/utils/forms.py
```diff
@@ -6,6 +6,44 @@
...
+        if "custom_error_messages" in kwargs:
+            user_added_messages = kwargs.pop("custom_error_messages")
+        super().__init__(*args, **kwargs)
+        for field_name, field in self.fields.items():
```
pi-sigma, line 21: 1. The logic in this loop is a bit hard to discern. If all you want to do is define the error message on `field` as the relevant entry from `custom_error_messages`, if it's supplied by the caller, otherwise fall back on `default_error_messages` defined on the class, it should be possible to simplify this.

2. I don't like that "required" and `required_message` are hard-coded. For every new field error that we add, we would have to update the logic as well. Try to make this more generic.

src/open_inwoner/accounts/views/profile.py
```diff
@@ -198,7 +217,12 @@ def post(self, request, *args, **kwargs):
 
 
 class EditProfileView(
-    LogMixin, LoginRequiredMixin, CommonPageMixin, BaseBreadcrumbMixin, UpdateView
+    LogMixin,
+    LoginRequiredMixin,
+    CommonPageMixin,
+    BaseBreadcrumbMixin,
+    ErrorMessageMixin,
```
pi-sigma, line 224: Not needed if the above class is removed

pi-sigma (changes requested):
The failing test is to be expected, since you're mixing in the newly defined `ErrorMessageMixin` into `UserForm`. Update the assertion in the test to reflect the new message.

src/open_inwoner/utils/forms.py
```diff
@@ -9,26 +9,30 @@
...
             for key in self.default_error_messages.keys():
                 if key in field_error_messages.keys():
-                    error_message = field_error_messages[key] % f'"{field.label}"'
+                    error_message = field_error_messages[key].format(
+                        field_name=f'"{field.label}"'
+                    )
                 else:
-                    error_message = (
-                        self.default_error_messages[key] % f'"{field.label}"'
+                    error_message = self.default_error_messages[key].format(
```
jiromaykin, line 34: @pi-sigma Great, this looks clean!

PR #1287, opened 2024-07-01
♿ [#2583] Make control of modal focusable in the right order

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/us/2402

1. accessibility requirement "When a dialog closes, focus returns to the element that invoked the dialog"
2. making sure the Escape button doesn't interfere with closing of the navigation etc.
3. moving the KCM feedback button to bottom of footer for more logical Tab order

Effect:
after closing a modal your Tab focus should land on the button that opened it, meaning you can Tab backward or forward starting from the modal-opening-control button.

clickmodalfocus

### DIFF
--- a/src/open_inwoner/js/components/modal/index.js
+++ b/src/open_inwoner/js/components/modal/index.js
@@ -11,6 +11,7 @@ export default class Modal {
       '.modal__confirm .inner-text'
     )
     this.closeTitle = this.node.querySelector('.modal__close-title')
+    this.openedBy = null // Track the element that opened the modal
 
     // This is for the prefilled modals so they will not be emptied
     if (!this.node.classList.contains('modal--no-reset')) {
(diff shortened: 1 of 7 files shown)


src/open_inwoner/js/components/modal/index.js
```diff
@@ -124,7 +125,16 @@ export default class Modal {
   show(refocusOnClose) {
     this.node.classList.add('modal--open')
     this.refocusOnClose = refocusOnClose
-    this.close.focus()
+
+    // Set focus on the close button if it exists
+    if (this.close) {
+      console.log('value of this close: ', this.close)
```
pi-sigma, line 131: Lingering `console.log`

pi-sigma (changes requested):

pi-sigma (approved):

PR #1295, opened 2024-07-04
[#2606] Lazy load synonyms for Elasticsearch

pi-sigma:
The synonyms for Elastic Search need to be loaded *after* all the migrations have been run.

Taiga: [issue #2606](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2606)

### DIFF
--- a/src/open_inwoner/search/utils.py
+++ b/src/open_inwoner/search/utils.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.db.utils import DatabaseError
+from django.utils.functional import lazy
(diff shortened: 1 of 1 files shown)


pi-sigma:
Some thoughts:
- Wrapping `load_synonyms` in `lazy` doesn't work (without further ado) because it causes a `SerializationError`
- We played with the idea of loading synonyms in the `ready` function and patch `ProductDocument`, but the solution is neither pretty nor straightforward.

PR #1296, opened 2024-07-05
Update docker_start.sh

alextreme:
Taken from open_zaak Dockerfile


pi-sigma (changes requested):
Can we exclude the possibility that fixtures need to be loaded in a certain order? Might not work for us even if it works for OpenZaak.

alextreme:
Afaik this only should be necessary for django-admin-index. The other fixtures in our project have unintended consequences.

Lets create a separate dir for the fixtures to be auto-loaded, and ensure that this fixture_dir is also copied over within our Dockerfile. Copying over the Open Zaak functionality is too simple and will break things

pi-sigma:
It seems that loading the django-admin-index fixtures through Django's post-migrate signal is broken (see https://taiga.maykinmedia.nl/project/dimpact-enschede-ssc-ict-1/issue/107). Running the migrations produces the error, but manually loading the fixture immediately afterwards works (after updating the fixture, that is). I could not figure out why this started happening, so this needs further investigation.

For immediate purposes, I updated the fixture and disabled automatic loading via the signal for manual builds. I also updated the docker startup script, so we have the autoload functionality when the app is build and run with Docker.

PR #1298, opened 2024-07-08
[#2605] Use NLDS classes for uitkeringen PDF styling

pi-sigma:
Taiga: [#2605](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2605)

### DIFF
--- a/src/open_inwoner/scss/pdf/_jaaropgave.scss
+++ b/src/open_inwoner/scss/pdf/_jaaropgave.scss
@@ -15,6 +15,7 @@
 
   .yearly-report__p,
   .yearly-report__td {
+    font-family: utrecht-paragraph;
     font-size: 13px;
   }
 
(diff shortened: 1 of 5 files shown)


src/open_inwoner/scss/pdf/_jaaropgave.scss
```diff
@@ -15,6 +15,7 @@
 
   .yearly-report__p,
   .yearly-report__td {
+    font-family: utrecht-paragraph;
```
jiromaykin, line 18: `font-family: utrecht-paragraph;` will never work because "utrecht-paragraph" is an HTML CLASS. 
So this is not CSS. 
_If_ this would be an NLDS project file, then the correct variable would be `var(--utrecht-document-font-family)` which eventually cascades to the end-value 'Body' font-family, but it is not an NLDS page , so i will revert all of these values and add more for the headings.

jiromaykin (approved):

PR #1313, opened 2024-07-17
[Dimpact #107] Fixed rebuild search index task not starting.

joeribekker:
Also explicitly indicated the result should not be kept, since there is not result from this task. The issue with the task not starting was essentially the translatable name messing things up but it's better to refer to tasks by Python path.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       - ./docker-init-db.sql:/docker-entrypoint-initdb.d/init_db.sql
       - db:/var/lib/postgresql/data
     networks:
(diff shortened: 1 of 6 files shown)


pi-sigma (approved):

pi-sigma:
@swrichards The `docker-compose` contains changes because it contained superfluous whitespaces at EOL. I don't know where these came from.

joeribekker:
Why was my CELERY_BEAT_SCHEDULER change undone?

I think Sidney has an interesting lead but I'm closing my PR since it no longer reflects much of what I originally did not did it fix the issue it seems.

PR #1314, opened 2024-07-17
Docker / Docker-compose improvements

joeribekker:
Changed Dockerfile to more closely follow Open Forms.

- Added os upgrade to be sure the libraries as up to date as possible everytime the image is built.
- Replaced vim with nano (I left `curl` in for you A)
- Used uv instead of pip (super fast drop in replacement)
- Compile translatable messages during build
- Added Celery beat script
- Added proper release scheme attributes

Added Celery beat, monitor and health check.

- There was 1 container reserved for all Celery related things:
worker, beat and monitoring. According to Celery documentation
it's best to have a separate Beat process so you can scale
separately. Also, combining monitoring with the events itself
could cause issues. Everything is separated now.

- Additional health checks and dependancies are introduced to
Docker-compose so Celery nicely waits for the database to be
prepped.


docker-compose.yml
```diff
@@ -60,6 +74,15 @@ services:
       - CELERY_RESULT_BACKEND=redis://redis:6379/0
       - CELERY_LOGLEVEL=DEBUG      
       - ES_HOST=elasticsearch
+      # Needed for Celery Flower to match the TIME_ZONE configured in the
+      # settings used by workers and beat containers.
+      - TZ=Europe/Amsterdam
+      # Legacy endpoints are deprecated, new installs should opt-out of using legacy
+      # settings (the default is to use legacy endpoints).
+      - USE_LEGACY_OIDC_ENDPOINTS=no
```
sergei-maertens, line 82: These envvars are not relevant for OIP

joeribekker: The OIDC endpoints? I thought these were part of the library?

Also the TZ can be gone? The comment seemed to indicate it was useful.

sergei-maertens: the `USE_LEGACY_*` envvars are Open Forms specific, they're not part of the library. The breaking change will be done in OIP instead of adding an opt-in flag since nobody is using this yet, which is not the case for OF

TZ needs to stay for celery I believe, but my comment is only about the OIDC envvars!

Dockerfile
```diff
@@ -86,15 +84,18 @@ WORKDIR /app
...
+COPY ./bin/check_celery_worker_liveness.py ./bin/
+
+# prevent writing to the container layer, which would degrade performance.
+# This also serves as a hint for the intended volumes.
+VOLUME ["/app/log", "/app/media", "/app/private_media"]
 
 # copy backend build deps
 COPY --from=backend-build /usr/local/lib/python3.11 /usr/local/lib/python3.11
 COPY --from=backend-build /usr/local/bin/uwsgi /usr/local/bin/uwsgi
-COPY --from=backend-build /app/src/ /app/src/
```
sergei-maertens, line 97: I suspect this will break because OIP uses non-PyPI dependencies: https://github.com/maykinmedia/open-inwoner/blob/574199acc11fb89785386043e7f30ee732f08ac6/requirements/base.in#L91

joeribekker: Hmm, Docker build worked locally and the image is built in the CI. Are there other things I need to worry about?

sergei-maertens: not that I can tell right now, but surprises may happen of course due to project specifics :P

bin/start_celery.sh
```diff
@@ -2,8 +2,4 @@
 
 . env/bin/activate
 
-LOGLEVEL=${CELERY_LOGLEVEL:-INFO}
-CONCURRENCY=${CELERY_WORKER_CONCURRENCY:-1}
-
-exec celery --workdir src -A "open_inwoner" events -l info --camera django_celery_monitor.camera.Camera --frequency=2.0 &
-exec celery --workdir src -A "open_inwoner" worker -l $LOGLEVEL -c $CONCURRENCY -E --max-tasks-per-child=50 -l info -B --scheduler django_celery_beat.schedulers:DatabaseScheduler
+celery_worker.sh
```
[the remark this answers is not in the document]
joeribekker: Doh, it doesn't. I'm removing it. These scripts make me shudder anyway, like bootstrap.py... and it was a copy of celery_worker + activate

bin/start_celery.sh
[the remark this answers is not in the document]
joeribekker: Alex i bet

docker-compose.yml
```diff
@@ -1,5 +1,17 @@
...
+#
+# You can use this docker-compose to spin up a local Open Forms stack for demo/try-out
+# purposes, or to get some insight in the various components involved (e.g. to build
+# your Helm charts from). Note that various environment variables are UNSAFE and merely
+# specified so that you can get up and running with the least amount of friction.
+#
+# Before deploying to production, please review the environment configuration reference:
+# https://open-inwoner.readthedocs.io/
+#
+version: '3.4'
```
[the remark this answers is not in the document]
joeribekker: Leaving for another time.

docker-compose.yml
```diff
@@ -1,5 +1,17 @@
-# Inspired by https://docs.docker.com/compose/django/
-version: '3'
+#
+# Docker-compose for full Open Inwoner stack.
+#
+# DISCLAIMER: THIS IS FOR DEVELOPMENT PURPOSES ONLY AND NOT SUITABLE FOR PRODUCTION.
+#
+# You can use this docker-compose to spin up a local Open Forms stack for demo/try-out
```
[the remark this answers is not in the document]
joeribekker: copy/pasta

Dockerfile
sergei-maertens: I remember what the deal is - when you `pip install -e ...`, then the code goes into `/app/src` so that's why it had to be copied over. While OIP uses non-PyPI dependencies, they aren't being installed in editable mode and go into site-packages, so this COPY instruction is now indeed obsolete.

bin/docker_start.sh
```diff
@@ -2,24 +2,16 @@
 
 set -ex
 
-# Wait for the database container
+# wait for required services
 # See: https://docs.docker.com/compose/startup-order/
-export PGHOST=${DB_HOST:-db}
-export PGPORT=${DB_PORT:-5432}
+${SCRIPTPATH}/wait_for_db.sh
```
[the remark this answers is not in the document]
sergei-maertens: this is at least infrastructure independent - health checks have different definitions in docker-compose/docker without compose/kubernetes and it makes the dependency explicit

I think it may work because the undefined/empty variable points to `/wait_for_db.sh` which likely happens to be exactly the file path where it's copied in the docker image?

sergei-maertens (approved):

bin/docker_start.sh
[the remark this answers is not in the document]
joeribekker: > likely happens to be exactly the file path where it's copied in the docker image?
correct :)

PR #1317, opened 2024-07-18
[#2483] Update cases listview to show the combined data from multiple endpoints

swrichards:
[Taiga issue](https://taiga.maykinmedia.nl/project/open-inwoner/us/2483).

### DIFF
--- a/src/open_inwoner/cms/cases/views/cases.py
+++ b/src/open_inwoner/cms/cases/views/cases.py
@@ -1,21 +1,41 @@
+import concurrent.futures
+import logging
+from dataclasses import dataclass
+
 from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import TemplateView
 
 from view_breadcrumbs import BaseBreadcrumbMixin
+from zgw_consumers.concurrent import parallel
 
 from open_inwoner.htmx.mixins import RequiresHtmxMixin
+from open_inwoner.openzaak.api_models import Zaak
 from open_inwoner.openzaak.cases import preprocess_data
-from open_inwoner.openzaak.clients import build_zaken_client
 from open_inwoner.openzaak.formapi import fetch_open_submissions
-from open_inwoner.openzaak.models import OpenZaakConfig
+from open_inwoner.openzaak.models import OpenZaakConfig, ZGWApiGroupConfig
 from open_inwoner.openzaak.types import UniformCase
 from open_inwoner.openzaak.utils import get_user_fetch_parameters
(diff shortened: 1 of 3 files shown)


src/open_inwoner/cms/cases/views/cases.py
```diff
@@ -55,17 +71,33 @@ class InnerCaseListView(
...
+        with parallel() as executor:
+            futures = [
+                executor.submit(self.get_cases_for_api_group, group)
+                for group in all_api_groups
+            ]
+
+            cases = []
+            for task in concurrent.futures.as_completed(futures):
+                try:
+                    task_group = all_api_groups[futures.index(task)]
```
pi-sigma, line 91: Maybe it's because I'm coming from my celery PR, but this variable name sounds strange (I think of `task_group` as a group of (celery) tasks). Perhaps `task_api_group`?

src/open_inwoner/openzaak/cases.py
```diff
@@ -102,30 +92,28 @@ def add_status_type_config(case: Zaak) -> None:
...
+        resolve_status_type(case, client=group.catalogi_client)
+        resolve_resultaat(case, client=group.zaken_client)
+        resolve_resultaat_type(case, client=group.catalogi_client)
         add_zaak_type_config(case)
         add_status_type_config(case)
 
     # TODO error handling if these are none?
     # use contextmanager to ensure the `requests.Session` is reused
-    with zaken_client, catalogi_client:
+    with group.catalogi_client, group.zaken_client:
```
pi-sigma, line 113: Is this comment still relevant?

src/open_inwoner/openzaak/cases.py
```diff
@@ -102,30 +92,28 @@ def add_status_type_config(case: Zaak) -> None:
...
 
 
-def preprocess_data(cases: list[Zaak]) -> list[Zaak]:
+def preprocess_data(cases: list[Zaak], group: ZGWApiGroupConfig) -> list[Zaak]:
     """
     Resolve zaaktype and statustype, add status type config, filter for visibility
 
     Note: we need to iterate twice over `cases` because the `zaak_type` must be
           resolved to a `ZaakType` object before we can filter by visibility
     """
```
pi-sigma, line 101: The docstring is out of date, now that `preprocess_case` is inside this function (though it's not the present commit). We're iterating several times over `cases`, not just twice.

pi-sigma (changes requested):

src/open_inwoner/openzaak/cases.py
pi-sigma: You're right, got it on second look.

pi-sigma (approved):

PR #1326, opened 2024-07-29
[#2643] make ZakenClient thread-safe

swrichards:


src/open_inwoner/openzaak/clients.py
```diff
@@ -95,18 +105,17 @@ def fetch_cases_by_bsn(
         user_bsn: str,
         max_requests: int | None = 4,
         identificatie: str | None = None,
+        max_confidentiality: VertrouwelijkheidsAanduidingen = VertrouwelijkheidsAanduidingen.openbaar,
```
pi-sigma, line 108: Passing `max_confidentiality` to `fetch_cases` is obligatory, so the default value here should be unnecessary.

src/open_inwoner/openzaak/clients.py
```diff
@@ -95,18 +105,17 @@ def fetch_cases_by_bsn(
...
     ) -> list[Zaak]:
         """
         retrieve cases for particular user with allowed confidentiality level
 
         :param:max_requests - used to limit the number of requests to list_zaken resource.
         :param:identificatie - used to filter the cases by a specific identification
```
pi-sigma, line 114: Docstring needs updating

src/open_inwoner/openzaak/clients.py
```diff
@@ -144,6 +153,7 @@ def fetch_cases_by_kvk_or_rsin(
         max_requests: int | None = 4,
         zaak_identificatie: str | None = None,
         vestigingsnummer: str | None = None,
+        max_confidentiality: VertrouwelijkheidsAanduidingen = VertrouwelijkheidsAanduidingen.openbaar,
```
pi-sigma, line 156: See above comment about default value

src/open_inwoner/openzaak/clients.py
```diff
@@ -144,6 +153,7 @@ def fetch_cases_by_kvk_or_rsin(
...
     ) -> list[Zaak]:
         """
         retrieve cases for particular company with allowed confidentiality level
```
pi-sigma, line 159: Docstring should be updated with `max_confidentiality`

pi-sigma (changes requested):

PR #1330, opened 2024-08-01
Disable loading of admin-index fixture in testing

swrichards:


src/open_inwoner/accounts/apps.py
```diff
@@ -6,10 +7,18 @@
...
     if "django_admin_index" not in settings.INSTALLED_APPS:
-        print("django_admin_index not installed, skipping update_admin_index()")
+        logger.info("django_admin_index not installed, skipping update_admin_index()")
+        return False
+
+    if not settings.LOAD_ADMIN_INDEX_FIXTURE_ON_STARTUP:
+        logger.info(
+            "LOAD_ADMIN_INDEX_FIXTURE_ON_STARTUP is set to False, skipping update_admin_index()"
+        )
         return False
```
pi-sigma, line 22: If you do:

```
if settings.SKIP_ADMIN_INDEX_FIXTURE:
    logger.info(...)
    return False
```

you only need to define the setting in `conf/ci.py`. Less clutter in our base settings. Also, it conveys directly that the default is to have the index loaded (with `if not settings.LOAD_...` you need to look at the settings file to find out what the default is).)

pi-sigma (changes requested):

pi-sigma (approved):

PR #1333, opened 2024-08-05
[#2648] handle legacy case detail URLs with redirect

swrichards:
The case detail URL now contains a reference to the
ZGW api group to be used to fetch the case. It is quite
likely there are various hard-coded URLs in the wild
(especially in notification emails) that point to the
old case detail URLs. This commit adds a redirect
handler for the now-deprecated case detail URL and
tries to handle the request gracefully, depending on
the case.

[Taiga 2648.](https://taiga.maykinmedia.nl/project/open-inwoner/task/2648)

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -7,7 +7,12 @@ from datetime import datetime
 from django.conf import settings
 from django.contrib import messages
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
-from django.http import Http404, StreamingHttpResponse
+from django.http import (
+    Http404,
+    HttpRequest,
+    HttpResponseRedirect,
+    StreamingHttpResponse,
+)
 from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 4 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -1029,3 +1029,47 @@ def get_context_data(self, **kwargs):
...
         )
         return context
+
+
+class LegacyCaseDetailHandler(View):
+    """Redirect the legacy case detail to the current version with ZGW API group ref."""
+
+    def get(
+        self,
+        request,
```
pi-sigma, line 1039:
```suggestion
        request: Request,
```
check that we have `from requests import Request` at the top

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -1029,3 +1029,47 @@ def get_context_data(self, **kwargs):
...
+        object_id: str,
+    ):
+        redirect_url = None
+        match ZGWApiGroupConfig.objects.count():
```
pi-sigma, line 1043: If we expect cases where `count > 1` and `count == 0` to be exceptional, then a `try/except` block would be more efficient. It's not a big deal though.

src/open_inwoner/openzaak/tests/test_case_detail_redirects.py
```diff
@@ -0,0 +1,74 @@
...
+    ROOT_URLCONF="open_inwoner.cms.tests.urls",
+)
+class LegacyCaseDetailUrlRedirectTest(TestCase):
+    def setUp(self):
+        self.url_patterns = [
+            "case_detail",
+        ]
+
+    def test_legacy_url_redirects_to_only_api_group_prefix(self):
+        first = ZGWApiGroupConfigFactory()
```
pi-sigma, line 19:
```suggestion
        api_group = ZGWApiGroupConfigFactory()
```
Same below for this test. `first` is misleading as there is no `second`

src/open_inwoner/openzaak/tests/test_case_detail_redirects.py
```diff
@@ -0,0 +1,74 @@
...
+            "cases:legacy_case_detail",
+            kwargs={"object_id": object_id},
+        )
+        target_url = reverse(
+            "cases:case_detail",
+            kwargs={"api_group_id": first.id, "object_id": object_id},
+        )
+
+        response = self.client.get(legacy_handler_url)
+        self.assertEqual(ZGWApiGroupConfig.objects.count(), 1)
```
pi-sigma, line 32: Is this necessary? You created one, so there exists one. The only way this assertion would add something if it was possible that additional `ZGWApiGroupConfig` objects get created during the action (i.e. `self.client.get()`), which is not the case.

src/open_inwoner/openzaak/tests/test_case_detail_redirects.py
```diff
@@ -0,0 +1,74 @@
...
+        legacy_handler_url = reverse(
+            "cases:legacy_case_detail",
+            kwargs={"object_id": object_id},
+        )
+        target_url = reverse("cases:index")
+
+        response = self.client.get(legacy_handler_url)
+        messages = [str(m) for m in get_messages(response.wsgi_request)]
+
+        self.assertEqual(ZGWApiGroupConfig.objects.count(), 2)
```
pi-sigma, line 49: See previous comment

src/open_inwoner/openzaak/tests/test_case_detail_redirects.py
```diff
@@ -0,0 +1,74 @@
...
+            kwargs={"object_id": object_id},
+        )
+
+        legacy_handler_url = reverse(
+            "cases:legacy_case_detail",
+            kwargs={"object_id": object_id},
+        )
+
+        response = self.client.get(legacy_handler_url)
+        self.assertEqual(ZGWApiGroupConfig.objects.count(), 0)
```
pi-sigma, line 73: See previous comment

pi-sigma (changes requested):

pi-sigma (approved):

PR #1348, opened 2024-08-14
Add a ZGW data import/export flow to the admin panel

swrichards:
Two notes:

1. Contrary to what I reported earlier, I've decided that the cost of doing this synchronously is acceptable (~6s on my machine for a 3k row dump). So no Celery complications here for now, we can re-introduce it if we run into issues in production.
2. This is only the case _if you turn `DEBUG` off_, that's what was causing the slowness. If you want to test this locally, make sure you set `DEBUG = False` (unless you have local configuration which addresses this, you won't see any assets, which is annoying but not blocking).


src/open_inwoner/openzaak/admin.py
```diff
@@ -90,6 +123,85 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
...
+        custom_urls = [
+            path(
+                "import-catalogus-dump/",
+                self.admin_site.admin_view(self.process_file_view),
+                name="upload_zgw_import_file",
+            ),
+        ]
+        return custom_urls + urls
+
+    @admin.action(description=_("Exporteer naar json export"))
```
pi-sigma, line 139:
```suggestion
    @admin.action(description=_("Exporteer naar json"))
```
2 x "export" is one too many?

src/open_inwoner/openzaak/admin.py
```diff
@@ -90,6 +123,85 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
...
+                            ),
+                            messages.SUCCESS,
+                        )
+
+                        return HttpResponseRedirect(
+                            reverse(
+                                "admin:openzaak_catalogusconfig_changelist",
+                            )
+                        )
+                    except Exception as e:
```
pi-sigma, line 183:
```suggestion
                    except Exception:
```

pi-sigma: Is it worth logging a failed import?

src/open_inwoner/openzaak/admin.py
```diff
@@ -90,6 +123,85 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
...
+                            _(
+                                "We were unable to process your upload. Please regenerate the file and try again."
+                            ),
+                            messages.ERROR,
+                        )
+                    finally:
+                        storage.delete(target_file_name)
+                else:
+                    self.message_user(
+                        request, _("No file was uploaded"), messages.WARNING
```
pi-sigma, line 196: Are the users not going to get an error message from the form?

src/open_inwoner/openzaak/admin.py
```diff
@@ -90,6 +123,85 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
...
+                    )
+            case _:
+                form = ImportZGWDumpForm()
+
+        return TemplateResponse(
```
pi-sigma, line 201: Should this be indented (as belonging to `case _` where `request.method == GET`?

src/open_inwoner/openzaak/tests/test_admin.py
```diff
@@ -109,3 +122,136 @@ def test_both_can_be_disabled(self):
...
+        messages = [str(msg) for msg in response.context["messages"]]
+        self.assertEqual(
+            messages,
+            ["Successfully processed 1 items"],
+        )
+        self.assertFalse(
+            PrivateMediaFileSystemStorage().exists(
+                "zgw_import_dump_2024-08-14-17-50-01.jsonl"
+            ),
+            msg="File should always deleted regardless of success or failure",
```
pi-sigma, line 211:
```suggestion
            msg="File should always be deleted regardless of success or failure",
```

src/open_inwoner/openzaak/tests/test_admin.py
```diff
@@ -109,3 +122,136 @@ def test_both_can_be_disabled(self):
...
+
+        self.assertEqual(m.call_count, 1)
+        self.assertEqual(m.call_args[0][0], "zgw_import_dump_2024-08-14-17-50-01.jsonl")
+        self.assertTrue(isinstance(m.call_args[0][1], PrivateMediaFileSystemStorage))
+
+        messages = [str(msg) for msg in response.context["messages"]]
+        self.assertEqual(
+            messages,
+            [
+                "We were unable to process your upload. Please regenerate the file and try again."
```
pi-sigma, line 243: Will this still be English when the msg has been translated? Otherwise needs translation as well.
```suggestion
                _("We were unable to process your upload. Please regenerate the file and try again.")
```

src/open_inwoner/openzaak/tests/test_admin.py
```diff
@@ -109,3 +122,136 @@ def test_both_can_be_disabled(self):
...
+            ],
+        )
+        self.assertFalse(
+            PrivateMediaFileSystemStorage().exists(
+                "zgw_import_dump_2024-08-14-17-50-01.jsonl"
+            ),
+            msg="File should always deleted regardless of success or failure",
```
pi-sigma, line 250: See previous comment

src/open_inwoner/templates/admin/catalogusconfig_change_list.html
```diff
@@ -0,0 +1,11 @@
+{% extends "admin/change_list.html" %}
+{% load i18n admin_urls %}
+
+{% block object-tools-items %}
+{{ block.super }}
+<li>
+    <a href="{% url 'admin:upload_zgw_import_file' %}" class="addlink">
+        {% trans "Create from json export" %}
```
pi-sigma, line 8:
```suggestion
        {% trans "Import from file" %}
```
"json export" is too specific IMHO.

pi-sigma (changes requested):

src/open_inwoner/openzaak/admin.py
pi-sigma: Nah, I missed that one.

src/open_inwoner/templates/admin/catalogusconfig_change_list.html
```diff
@@ -0,0 +1,12 @@
...
```
pi-sigma, line 8: Still the same?

pi-sigma (approved):
A template was supposedly changed but is still the same. It's a minor thing though.

alextreme (approved):
can be merged after CI passes

PR #1349, opened 2024-08-15
[#2563] Fix enable variables for setup configuration

pi-sigma:
The `enable_config_setting` variables had the wrong type of default value, preventing the configuration steps from being disabled because `enable_setting` would always evaluate to `True`.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2563


alextreme:
@pi-sigma check the failing test

pi-sigma:
@alextreme good to go

PR #1351, opened 2024-08-15
Implement zgw natural key support

swrichards:
This is a preliminary PR to support the ZGW import/export logic, which will rely on natural keys for (de)serialization.

### DIFF
--- a/src/open_inwoner/openzaak/models.py
+++ b/src/open_inwoner/openzaak/models.py
@@ -17,10 +17,12 @@ from zgw_consumers.constants import APITypes
 from zgw_consumers.models import Service
 
 from open_inwoner.openzaak.managers import (
+    CatalogusConfigManager,
     UserCaseInfoObjectNotificationManager,
     UserCaseStatusNotificationManager,
     ZaakTypeConfigQueryset,
     ZaakTypeInformatieObjectTypeConfigQueryset,
+    ZaakTypeResultaatTypeConfigManger,
     ZaakTypeStatusTypeConfigQuerySet,
 )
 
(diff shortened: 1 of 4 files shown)


src/open_inwoner/openzaak/models.py
```diff
@@ -623,6 +636,11 @@ def informatieobjecttype_uuid(self):
     def __str__(self):
         return f"{self.omschrijving} [{self.zaaktype_config.catalogus.base_url}]"
 
+    def natural_key(self):
+        return (self.informatieobjecttype_url,) + self.zaaktype_config.natural_key()
```
pi-sigma, line 640:
```suggestion
        return (self.informatieobjecttype_url, self.zaaktype_config.natural_key())
```
Same below a few times

src/open_inwoner/openzaak/tests/factories.py
```diff
@@ -82,7 +83,8 @@ class Meta:
 
 
 class ZaakTypeConfigFactory(factory.django.DjangoModelFactory):
-    urls = [factory.Faker("url")]
+
+    urls = factory.LazyFunction(lambda: [faker.Faker().url()])
```
pi-sigma, line 87: Which change made the `LazyFunction` necessary?

src/open_inwoner/openzaak/tests/factories.py
```diff
@@ -82,7 +83,8 @@ class Meta:
 
 
 class ZaakTypeConfigFactory(factory.django.DjangoModelFactory):
-    urls = [factory.Faker("url")]
+
```
pi-sigma, line 86: Empty line

pi-sigma (changes requested):

src/open_inwoner/openzaak/models.py
pi-sigma: That's fine, thanks for the explanation!

pi-sigma (approved):

PR #1354, opened 2024-08-15
Test behavior of case search with multiple results

swrichards:
This commit adds a simple test to ensure we cover the (expectedly) rare case when cases are found in multiple backends. We can't currently cover the error cases well because we silence all client errors. Ideally we'd test this without error-silencing and using VCR, but that's for a later iteration.

### DIFF
--- a/src/open_inwoner/search/tests/test_views.py
+++ b/src/open_inwoner/search/tests/test_views.py
@@ -11,7 +11,12 @@ from open_inwoner.accounts.tests.factories import DigidUserFactory
 from open_inwoner.configurations.models import SiteConfiguration
 from open_inwoner.openzaak.models import OpenZaakConfig
 from open_inwoner.openzaak.tests.factories import ZGWApiGroupConfigFactory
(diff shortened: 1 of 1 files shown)


src/open_inwoner/search/tests/test_views.py
```diff
@@ -294,3 +299,40 @@ def test_search_case_empty_query(self, m):
...
+
+        # In case of an exact identificatie match, user should be redirected to
+        # the status page of that Zaak
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(
+            response.url,
+            reverse(
+                "cases:case_detail",
+                kwargs={
+                    "object_id": "d8bbdeb7-770f-4ca9-b1ea-77b4730bf67d",
```
pi-sigma, line 334: We have a utility for extracting UUIDs from URLs to avoid hard-coding like this:

```
from open_inwoner.utils.test import uuid_from_url

"object_id": uuid_from_url(self.zaak1["url"])
```

pi-sigma (approved):
Minor suggestion regarding a hard-coded UUID, but I'll approve since we're planning to refactor these tests anyway.

PR #1356, opened 2024-08-16
Check for missing ZGW backend config in zgw_import_data command

swrichards:
Not having a ZGW backend configured is something that happens in the wild, and more broadly, it would be a good thing if the management command does not fail hard on a missing config (as now) but gracefully exits with an error message. Better reporting to users (including our Celery task wrapper for this command), and cuts down on false-positive Sentry pings.

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_zgw_imports_command.py
+++ b/src/open_inwoner/openzaak/tests/test_zgw_imports_command.py
@@ -1,5 +1,6 @@
 import inspect
 from io import StringIO
+from unittest import mock
 
 from django.core.management import call_command
 from django.test import TestCase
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/tests/test_zgw_imports_command.py
```diff
@@ -134,3 +136,21 @@ def test_zgw_import_data_command(self, m):
         ).strip()
 
         self.assertEqual(stdout, expected)
+
+
+class ZGWImportComamndWithoutConfigTest(TestCase):
```
pi-sigma, line 141:
```suggestion
class ZGWImportCommandWithoutConfigTest(TestCase):
```

pi-sigma (changes requested):

pi-sigma (approved):

alextreme (approved):

PR #1362, opened 2024-08-20
[#2684] Remove/replace display_name (roepnaam)

jiromaykin:
Taiga: 
https://taiga.maykinmedia.nl/project/open-inwoner/task/2684

See also previous issue: 
https://github.com/maykinmedia/open-inwoner/pull/1270


jiromaykin:
@pi-sigma This was much more back-end than I expected, so hopefully I didn't remove too much from the tests... 
I'm mostly replacing `display_name` with `first_name`, 
but we also used to have have a `short_name` here and there, so not sure if we should bring that back again to revert the change from #1270 .

src/open_inwoner/accounts/models.py
```diff
@@ -367,7 +360,7 @@ def get_short_name(self):
 
     @property
     def call_name(self):
-        return self.display_name or self.first_name
+        return self.first_name or self.first_name
```
pi-sigma, line 363: "return a or a" is redundant. Context: I introduced `call_name` to show either the `display_name` or the `first_name`, in case the `display_name` was not set. Now that we're getting rid of the model field `display_name`, I suggest renaming the propery to "display_name" and return the first name for now. If some municipalities decide they want to display user names in a different way, we can easily customize it:

```
@property
def display_name(self)
    return self.first_name
```
You'll need to change it in `profile/me.html` as well

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -139,7 +137,7 @@ def test_user_information_profile_page(self):
         response = self.app.get(self.url, user=self.user)
 
         self.assertContains(response, self.user.first_name)
-        self.assertContains(response, f"Welkom, {self.user.display_name}")
+        self.assertContains(response, f"Welkom, {self.user.first_name}")
```
pi-sigma, line 140: If you use `display_name` to return `first_name`, this change is unnecessary.

src/open_inwoner/components/templates/components/Header/Header.html
```diff
@@ -88,7 +88,7 @@
 
                         {% if request.user.is_authenticated %}
                             <p class="utrecht-paragraph">
-                                {% icon icon="person" icon_position="before" outlined=True %}{% trans "Ingelogd als" %} {{ request.user.display_name }}
+                                {% icon icon="person" icon_position="before" outlined=True %}{% trans "Ingelogd als" %} {{ request.user.first_name }}
```
pi-sigma, line 91: See previous comment

src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
```diff
@@ -7,7 +7,7 @@
     <nav class="primary-navigation primary-navigation--desktop primary-navigation__authenticated" aria-label="Navigatie na inloggen">
         <ul class="primary-navigation__list">
             <li class="primary-navigation__list-item">
-            {% button text=_('Welkom ')|addstr:request.user.display_name type="button" icon="expand_more" icon_position="after" icon_outlined=True transparent=True extra_classes="primary-navigation--toggle" %}
+            {% button text=_('Welkom ')|addstr:request.user.get_short_name type="button" icon="expand_more" icon_position="after" icon_outlined=True transparent=True extra_classes="primary-navigation--toggle" %}
```
pi-sigma, line 10: See previous comment.

Additionally, if you follow the suggestion about `display_name`, you can delete `get_short_name` from the User model

pi-sigma (changes requested):

jiromaykin:
@pi-sigma I think this looks good now, thanks 😊

pi-sigma (approved):

PR #1364, opened 2024-08-21
Sync user's notification preferences from klant upon login

swrichards:
Note that this PR also contains a fix for the misnamed `toestemming_zaak_notificaties_alleen_digitaal` on the `Klant` API model (was camel-cased) and some slight refactoring of the hooks for readability and composability.

It might be best to go commit-for-commit: the actual change is in the last commit.

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1,6 +1,8 @@
 import logging
 
+from open_inwoner.accounts.choices import NotificationChannelChoice
 from open_inwoner.accounts.models import User
+from open_inwoner.configurations.models import SiteConfiguration
 from open_inwoner.openklant.api_models import Klant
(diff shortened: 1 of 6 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -35,6 +39,37 @@ def update_user_from_klant(request):
...
+
+        elif (
+            klant.toestemming_zaak_notificaties_alleen_digitaal is False
+            and user.case_notification_channel
+            != NotificationChannelChoice.digital_and_post
+        ):
+            update_data[
+                "case_notification_channel"
+            ] = NotificationChannelChoice.digital_and_post
+        elif klant.toestemming_zaak_notificaties_alleen_digitaal is None:
```
pi-sigma, line 60: Could also be just `else`

pi-sigma (approved):

PR #1368, opened 2024-08-22
[#2698] Remove default Django authentication backend

pi-sigma:
- The default Django auth backend is not compatible with our practice of (a) using email as username and (b) allowing duplicate emails for DigiD users, because it checks for duplicate emails regardless of login type
- The auth backend overrides in the dev settings are redundant and potentially confusing (in case someone overrides base.py for local testing purposes)

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2698


pi-sigma:
@alextreme The PR assumes that we have no use case of DigiD/eHerkenning users attempting to log in with username + password. Can you confirm this?

alextreme:
> @alextreme The PR assumes that we have no use case of DigiD/eHerkenning users attempting to log in with username + password. Can you confirm this?

I certainly hope not. Hereby confirmed.

PR #1369, opened 2024-08-22
:bug: [#2665] Fix phone-parsing Safari (iOS only)

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2665 

The _turn-numbers-into-phone-links_ feature should be disabled with the format-detection meta-tag, but... on every Safari and iOS update this breaks again 😡; many users on new devices report the meta tag does not solve the problem, so we need to create a hack with pointer-events is zero and add a zero-width joiner `&zwj;` to make the numbers unrecognizable.
I've also added these to the Zaaknummers e-mail templates, since those are also parsed in iOS.

While I was reproducing all this in [Browserstack](https://live.browserstack.com/) I found a bug in the skip-link specific for Safari only, which is also solved here.

### DIFF
--- a/src/open_inwoner/scss/views/_view.scss
+++ b/src/open_inwoner/scss/views/_view.scss
@@ -62,4 +62,14 @@
       display: inline;
     }
   }
+
+  // Remove Safari iOS phone-parsing from specific elements
+  .phone-parsing {
+    pointer-events: none;
+  }
+  // Hide style for iOS-rendered anchor element
+  .phone-parsing > a {
+    text-decoration: none;
+    color: inherit;
+  }
 }
--- a/src/open_inwoner/scss/views/_view.scss
+++ b/src/open_inwoner/scss/views/_view.scss
@@ -64,11 +64,11 @@
   }
 
   // Remove Safari iOS phone-parsing from specific elements
-  .phone-parsing {
+  .skip-phone-parsing {
     pointer-events: none;
   }
   // Hide style for iOS-rendered anchor element
-  .phone-parsing > a {
+  .skip-phone-parsing > a {
     text-decoration: none;
     color: inherit;
   }
--- a/src/open_inwoner/static/mailcss/email.css
+++ b/src/open_inwoner/static/mailcss/email.css
@@ -260,6 +260,17 @@ strong .color--primary, strong .color--secondary {
     border-top: 3px solid #999999;
 }
 
+/* iOS specific */
+
+.phone-parsing {
+    pointer-events: none;
+}
+
+.phone-parsing > a {
+    text-decoration: none;
+    color: inherit;
+}
+
 /* CKEditor only */
 
 .cke_editable a {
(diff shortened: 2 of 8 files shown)


src/open_inwoner/scss/views/_view.scss
```diff
@@ -62,4 +62,14 @@
       display: inline;
     }
   }
+
+  // Remove Safari iOS phone-parsing from specific elements
+  .phone-parsing {
```
pi-sigma:
```suggestion
  .skip-phone-parsing {
```
or something like it so that the function of the class is clear from the name alone (otherwise you have to come hear to look at the comment)

src/open_inwoner/scss/views/_view.scss
```diff
@@ -62,4 +62,14 @@
...
+    pointer-events: none;
+  }
+  // Hide style for iOS-rendered anchor element
+  .phone-parsing > a {
```
pi-sigma:
```suggestion
  .skip-phone-parsing > a {
```

src/open_inwoner/static/mailcss/email.css
```diff
@@ -260,6 +260,17 @@ strong .color--primary, strong .color--secondary {
     border-top: 3px solid #999999;
 }
 
+/* iOS specific */
+
+.phone-parsing {
+    pointer-events: none;
+}
+
+.phone-parsing > a {
```
pi-sigma:
```suggestion
.skip-phone-parsing > a {
```

src/open_inwoner/static/mailcss/email.css
```diff
@@ -260,6 +260,17 @@ strong .color--primary, strong .color--secondary {
     border-top: 3px solid #999999;
 }
 
+/* iOS specific */
+
+.phone-parsing {
```
pi-sigma:
```suggestion
.skip-phone-parsing {
```

pi-sigma (commented):
Small suggestion for the new CSS class name, otherwise looking good. Browserstack is currently not working for me, I'll need to try again to test this.

src/open_inwoner/scss/views/_view.scss
jiromaykin: Ah, yeah, good one, will do.

jiromaykin:
> ...Browserstack is currently not working for me, I'll need to try again to test this.

@pi-sigma Actually I was not able to replicate this problem in Browserstack. Meaning: if I keep the meta-tag in the Master template as is `<meta name="format-detection" content="telephone=no">`, then it doesn't really matter which iphone or iOS version you choose in Browserstack, the formatting there is correctly turned off for any number. So the bug actually doesn't happen there, but I am guessing this is because Browserstack probably is not a full 100% native device. This is quite obscure.
But Hendrik specifically said the formatting problem comes back on his specific device (an iPhone mini with iOS 12 Safari) so for testing I took out the meta-tag and looked at what happens with the other code changes in place.

PR #1389, opened 2024-09-09
Feature/2687 selected multiselect dropdown

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2687

- [ ] `Implement the “selected" status of each filter option,
- [ ] Add the Sum of all Selected checkboxes: + “Als er in een multi-select 1 filter geselecteerd is, dan kunnen we die direct tonen.”
- [ ] “Heeft de gebruiker meer dan 1 filter geselecteerd, dan kunnen wij een label "X actieve filters" tonen.”
- [ ] Perhaps close the dropdown by clicking outside of element instead of toggle
- [ ] keep checked/selected state on refresh


src/open_inwoner/js/components/FilterBar/multiselect_listbox_checkbox.js
```diff
@@ -2,134 +2,225 @@
...
+          checkbox.checked = false
+        })
+
+        calculateAndDisplayCheckedSum() // Update the button and reset state after clearing
+
+        // Optionally, resubmit the form after resetting filters,
+        // may need to be added when submit button is removed in future
+        // const form = document.querySelector('#filterBar .form')
+        // if (form) {
+        //   form.submit()
```
jiromaykin, line 213: @pi-sigma  I kept this out for now, but see how you feel about this, once you start interacting with the component.

pi-sigma: Currently there is no way to re-select all aanvragen after filtering through the form (you have to deselect the filters and then refresh the page). So this should definitely be added.

Alternatively, always show the form button. In this case, the button also needs the correct number (currently it has "Toon 0 resultaten" if I manually change the relevant conditional in `multiselect_listbox_checkbox.js`), or the text could just be "Toon alle resultaten" if no filter is selected.

The `console.log` from below though can be deleted of course

src/open_inwoner/js/components/FilterBar/multiselect_listbox_checkbox.js
```diff
@@ -2,134 +2,225 @@
...
+      const checkboxes = document.querySelectorAll(
+        '.filter-bar .checkbox__input'
+      )
+      let sum = 0
+      let selectedFilters = []
+
+      /**
+       * Note: dynamic text from selectedFilters is inserted using textContent,
+       * while static elements like icons or spans are appended separately using createElement and appendChild.
+       * ensuring HTML injection is safe and prevents cross-site scripting vulnerability.
```
pi-sigma, line 37:
```suggestion
       * This ensures HTML injection is safe and prevents cross-site scripting vulnerability.
```
(new sentence)

pi-sigma (changes requested):

pi-sigma:
One more thing: the `x` on the filter form currently has no function, which is a bit surprising (I expect to be able to close windows, modals etc. by clicking on the `x` in the corner).

src/open_inwoner/js/components/FilterBar/multiselect_listbox_checkbox.js
```diff
@@ -2,134 +2,225 @@
...
+
+      if (selectedFilters.length === 0) {
+        // Show 'Status' text and 'expand_more' icon when no filters are selected
+        selectButton.textContent = 'Status '
+        selectButton.appendChild(expandIcon)
+        selectButton.classList.remove('active')
+      } else {
+        if (selectedFilters.length === 1) {
+          // Show only the selected filter text and close icon
+          selectButton.textContent = ''
```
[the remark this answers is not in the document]
jiromaykin: In the design the 'status' phrase for the situation when only 1 is selected, was left out intentionally - perhaps because some of them have a very long name. So I'm keeping this out of the displayed selection for now. This component has not been tested in real life and I feel like we could benefit from user feedback.

PR #1397, opened 2024-09-17
:sparkles: [#2752] Add accessible captcha

jiromaykin:
~~Adding a package for captcha that is more accessible than 'mangled' images etc.~~
https://taiga.maykinmedia.nl/project/open-inwoner/task/2752

New PR: #1417 

Screenshot 2024-09-26 at 09 54 17

~~Good to know:~~

~~1. For users with visual impairments, image-based CAPTCHAs present multiple challenges. Text within these CAPTCHAs may be distorted or overlaid on complex backgrounds, making them hard to read
2. Audio CAPTCHAs, which are meant to offer an alternative to visual CAPTCHAs, also have their issues. Often, the audio quality is poor
3. Interactive CAPTCHAs requiring precise mouse movements or quick responses can be a barrier for individuals with motor impairments
4. CAPTCHAs, which often involve complex instructions, problem-solving, deciphering ambiguous images, or solving puzzles, can impose an unnecessary cognitive load and be taxing for users with cognitive disabilities.
5. Many CAPTCHAs are not compatible with assistive technologies, such as screen readers~~


src/open_inwoner/conf/base.py
```diff
@@ -158,6 +157,7 @@
     "two_factor",
     "two_factor.plugins.webauthn",
     "maykin_2fa",
+    "simplemathcaptcha",
```
jiromaykin, line 160: Due to this change my linter reformatted the file.

src/open_inwoner/openklant/forms.py
```diff
@@ -45,6 +47,12 @@ class ContactForm(Form):
         widget=forms.Textarea(attrs={"rows": "5"}),
         required=True,
     )
+    captcha = MathCaptchaField(
+        label=_("Beantwoord deze rekensom"),
```
jiromaykin, line 51: This sentence has to be in "actief taalgebruik" and it also has to be "B1 niveau" so this is the best I could come up with.

requirements/ci.txt
```diff
@@ -447,6 +447,10 @@ django-simple-certmanager==1.4.1
     #   -r requirements/base.txt
     #   django-digid-eherkenning
     #   zgw-consumers
+django-simple-math-captcha==2.0.1
```
jiromaykin, line 450: Not sure if this is enough for adding package.

PR #1398, opened 2024-09-17
:lipstick: [#2741] Add multiselect listbox mobile design

jiromaykin:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2742

Design (multiple screens and multiple States here): https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?node-id=3196-14280&node-type=frame&m=dev


pi-sigma (approved):
The error message for 0 results after filtering ("U heeft op dit moment nog geen lopende aanvragen.") is misleading. Not part of this PR, I know, but IMHO we should consider adding a suggetion for possible actions ("reset uw filters of kom later terug" or something like this). Leaving the final call to you, though.

PR #1403, opened 2024-09-18
Increase uwsgi concurrency

alextreme:


alextreme:
Looking at sandra, I see 8 ES instances which is eating a lot of RAM. Lets hold off on this change, otherwise we'll cause ourselves a bit of a mess

PR #1407, opened 2024-09-19
[#2764] Resolve cases concurrently when loading the case list

swrichards:
[Taiga 2764](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2764).


pi-sigma (changes requested):
As discussed `CaseListService` could go in a dedicated service module, now that we're also collecting the service functions from the deleted `openzaak.cases`.

pi-sigma (approved):

PR #1410, opened 2024-09-26
[#2779] Only filter case list by open and closed status

swrichards:
[Taiga #2779](https://taiga.maykinmedia.nl/project/open-inwoner/task/2779).

### DIFF
--- a/src/open_inwoner/cms/cases/views/cases.py
+++ b/src/open_inwoner/cms/cases/views/cases.py
@@ -1,4 +1,5 @@
 import concurrent.futures
+import enum
 import logging
 from dataclasses import dataclass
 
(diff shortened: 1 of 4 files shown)


src/open_inwoner/cms/cases/views/cases.py
```diff
@@ -144,38 +156,43 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         config = OpenZaakConfig.get_solo()
         case_service = CaseListService(self.request)
-        form = None
+        context["filter_form_enabled"] = config.zaken_filter_enabled
 
         # update ctx with open submissions, cases, form for filtering
```
pi-sigma, line 161:
```suggestion
        # update ctx with open submissions and cases (possibly fitered)
```
we no longer add a form to the context

pi-sigma (commented):
Looks good, I'll take another look when the submit button is back

pi-sigma (approved):

alextreme (approved):

PR #1413, opened 2024-09-26
Disambiguate missing cache values from None

swrichards:
We have observed an issue with spurious cache hits in development, that we've been unable to reproduce so far. Looking into this did however prompt an audit of the caching decorator, which reveals we potentially have an issue where a cached "None" value is mis- interpreted as "not cached" rather than "cached as None". This commit clarifies the expected behaviour with a test and a more robust cache check.

### DIFF
--- a/src/open_inwoner/utils/tests/test_utils.py
+++ b/src/open_inwoner/utils/tests/test_utils.py
@@ -55,11 +55,11 @@ class DynamicCacheKeyTest(DjangoTestCase):
 
         self.cache.get.assert_has_calls(
             [
-                mock.call("alpha:charlie:bravo:42"),
-                mock.call("alpha:bar:bravo:baz"),
-                mock.call("alpha:bar:bravo:5"),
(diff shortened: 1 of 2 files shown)


src/open_inwoner/utils/tests/test_utils.py
```diff
@@ -201,3 +201,19 @@ def foo():
...
+    def test_returning_None_is_not_treated_as_a_cache_miss(self):
+        m = mock.Mock()
+
+        @cache("foo")
+        def returns_none():
+            m()
+            return None
+
+        # The second call should return the cached "None" from the first call,
+        # which the cache decorator should interpreted as a valid cached value,
```
pi-sigma, line 214:
```suggestion
        # which the cache decorator should interpret as a valid cached value,
```

pi-sigma (approved):
Typo in comment, otherwise looking good. Now we have to figure out the issue with wrong cache hits...

PR #1414, opened 2024-09-27
[#2782+#2780] Disable first name editing for DigiD user with BRP

pi-sigma:
Users should not be able to edit their first name in OIP if the Hall Centraal API is configured to pull data from BRP

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2782 and https://taiga.maykinmedia.nl/project/open-inwoner/issue/2780

### DIFF
--- a/src/open_inwoner/accounts/tests/test_profile_views.py
+++ b/src/open_inwoner/accounts/tests/test_profile_views.py
@@ -463,6 +463,10 @@ class EditProfileTests(AssertTimelineLogMixin, WebTest):
         response = self.app.get(self.url, user=user)
         form = response.forms["profile-edit"]
 
+        # check that first_name field is not rendered for digid_brp_user
+        with self.assertRaises(AssertionError):
+            form["first_name"] = "test"
+
         form["email"] = "user@example.com"
         form["phonenumber"] = "0612345678"
         response = form.submit()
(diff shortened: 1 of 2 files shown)


pi-sigma:
@alextreme 

There was indeed a mistake in our form + template. However, our logic is generally based on assumptions that are a bit at odds with the description in the Taiga issue:

- Our code displays/hides certain things based on whether data has been retrieved from BRP. We check for the field `is_prepopulated` on the user model, which is set by `update_brp_data_in_db`. As you can see from the second, the field is only set to `True` if the API request is successful. Otherwise, the fields remains `False` (the default) and we will still show `first_name` in the profile edit.

  https://github.com/maykinmedia/open-inwoner/blob/2835a2805c5a22946c361acde3371f40ac5f2c97/src/open_inwoner/accounts/models.py#L499

  https://github.com/maykinmedia/open-inwoner/blob/2835a2805c5a22946c361acde3371f40ac5f2c97/src/open_inwoner/haalcentraal/utils.py#L31

- The taiga issue(s) state that `first_name` should be gone as soon as the API service is configured, irrespective of whether data has been requested or if the request was successful.

The latter would require a few more changes.

alextreme (approved):

PR #1420, opened 2024-10-03
Serialize list params in an OK2 supported format

swrichards:
This is necessary to deal with https://github.com/maykinmedia/open-klant/issues/250


src/openklant2/_resources/base.py
```diff
@@ -82,14 +80,31 @@ def process_response(response: requests.Response) -> TypeGuard[JSONValue]:
 
         raise ResponseError(response, msg="Error response")
 
+    @staticmethod
+    def process_params(params: Mapping | None):
```
pi-sigma, line 84:
```suggestion
    def _process_params(params: Mapping | None) -> Mapping | None:
```

pi-sigma (approved):

PR #1421, opened 2024-10-03
[#2795] More fine-grained error handling in concurrent case fetching

swrichards:
[Taiga 2795](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2795).

These changes are necessary because we do not want a failure to resolve a single case to abort the thread for the whole API group. If one case fails, we should still try and return the cases that properly resolve.

### DIFF
--- a/src/open_inwoner/cms/cases/views/services.py
+++ b/src/open_inwoner/cms/cases/views/services.py
@@ -186,31 +186,51 @@ class CaseListService:
 
         return cases_with_api_group
 
-    def _get_cases_for_api_group(self, group: ZGWApiGroupConfig):
+    def _get_cases_for_api_group(self, group: ZGWApiGroupConfig) -> list[Zaak]:
         raw_cases = group.zaken_client.fetch_cases(
             **get_user_fetch_parameters(self.request)
         )
-        preprocessed_cases = self.resolve_cases(raw_cases, group)
-        return preprocessed_cases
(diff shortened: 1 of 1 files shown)


src/open_inwoner/cms/cases/views/services.py
```diff
@@ -186,31 +186,41 @@ def get_cases(self) -> list[ZaakWithApiGroup]:
...
     def resolve_cases(
         self,
         cases: list[Zaak],
         group: ZGWApiGroupConfig,
     ) -> list[Zaak]:
+        resolved_cases = []
         with parallel(max_workers=self._thread_limits["resolve_case_list"]) as executor:
             futures = [
                 executor.submit(self.resolve_case, case, group) for case in cases
             ]
```
pi-sigma, line 210: It could be useful to log the identity of the case that failed to resolve. To this end, we could use a `dict` comprehension instead of a `list` as `futures`:

```
futures = {
    executor.submit(self.resolve_case, case, group): {"case": case, "api_group": group} for case in cases
}
```

Then the log could be:

```
except BaseException:
    case = task["case"]
    group = task["api_group"]
    logger.exception("Error while resolving case {case} with API group {group}".format(case=case, group=group))
```

pi-sigma (approved):
Minor suggestion about logging.

pi-sigma (approved):

PR #1433, opened 2024-10-09
Sync message read statuses

swrichards:


alextreme (approved):
Issue created: https://github.com/maykinmedia/objects-api/issues/466

PR #1445, opened 2024-10-15
Enable log level filtering on the TimelineLogAdmin page

swrichards:
The log level is actually included in our timeline logger, although (mostly) the level defaults to `INFO`: that's something I'd like to start changing in future, but for now, we can actually already filter on logging level without adding a special flag, by using Postgres JSON field lookup functionality.

Screenshot from 2024-10-15 12-10-22

### DIFF
--- a/src/open_inwoner/utils/admin.py
+++ b/src/open_inwoner/utils/admin.py
@@ -4,6 +4,7 @@ import logging
 from django.contrib import admin
 from django.contrib.admin.models import ADDITION, CHANGE, DELETION
 from django.contrib.contenttypes.models import ContentType
+from django.db.models import Q
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import NoReverseMatch, path, reverse
 from django.utils.html import escape, format_html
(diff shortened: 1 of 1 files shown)


src/open_inwoner/utils/admin.py
```diff
@@ -62,6 +63,32 @@ def queryset(self, request, queryset):
...
+
+    def queryset(self, request, queryset):
+        if not (value := self.value()):
+            return queryset
+
+        if value == "unknown":
+            return queryset.filter(
+                ~Q(extra_data__has_key="log_level")
+                | Q(extra_data__log_level__isnull=True)
+            )
```
pi-sigma, line 83: Selecting `unknown` will return all logs which either (a) have the default log level `Info` (or whatever we choose in the future) because there's no custom log-level specified, or (b) have no log level at all. I wonder what the use case is for this kind of filter? Wouldn't it make more sense to _not_ filter and just show all the logs in this case?

swrichards: I don't think your assumption under (a) is right: if they actually set `INFO` those rows are excluded, and there are a few calls where we do use the log level.

The use-case is simply that, because we didn't enforce much in the past, we simply have many rows with a missing log level. I could perhaps rename this to "Empty" to reflect this.

Don't forget the option of showing everything is available as the generic Django "Show everything" link, which clears all filters.

PR #1446, opened 2024-10-15
[Venray #121] Support sorting the statusses on date set (Rx.mission, Open Zaak)

alextreme:


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -189,11 +189,14 @@ def get_context_data(self, **kwargs):
...
             )
 
-            # NOTE we cannot sort on the Status.datum_status_gezet (datetime) because eSuite
+            # NOTE we cannot always sort on the Status.datum_status_gezet (datetime) because eSuite
             # returns zeros as the time component of the datetime, so we're going with the
             # observation that on both OpenZaak and eSuite the returned list is ordered 'oldest-last'
             # here we want it 'oldest-first' so we reverse() it instead of sort()-ing
-            statuses.reverse()
+            if config.order_statusses_by_date_set:
+                statuses.sort()
```
alextreme, line 197: I'm not sure this actually works, trusting on the comment above...

src/open_inwoner/openzaak/models.py
```diff
@@ -412,6 +412,17 @@ def form_service(self, service):
         default=False,
     )
 
+    order_statusses_by_date_set = models.BooleanField(
+        verbose_name=_(
+            "On the detail page of the case, order the statusses on the date they have been set"
+        ),
+        help_text=_(
+            "If enabled, order the statusses of a case based on 'datum_status_gezet'. "
+            "If not enabled, we show the statusses in the reverse order they are returned via the API, this because the eSuite does not return the timestamps of the statusses (eSuite, but also works for Open Zaak)."
```
pi-sigma, line 421: Please break the string after 100 chars, it's annoying to read both on GH and on monitors with less than 28" or so because you can't split the screen.

pi-sigma (changes requested):
Test needed

src/open_inwoner/cms/cases/views/status.py
alextreme: Trying this out in tag venray-121-pr-1446-v2

alextreme: This updated version works as expected for HadM and Venray, thanks for the fix!

alextreme:
@pi-sigma PR has been confirmed to work for the 2 Rx.mission instances with the fix from Sidney using tag venray-121-pr-1446-v2, please resolve the remaining issues so we can include it in a regular release

pi-sigma (approved):
@swrichards should take another look as well

pi-sigma (approved):

alextreme:
thanks!

PR #1297, opened 2024-07-05
WIP - replace digid_eherkenning_oidc_generics with library code

sergei-maertens:
Doing this in a backwards compatible way :)

Questions:

* is the `error_message_mapping` really necessary? Error code + prescribed error messages are specified by OIDC standard and Logius, no?
* ...

PR #1448, opened 2024-10-16
[#2618] Replace digid_eherkenning_oidc_generics with library

pi-sigma:
After discussion we decided to replace our code with the library without copying the configuration; this has to be done manually by the client.

Cherry-picked/adapted from https://github.com/maykinmedia/open-inwoner/pull/1297

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/2617?milestone=511


alextreme:
We need to make very clear that OIDC needs to be reconfigured with this change but glad that we were able to get this done

PR #1449, opened 2024-10-16
[#2816] Split user/password auth from token verification

swrichards:
Previously, the UserModelEmailBackend was overloaded to support both username/password authentication, as well as TOTP validation. It did this, in part, by verifying the 2FA flag on the site configuration.

This was both confusing (an authentication backend should do one thing only, multiple logics means multiple backends), as well as mixing concerns (the _view_ should decide which arguments to pass to to authentication backend based on the site configuration, authentication backends should only authenticate).

This commit separates both concerns into independent backends, and adds some tests to ensure that they are properly invoked.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_backends.py
+++ b/src/open_inwoner/accounts/tests/test_backends.py
@@ -1,13 +1,18 @@
+import datetime
 from unittest.mock import patch
 
+from django.conf import settings
 from django.contrib import auth
 from django.test import RequestFactory, TestCase, override_settings
 from django.urls import reverse
 
+from freezegun import freeze_time
 from furl import furl
 from mozilla_django_oidc_db.config import store_config
+from oath import totp
 
-from .factories import UserFactory
+from open_inwoner.accounts.choices import LoginTypeChoices
+from open_inwoner.accounts.tests.factories import UserFactory
 
 
 class OIDCBackendTestCase(TestCase):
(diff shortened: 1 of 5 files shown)


src/open_inwoner/accounts/tests/test_backends.py
```diff
@@ -69,3 +75,108 @@ def test_admin_oidc_use_correct_backend(
...
+    AUTHENTICATION_BACKENDS=[
+        "open_inwoner.accounts.backends.UserModelEmailBackend",
+    ]
+)
+class UserModelEmailBackendTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        cls.password = "keepitsecert"
```
pi-sigma, line 90:
```suggestion
        cls.password = "keepitsecret"
```

src/open_inwoner/accounts/tests/test_backends.py
```diff
@@ -69,3 +75,108 @@ def test_admin_oidc_use_correct_backend(
...
+    ]
+)
+class Verify2FATokenBackendTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        super().setUpTestData()
+
+        cls.user = UserFactory(login_type=LoginTypeChoices.default)
+        cls.expires_in = getattr(settings, "ACCOUNTS_USER_TOKEN_EXPIRE_TIME", 300)
+        cls.make_token = lambda: totp(cls.user.seed, period=cls.expires_in)
```
pi-sigma, line 152:
```suggestion
        cls.token = lambda: totp(cls.user.seed, period=cls.expires_in)
```
The value for this will be the same for all tests, so it's a bit misleading to call it `make_token` (which to me suggests that a new token is made every time this is called)

pi-sigma (approved):
Looking good, couple stylistic suggestiions

alextreme:
This doesn't seem to check config.login_2fa_sms anymore? This is a PR we should discuss on monday, I may be overlooking something on how this works

alextreme:
Discussed and change is okay after tests pass

PR #1452, opened 2024-10-18
[#2821] Implement provisional configuration for OpenKlant2 service

pi-sigma:
The PR adds a provisional configuration for the OpenKlant2 service + tests so we can start integrating OpenKlant2 into our code. To be replaced with a proper configuration model.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2821


alextreme (approved):
As long as we use the Services for the actual configuration later on, this is fine for now

PR #1453, opened 2024-10-18
[#2825] Venray/Rx.mission - adding openzaak config option to use zaak.omschrijving if available instead of zaaktype.omschrijving

alextreme:


alextreme:
Thanks for the improvements Paul, appreciated!

PR #1456, opened 2024-10-21
[#2832] Rx.mission/Venray: Also show documents with the status archived

alextreme:


alextreme:
Thanks Paul for the tests, appreciated!

PR #1462, opened 2024-10-23
[#2831] Fix contactform math captcha

pi-sigma:
The `MathCaptcha` field on the contactform is replaced with a simple `IntegerField` and the captcha logic is moved to the contactform view.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2831


alextreme:
I haven't heard complaints about the captcha, but is this issue worthwhile to release a 1.22.2? If not I'll remove the label

PR #1470, opened 2024-10-24
♿  [#2625] Add mobile design for keeping accessibility functions available on small screens

jiromaykin:
issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2625

- functionality visible on desktop must remain available when zooming in 200%
- skip links are not allowed in Navs, so I moved it higher up
- when desktop users zoom in > 200% they will see the mobile view, but usually do not have typical mobile touch functions like "pinching" etc. so they still need the 'Vergroten' function

Design: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?node-id=1664-13773&m=dev

A11Y docs from NL Design-System: : https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-skip-link--docs 

Skiplink on Desktop and Mobile becomes visible as very first Tabbable focus-visible element:

Screenshot 2024-11-05 at 10 12 28


jiromaykin:
@pi-sigma @swrichards  How to review this PR:
- terminate all your NPM processes
- do a `npm ci --legacy-peer-deps` = this will retrieve the latest packages, like the updated Utrecht Components and including my [updated package with new design-tokens](https://github.com/maykinmedia/open-inwoner-design-tokens/pull/11)
- `npm run build`
- runserver
- clear browser caches
- click and see if accessibility functionality is working in the opened mobile menu, like: if Vergroten/Verkleinen and its attributes are still toggling correctly (for the Speech stuff: either use Browserstack or use official Google Chrome browsers on Linux)
- you can reach the Skiplink by reverse(Shift) Tabbing, (if it is not the first element in your visible focus anymore).
- +codereview: I had to change some of the Tests, since it's no longer needed to show a Search bar when mobile menu is opened.

swrichards:
Thanks for the review steps @jiromaykin -- very helpful!

I see the a11y buttons and they're functional (enlarge, read aloud work as expected, at least for the latter with the known caveats for Linux). I can also reach the skiplink. The styling of "Inloggen" en "Zoeken" does look slightly off to me compared to the design:

Screenshot from 2024-11-06 09-25-47

Is that expected? I went through the steps outlined above (and did another `collectstatic` for good measure).

src/open_inwoner/components/tests/test_header.py
```diff
@@ -161,23 +161,41 @@ def test_search_bar_hidden_from_anonymous_users(self):
...
 
         response = self.client.get("/")
-
         doc = PyQuery(response.content)
 
-        search_buttons = doc.find("[title='Zoeken']")
+        search_forms = [
+            form
+            for form in doc.find("form").items()
+            if form.find("button[title='Zoeken']")
```
swrichards, line 169: Nitpick I guess, but: can't we simply do something like `doc.find("form:has(> button[title='Zoeken'])")`?

INSTALL.rst
```diff
@@ -60,12 +60,6 @@ development machine.
 
        git clone git@github.com:maykinmedia/open-inwoner.git
        cd open-inwoner
-       # initialize submodules
-       git submodule update --init --recursive
-
-This will include the `Open-Inwoner-Design-Tokens`_ subdirectory. When all is built and run this is where the OIP design tokens CSS will be generated. When this repository gets updated, it needs to be pulled again.
-
-.. _Open-Inwoner-Design-Tokens: https://github.com/maykinmedia/open-inwoner-design-tokens
```
swrichards, line 68: Might not be a bad idea to still include a reference to this repo? Not in these steps indeed, but just as a note (perhaps in step 5, "Note the project makes use of the [`open-inwoner-design-tokens`](https://github.com/maykinmedia/open-inwoner-design-tokens) package, which contains the project's design tokens for the NL Design System."

src/open_inwoner/components/tests/test_header.py
jiromaykin: I did at first! But then I got the error "_cssselect.parser.SelectorSyntaxError: Expected an argument, got <DELIM '[' at 15>..._" and it seems due to `cssselect`, which doesn’t support all the same selectors as the jQuery-like PyQuery. Specifically, it seems to not support :has() with attribute selectors directly.

To get around this limitation, I tried an alternative approach: first find all forms on the page, then filter them programmatically to check if they contain a button[title='Zoeken'] with a simple CSS attribute selector that would work in all browser engines.

INSTALL.rst
jiromaykin: Oh yeah, I could, though programmatically it has the same importance as all other NPM packages needed for this project to work, so NLDS package is just another Node module now. 
This Readme contained a remnant from when we were still using some subdirectory packages.

jiromaykin:
> [...]The styling of "Inloggen" en "Zoeken" does look slightly off to me compared to the design: [...]
> Is that expected? I went through the steps outlined above (and did another `collectstatic` for good measure).

@swrichards  Oh no's that screenshot doesn't look correct at all.
If you did a proper NPM build and (perhaps multiple) browser cache reloads, it should look like this:
in your screenshot e.g. I am missing the new Home icon, FAQ icon, and also the new Close button:

a11y-close

src/open_inwoner/components/tests/test_header.py
swrichards: Shame that the selector support appears to be limited, but no worries, thanks for the explanation!

swrichards:
> > [...]The styling of "Inloggen" en "Zoeken" does look slightly off to me compared to the design: [...]
> > Is that expected? I went through the steps outlined above (and did another `collectstatic` for good measure).
> 
> @swrichards Oh no's that screenshot doesn't look correct at all. If you did a proper NPM build and (perhaps multiple) browser cache reloads, it should look like this: in your screenshot e.g. I am missing the new Home icon, FAQ icon, and also the new Close button:
> a11y-close

This is now fixed (presumably due to working from a stale branch ref on my end).

pi-sigma (approved):

PR #1472, opened 2024-10-24
[#2838] Make OpenKlant2 client compatible with zgw_consumers logic

swrichards:
[Taiga 2838](https://taiga.maykinmedia.nl/project/open-inwoner/task/2838)

### DIFF
--- a/src/open_inwoner/openklant/models.py
+++ b/src/open_inwoner/openklant/models.py
@@ -1,7 +1,9 @@
 import uuid
 from dataclasses import dataclass
+from typing import Self
 from urllib.parse import urljoin
 
+from django.core.exceptions import ImproperlyConfigured
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


src/open_inwoner/openklant/models.py
```diff
@@ -208,8 +209,12 @@ def api_url(self):
...
 
     @classmethod
-    def from_django_settings(cls):
+    def from_django_settings(cls) -> Self:
         from django.conf import settings
 
-        if config := getattr(settings, "OPENKLANT2_CONFIG", None):
-            return cls(**config)
+        if not (config := getattr(settings, "OPENKLANT2_CONFIG", None)):
+            raise RuntimeError(
```
pi-sigma, line 216: This makes branching on eSuite/OpenKlant2 a bit more tedious. We then need an additional flag to check if one or the other should be used; otherwise tests fail with this `RuntimeEror`. See my PR link below. Since this is provisional, not a big thing

https://github.com/maykinmedia/open-inwoner/pull/1474/files

pi-sigma (approved):

src/open_inwoner/openklant/models.py
pi-sigma: Yeah, `ImproperlyConfigured` sound more on target.

PR #1473, opened 2024-10-24
[#2840] Refactor contactmomenten list and detail to use a protocol for data fetching

swrichards:
[Taiga 2840](https://taiga.maykinmedia.nl/project/open-inwoner/task/2840)

This is in preparation for adding OpenKlant2 support.


src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -70,53 +76,57 @@ def handle_no_permission(self):
...
-            # eSuite extra
-            "identificatie": kcm.contactmoment.identificatie,
-            "type": kcm.contactmoment.type,
-            "status": Status.safe_label(kcm.contactmoment.status, _("Onbekend")),
-            "antwoord": kcm.contactmoment.antwoord,
-            "new_answer_available": contactmoment_has_new_answer(
-                kcm.contactmoment, local_kcm_mapping=local_kcm_mapping
-            ),
-        }
+class ContextQuestion(Question):
```
pi-sigma: What is the difference between `Question` and `ContextQuestion` (e.g. why does the latter have `identification` as field when this is set on the parent)? Perhaps this becomes more clear when looking elsewhere, but it would be nicer to make this transparent right here, either by choosing a different name or adding a docstring

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -70,53 +76,57 @@ def handle_no_permission(self):
...
-        return data
+QuestionValidator = TypeAdapter(Question)
 
-    def get_context_data(self, **kwargs):
-        ctx = super().get_context_data(**kwargs)
-        ctx["anchors"] = self.get_anchors()
-        return ctx
 
-    def get_kcm_subject(
+class MijnVragenService(Protocol):
```
pi-sigma, line 100: "MijnVragenService" might be too narrow, the service could be used elsewhere to fetch `vragen`. Perhaps just `VragenService`?

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -70,53 +76,57 @@ def handle_no_permission(self):
...
+    def retrieve_question(
+        self,
+        fetch_params: FetchParameters,
+        question_uuid: str,
+        user: User,
+    ) -> tuple[Question | None, ZaakWithApiGroup | None]:  # noqa: E704
+        ...
+
+
+class ContactMomentenVragenService(MijnVragenService):
```
pi-sigma, line 117: In my PR chose to name the services with regard to the API they are talking to, so I would do something like `eSuiteVragenService(VragenService)`. I find that clearer than `ContactMomentenVragenService`

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -33,6 +34,171 @@
...
+            with self.subTest(f"{user=} {params=} {use_rsin=}"):
+                config.use_rsin_for_innNnpId_query_parameter = use_rsin
+                config.save()
+                questions = list(self.service.list_questions(params, user))
+
+                detail_url = reverse(
+                    "cases:contactmoment_detail",
+                    kwargs={"kcm_uuid": uuid_from_url(expected_klantcontact["url"])},
+                )
+                self.assertEqual(
```
pi-sigma, line 109: Could use an additional `self.assertEqual(len(questions), 1)` to ensure that we don't have spurious stuff in there we're missing out on

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -33,6 +34,171 @@
...
+                        "identification": expected_contactmoment["identificatie"],
+                        "status": str(Status.afgehandeld.label),
+                        "case_detail_url": detail_url,
+                        "new_answer_available": False,
+                    },
+                )
+                m.reset_mock()
+                data = MockAPIReadData().install_mocks(m)
+
+    def test_retrieve_questions_returns_expected_result(self, m):
```
pi-sigma, line 128:
```suggestion
    def test_retrieve_question_returns_expected_result(self, m):
```

src/open_inwoner/templates/pages/contactmoment/detail.html
```diff
@@ -5,7 +5,7 @@
     {% if contactmoment %}
         {% render_grid %}
             {% render_column span=12 %}
-                <h1 class="utrecht-heading-1" id="title">{{ contactmoment.onderwerp }}</h1>
+                <h1 class="utrecht-heading-1" id="title">{{ contactmoment.subject }}</h1>
```
pi-sigma, line 8: Our API dataclasses have Dutch expressions for the fields (`contactmoment.identificatie`, `contactmoment.onderwerp` etc.). We should keep these consistent, so either we stick to Dutch (preference) or move the dataclasses to English as well. Unless there is some reason for and benefit from this divergence.

pi-sigma (changes requested):

pi-sigma (approved):

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -70,84 +59,53 @@ def handle_no_permission(self):
...
-        except ContactFormSubject.DoesNotExist as exc:
-            logger.warning(
-                "Could not determine OIP subject for contactmoment %s; "
-                "falling back on e-suite subject code ('onderwerp')",
-                kcm.contactmoment.url,
-                exc_info=exc,
+    @property
+    def fetch_params(self):
+        if not (
+            fetch_params := self.get_service().get_fetch_parameters(
```
pi-sigma, line 102: We end up building the service multiple times: once explicitly in `get_context_data` of the list view, then again implicitly by accessing `self.fetch_params` (lines 164, 167). Perhaps we can do the `get_service` in the `__init__` of the view to avoid this?

src/open_inwoner/openklant/tests/test_esuite_vragen_service.py
```diff
@@ -0,0 +1,179 @@
...
+from open_inwoner.openklant.constants import Status
+from open_inwoner.openklant.models import ContactFormSubject, OpenKlantConfig
+from open_inwoner.openklant.services import eSuiteVragenService
+from open_inwoner.openklant.tests.data import MockAPIReadData
+from open_inwoner.utils.test import uuid_from_url
+
+
+@requests_mock.Mocker()
+@override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
+class ContactMomentenVragenServiceTestCase(TestCase):
```
pi-sigma, line 18:
```suggestion
class eSuiteVragenServiceTestCase(TestCase):
```
In keeping with the name of the class itself (and the module).

pi-sigma (changes requested):

pi-sigma (approved):

PR #1478, opened 2024-10-30
[#2843] Refactor error handling for vragen/aanvragen list views

pi-sigma:
Instead of passing silently over errors which occurred when trying to fetch vragen/aanvragen, we re-raise the exception and display an informative message to the user.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2843


pi-sigma:
See https://github.com/maykinmedia/open-inwoner/pull/2503

PR #1481, opened 2024-10-31
[#2851] Making product admin relation to categories clearer

alextreme:


alextreme:
@pi-sigma could use a double-check on the test which is failing here, I set required=True to ensure that the label in the admin is bold

pi-sigma (approved):

alextreme:
Thanks!

PR #1483, opened 2024-11-01
[#2852] Integrate OpenKlant2 service with list view

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2852

### DIFF
--- a/src/open_inwoner/openklant/constants.py
+++ b/src/open_inwoner/openklant/constants.py
@@ -1,3 +1,5 @@
+import enum
+
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
@@ -17,3 +19,8 @@ class Status(models.TextChoices):
             if default:
                 return default
             return str(value).replace("_", " ").title()
+
+
+class KlantenServiceType(enum.Enum):
+    ESUITE = "esuite"
+    OPENKLANT2 = "openklant2"
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1,8 +1,10 @@
 import datetime
 import logging
 import uuid
+from datetime import timedelta
 from typing import Iterable, Literal, NotRequired, Protocol, Self
 
+from django.conf import settings
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
@@ -10,7 +12,7 @@ from django.utils.translation import gettext_lazy as _
 import glom
 from ape_pie.client import APIClient
 from attr import dataclass
-from pydantic import TypeAdapter
+from pydantic import BaseModel, ConfigDict, TypeAdapter
 from requests.exceptions import RequestException
 from typing_extensions import TypedDict
 from zgw_consumers.api_models.base import factory
@@ -31,7 +33,7 @@ from open_inwoner.openklant.api_models import (
     KlantCreateData,
     ObjectContactMoment,
 )
-from open_inwoner.openklant.constants import Status
+from open_inwoner.openklant.constants import KlantenServiceType, Status
 from open_inwoner.openklant.models import (
     ContactFormSubject,
     KlantContactMomentAnswer,
@@ -49,6 +51,7 @@ from open_inwoner.openzaak.clients import MultiZgwClientProxy
 from open_inwoner.openzaak.models import ZGWApiGroupConfig
 from open_inwoner.utils.api import ClientError, get_json_response
 from open_inwoner.utils.logentry import system_action
+from open_inwoner.utils.time import instance_is_new
 from openklant2.client import OpenKlant2Client
 from openklant2.types.resources.digitaal_adres import DigitaalAdres
 from openklant2.types.resources.klant_contact import (
@@ -80,15 +83,17 @@ class ZaakWithApiGroup:
 
 class Question(TypedDict):
     identification: str
+    source_url: str  # points to contactmoment or klantcontact
     subject: str
     registered_date: datetime.datetime
     question_text: str
     answer_text: str | None
     status: str
-
-    new_answer_available: bool
-    case_detail_url: str
     channel: str
+    case_detail_url: str | None = None
+
+    api_service: KlantenServiceType
+    new_answer_available: bool = False
 
 
 QuestionValidator = TypeAdapter(Question)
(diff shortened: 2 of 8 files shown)


src/open_inwoner/openklant/constants.py
```diff
@@ -17,3 +19,9 @@ def safe_label(cls, value, default=""):
             if default:
                 return default
             return str(value).replace("_", " ").title()
+
+
+@dataclass
+class KlantenServiceType:
+    esuite = "esuite"
+    openklant2 = "openklant2"
```
swrichards: Seems like an `enum` to me:

```suggestion

class KlantenServiceType(enum.Enum):
    ESUITE = "esuite"
    OPENKLANT2 = "openklant2"
```

src/open_inwoner/openklant/services.py
```diff
@@ -700,6 +708,9 @@ class OpenKlant2Answer:
     nummer: str
     plaatsgevonden_op: datetime.datetime
 
+    # metadata
+    viewed: bool = False
```
swrichards, line 712: This class was mainly meant as an API model: if `viewed` is not provided by the API, I think it might belong elsewhere, maybe on the `Question`?

src/openklant2/_resources/base.py
```diff
@@ -152,9 +152,10 @@ def _get(
         params: Mapping | None = None,
     ) -> requests.Response:
 
-        return self.http_client.request(
+        res = self.http_client.request(
             "get", path, headers=headers, params=self._process_params(params)
         )
+        return res
```
swrichards: Nitpick, but: the assignment doesn't add anything and breaks the convention of the other methods.

src/openklant2/_resources/partij.py
```diff
@@ -17,7 +17,7 @@
 
 class PartijResource(ResourceMixin):
     http_client: APIClient
-    base_path: str = "/partijen"
+    base_path: str = "partijen"
```
swrichards: Why here and not elsewhere?

src/open_inwoner/openklant/services.py
```diff
@@ -716,14 +727,21 @@ def from_klantcontact(cls, klantcontact: KlantContact) -> Self:
 
 
 @dataclass(frozen=True)
-class OpenKlant2Question:
+class OpenKlant2Question(BaseModel):
```
swrichards, line 730: Perhaps we can use [the Pydantic flavoured dataclass](https://docs.pydantic.dev/latest/concepts/dataclasses/).

src/open_inwoner/openklant/services.py
```diff
@@ -1237,3 +1243,43 @@ def questions_for_partij(self, partij_uuid: str) -> list[OpenKlant2Question]:
...
+    def list_questions(self, fetch_params: FetchParameters) -> list[Question]:
+        if bsn := fetch_params.get("user_bsn"):
+            partij = self.find_persoon_for_bsn(bsn)
+        elif kvk_or_rsin := fetch_params.get("user_kvk_or_rsin"):
+            # TODO: does rsin make sense here?
+            partij = self.find_organisatie_for_kvk(kvk_or_rsin)
+
+        questions = self.questions_for_partij(partij_uuid=partij["uuid"])
+        questions_uniform = self._reformat_questions(questions)
+        return [QuestionValidator.validate_python(q) for q in questions_uniform]
```
swrichards: I think this might be better in `_reformat_question`: it's the source where it needs to be correct, and the view also does its own validation for good measure.

src/open_inwoner/openklant/services.py
```diff
@@ -650,7 +658,7 @@ def list_questions(
             )
             for kcm in kcms
         ]
-        return contactmomenten
+        return [QuestionValidator.validate_python(q) for q in contactmomenten]
```
swrichards: As below: perhaps better in `self._get_question_data`

src/open_inwoner/openklant/tests/test_views.py
```diff
@@ -34,8 +34,50 @@
 from .factories import KlantContactMomentAnswerFactory
 
 
+class MockOpenKlant2Service:
```
swrichards, line 37: Like this pattern, very readable!

src/openklant2/_resources/partij.py
pi-sigma: Left over from testing. The api client is sensitive to trailing slashes. Changed it back.

src/open_inwoner/openklant/services.py
pi-sigma: Even with the pydantic dataclass I get:

```
PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'open_inwoner.openklant.services.OpenKlant2Answer'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
```

src/open_inwoner/openklant/services.py
pi-sigma: Sounds right. However, I think that we can't determine if a question/answer has been viewed without persisting state across sessions. For the eSuite, we store the relevant information in a dedicated model (`KlantContactMomentAnswer`). We would need something similar for the `OpenKlant2Service`. I removed the field and for now set `new_answer_available` to `False` for `OpenKlant2` questions. I suggest to discuss and revisit this next week, unless you have a solution.

swrichards: I don't think `KlantContactMomentAnswer` has any functionality specific to eSuite, it just stores a canonical URL reference to the contactmoment. We could re-use the same functionality if we can produce a similar, stable URL for OpenKlant2 answers.

What if we expand the `Question` DTO with a `api_url` field, which presents the canonical path the API resource that built it: `contactmomenten.url` and, for OpenKlant2, the `klantcontact` object that holds the question (though not the answer). Then we have a common field we can use for the `KlantContactMomentAnswer` functionality.

The other alternative would be to not use a URL in `KlantContactMomentAnswer`, but the combination of backend type and `uuid`, which would effectively do the same thing, but be a bit neater imo.

pi-sigma: Thanks for the suggestions! Second idea sounds nice, but would require a data migration to deal with existing instances, so I'm rolling with the first approach.

PR #1486, opened 2024-11-06
[#2859] Filter zaken search by visibility

pi-sigma:
Taiga https://taiga.maykinmedia.nl/project/open-inwoner/issue/2859

### DIFF
--- a/src/open_inwoner/search/tests/test_views.py
+++ b/src/open_inwoner/search/tests/test_views.py
@@ -98,12 +98,12 @@ class TestSearchView(ESMixin, TransactionTestCase):
             "schemas/Zaak",
             url=f"{ZAKEN_ROOT}zaken/e4d469b9-6666-4bdd-bf42-b53445298102",
             zaaktype=self.zaaktype["url"],
-            identificatie="ZAAK-2022-0008800002",
+            identificatie="ZAAK-2024-0000000000",
             omschrijving="Coffee zaak2",
             startdatum="2022-01-12",
             einddatum=None,
             status=f"{ZAKEN_ROOT}statussen/3da81560-c7fc-476a-ad13-beu760sle929",
-            vertrouwelijkheidaanduiding=VertrouwelijkheidsAanduidingen.openbaar,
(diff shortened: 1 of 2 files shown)


src/open_inwoner/search/tests/test_views.py
```diff
@@ -242,6 +254,16 @@ def test_search_case_identificatie_success(self, m):
...
 
+    def test_search_case_not_visible(self, m):
+        self._setUpMocks(m)
+
+        self.client.force_login(self.user)
+        params = urlencode({"query": "ZAAK-2024-0000000000"}, doseq=True)
+        response = self.client.get(f'{reverse("search:search")}?{params}')
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotIn(self.zaak2["omschrijving"], str(response.content))
```
swrichards, line 268: Should we use `zaak2["identificatie"]` for narrowing the risk of false positives? (Description changes, identificatie stays the same).

src/open_inwoner/search/tests/test_views.py
```diff
@@ -242,6 +254,16 @@ def test_search_case_identificatie_success(self, m):
...
         )
 
+    def test_search_case_not_visible(self, m):
+        self._setUpMocks(m)
+
+        self.client.force_login(self.user)
+        params = urlencode({"query": "ZAAK-2024-0000000000"}, doseq=True)
+        response = self.client.get(f'{reverse("search:search")}?{params}')
+
+        self.assertEqual(response.status_code, 200)
```
swrichards, line 266: I think here and below for the other tests: what we're really testing is not only that there is no redirect, but that the resultant url is the search results page, not the zaak detail page. Arguably that should be reflected in the assertions as well: that the current url _is_ equal to the search results page, but _is not_ equal to the case detail page.

swrichards (changes requested):

src/open_inwoner/search/tests/test_views.py
pi-sigma: Unfortunately not possible, the zaak number (the value of `zaak["identificatie"]` shows up in the html (which makes sense: it is the expression we used for searching).

swrichards (approved):

PR #1487, opened 2024-11-06
[#2825] Ensure we show case.omschrijving on the detailview if we configured to use this on the listview

alextreme:
Earlier PR worked for the listview, however on the case/status detailview zaaktype.description was always used. To be verified if this is okay...


alextreme:
Confirmed to be working as expected on the venray instance, ready for review

image

image

pi-sigma:
@alextreme ~~I updated the code. We don't have `description` directly available on the case in the detail view, so we need an additional check and fetch it.~~

My mistake. This works as expected.

pi-sigma (approved):

PR #1490, opened 2024-11-07
[#2862] Integrate OpenKlant2 service with detail views

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2862


src/open_inwoner/openklant/services.py
```diff
@@ -84,16 +84,19 @@ class ZaakWithApiGroup:
 
 class Question(TypedDict):
     identification: str
-    source_url: str  # points to contactmoment or klantcontact
+    # url: str
```
swrichards: Legacy?

src/open_inwoner/openklant/services.py
```diff
@@ -592,6 +595,8 @@ def get_kcm_answer_mapping(
     ) -> dict[str, KlantContactMomentAnswer]:
         return get_kcm_answer_mapping(contactmomenten, user)
 
+    import uuid
+
```
swrichards: Hello!

src/open_inwoner/openklant/services.py
```diff
@@ -602,21 +607,28 @@ def _get_question_data(
...
         question_data = {
-            "answer_text": kcm.contactmoment.antwoord,
             "identification": kcm.contactmoment.identificatie,
+            # "url": reverse(
+            #     "cases:contactmoment_detail",
+            #     kwargs={
+            #         "api_service": KlantenServiceType.ESUITE.value,
+            #         "kcm_uuid": kcm.uuid,
+            #     },
+            # ),
```
```suggestion

```

src/open_inwoner/openklant/services.py
```diff
@@ -602,21 +607,28 @@ def _get_question_data(
...
+            "subject": self._get_kcm_subject(kcm) or "",
+            "registered_date": kcm.contactmoment.registratiedatum,
             "question_text": kcm.contactmoment.tekst,
+            "answer_text": kcm.contactmoment.antwoord,
+            # "status": str(Status.safe_label(kcm.contactmoment.status, _("Onbekend"))),
```
```suggestion
```

src/open_inwoner/openklant/services.py
```diff
@@ -703,6 +715,50 @@ def retrieve_question(
...
+        questions = []
+        for contactmoment in contactmomenten:
+            new_answer_available = contactmoment_has_new_answer(
+                contactmoment, kcm_answer_mapping
+            )
+            questions.append(
+                Question(
+                    identification=contactmoment.identificatie,
+                    # TODO: do we need a separate field for question url? (we reverse via source_uuid)
+                    # url="",
```
swrichards: Indeed, I guess it makes sense for the views to handle that: they decide which view names they want to use, this just provides the parameters and doesn't have knowledge of how it will be used in the view layer.

src/open_inwoner/openklant/services.py
```diff
@@ -1254,40 +1304,66 @@ def list_questions(
...
+        ]
+
+    def _reformat_question(
+        self,
+        question_ok2: OpenKlant2Question,
+        user: User,
+    ) -> Question:
+        answer_metadata = KlantContactMomentAnswer.objects.get_or_create(
+            user=user, contactmoment_url=question_ok2.url
+        )
```
swrichards, line 1352: I feel a bit uncomfortable having this here in the service: this model is only needed to support logic in the view, and is thus a purely a presentation-layer concern. Would there be a way to omit the "new answer available" key from the DTO, and add that only in the view?

It seems to me we should be able to use  `KlantContactMomentAnswer` in a generic way (e.g. it only needs a url and user, both of which are available in the view without having to know much more).

src/open_inwoner/openklant/services.py
```diff
@@ -1254,40 +1304,66 @@ def list_questions(
...
-                "subject": q.onderwerp,
-                "registered_date": q.plaatsgevonden_op,
-                "question_text": q.question,
-                "answer_text": getattr(q.answer, "answer", None),
+        return [
+            self._reformat_question(questions_ok2, user=user)
+            for question in questions_ok2
+        ]
+
+    def _reformat_question(
```
swrichards: Here and above, I'd prefer a name such as `_build_question_dto` (we could even, if we don't have to use `KlantContactMomentAnswer` here, make this a method on `OpenKlant2Question`, which might be even neater, so you get:

```
question_ok2.to_question_dto()
```

src/open_inwoner/openklant/tests/factories.py
```diff
@@ -13,8 +15,27 @@ class Meta:
     subject_code = factory.Faker("word")
 
 
-def make_contactmoment(contact_moment_data: dict):
-    return zgw_factory(ContactMoment, contact_moment_data)
+def make_question_from_contactmoment(
+    contact_moment_data: dict,
+    case_detail_url: str | None = None,
+    new_answer_available: bool = False,
+) -> dict:
```
swrichards: We could type this as a `Question` and validate here as well.

src/open_inwoner/templates/pages/questions/questions.html
swrichards, line 1: As discussed: perhaps move these to `src/open_inwoner/templates/pages/partials`

src/open_inwoner/utils/url.py
```diff
@@ -58,3 +60,11 @@ def prepend_next_url_param(url: str, next_url: str) -> str:
...
     u.args.set("next", next_url)
     return u.url
+
+
+def uuid_from_url(url: str) -> str:
+    """
+    Extract UUID string from `url`, raise `ValueError` via UUID lib for invalid uuids
+    """
+    uuid_str = url.rstrip("/").split("/")[-1]
+    return uuid.UUID(uuid_str)
```
swrichards: This seems to assume a URL has a very specific shape: `http://foo/bar/{uuid}/` but might fail for  `http://foo/bar/{uuid}/bar-detail`.

I think a full regex would be better here, to just match the UUID directly and perhaps raise if more than one UUID is found to avoid subtle bugs where we match the wrong uuid, as we might in nested resources. (Alternatively, return all found UUIDs and let the caller handle the >1 case).

src/open_inwoner/openklant/services.py
```diff
@@ -1254,40 +1304,66 @@ def list_questions(
...
         return self._reformat_questions(questions, user)
 
-    # TODO: handle `status` + `new_answer_available`
-    # `case_detail_url`: will be handled in integration of detail view
-    # `status`: eSuite has three: "nieuw", "in behandeling", "afgehandeld"
+    def retrieve_question(
+        self,
+        fetch_params: FetchParameters,
+        question_uuid: str,
+        user: User,  # is this needed?
```
swrichards: I think you're right that we don't: we just use the fetch params. I vaguely recall we might have needed it for cases to check the confidentiality level or some such, but we can always reintroduce it (and then just pass that directly, rather than the user). Happy to remove this from the protocol.

src/open_inwoner/openklant/tests/test_esuite_vragen_service.py
```diff
@@ -81,26 +80,31 @@ def test_list_questions_returns_expected_rows(self, m):
...
-                    kwargs={"kcm_uuid": uuid_from_url(expected_klantcontact["url"])},
-                )
+                # TODO: do we need detail_url?
+                # detail_url = reverse(
+                #     "cases:contactmoment_detail",
+                #     kwargs={
+                #         "api_service": KlantenServiceType.ESUITE.value,
+                #         "kcm_uuid": uuid_from_url(expected_klantcontact["url"]),
+                #     },
+                # )
```
swrichards: As per other comment, here and elsewhere: no indeed, let the view build the URLs, it's not a service level concern. The service should just provide the identifiers.

swrichards (changes requested):
Looks good, big push here. Mostly small'ish comments.

src/open_inwoner/openklant/services.py
pi-sigma: Indeed

src/open_inwoner/utils/url.py
pi-sigma: I added a parameter to allow multiple matches which defaults to `False` and raises `ValueError`. Not 100% sure what to do if no match is found. Raise or return `None`? I went for the second option, but could go the other way as well.

src/open_inwoner/openklant/services.py
pi-sigma: On the other hand, the eSuite service needs the `User` to fetch the relevant data. Although `OpenKlant2Service` doesn't need it for `retrieve_question`, we would need to change the logic and distinguish between the two services whenever we retrieve a single question. Though perhaps that's the way to go if the methods are different with regard to (not) requiring the `User` as argument?

src/open_inwoner/utils/url.py
```diff
@@ -58,3 +61,23 @@ def prepend_next_url_param(url: str, next_url: str) -> str:
...
     return u.url
+
+
+def uuid_from_url(
+    url: str, allow_multiple: bool = False
+) -> uuid.UUID | list[uuid.UUID] | None:
+    """Extract UUID(s) from a given URL"""
+
+    pattern = r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
+    matches = re.findall(pattern, url, re.IGNORECASE)
```
swrichards: Given that this is likely to be called in loops, we could pre-compile:

```suggestion

_UUID_PATTERN = re.compile(r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}")

def uuid_from_url(
    url: str, allow_multiple: bool = False
) -> uuid.UUID | list[uuid.UUID] | None:
    """Extract UUID(s) from a given URL"""

    matches = _UUID_PATTERN.findall(pattern, url, re.IGNORECASE)
```

src/open_inwoner/utils/url.py
```diff
@@ -58,3 +61,23 @@ def prepend_next_url_param(url: str, next_url: str) -> str:
...
+
+    if not matches:
+        return None
+    if len(matches) > 1 and allow_multiple:
+        return [uuid.UUID(match) for match in matches]
+    elif len(matches) > 1:
+        raise ValueError(
+            "url contains more than 1 UUID (use allow_multiple=True if you expect multiple results)"
+        )
```
swrichards: The cases seem a bit confusing to me, as both branches are special cases if `len(matches) > 1`:

```suggestion
    if len(matches) > 1:
        if allow_multiple:
          return [uuid.UUID(match) for match in matches]

        raise ValueError(
            "url contains more than 1 UUID (use allow_multiple=True if you expect multiple results)"
        )
```

src/open_inwoner/utils/url.py
swrichards: Good solution, just a small nitpick about the conditional.

src/open_inwoner/openklant/services.py
swrichards: I think its fine to provide it as context to both, I'd say the general approach is maximalist: we provide as much context as required by the most demanding implementation.

swrichards (changes requested):
LGTM, just one or two nitpicks, and the outstanding issue about integrating the seen status into the service.

src/open_inwoner/openklant/services.py
pi-sigma: Ticket: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2901

swrichards (approved):

alextreme:
Can be merged after rebase from Paul and final check

swrichards (approved):

PR #1491, opened 2024-11-11
[#2757] Add home view styling for registratiemelding notifications

jiromaykin:
userstory:  https://taiga.maykinmedia.nl/project/open-inwoner/us/2475

Figma design: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?t=h3bwJSSWV49cLKFB-0

Testing: 
Login and go to URL http://localhost:8000/accounts/register/ to see Notification

IF notification exists, margins and padding will change on Home page only: so the Welcome message drops lower and becomes readable again, also there is a difference between the margins on mobile and desktop:

home-margins

When the user closes the notification, 
styling should jump back to normal.

Screenshot 2024-11-14 at 13 50 26

### DIFF
--- a/src/open_inwoner/js/components/index.js
+++ b/src/open_inwoner/js/components/index.js
@@ -36,6 +36,7 @@ import './siteimprove/tracking'
 import './twofactor-sms'
 import { FileInput } from './form/FileInput'
 import { ToggleHide } from './card/ToggleHide'
+import './views'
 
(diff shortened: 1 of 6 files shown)


pi-sigma (approved):

alextreme:
Piet approves

PR #1494, opened 2024-11-13
Improve error handling for use of klanten services

pi-sigma:


src/open_inwoner/openklant/services.py
```diff
@@ -770,7 +773,9 @@ class OpenKlant2Service(KlantenService):
     def __init__(self, config: OpenKlant2Config | None = None):
         self.config = config or OpenKlant2Config.from_django_settings()
         if not self.config:
-            raise RuntimeError("OpenKlant2Service instance needs a configuration")
+            raise ImproperlyConfigured(
```
swrichards, line 776: Agree with using this exception here: in fact, I also think we can remove the lines at `l.786`:

```
  if not self.client:
            raise RuntimeError("OpenKlant2Service instance needs a client")
```

There is really no reason we shouldn't be able to build the client, it's a simple constructor which doesn't do any complex validation.

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -89,11 +90,21 @@ def get_fetch_parameters(
...
+            except RuntimeError:
+                logger.error("Failed to build eSuiteVragenService")
         elif service_type == KlantenServiceType.OPENKLANT2:
-            return OpenKlant2Service()
+            try:
+                return OpenKlant2Service()
+            except ImproperlyConfigured:
+                logger.error("OpenKlant2 configuration missing")
+            except RuntimeError:
+                logger.error("Failed to build OpenKlant2Service")
```
swrichards, line 105: See below: I think we can probably do without this.

src/open_inwoner/accounts/signals.py
```diff
@@ -37,27 +36,26 @@ def update_user_from_klant_on_login(sender, user, request, *args, **kwargs):
...
+        logger.error("OpenKlant2 configuration missing")
+    except RuntimeError:
+        logger.error("Failed to build OpenKlant2Service")
+    else:
+        if fetch_params := service.get_fetch_parameters(request=request):
+            partij, created = service.get_or_create_partij_for_user(
+                fetch_params=fetch_params, user=user
+            )
+            if partij and not created:
+                service.update_user_from_partij(partij_uuid=partij.uuid, user=user)
```
swrichards: I find the `else` a bit hard to read here as the flow is not exceptional, couldn't we just continue the code block after initializing the service? (Alternative we could try to initialize both services at the top, and move the actual logic into separate functions, so the signal handler itself really is just a controller).

src/open_inwoner/openklant/models.py
```diff
@@ -213,13 +212,11 @@ def api_url(self):
...
     @classmethod
-    def from_django_settings(cls) -> Self:
+    def from_django_settings(cls) -> Self | None:
         from django.conf import settings
 
         if not (config := getattr(settings, "OPENKLANT2_CONFIG", None)):
-            raise ImproperlyConfigured(
-                "Please set OPENKLANT2_CONFIG in your settings to configure OpenKlant2"
-            )
+            return None
```
swrichards: I think it's still clearer to keep this: it's the core purpose of the method, and if the settings are wrong, we should raise here and the caller should handle it (in this case, `OpenKlant2Service` could just let it bubble up I think without catching it explicitly).

swrichards (changes requested):
Nice, thanks, just small things.

src/open_inwoner/openklant/models.py
pi-sigma: I was thinking about bringing the `OpenKlant2Service` in line with the `eSuiteService` (the config for the latter does not raise an exception, it's the job of the service). However, it's probably better to leave as is and rethink this when we swap out the config-via-settings with a proper config model

src/open_inwoner/accounts/signals.py
pi-sigma: Can't skip the `else` block for the first service because we wouldn't be able to do `service.get_fetch_parameters`. I did export the code in the `else` block to a separate function.

src/open_inwoner/openklant/models.py
swrichards: Sounds good.

src/open_inwoner/openklant/models.py
```diff
@@ -213,7 +213,7 @@ def api_url(self):
         return urlunparse((scheme, netloc, path, params, query, fragment))
 
     @classmethod
-    def from_django_settings(cls) -> Self:
+    def from_django_settings(cls) -> Self | None:
```
```suggestion
    def from_django_settings(cls) -> Self:
```

swrichards (changes requested):
One more nitpick

swrichards (approved):

PR #1496, opened 2024-11-14
[#2755] Expand speechsynthesis

jiromaykin:
issues: https://taiga.maykinmedia.nl/project/open-inwoner/task/2755

- [x] remove asterisk from Mijn Uitkeringen Select input
- [x] add Headings (was already being read, no change)
- [x] HTMX elements (was already being read, no change)
- [x] add Waarschuwingsbanner to speech loop
- [x] remove Select contents from speech / only read an actual selected input inside a Select

### DIFF
--- a/src/open_inwoner/js/components/accessibility/read.js
+++ b/src/open_inwoner/js/components/accessibility/read.js
@@ -58,8 +58,34 @@ export class ReadOut {
         main = document.querySelector('main')
       }
 
+      // Select the warning header element and get its text content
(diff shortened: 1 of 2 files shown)


pi-sigma (approved):
Two `consoloe.log`s from debugging (?) that should be removed, otherwise looking good.

jiromaykin:
> Two `consoloe.log`s from debugging (?) that should be removed

None of those console.logs are active since they are all commented out, and they were put in there by Alex, not by me. They are nice to have so I kept them in.

PR #1497, opened 2024-11-14
[#2871] Refactor ZGW imports to work on different environments

pi-sigma:
- [x] Use composite natural key for `CatalogusConfig` and a combination of `omschrijving` + `zaaktype_identificatie` for nested ZGW objects like `ZaakTypeStatusTypeConfig` to support import/export of ZGW objects across different environments (where the url is different)
 - [x] Update ZGW import/export logic to work with new natural keys
 - [x] Update success and error messages displayed in the admin

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2871


src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -69,10 +69,10 @@ def import_catalog_configs() -> list[CatalogusConfig]:
     create = []
 
     with transaction.atomic():
-        known = set(CatalogusConfig.objects.values_list("url", flat=True))
+        known = set(CatalogusConfig.objects.values_list("domein", "rsin"))
         for response in result:
             for catalog in response.result:
-                if catalog.url in known:
+                if (catalog.domein, catalog.rsin) in known:
                     continue
```
swrichards: As discussed, I don't think there is a need to change these: the actual API import _should_ use the URL, UUID, etc.

src/open_inwoner/openzaak/tests/test_zgw_imports_command.py
```diff
@@ -35,6 +35,8 @@ def setUpTestData(cls):
```
swrichards: These changes might now be redundant too.

src/open_inwoner/openzaak/import_export.py
```diff
@@ -22,6 +25,173 @@
...
+            items = [
+                f"Identificatie = {fields['identificatie']}",
+                f"Catalogus domein = {fields['catalogus'][0]}",
+                f"Catalogus rsin = {fields['catalogus'][1]}",
+            ]
+        if source_config in [
+            ZaakTypeStatusTypeConfig,
+            ZaakTypeResultaatTypeConfig,
+            ZaakTypeInformatieObjectTypeConfig,
+        ]:
```
swrichards:
```suggestion
        if source_config in {
            ZaakTypeStatusTypeConfig,
            ZaakTypeResultaatTypeConfig,
            ZaakTypeInformatieObjectTypeConfig,
        }:
```

src/open_inwoner/openzaak/import_export.py
```diff
@@ -22,6 +25,173 @@
...
+    else:
+        exclude_fields = [
+            "id",
+            "catalogus",
+        ]
+        for field in source_config._meta.fields:
+            field_name = field.name
+
+            if field_name in exclude_fields:
+                pass
```
swrichards:
```suggestion
            if field_name in exclude_fields:
                continue
```

src/open_inwoner/openzaak/import_export.py
```diff
@@ -22,6 +25,173 @@
...
+
+            val = getattr(source_config, field_name, None)
+            setattr(target, field_name, val)
+            target.save()
```
swrichards: This could perhaps itself be a helper method, given we use it a few times?

src/open_inwoner/openzaak/import_export.py
```diff
@@ -22,6 +25,173 @@
...
+        target = source_config.__class__.objects.get_by_natural_key(
+            omschrijving=source_config.omschrijving,
+            zaaktype_config_identificatie=zaaktype_config_identificatie,
+            catalogus_domein=catalogus_domein,
+            catalogus_rsin=catalogus_rsin,
+        )
+    except (source_config.MultipleObjectsReturned) as exc:
+        raise ZGWImportError.from_exception_and_jsonl(exc, jsonl)
+    except source_config.DoesNotExist as exc:
+        raise ZGWImportError.from_exception_and_jsonl(exc, jsonl)
```
swrichards:
```suggestion
    except (source_config.MultipleObjectsReturned, source_config.DoesNotExis) as exc:
        raise ZGWImportError.from_exception_and_jsonl(exc, jsonl)
```

src/open_inwoner/openzaak/import_export.py
swrichards: I generally feel there is a fair bit of duplication in these functions. I wonder whether we couldn't simply inline all of these functions in the `match` statement of the main loop, whereby each branch would define the target and the exclude fields, and then the loop would only do the updating logic once.

swrichards: (I guess the related object handling is slightly different, no worries if it's too much hassle, but it might make sense to mark the functions as private with the `_` prefix).

src/open_inwoner/openzaak/admin.py
```diff
@@ -174,11 +175,26 @@ def process_file_view(self, request):
                     self.message_user(
                         request,
                         _(
-                            "Successfully processed %(num_rows)d items"
+                            "%(num_rows)d item(s) processed in total. Any rows that could not be "
+                            "imported are reported below."
```
swrichards:
```suggestion
                            "%(num_rows)d item(s) processed in total, with %(erroring_rows)d failing rows.
```

swrichards: We could then perhaps vary `messages.SUCCESS` to be a warning if there are `>0` failing rows, so it's immediately obvious that something went _kind of_ ok, but not completely.

src/open_inwoner/openzaak/admin.py
```diff
@@ -174,11 +175,26 @@ def process_file_view(self, request):
...
                     )
+                    if errors := import_result.import_errors:
+                        msgs_deduped = set(error.text for error in errors)
+                        error_msg_iterator = ([msg] for msg in msgs_deduped)
+
+                        error_msg_html = format_html_join(
+                            "\n", "<li>{}</li>", error_msg_iterator
+                        )
+                        error_msg_html_ordered = format_html(
+                            f"<ol> {error_msg_html} </ol>"
```
swrichards:
```suggestion
                            _("It was not possible to import the following items:") + f"<ol> {error_msg_html} </ol>"
```

src/open_inwoner/openzaak/import_export.py
```diff
@@ -22,6 +25,173 @@
...
+            "%(source_config_name)s not found in target environment: %(info)s"
+        )
+        if error_data["error_type"] is MultipleObjectsReturned:
+            error_template = "Got multiple results for %(source_config_name)s: %(info)s"
+
+        return cls(error_template % error_data)
+
+    @property
+    def text(self):
+        return self.__str__()
```
swrichards: Not sure the `text` property adds much over the default `__str__` that's already implemented.

src/open_inwoner/openzaak/tests/test_import_export.py
```diff
@@ -261,252 +263,260 @@ def test_export_catalogus_configs_as_jsonl(self):
...
+            zaaktype_informatie_objecttype_config.document_upload_enabled, True
+        )
+        self.assertEqual(
+            zaaktype_informatie_objecttype_config.document_notification_enabled, True
         )
 
-    def test_bad_import_types(self):
-        for bad_type in (set(), list(), b""):
-            with self.assertRaises(ValueError):
-                CatalogusConfigImport.from_jsonl_stream_or_string(bad_type)
```
swrichards, line 332: I think this test is still valid?

src/open_inwoner/openzaak/tests/test_import_export.py
```diff
@@ -261,252 +263,260 @@ def test_export_catalogus_configs_as_jsonl(self):
...
-                    import_result,
-                    CatalogusConfigImport(
-                        total_rows_processed=10,
-                        catalogus_configs_imported=2,
-                        zaaktype_configs_imported=2,
-                        zaak_inormatie_object_type_configs_imported=2,
-                        zaak_status_type_configs_imported=2,
-                        zaak_resultaat_type_configs_imported=2,
-                    ),
-                )
```
swrichards, line 352: Technically still a valid test, though we may want to assert that all the returned imports are the same.

src/open_inwoner/openzaak/tests/test_import_export.py
```diff
@@ -261,252 +263,260 @@ def test_export_catalogus_configs_as_jsonl(self):
...
+        self.assertEqual(ZaakTypeConfig.objects.count(), 1)
+        self.assertEqual(ZaakTypeInformatieObjectTypeConfig.objects.count(), 1)
+        self.assertEqual(ZaakTypeStatusTypeConfig.objects.count(), 1)
+        self.assertEqual(ZaakTypeResultaatTypeConfig.objects.count(), 1)
 
-    def test_rewrite_target_diverges_from_existing_objects(self):
-        self._create_fixtures("http://one.maykinmedia.nl/")
-        self.service.api_root = "http://two.maykinmedia.nl"
-        self.service.save()
+    def test_import_jsonl_update_report_duplicates(self):
```
swrichards:
```suggestion
    def test_import_jsonl_update_reports_duplicates(self):
```

src/open_inwoner/openzaak/tests/test_import_export.py
```diff
@@ -261,252 +263,260 @@ def test_export_catalogus_configs_as_jsonl(self):
...
+    def test_import_is_atomic(self):
+        bad_line = '{"model": "openzaak.zaaktyperesultaattypeconfig", "fields": {}}\n'
+        bad_jsonl = self.jsonl + "\n" + bad_line
+
+        try:
+            CatalogusConfigImport.from_jsonl_stream_or_string(
+                stream_or_string=bad_jsonl
+            )
+        except Exception:
+            pass
```
swrichards: maybe a `self.assertRaises` to actualy ensure we got an error.

src/open_inwoner/openzaak/tests/test_import_export.py
swrichards, line 1: As discussed, I think we need a test that validates a source and target scenario with (1) corresponding natural keys (e.g. all objects are found in the target) but (2) _differing_ values for URLs, zaaktype UUIDs, etc.,, and that the (3) latter values are not amended in the update (because those will point to the targets ZGW backend and should remain the same).

swrichards (changes requested):
Nice work, few suggestions here and there.

src/open_inwoner/openzaak/import_export.py
pi-sigma: I moved the `exclude_fields` into the match statement, which makes the helper functions slimmer. I'm not sure more inlining would improve readability further, due to the difference between `CatalogusConfig`, `ZaaktypeConfig`, and the nested configs (the match statement would get pretty crowded).

swrichards (approved):

PR #1498, opened 2024-11-18
:chart_with_upwards_trend: [#2870] Add 'Mijn aanvragen' Siteimprove tracking (static) 

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/2870

- [x] hoeveel mensen naar Mijn Aanvragen via menu gaan 
- [x] hoeveel mensen naar een specifieke aanvraag gaan door op tegel van de aanvraag te klikken 
- [x] hoeveel mensen naar een specifieke aanvraag gaan door op de tekst 'Bekijk aanvraag' te klikken 
- [x] hoe vaak een document word gedownload 
- [x] hoe vaak mensen een document (proberen) te uploaden 
- [x] hoe vaak mensen klikken op de knop 'scroll omlaag'
- [x] Ik wil weten hoevaak mensen al voltooide statussen terugkijken dmv het uitklappen van een status. Hierbij zijn er verschillende statussen mogelijk
- [ ] ~~Ik wil weten hoeveel mensen een verkeerd bestand uploaden en daardoor een error krijgen. Dit uitgesplitst in een error van een te groot bestand en een error van een verkeerd bestand~~
- [x] Ik wil weten hoe vaak mensen een geupload bestand verwijderen

The tracking of dynamic errors are more complicated and will be tackled separately in a different issue (#1563 ).

### DIFF
--- a/src/open_inwoner/js/components/siteimprove/tracking.js
+++ b/src/open_inwoner/js/components/siteimprove/tracking.js
@@ -1,10 +1,11 @@
-// Mock _sz object for testing
+// // Mock _sz object for testing
 // if (typeof _sz === 'undefined') {
 //   var _sz = {
 //     push: function (data) {
 //       try {
 //         console.log('Event pushed to _sz:', data)
 //       } catch (error) {
+//         // Log the error
 //         console.error('Error occurred while pushing event data:', error)
 //       }
 //     },
@@ -326,7 +327,7 @@ let isEventTrackerInitialized = false
         '#cases-content > div.card__grid > div > div > a > div > span > span.link__text':
           [
             'event',
-            'Mijn aanvragen overzicht',
+            'Mijn aanvragen',
             'Click',
             "Open aanvraag via 'Bekijk aanvraag' link",
           ],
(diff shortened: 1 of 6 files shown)


jiromaykin:
@pi-sigma @swrichards We do not have a Siteimprove test environment, so the only way to 'test' this code, is by locally uncommenting the `_sz` mock object in top of the JS code.
(Do not commit this change...).
The rather insane looking event tracker contains a plethora of components and elements for which I cannot write a simple general tracker because Groningen needs to have very specifically named 'categories' and unique 'labels' to appear in their Siteimprove Dashboard.

Here is how a Siteimprove 'onClick' push event is constructed, and why it has consequences for their dashboard: https://help.siteimprove.com/support/solutions/articles/80000863895-getting-started-with-event-tracking#Setting-up-event-tracking

The Siteimprove dashboard looks like this (screenshot in Anneke's comment from _15 Jan 2025_): https://taiga.maykinmedia.nl/project/open-inwoner/us/2701

I have uploaded a new spreadsheet in Taiga (https://taiga.maykinmedia.nl/project/open-inwoner/task/2870) that shows which elements should be tracked, so if you click on any of those, you should be able to see a console log that indicates an event is pushed to the non-existing `_sz` object.

Or you could just do a review on code quality.

pi-sigma (commented):
When I search for something, e.g. "yolo", I get the intended log message: `Event pushed to _sz: Array(5) [ "event", "Header", "Zoeken", "Enter click", "yolo" ]`. So the mock `_sz` object works. However, I don't see any logs when navigating to `Mijn Aanvragen` via the menu.

jiromaykin:
> When I search for something, e.g. "yolo", I get the intended log message: `Event pushed to _sz: Array(5) [ "event", "Header", "Zoeken", "Enter click", "yolo" ]`. So the mock `_sz` object works. However, I don't see any logs when navigating to `Mijn Aanvragen` via the menu.

That's odd @pi-sigma , but maybe your logs are cleared when you're (re)loading the page.
In Firefox Devtools you need to check the "_persist logs_" checkbox (in Chrome it's "_preserve log_") so you can still see all of the logs that were spat out.
And then, between all the other logs, you should see something like `Event pushed to _sz: 
Array(5) [ "event", "Header", "Click", "Open Mijn aanvragen", "Mijn aanvragen" ]`

pi-sigma:
Still no luck, same result (tried Firefox and Chrome with the preserver/persist logs option enabled). @swrichards does this work on your end?

swrichards:
> Still no luck, same result (tried Firefox and Chrome with the preserver/persist logs option enabled). @swrichards does this work on your end?

Same, I can see the events for e.g. entering a search query, but not for menu page navigations. See this dump (with persist logs), the click to "Mijn Vragen" is a few lines in, and I ended by typing something in the search box to show the logged event:

Screenshot from 2025-01-20 14-54-57

src/open_inwoner/js/components/siteimprove/tracking.js
```diff
@@ -192,133 +194,331 @@ let isEventTrackerInitialized = false
...
+          ],
+        '.header > div > div.header--mobile.header__submenu > nav > ul > li > a[aria-label="Mijn profiel"] .material-icons-outlined':
+          [
+            'event',
+            'Mijn Profiel',
+            'Click mijn profiel ',
+            'Open Mijn profiel mobiel',
+          ],
+        // Desktop authenticated menu
+        '.header > div > nav.primary-navigation.primary-navigation--desktop.primary-navigation__authenticated > ul > li > ul > li > a[aria-label="Mijn profiel"]':
```
swrichards: @jiromaykin as far as I can tell, this selector does not select the menu items for me. I do not that the `aria-label` on the final `A` is `Mijn Profiel` (note the casing).

More generally I was wondering: would it be an idea to move towards more targeted selectors? These feel quite fragile to me, they seem easy to break in a refactor (or even in an i18n change, as perhaps happened with the label), which wouldn't I think be covered by tests. We could maybe do something like:

```
<a href="/mijn-profiel/" class="link link--icon link--icon-position-before" aria-label="Mijn Profiel" title="Mijn Profiel" data-sz-id="mijn-profiel-link">
    <span class="link__text">Mijn Profiel</span>
    <span aria-hidden="true" class="material-icons-outlined ">apps</span>
</a>
```

And then use the `data-sz-id` selector to more directly target the element, which hopefully in would be more resistant to other changes, as well as being a clearer marker during refactors of the HTML (and fewer elements to bind handlers to, which helps performance).

jiromaykin: @pi-sigma @swrichards Ah, I see.... I did not know people would use different menu titles (which they can indeed change in the CMS-admin as well) so it should work but only if wording/casing is 100% the same.
i can easily change it into a case-insensitive thing.

The data-attributes idea seems like a great idea, but we have a CMS menu in **open_inwoner/templates/cms/menu/primary.html** with link components that optionally already have a data-attribute (like so: `"<li class="primary-navigation__list-item"> {% link text=child.get_menu_title href=url data_text=child.get_menu_title data_alt_text=child.get_menu_title data_icon=child.common.menu_icon data_alt_icon=child.common.menu_icon icon=child.common.menu_icon icon_outlined=True icon_position="before" %}"` which does not solve the javascript 'event bubbling' problem for all individual child elements, which could mean I'll have to add more listeners anyway. 
Fragility will be less, but slightly unavoidable due to how specific Groningen needs the Siteimprove Dashboard to look like.

So I'm going to have to do quite some experimenting with this, and also not sure yet if this will make the DOM noticeably slower - since we have to use multiple 'pollers/MutationObservers', but it will probably be okay.

swrichards: > that optionally already have a data-attribute 

Quick question just on this point: why would that be an issue? HTML elements can validly have multiple `data-*` tags right?

jiromaykin: Quick answer: no, that's not the problem, that's why, as you can see, they already have optional data attributes in the template tags. It's what you do with them that could be a potential problem. I meant: separate problem here is that data-attributes would not solve the fragility 100%.

swrichards: Understood, thanks!

swrichards (approved):

PR #1500, opened 2024-11-19
Implement frontchannel OIDC logout flow

swrichards:
[Taiga 2818](https://taiga.maykinmedia.nl/project/open-inwoner/us/2818).

### DIFF
--- a/src/open_inwoner/accounts/tests/test_oidc_views.py
+++ b/src/open_inwoner/accounts/tests/test_oidc_views.py
@@ -1,7 +1,9 @@
 from hashlib import md5
 from typing import Literal
 from unittest.mock import patch
+from urllib.parse import urlencode
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import ValidationError
 from django.test import TestCase, modify_settings, override_settings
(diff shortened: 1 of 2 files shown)


src/open_inwoner/accounts/views/auth_oidc.py
```diff
@@ -97,16 +98,34 @@ def get_success_url(self):
...
+        id_token = request.session.get("oidc_id_token")
         if "oidc_login_next" in request.session:
             del request.session["oidc_login_next"]
 
+        # Always destroy our session, having obtained the OIDC artifacts from the session
         auth.logout(request)
 
+        # Try to initiate a frontchannel redirect
+        if id_token:
+            if logout_endpoint:
```
pi-sigma:
```suggestion
            if id_token and logout_endpoint:
```

swrichards: We need the separate conditionals because we only want to trigger `do_op_logout` if there's no logout endpoint: refactored to make that clearer.

alextreme:
@swrichards can we set this PR to ready-for-review?

swrichards:
> @swrichards can we set this PR to ready-for-review?

Not yet, it needs:

1. Updated tests
2. We should run another small experiment to see if this works using `POST` entirely in the backend, where we follow the redirect chain. Technically this could be done separately from this PR though.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -643,20 +645,64 @@ def test_logout(self, mock_get_solo):
...
+                    id_token_hint="foo",
+                    post_logout_redirect_uri=f"http://testserver{settings.LOGOUT_REDIRECT_URL}",
+                )
+            ),
+            fetch_redirect_response=False,
+        )
+
+        self.assertNotIn("oidc_states", self.client.session)
+        self.assertNotIn("oidc_id_token", self.client.session)
+        self.assertFalse(logout_response.wsgi_request.user.is_authenticated)
```
pi-sigma, line 665: Here and below: why assert about `wsgi_request.user` as opposed to `user` (defined earlier in the function?)

src/open_inwoner/accounts/views/auth_oidc.py
```diff
@@ -97,16 +97,34 @@ def get_success_url(self):
...
             del request.session["oidc_login_next"]
 
+        # Always destroy our session first before trying to initiate single-sign out
         auth.logout(request)
 
+        # Try to initiate a frontchannel redirect
+        if logout_endpoint:
+            params = {
+                # The value MUST have been previously registered with the
+                # OP, either using the post_logout_redirect_uris
```
pi-sigma:
```suggestion
                # OP, either using the post_logout_redirect_uri
```

src/open_inwoner/accounts/views/auth_oidc.py
```diff
@@ -97,16 +97,34 @@ def get_success_url(self):
 
     def get(self, request):
         assert self.config_class is not None
+        config = self.config_class.get_solo()
 
-        if id_token := request.session.get("oidc_id_token"):
-            config = self.config_class.get_solo()
-            do_op_logout(config, id_token)
+        if not (logout_endpoint := config.oidc_op_logout_endpoint):
+            logger.warning("No OIDC logout endpoint defined")
```
pi-sigma: You do another check `if logout_endpoint` at l 113. So instead of checking here, you could do `logout_endpoint = config.oidc_op_logout_endpoint` at l 112, continue with `if logout_endpoint`, and add the `logger.warning` outside the conditional right before the response.

pi-sigma (approved):

src/open_inwoner/accounts/tests/test_oidc_views.py
swrichards: I want to make sure the logout endpoint actually destroyed our session as well.

PR #1503, opened 2024-11-26
[#2903] Refactor import/export for ZaakType configs

pi-sigma:
- [x] Support export of only select ZaakType configs
- [x] Support import of ZaakType configs without library

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2903

### DIFF
--- a/src/open_inwoner/openzaak/admin.py
+++ b/src/open_inwoner/openzaak/admin.py
@@ -17,7 +17,7 @@ from privates.storages import PrivateMediaFileSystemStorage
 from solo.admin import SingletonModelAdmin
 
 from open_inwoner.ckeditor5.widgets import CKEditorWidget
-from open_inwoner.openzaak.import_export import CatalogusConfigImport, ZGWConfigExport
+from open_inwoner.openzaak.import_export import ZGWConfigExport, ZGWConfigImport
 from open_inwoner.utils.forms import LimitedUploadFileField
 
 from .models import (
@@ -133,7 +133,7 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
             path(
                 "import-catalogus-dump/",
                 self.admin_site.admin_view(self.process_file_view),
-                name="upload_zgw_import_file",
+                name="upload_catalogus_import_file",
             ),
         ]
         return custom_urls + urls
@@ -163,7 +163,7 @@ class CatalogusConfigAdmin(admin.ModelAdmin):
 
                 try:
                     import_result = (
-                        CatalogusConfigImport.import_from_jsonl_file_in_django_storage(
+                        ZGWConfigImport.import_from_jsonl_file_in_django_storage(
                             target_file_name,
                             storage,
                         )
@@ -371,6 +371,7 @@ class ZaakTypeResultaattypeConfigInline(admin.StackedInline):
 
 @admin.register(ZaakTypeConfig)
 class ZaakTypeConfigAdmin(admin.ModelAdmin):
+    change_list_template = "admin/zaaktypeconfig_change_list.html"
     inlines = [
         ZaakTypeInformatieObjectTypeConfigInline,
         ZaakTypeStatusTypeConfigInline,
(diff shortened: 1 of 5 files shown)


src/open_inwoner/openzaak/import_export.py
```diff
@@ -152,7 +152,65 @@ def _update_nested_zgw_config(
 
 
 @dataclasses.dataclass(frozen=True)
-class CatalogusConfigExport:
+class ZGWConfigExport:
```
swrichards, line 169: I am not sure we need to have inheritance here. I think the more obvious change would to:

1. Rename the existing `CatalogusConfigExport` to `ZGWConfigExport`
2. Add a new factory classmethod, `from_zaaktype_configs`, next to the existing `from_catalogus_configs`, that basically does the same thing, except return `CatalogusConfig.objects.empty()` for the catalogus configs.

Perhaps I am missing something, but I don't think the export logic actually depends on catalogus config exports being present (beyond the constructor method, where you use the catalogus configs to limit which zaaktype configs to select).

swrichards (changes requested):
An architectural question before I proceed, but TLDR: I think we can perhaps make this simpler by just adding another factory method to `CatalogusConfigExport` (and renaming it).

src/open_inwoner/openzaak/import_export.py
swrichards: So we'd get something like:

```python
@@ -210,6 +212,36 @@ class CatalogusConfigExport:
             zaak_resultaat_type_configs=zaak_resultaat_type_configs,
         )
 
+    @classmethod
+    def from_zaaktype_configs(cls, zaak_type_configs: QuerySet) -> Self:
+        if not isinstance(zaak_type_configs, QuerySet):
+            raise TypeError(
+                f"`zaaktype_configs` is not a QuerySet, but a {type(zaak_type_configs)}"
+            )
+
+        if zaak_type_configs.model != ZaakTypeConfig:
+            raise ValueError(
+                f"`zaaktype_configs` is of type {zaak_type_configs.model}, not ZaakTypeConfig"
+            )
+
+        informatie_object_types = ZaakTypeInformatieObjectTypeConfig.objects.filter(
+            zaaktype_config__in=zaak_type_configs
+        )
+        zaak_status_type_configs = ZaakTypeStatusTypeConfig.objects.filter(
+            zaaktype_config__in=zaak_type_configs
+        )
+        zaak_resultaat_type_configs = ZaakTypeResultaatTypeConfig.objects.filter(
+            zaaktype_config__in=zaak_type_configs
+        )
+
+        return cls(
+            catalogus_configs=CatalogusConfig.objects.none(),
+            zaak_type_configs=zaak_type_configs,
+            zaak_informatie_object_type_configs=informatie_object_types,
+            zaak_status_type_configs=zaak_status_type_configs,
+            zaak_resultaat_type_configs=zaak_resultaat_type_configs,
+        )
+
```

pi-sigma: Nice pattern, ~except for `from_zaaktype_configs` we need to return `objects.none()` for the other ZGW objects as well~. Misunderstanding on my part

src/open_inwoner/openzaak/admin.py
```diff
@@ -13,15 +13,11 @@
...
-from import_export.admin import ImportExportMixin
 from privates.storages import PrivateMediaFileSystemStorage
 from solo.admin import SingletonModelAdmin
 
 from open_inwoner.ckeditor5.widgets import CKEditorWidget
-from open_inwoner.openzaak.import_export import (
-    CatalogusConfigExport,
-    CatalogusConfigImport,
-)
+from open_inwoner.openzaak.import_export import CatalogusConfigImport, ZGWConfigExport
```
swrichards: I guess we could propagate the rename to the import container as well:

```suggestion
from open_inwoner.openzaak.import_export import ZGWConfigImport, ZGWConfigExport
```

src/open_inwoner/openzaak/admin.py
```diff
@@ -437,6 +435,94 @@ class ZaakTypeConfigAdmin(ImportExportMixin, admin.ModelAdmin):
...
+        response = StreamingHttpResponse(
+            export.as_jsonl_iter(),
+            content_type="application/json",
+        )
+        response[
+            "Content-Disposition"
+        ] = 'attachment; filename="zgw-zaaktype-export.json"'
+        return response
+
+    def process_file_view(self, request):
```
swrichards, line 461: The only thing that springs to mind is whether to do this as a mixin, given that there's non trivial logic here. But I would say we can add that to the technical debt list (because I think we might also want to add a Celery option given how long these imports take, but it shouldn't delay the current PR and patch release).

swrichards (approved):

src/open_inwoner/openzaak/admin.py
pi-sigma: Noted, I'll create a Taiga issue for this.

PR #1508, opened 2024-11-28
[#2913] Check for duplicate keys in ZGW imports

pi-sigma:
Follow-up to https://github.com/maykinmedia/open-inwoner/pull/1497

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2913

### DIFF
--- a/src/open_inwoner/openzaak/import_export.py
+++ b/src/open_inwoner/openzaak/import_export.py
@@ -27,16 +27,19 @@ logger = logging.getLogger(__name__)
 
 class ZGWImportError(Exception):
     @classmethod
-    def extract_error_data(cls, exc: Exception, jsonl: str):
-        exc_source = type(exc.__context__)
-        data = json.loads(jsonl) if jsonl else {}
+    def extract_error_data(
+        cls, exception: Exception | None = None, jsonl: str | None = None
+    ):
+        data = json.loads(jsonl) if jsonl is not None else {}
         source_config = apps.get_model(data["model"])
 
-        # error type
-        if exc_source is CatalogusConfig.DoesNotExist or source_config.DoesNotExist:
-            error_type = ObjectDoesNotExist
-        if exc_source is source_config.MultipleObjectsReturned:
-            error_type = MultipleObjectsReturned
+        error_type = None
+        if exception:
+            exc_source = type(exception.__context__)
+            if exc_source is CatalogusConfig.DoesNotExist or source_config.DoesNotExist:
+                error_type = ObjectDoesNotExist
+            if exc_source is source_config.MultipleObjectsReturned:
+                error_type = MultipleObjectsReturned
 
         # metadata about source_config
         items = []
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/import_export.py
```diff
@@ -27,16 +27,17 @@
 
 class ZGWImportError(Exception):
     @classmethod
-    def extract_error_data(cls, exc: Exception, jsonl: str):
-        exc_source = type(exc.__context__)
+    def extract_error_data(cls, exception: Exception | None = None, jsonl: str = ""):
         data = json.loads(jsonl) if jsonl else {}
```
swrichards: Let's use `None` in case an empty string is actually passed and we want to error on _that_ as well:

```suggestion
    def extract_error_data(cls, exception: Exception | None = None, jsonl: str | None= None):
        data = json.loads(jsonl) if jsonl is not None else {}
```

src/open_inwoner/openzaak/import_export.py
```diff
@@ -82,16 +81,25 @@ def from_exception_and_jsonl(cls, exception: Exception, jsonl: str) -> Self:
 
         return cls(error_template % error_data)
 
+    @classmethod
+    def from_jsonl(cls, jsonl: str) -> Self:
+        error_data = cls.extract_error_data(jsonl=jsonl)
+        error_template = (
+            "%(source_config_name)s was processed multiple times because it contains "
+            "duplicate natural keys: %(info)s"
```
swrichards, line 91: Perhaps something a bit more contextual:
```suggestion
            "%(source_config_name)s was processed multiple times because it shares "
            "defining attributes (such as `rsin`, `identificatie`, and `omschrijving`) with multiple objects. This may lead to unexpected results, as only the last row will have an effect on the target: %(info)s"
```

On the other hand, maybe your shorter and punchier version is better, as the most likely course is now that people fix it but escalate to us, and the main thing is that _we_ know what is going on. What do you think?

src/open_inwoner/openzaak/import_export.py
```diff
@@ -114,9 +122,9 @@ def _update_zaaktype_config(source_config: ZaakTypeConfig, jsonl: str):
             catalogus_rsin=source_config.catalogus.rsin,
         )
     except ZaakTypeConfig.MultipleObjectsReturned as exc:
-        raise ZGWImportError.from_exception_and_jsonl(exc, jsonl)
+        raise ZGWImportError.from_exception_and_jsonl(exception=exc, jsonl=jsonl)
     except (CatalogusConfig.DoesNotExist, ZaakTypeConfig.DoesNotExist) as exc:
-        raise ZGWImportError.from_exception_and_jsonl(exc, jsonl)
+        raise ZGWImportError.from_exception_and_jsonl(exception=exc, jsonl=jsonl)
```
swrichards: This can be a single except block I think?

src/open_inwoner/openzaak/import_export.py
swrichards: Ignore more comment: I think your error message looks fine tbh.

swrichards (approved):
LGTM, just a few small nitpicks.

swrichards (approved):

PR #1509, opened 2024-12-02
[#2917] Increase number of simultaneous threads per worker

alextreme:


swrichards (approved):
This seems fine, with the usual caveats of increased resource consumption and database load and such, we should keep an eye on that when we deploy.

pi-sigma (approved):

PR #1510, opened 2024-12-03
[#2894] Remove grey border-bottom from search bar

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2894


swrichards (approved):
LGTM, but tagging @jiromaykin for when he's back in case there's cascading effects here we're not aware of that might need follow-up action after this PR.

jiromaykin:
@pi-sigma @swrichards 
This is fine and correct - that line was explicitly added to the design in one place only, so no worries for side effects when removed, except for the collapsing margins (often happens when border is removed) which seem to be okay here.

PR #1511, opened 2024-12-03
[#2893] Remove search from mobile menu

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2893

### DIFF
--- a/src/open_inwoner/scss/components/Header/Header.scss
+++ b/src/open_inwoner/scss/components/Header/Header.scss
@@ -512,6 +512,10 @@ $vm: var(--spacing-large);
               gap: var(--spacing-large);
               vertical-align: bottom;
             }
+
+            [title='Zoeken'] {
+              display: none;
+            }
           }
 
           .form--inline {


pi-sigma:
@jiromaykin should take a quick look as well when he's back

swrichards (approved):

alextreme:
Fine if this works, however a neater approach would be to remove it from components/Header/Header.html. But considering the complexity of this template I agree with doing this via CSS for now

jiromaykin:
@pi-sigma @alextreme  Oh dear.... This is not the way to do this.
Instead two blocks should have been removed from the Header HTML template, accounting for authenticated/non-authenticated:

```{% if cms_apps.products and request.user.is_authenticated or not config.hide_search_from_anonymous_users %}
 <li class="header__list-item">
  {% trans "Zoeken" as search %}
  {% link text=search href="/search" icon="search" icon_position="before" %}
 </li>
{% endif %}```


*** and also: ***


```{% elif config.login_show %}
        [ ... ]
       <li class="header__list-item">
        {% trans "Zoeken" as search %}
        {% link text=search href="/search" icon="search" icon_position="before" %}
       </li>```

alextreme:
> @pi-sigma @alextreme Oh dear.... This is not the way to do this. Instead two blocks should have been removed from the Header HTML template, accounting for authenticated/non-authenticated:
> 
> ```
>  <li class="header__list-item">
>   {% trans "Zoeken" as search %}
>   {% link text=search href="/search" icon="search" icon_position="before" %}
>  </li>
> {% endif %}```
> 
> 
> *** and also: ***
> 
> 
> ```{% elif config.login_show %}
>         [ ... ]
>        <li class="header__list-item">
>         {% trans "Zoeken" as search %}
>         {% link text=search href="/search" icon="search" icon_position="before" %}
>        </li>```
> ```

feel free to create an issue/task and correct this

PR #1516, opened 2024-12-04
[#2860] Additional header for Haal Centraal BRP via Yenlo

alextreme:


alextreme:
Testing this via tag hoorn-hc-brp-yenlo

alextreme:
Discussed and for the short term this header will be necessary for MijnHoorn. Paul if you could add this change to the tests then it should be ready for review

pi-sigma (approved):

PR #1517, opened 2024-12-04
Patch release 1.23.1 (ZGW import/export changes)

swrichards:


swrichards:
Released as https://github.com/maykinmedia/open-inwoner/releases/tag/v1.23.1 (by mistake, it should've been backported to `v1.21.x`, which was done in https://github.com/maykinmedia/open-inwoner/releases/tag/v1.21.4).

PR #1519, opened 2024-12-05
Release/1.21.4

swrichards:


swrichards:
Released as https://github.com/maykinmedia/open-inwoner/releases/tag/v1.21.4

This required removing a migration that was merged into mainline (but not strictly needed for this patch release), so we should be doubly sure never to build upon this release without taking additional care to ensure the migration history integrity.

PR #1520, opened 2024-12-05
[#2899] Integrate OpenKlantConfig and OpenKlant2Config

pi-sigma:


src/open_inwoner/openklant/models.py
```diff
@@ -223,4 +230,45 @@ def from_django_settings(cls) -> Self:
 
         return cls(**config)
 
-    # TODO: add from_openklant_config_model or similar
+
+# TODO: rename after removing dataclass config
+class OpenKlant2Config2(SingletonModel):
```
swrichards: Changing the names of models after the fact is a bit of a pain: easier to refactor the dataclass name first.

pi-sigma:
@swrichards Keeping the `service` field on the `OpenKlant2Config` non-nullable is problematic when attempting to update the config for the first time: the `get_or_create` in the admin's `change_view` raises an `IntegrityError` because no zgw service is provided. I worked around this by creating a default API service and displaying a message to the user that they should update the service. However, it might just be simpler to make the field nullable, as in the old `OpenKlantConfig`. Unless you can think of some other way around this?

swrichards:
> @swrichards Keeping the `service` field on the `OpenKlant2Config` non-nullable is problematic when attempting to update the config for the first time: the `get_or_create` in the admin's `change_view` raises an `IntegrityError` because no zgw service is provided. I worked around this by creating a default API service and displaying a message to the user that they should update the service. However, it might just be simpler to make the field nullable, as in the old `OpenKlantConfig`. Unless you can think of some other way around this?

@pi-sigma I am happy making the field nullable.

The general problem we have is that we want to know if the eSuite/OK2 service is correctly configured, and that does not play nice with `SingletonModel` as you say, because you have to be able to create it with sane defaults without user input. For now, I think we have to tackle this in our Service constructors: that is where we should fail loudly if:

1. OpenKlant2 is configured as the primary
2. But there is no service defined

Arguably we could also do this in the validation step for the global ok admin: if ok2 is selected there, you get a warning to first configure a service.

src/open_inwoner/templates/admin/openklant/klanteninteractiesconfig/change_form.html
```diff
@@ -0,0 +1,10 @@
+{% extends "admin/solo/change_form.html" %}
+
+{% block content %}
+    <p>Configure the backends for retrieving Klanten data:</p>
+    <ul>
+        <li><a href="{% url 'admin:openklant_openklantconfig_change' %}">Configure eSuite backend</a></li>
+        <li><a href="{% url 'admin:openklant_openklant2config_change' %}">Configure OpenKlant2 backend</a></li>
+    </ul>
```
swrichards: I suspect we could solve this by using a combination of the `@admin.display`/`readonly_fields` attribute ([docs](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.readonly_fields)): there will be no list page (because singleton), but the read only fields should still be display in the singleton detail page. Arguably less complex than tweaking the template.

src/open_inwoner/openklant/models.py
```diff
@@ -229,3 +185,59 @@ class Meta:
...
+        verbose_name=_("Interne taak toelichting"),
+        default="",
+        blank=True,
+    )
+
+    class Meta:
+        verbose_name = _("OpenKlant2 configuration")
+
+
+class KlantenInteractiesConfig(SingletonModel):
```
swrichards: Naming things :/ 

Tricky: `KlantInteracties` is specifically an OpenKlant2 concept, not a cross-service concept. I would be inclined to just go for `KlantenSysteemConfig` (`systeem` being generic, but not ambiguous with `service`).

src/open_inwoner/openklant/admin.py
```diff
@@ -108,3 +117,77 @@ class KlantContactMomentAnswerAdmin(admin.ModelAdmin):
...
+                    "was automatically created. Please update the Klanten API service  "
+                    "configuration as necessary.",
+                )
+
+        return super(SingletonModelAdmin, self).change_view(
+            request,
+            object_id,
+            form_url=form_url,
+            extra_context=extra_context,
+        )
```
swrichards: As discussed in the other comment: I think making the field nullable is preferable, and [moving the validation to admin layer](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#adding-custom-validation-to-the-admin) in `KlanteInteractiesConfigAdmin` below (possibly in the `clean_primary_backend` method, which can then check if the service is set for OpenKlant2).

src/open_inwoner/openklant/models.py
swrichards: Would it be possible to tweak the admin index in this PR so this admin is the one accessible from the menu?

swrichards (changes requested):

src/open_inwoner/templates/admin/openklant/klanteninteractiesconfig/change_form.html
pi-sigma: I played with `@admin.display` but couldn't get it to do what's needed without modifying the templates anyway. Also, compare:

- want to customize detail page -> customize template for detail page
- want to customize detail page -> use `@admin.display` to customize list page; although there is no list page, the strategy can still be used to change detail page indirectly

The first is less complex. Unless I'm missing something about the use of the decorator?

swrichards (approved):
Good job, looks a lot better IMO.

PR #1522, opened 2024-12-06
[#2898] Fix update_user_from_partij with OpenKlant2

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2898

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -63,7 +63,7 @@ def _update_user_from_openklant2(
             fetch_params=fetch_params, user=user
         )
         if partij and not created:
-            service.update_user_from_partij(partij_uuid=partij.uuid, user=user)
+            service.update_user_from_partij(partij_uuid=partij["uuid"], user=user)
 
 
 def _update_user_from_esuite(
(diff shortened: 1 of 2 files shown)


pi-sigma:
@swrichards `partij` can be a `Partij` instance or a `dict` with information about a `Partij` instance. This is confusing, every time we see `partij` we have to wonder whether it's one or the other. Should we take the time to make this consistent (I don't remember why we're dealing with `dict` instead of `Partij` in the signals)

swrichards:
> @swrichards `partij` can be a `Partij` instance or a `dict` with information about a `Partij` instance. This is confusing, every time we see `partij` we have to wonder whether it's one or the other. Should we take the time to make this consistent (I don't remember why we're dealing with `dict` instead of `Partij` in the signals)

Objects coming directly from the underlying client should always be dicts (we decided on that over e.g. dataclasses to avoid overly excessive validation). I don't believe we actually have any classes for `Partij`, though our services do use dataclasses as DTOs, so you can get mixups there on the nested objects.

I think the ideal solution here would be better type checking as part of the pipeline, but that's a longer term project. More short term I am on board with any improvements you have in mind here!

swrichards (approved):

pi-sigma:
Thanks for the clarification! Let's not delay this, we can improve type checking later.

PR #1524, opened 2024-12-09
[#2931] Update company data on login

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2931

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -9,6 +9,8 @@ from django.utils.translation import gettext as _
 from open_inwoner.accounts.models import User
 from open_inwoner.haalcentraal.models import HaalCentraalConfig
 from open_inwoner.haalcentraal.utils import update_brp_data_in_db
+from open_inwoner.kvk.client import KvKClient
+from open_inwoner.kvk.signals import company_branch_selected
 from open_inwoner.openklant.services import OpenKlant2Service, eSuiteKlantenService
 from open_inwoner.utils.logentry import user_action
 
@@ -24,6 +26,35 @@ MESSAGE_TYPE = {
     "frontend_oidc": _("user was logged in via frontend using OpenIdConnect"),
     "logout": _("user was logged out"),
 }
+
+
+@receiver(company_branch_selected)
+def update_company_from_kvk_api(sender, *args, **kwargs):
+    request = kwargs["request"]
+    user = request.user
+    kvk = user.kvk
+    vestigingsnummer = kwargs["vestigingsnummer"]
+
+    kvk_client = KvKClient()
+
+    vestiging = kvk_client.get_vestiging(kvk=kvk, vestigingsnummer=vestigingsnummer)
+
+    binnenlandsadres = vestiging.get("adres", {}).get("binnenlandsAdres", {})
+
+    fields = {
+        "company_name": vestiging.get("naam"),
+        "city": binnenlandsadres.get("plaats"),
+        "postcode": binnenlandsadres.get("postcode"),
+        "street": binnenlandsadres.get("straatnaam"),
+        "housenumber": binnenlandsadres.get("huisnummer"),
+    }
+
+    for field, new_value in fields.items():
+        current_value = getattr(user, field, None)
+        if new_value and new_value != current_value:
+            setattr(user, field, new_value)
+
+    user.save()
 
 
 @receiver(user_logged_in)
--- a/src/open_inwoner/kvk/client.py
+++ b/src/open_inwoner/kvk/client.py
@@ -1,13 +1,15 @@
 import logging
 from functools import cached_property
-from typing import cast
 from urllib.parse import urlencode
 
 import requests
 from requests.exceptions import JSONDecodeError
 
-from .constants import Bedrijf, CompanyType
+from open_inwoner.utils.decorators import cache
+
+from .constants import CompanyType
 from .models import KvKConfig
+from .types import BedrijfValidator
 
 logger = logging.getLogger(__name__)
 
@@ -123,10 +125,29 @@ class KvKClient:
 
         return branches
 
-    def get_company_branch(self, vestigingsnummer: str, **kwargs) -> dict:
-        kwargs.update({"vestigingsnummer": vestigingsnummer})
-        vestiging = self.search(**kwargs).get("resultaten", {})
-        return cast(Bedrijf, vestiging)
+    @cache("kvk:{kvk}vestiging:{vestigingsnummer}")
+    def get_vestiging(self, kvk: str, vestigingsnummer: str) -> dict:
+        response = self.search(vestigingsnummer=vestigingsnummer)
+
+        results = response.get("resultaten")
+        if not results:
+            logger.error(
+                "No company branch with vestigingsnummer %s found", vestigingsnummer
+            )
+            return {}
+
+        # the same vestigingsnummer could be associated with different kvk numbers
+        try:
+            vestiging = next(r for r in results if r["kvkNummer"] == kvk)
+        except StopIteration:
+            logger.error(
+                "Company branch with vestigingsnummer %s and kvk number %s not found",
+                vestigingsnummer,
+                kvk,
+            )
+            return {}
+
+        return BedrijfValidator.validate_python(vestiging)
 
     def retrieve_rsin_with_kvk(self, kvk, **kwargs) -> str | None:
         basisprofiel = self._request(
--- a/src/open_inwoner/kvk/client.py
+++ b/src/open_inwoner/kvk/client.py
@@ -1,11 +1,12 @@
 import logging
 from functools import cached_property
+from typing import cast
 from urllib.parse import urlencode
 
 import requests
 from requests.exceptions import JSONDecodeError
 
-from .constants import CompanyType
+from .constants import Bedrijf, CompanyType
 from .models import KvKConfig
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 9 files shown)


src/open_inwoner/kvk/views.py
```diff
@@ -90,6 +91,34 @@ def post(self, request):
...
+
+        # data for update company signal
+        company_branch = (
+            next(
+                branch
+                for branch in form.company_branches
+                if branch.get("vestigingsnummer", None) == branch_number
+            )
+            or form.company_branches[0]
+        )
```
swrichards: Should raise if `branch_number` is not found, with a meaningful error message (of the Not authorized variety).

src/open_inwoner/kvk/views.py
```diff
@@ -90,6 +91,34 @@ def post(self, request):
...
+        company_data = {
+            "name": company_branch["naam"],
+            "address": {
+                "city": company_address.get("plaats", None),
+                "postcode": company_address.get("postcode", None),
+                "street": company_address.get("straatnaam", None),
+                "house_number": company_address.get("huisnummer", None),
+                "house_letter": company_address.get("huisletter", None),
+            },
+        }
```
swrichards: Let's only pass the branch number and let the signal handler fetch this (will indeed lead to multiple requests, but we can handle that with a caching op if it becomes an issue).

src/open_inwoner/accounts/signals.py
```diff
@@ -26,6 +27,31 @@
 }
 
 
+@receiver(company_branch_selected)
+def update_company_from_kvk_api(sender, *args, **kwargs):
```
swrichards, line 32: I would spell out the kwargs specifically, and accept the branch number here, so this handler can fetch the vestigingen (see comment below).

src/open_inwoner/kvk/views.py
```diff
@@ -90,6 +91,34 @@ def post(self, request):
             # Directly calling `super().form_invalid(form)` would override the error
             return self.render_to_response(context)
 
-        request.session[KVK_BRANCH_SESSION_VARIABLE] = request.POST["branch_number"]
+        branch_number = request.POST["branch_number"]
+        request.session[KVK_BRANCH_SESSION_VARIABLE] = branch_number
```
swrichards: On this: we have some helpers in `open_inwoner.kvk.branches` for this. I'd say we re-use them, or get rid of the helpers.

But in light of what we said: we might want to keep the helpers, and add a `get_company_branches_cached` helper, to avoid the multiple lookups.

src/open_inwoner/accounts/signals.py
```diff
@@ -26,6 +27,31 @@
...
+    if postcode and postcode != user.postcode:
+        user.postcode = postcode
+    street = address.get("street", None)
+    if street and street != user.street:
+        user.street = street
+    house_number = address.get("house_number", None)
+    if house_number and house_number != user.housenumber:
+        user.housenumber = house_number
+
+    user.save()
```
swrichards: We can possibly make this easier by defining the expected in a TypedDict, with (most) of the field optional.

swrichards (changes requested):
Discussed off-line, we'll do this PR first before the selection screen skipper.

src/open_inwoner/kvk/views.py
pi-sigma: The form raises a ValidationError if the branch number from the POST doesn't correspond to any branch number returned by the KVK API

src/open_inwoner/accounts/signals.py
```diff
@@ -26,6 +28,34 @@
...
+        "city": binnenlandsadres.get("plaats"),
+        "postcode": binnenlandsadres.get("postcode"),
+        "street": binnenlandsadres.get("straatnaam"),
+        "housenumber": binnenlandsadres.get("huisnummer"),
+    }
+
+    for field, new_value in fields.items():
+        current_value = getattr(user, field, None)
+        if new_value and new_value != current_value:
+            setattr(user, field, new_value)
```
swrichards, line 55: I am not sure these checks have added value: the kvk is the source of truth here I think, we can just overwrite all the fields no (assuming there's a `binnenlandsAdres`)?

src/open_inwoner/kvk/client.py
```diff
@@ -122,6 +125,24 @@ def get_all_company_branches(self, kvk: str, **kwargs) -> list[dict | None]:
 
         return branches
 
+    @cache("vestigingsnummer")
```
swrichards:
```suggestion
    @cache("vestiging:{vestigingsnummer}")
```

I am a bit concerned by the `**kwargs` though, I don't think you can currently add that as a parameter in the cache key, but that means you'll always get a cached result even if `kwargs` changes. Is there any way to actually name the `kwargs`?

src/open_inwoner/kvk/client.py
```diff
@@ -122,6 +125,24 @@ def get_all_company_branches(self, kvk: str, **kwargs) -> list[dict | None]:
...
+        if not vestigingen:
+            logger.warning(
+                "Company branch with vestigingsnummer %s not found", vestigingsnummer
+            )
+            return {}
+
+        # TODO: could we have multiple results for single vestigingsnummer? API seems to allow it:
+        # https://developers.kvk.nl/documentation/testing/swagger-zoeken-api
+        if len(vestigingen) > 1:
+            return {}
```
swrichards: This could be exactly the case that's been posing challenges in OpenKlant: you can have the same vestigingsnummer under two different kvkNummers, basically. But perhaps that's narrowed down by the `kwargs`.

swrichards: This would be a strong argument for including `kvkNummer` here, because we'll face that problem eventually, might as well be explicit now.

src/open_inwoner/kvk/views.py
```diff
@@ -92,4 +93,8 @@ def post(self, request):
 
         request.session[KVK_BRANCH_SESSION_VARIABLE] = request.POST["branch_number"]
 
+        company_branch_selected.send(
+            sender=self, request=request, vestigingsnummer=request.POST["branch_number"]
+        )
+
```
swrichards, line 99: For reasons noted above (that we're likely to need a `(kvk, vestiging)` tuple because that's what OpenKlant expects), I'd add the `kvk_nummer` too.

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -37,16 +37,38 @@ def test_post_branches_page_without_kvk_unauthenticated_throws_401(self):
...
+            "kvkNummer": "12345678",
+            "naam": "Test BV Donald",
+            "adres": {
+                "binnenlandsAdres": {
+                    "plaats": "Lollum",
+                    "postcode": "1234",
+                    "straatnaam": "Fantasielaan",
+                }
+            },
+        }
```
swrichards, line 57: Running theme: I think having the same vestiging multiple times (With different `kvkNummer`) is worth adding to the mock.

src/open_inwoner/kvk/client.py
```diff
@@ -1,11 +1,14 @@
...
+from typing import cast
 from urllib.parse import urlencode
 
 import requests
 from requests.exceptions import JSONDecodeError
 
-from .constants import CompanyType
+from open_inwoner.utils.decorators import cache
+
+from .constants import Bedrijf, CompanyType
```
swrichards: I think technically it's a `type` and not a `constant`.

src/open_inwoner/kvk/constants.py
```diff
@@ -6,3 +8,40 @@ class CompanyType(models.TextChoices):
...
+    postcodeWoonplaats: Optional[str]
+    land: str
+
+
+class BedrijfsAdres(TypedDict, total=False):
+    binnenlandsAdres: Optional[BedrijfsBinnenlandsAdres]
+    buitenlandsAdres: Optional[BedrijfsBuitenlandsAdres]
+
+
+class Bedrijf(TypedDict, total=False):
```
swrichards: I think we should also add a pydantic `BedrijfValidator = TypeValidator(Bedrijf)` here, and use `BedrijfValidator.validate_python` in the handler to ensure the contract is being respected.

swrichards (changes requested):
A few issues to do with vestiging/kvk pairs, but maybe we can tackle that later, let's discuss on monday.

src/open_inwoner/accounts/signals.py
pi-sigma: I wouldn't be enough to check for the presence of `binnenlandAdres`, because each field in `binnenlandAdres` is potentially null and would have to be checked as well. Some of the fields may actually be required (an address without at least city name would be strange), but I don't know which ones (the spec doesn't say). I find it simpler therefore to `get` the values from the `dict` and check if they are different from what's stored on the user.

src/open_inwoner/kvk/client.py
pi-sigma: That's not a use case because the only `kwarg` passed on is the `vestigingsnummer`. The code is misleading though; `kwargs.update` suggests that there is other stuff in there. I changed the function signature + body to make clear that we're only passing on the `vestigingsnummer`.

pi-sigma:
At the time of opening this PR we are not able to support separate updates for users with the same `kvkNummer` but different `vesttigingsnumme`r because our user model treats users with the same `kvk` as identical. For now, we will only update the company name, treating the `hoofdvestiging` as the source of truth. Part of the code can be re-used once we decide how to deal with the kvk/vestigingsnummer situation and (potentially) refactor our user model to deal with it.

PR #1525, opened 2024-12-10
Fix assorted issues in the OpenKlant2 service

swrichards:


src/open_inwoner/openklant/services.py
```diff
@@ -1133,7 +1133,7 @@ def create_question(
         if len(question.rstrip()) == 0:
             raise ValueError("You must provide a question")
 
-        if self.mijn_vragen_actor is None:
+        if self.config.mijn_vragen_actor is None:
```
pi-sigma, line 1136: We check for and set the `mijn_vragen_actor` in the `__init__` of the service, although we didn't provide for the case when the actor is not set in the config, hence the old cold would give an `AttributeError`. Alternatively, you could update the class attributes with `mijn_vragen_actor: str | uuid.UUID | None = None`, then the old check `if self.mijn_vragen_actor is None` could remain.

pi-sigma (approved):

src/open_inwoner/openklant/services.py
swrichards: Fair point, but I suggest we keep this for now. I wanted to stay close to parity and the extra check can't hurt, but when we revamp the config we should reconsider these constraints.

PR #1526, opened 2024-12-10
[#2932] Skip KVK branch selection if vestigingsnummer already selected

pi-sigma:
The KVK branch selection page should be skipped for users who have already selected a branch on a different platform

- [x] Update eHerkenning OIDC flow: get vestigingsnummer from OIDC claim (if present) and store in session
- [x] Update eHerkenning SAML flow: get vestigingsnummer from SAML attributes (if present) and store in session
- [x] Skip KVK branch selection if vestigingsnummer already in session

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2932

### DIFF
--- a/src/open_inwoner/accounts/backends.py
+++ b/src/open_inwoner/accounts/backends.py
@@ -6,7 +6,6 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.hashers import check_password
 from django.contrib.auth.models import AbstractUser
-from django.core.exceptions import SuspiciousOperation
 from django.urls import reverse, reverse_lazy
 
 from axes.backends import AxesBackend
--- a/src/open_inwoner/accounts/backends.py
+++ b/src/open_inwoner/accounts/backends.py
@@ -5,16 +5,21 @@ from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.hashers import check_password
+from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import SuspiciousOperation
 from django.urls import reverse, reverse_lazy
 
 from axes.backends import AxesBackend
 from digid_eherkenning.oidc.backends import BaseBackend
 from mozilla_django_oidc_db.backends import OIDCAuthenticationBackend
 from mozilla_django_oidc_db.config import dynamic_setting
+from mozilla_django_oidc_db.typing import JSONObject
 from oath import accept_totp
 
 from open_inwoner.configurations.models import SiteConfiguration
+from open_inwoner.kvk.branches import KVK_BRANCH_SESSION_VARIABLE
 from open_inwoner.utils.hash import generate_email_from_string
+from open_inwoner.utils.views import LogMixin
 
 from .choices import LoginTypeChoices
 from .models import OpenIDDigiDConfig, OpenIDEHerkenningConfig
@@ -147,7 +152,7 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
         return self.UserModel.objects.filter(**{"oidc_id__iexact": unique_id})
 
 
-class DigiDEHerkenningOIDCBackend(BaseBackend):
+class DigiDEHerkenningOIDCBackend(LogMixin, BaseBackend):
     OIP_UNIQUE_ID_USER_FIELDNAME = dynamic_setting[Literal["bsn", "kvk"]]()
     OIP_LOGIN_TYPE = dynamic_setting[LoginTypeChoices]()
 
@@ -156,6 +161,26 @@ class DigiDEHerkenningOIDCBackend(BaseBackend):
         return parent and self.config_class in (
             OpenIDDigiDConfig,
             OpenIDEHerkenningConfig,
+        )
+
+    def _store_vestigingsnummer_in_session(self, claims: JSONObject):
+        """Get company vestigingsnummer from OIDC claims & store in session"""
+
+        eherkenning_config = self.config_class.get_solo()
+
+        branch_number_claim = eherkenning_config.branch_number_claim[0]
+        if not (vestigingsnummer := claims.get(branch_number_claim)):
+            return
+
+        self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
+        self.request.session.save()
+
+        identifier_claim = eherkenning_config.identifier_type_claim[0]
+        kvk_or_rsin = claims.get(identifier_claim)
+
+        self.log_system_action(
+            f"Vestigingsnummer {vestigingsnummer} retrieved from IdP for "
+            f"eHerkenning user (KVK/RSIN: {kvk_or_rsin})"
         )
 
     def filter_users_by_claims(self, claims):
@@ -169,7 +194,11 @@ class DigiDEHerkenningOIDCBackend(BaseBackend):
         )
 
     def create_user(self, claims):
-        """Return object for a newly created user account."""
+        """
+        Return object for a newly created user account.
+
+        Get vestigingsnummer from OIDC claims & store in session
+        """
 
         unique_id = self._extract_username(claims)
 
@@ -185,4 +214,12 @@ class DigiDEHerkenningOIDCBackend(BaseBackend):
             }
         )
 
+        if self.config_class is OpenIDEHerkenningConfig:
+            self._store_vestigingsnummer_in_session(claims)
+
         return user
+
+    def update_user(self, user: AbstractUser, claims: JSONObject):
+        if self.config_class is OpenIDEHerkenningConfig:
+            self._store_vestigingsnummer_in_session(claims)
+        return super().update_user(user, claims)
--- a/src/open_inwoner/accounts/tests/test_oidc_views.py
+++ b/src/open_inwoner/accounts/tests/test_oidc_views.py
@@ -1828,3 +1828,66 @@ class eHerkenningOIDCFlowTests(WebTest):
         profile_response = self.app.get(profile_response.url)
 
         self.assertEqual(profile_response.status_code, 200)
+
+    @patch("open_inwoner.kvk.client.KvKClient.get_all_company_branches")
+    @patch("open_inwoner.utils.context_processors.SiteConfiguration")
+    @patch("mozilla_django_oidc_db.backends.OIDCAuthenticationBackend.get_userinfo")
+    @patch("mozilla_django_oidc_db.backends.OIDCAuthenticationBackend.store_tokens")
+    @patch("mozilla_django_oidc_db.backends.OIDCAuthenticationBackend.verify_token")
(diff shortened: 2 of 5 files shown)


src/open_inwoner/accounts/backends.py
```diff
@@ -186,3 +188,39 @@ def create_user(self, claims):
         )
 
         return user
+
+    def get_or_create_user(self, access_token, id_token, payload):
+        """Override to store vestigingsnummer (if already present) in session"""
```
swrichards: I think this is very much worth adding to the library, especially if the PR Is small (we can keep it here for now to test our flow, but it would be good to ensure we upstream our changes and eventually revert our override when the library is fixed).

swrichards:
Braindump: @pi-sigma this is technically orthogonal to the issue here (skipping the branch if already selected during the eHerkenning flow), but it's also important that users can see _only_ vestigingen they're authorized to see according to the eHerkenning scope. I am not sure if that flows from this claim (Selected branch) versus some other claim that may or may not exist (all authorized branches).

pi-sigma:
@swrichards Good point. Is this already covered by our use of `get_kvk_branch_number` to scope the retrieval of roles, zaken, and vragen to a particular vestiging, or are we missing something?

src/open_inwoner/accounts/backends.py
swrichards: As discussed off-line, we'll try something that requires a bit less overloading of third party libraries:

```python
def authenticate(self, *args, **kwargs):
    result = super().authenticate(*args, **kwargs)
    if result is not None:
        # TODO: Ensure the setting to persist the ID token in the session is set
        id_token = self.request.session["oidc_id_token"]
        user_info = self.verify_token(id_token)
        if vestigingsnummer := user_info.get("vestigingsnummer", None):
            self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
            self.request.session.save()
    
    return result
```

swrichards: Alternative no. 2, nicer because it stays closer to the public API:

```python
# Misusing the public interface to gain access to claims, which are not
# otherwise exposed but are needed to access the vestigingsnummer.
def create_user(self, claims):
    user = super().create_user(claims)
    if vestigingsnummer := claims.get("vestigingsnummer", None):
        self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
        self.request.session.save()

    return user

def update_user(self, user, claims):
    user = super().update_user(claims)
    if vestigingsnummer := claims.get("vestigingsnummer", None):
        self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
        self.request.session.save()
    
    return user
```

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -1828,3 +1829,69 @@ def test_redirect_after_login_no_registration_and_no_branch_selection(
...
+            enabled=True,
+            legal_subject_claim=["kvk"],
+            oidc_op_authorization_endpoint="http://idp.local/auth",
+        ),
+    )
+    def test_redirect_after_login_branch_already_selected(
+        self,
+        mock_get_solo,
+        mock_get_token,
+        mock_verify_token,
```
swrichards, line 1851: We should make sure that your overrides are actually hit: currently, I don't think the bit of code that actually sets the vestigingsnummer in the session is being tested.

src/open_inwoner/accounts/backends.py
```diff
@@ -185,4 +189,14 @@ def create_user(self, claims):
             }
         )
 
+        if vestigingsnummer := claims.get("vestigingsnummer", None):
+            self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
+            self.request.session.save()
```
swrichards: Note the helpers we have for this in `open_inwoner.kvk.brances`.

src/open_inwoner/accounts/backends.py
```diff
@@ -185,4 +189,14 @@ def create_user(self, claims):
...
+
+        return user
+
+    def update_user(self, user: AbstractUser, claims: JSONObject):
+        if vestigingsnummer := claims.get("vestigingsnummer", None):
+            self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
+            self.request.session.save()
         return user
```
swrichards: Shouldn't we still call the `super().update_user`?

swrichards (changes requested):

src/open_inwoner/accounts/tests/test_oidc_views.py
pi-sigma: Updated the test

src/open_inwoner/accounts/backends.py
pi-sigma: Went for #2

src/open_inwoner/accounts/backends.py
pi-sigma: As discussed, we don't have a helper function for storing the value in the session, and since the functions are quite simple, adding another one doesn't seem to be of much use. Let's continue importing and using `KVK_BRANCH_SESSION_VARIABLE`

src/open_inwoner/accounts/backends.py
pi-sigma: The parent class in in `digid-eherkenning/oidc/backends` overrides the `update_user` from `mozilla_django_oidc_db` method to do nothing, so I though there's no point in calling `super`. I added it back in, though.

src/open_inwoner/accounts/backends.py
```diff
@@ -158,6 +162,27 @@ def _check_candidate_backend(self) -> bool:
...
         )
 
+    def _store_vestigingsnummer_in_session(
+        self, claims: JSONObject, user_created: bool = False
+    ):
+        """Get company vestigingsnummer from OIDC claims & store in session"""
+
+        eherkenning_config = self.config_class.get_solo()
+
+        vestigingsnummer = claims.get(eherkenning_config.branch_number_claim[0], None)
```
swrichards: I am still somewhat worried by the fact that `branch_number_claim` is a list. Can we assume it always has 1 item? If not, should we iterate over the list and return the first vestigingsnummer found to be future-proof?

swrichards (approved):
Just one question about the claims: I'll leave it up to you whether we need to fix this now or whether it needs fixing at all.

src/open_inwoner/accounts/backends.py
pi-sigma: Fair enough, I don't have any experience with the format of these claims either. Will change as suggested.

src/open_inwoner/accounts/backends.py
```diff
@@ -158,6 +162,31 @@ def _check_candidate_backend(self) -> bool:
...
+        self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
+        self.request.session.save()
+
+        if not (identifier_type_claims := eherkenning_config.identifier_type_claim):
+            return
+
+        kvk_or_rsin = identifier_type_claims[0]
+        msg = f"eHerkenning user (KVK/RSIN = {kvk_or_rsin}) retrieved with vestigingsnummer from IdP"
+        if user_created:
+            msg = f"eHerkenning user (KVK/RSIN = {kvk_or_rsin}) created with vestigingsnummer from IdP"
```
swrichards: Should this be in this method? It's strictly speaking about the vestigingsnummer. If we're logging anything here, I'd say we log that we did (or did not) find a vestigingsnummer, and leave the "we created or updated a user" in the function where that happens.

src/open_inwoner/accounts/backends.py
```diff
@@ -158,6 +162,31 @@ def _check_candidate_backend(self) -> bool:
...
+    def _store_vestigingsnummer_in_session(
+        self, claims: JSONObject, user_created: bool = False
+    ):
+        """Get company vestigingsnummer from OIDC claims & store in session"""
+
+        eherkenning_config = self.config_class.get_solo()
+
+        if not (branch_number_claims := eherkenning_config.branch_number_claim):
+            return
+        if not (vestigingsnummer := claims.get(branch_number_claims[0])):
```
swrichards: Just out of interest: did we figure out if there are ever multiple claims? No need to hold up the PR for that, just curious.

src/open_inwoner/accounts/backends.py
```diff
@@ -158,6 +162,31 @@ def _check_candidate_backend(self) -> bool:
...
+            return
+
+        self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
+        self.request.session.save()
+
+        if not (identifier_type_claims := eherkenning_config.identifier_type_claim):
+            return
+
+        kvk_or_rsin = identifier_type_claims[0]
```
swrichards: Are we fetching this only for the logging? In that case I'd also say we log this earlier in the chain, there's a lot of escape hatches in this function making it slightly harder to follow, if we can simplify it that'd be better.

src/open_inwoner/accounts/backends.py
```diff
@@ -158,6 +162,31 @@ def _check_candidate_backend(self) -> bool:
...
 
+    def _store_vestigingsnummer_in_session(
+        self, claims: JSONObject, user_created: bool = False
+    ):
+        """Get company vestigingsnummer from OIDC claims & store in session"""
+
+        eherkenning_config = self.config_class.get_solo()
+
+        if not (branch_number_claims := eherkenning_config.branch_number_claim):
+            return
```
swrichards: My sense is this is an optional in a "you might not have eHerkenning set up" kind of way, but if we make it to this function, eHerkenning is evidently configured, and then I feel a Sentry ping would be in order to at least alert us to the incomplete configuration.

swrichards (changes requested):
Very small nitpicks, otherwise GTG!

src/open_inwoner/accounts/backends.py
pi-sigma: You're right, that "retrieve/create" dichotomy indicates that the message is rather specific to the calling functions. I'll change and simplify accordingly.

src/open_inwoner/accounts/backends.py
pi-sigma: No. I'm not sure what the use case would be, but the library definitely makes room for this.

src/open_inwoner/accounts/backends.py
pi-sigma: I think the escape hatch for the `identifier_type_claim` is unnecessary. If you're logging in with eHerkenning, the KVK or RSIN needs to be in the claims

src/open_inwoner/accounts/backends.py
pi-sigma: In fact, the `_check_candidate_backend` already takes care of this, and since the config class provides a default for the claim field, this is guaranteed to exist. Will simplify

swrichards (approved):

alextreme (approved):

PR #1528, opened 2024-12-11
[Venray #133] Adding a featureflag to allow showing cases without a status

alextreme:


alextreme:
Verified via Venray 133

pi-sigma (approved):

alextreme:
Note that this setting isn't yet configurable via the admin, so it's only toggable via the shell

PR #1529, opened 2024-12-16
[#2940] Add OpenKlant2 configuration model

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2940


src/open_inwoner/accounts/views/signals.py
```diff
@@ -23,9 +23,7 @@ def get_or_create_klant_for_new_user(
     user = instance
 
     # OpenKlant2
-    # TODO: replace with proper config and refactor branching
-    use_ok2 = getattr(settings, "OPENKLANT2_CONFIG", None)
-    if use_ok2 and (openklant2_config := OpenKlant2Config.from_django_settings()):
+    if openklant2_config := OpenKlant2Config.get_solo():
```
swrichards: @pi-sigma I think this is a reason not to use a singleton for the OK2 config model, because this will never be null, it will always return something.

I wonder if it isn't preferable to not make it a singleton, and instead actually make some of the fields (such as the service and mijn vragen actor) required, so that we know _if_ we have a config, it'll be valid.

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -17,10 +21,30 @@
 class PartijGetOrCreateTestCase(Openklant2ServiceTestCase):
     def setUp(self):
         super().setUp()
+
+        server_certificate, _ = Certificate.objects.get_or_create(
+            label="test-server-cert",
+            defaults={
+                "type": CertificateTypes.cert_only,
+            },
+        )
```
swrichards: Curious: this was required?

pi-sigma: ~The `zgw_service` needs a certificate, can't get away with a string~

Double-checked and the service is indeed not necessary. Indeed, in many places the mock config itself is not needed if we disable the `user_logged_in` signal in the test `setUp` instead

src/open_inwoner/accounts/views/signals.py
pi-sigma: Fair point, will get back to this once I have the tests passing.

src/open_inwoner/openklant/models.py
```diff
@@ -138,6 +132,54 @@ def has_api_configuration(self):
...
 
 
+class OpenKlant2ConfigManager(models.Manager):
+    def get_queryset(self):
+        qs = super().get_queryset()
+        return qs.select_related("service_config")
+
+
+class OpenKlant2Config(SingletonModel):
+    zgw_service = models.OneToOneField(
```
swrichards: I think `service` would be slightly preferable here, because (I believe) the `zgw` part is apparently more of a legacy name, we use the `Service` for all kinds of backend.

src/open_inwoner/openklant/models.py
```diff
@@ -138,6 +132,54 @@ def has_api_configuration(self):
...
+        "zgw_consumers.Service",
+        verbose_name=_("Klanten API"),
+        on_delete=models.PROTECT,
+        limit_choices_to={"api_type": APITypes.kc},
```
swrichards: I'd like to remove this constraint for now: as above, I believe the api types are also somewhat redundant, and there's no specific constant to distinguish openklant2 for klanten/contactmomenten.

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -17,10 +19,18 @@
 class PartijGetOrCreateTestCase(Openklant2ServiceTestCase):
     def setUp(self):
         super().setUp()
+
+        zgw_service, _ = Service.objects.update_or_create(
```
swrichards: Curious: why not just use `ServiceFactory` here and throughout? We do that elsewhere in tests when we need services.

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -128,6 +132,24 @@ def setUp(self):
...
+            },
+        )
+        self.openklant2_config, _ = OpenKlant2Config.objects.get_or_create(
+            zgw_service=zgw_service,
+            mijn_vragen_kanaal="oip_mijn_vragen",
+            mijn_vragen_organisatie_naam="Open Inwoner Platform",
+            mijn_vragen_actor="ca0783a1-1d74-4e07-b3e0-185b1d2fccd4",
+            interne_taak_gevraagde_handeling="Beantwoorden vraag Mijn Omgeving",
+            interne_taak_toelichting="Beantwoorden vraag",
+        )
```
swrichards: We could consider also adding an `OpenKlant2ConfigFactory` given that a lot of the fields are basically static (it could even then use a SubFactory to create the service by default if no specific service is needed, possibly with a default api root).

src/open_inwoner/openklant/services.py
```diff
@@ -823,16 +823,16 @@ class OpenKlant2Service(KlantenService):
     client: OpenKlant2Client
 
     def __init__(self, config: OpenKlant2Config | None = None):
-        self.config = config or OpenKlant2Config.from_django_settings()
+        self.config = config or OpenKlant2Config.get_solo()
```
swrichards: As per an earlier comment, doesn't `get_solo` always return a model, i.e. it creates it if not defined, or does this return `None` if not defined?

The reason is that the check on `l.827` will then never fail, but we might not have the fields we need.

src/open_inwoner/openklant/models.py
```diff
@@ -138,6 +132,54 @@ def has_api_configuration(self):
         return all(getattr(self, f, "") for f in self.register_api_required_fields)
 
 
+class OpenKlant2ConfigManager(models.Manager):
+    def get_queryset(self):
+        qs = super().get_queryset()
+        return qs.select_related("service_config")
```
swrichards: Should this be:

```suggestion
        return qs.select_related("zgw_service")
```

src/open_inwoner/openklant/models.py
```diff
@@ -138,6 +132,54 @@ def has_api_configuration(self):
...
+        blank=True,
+    )
+    interne_taak_toelichting = models.CharField(
+        verbose_name=_("Interne taak toelichting"),
+        default="",
+        blank=True,
+    )
+
+    class Meta:
+        verbose_name = _("Open Klant configuration")
```
```suggestion
        verbose_name = _("OpenKlant2 configuration")
```

swrichards (changes requested):
Looks good, just a few minor points which might require attention (especially the singleton issue).

src/open_inwoner/openklant/services.py
pi-sigma: You're right, that's from our earlier implementation. I'm inclined to check for `config.service` and raise `ImproperlyConfigured` if it's not there (`get_solo` will return an object without service). What do you think?

swrichards: I am a bit concerned that we would need a lot of ad-hoc checks all over the place for the specific fields we expect in a specific context. For instance, the mijn vragen settings are in fact required to be able to use the "mijn vragen" page with OK2. It is in a way easier to just say "OK2 is configured or not configured", but if it is configured, the database has enforced the presence of the fields we care about.

That being said, that would in this case mean you cannot configure OK2 without knowing in advance (particularly) the Mijn Vragen Actor, which is annoying. So perhaps we can't avoid doing field specific checks.

TLDR: I am not sure, but let's then check the service for now as you suggest. I suspect what we'll have to do later is to make the "interne taak" part of the question/answer bit conditional upon actually have a mijn vragen actor.

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -89,6 +91,8 @@ class CaseListAccessTests(AssertRedirectsMixin, ClearCachesMixin, TransactionWeb
     def setUp(self):
         super().setUp()
 
+        user_logged_in.disconnect(update_user_from_klant_on_login)
```
swrichards: Are these disconnects still needed now that we properly check the errors everywhere?

src/open_inwoner/openklant/tests/factories.py
```diff
@@ -24,6 +25,19 @@ class Meta:
     user = factory.SubFactory(UserFactory)
 
 
+class OpenKlant2ConfigFactory(factory.django.DjangoModelFactory):
+    service: Service
```
swrichards: This can be a `factory.SubFactory(ServiceFactory, api_root=KLANTEN_ROOT_VCR, api_type=...)`, then you don't have to repeat this in your invocations.

src/open_inwoner/openklant/tests/data.py
```diff
@@ -16,6 +16,7 @@
 from open_inwoner.utils.test import paginated_response
 
 KLANTEN_ROOT = "https://klanten.nl/api/v1/"
+KLANTEN_ROOT_VCR = "http://localhost:8338/klantinteracties/api/v1"
```
swrichards: I'd like to make this `OPEN_KLANT2_ROOT` because `KLANTEN` is just a bit too close to the eSuite "klanten" to cause confusion.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -85,6 +86,7 @@ class OIDCFlowTests(TestCase):
     @classmethod
     def setUpClass(cls):
         super().setUpClass()
+        signals.user_logged_in.disconnect(update_user_from_klant_on_login)
```
swrichards: Same question as below: are these disconnects still needed?

swrichards (changes requested):

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -1139,7 +1138,6 @@ def test_pass_endstatus_type_data_if_endstatus_not_reached(
                 "questions": [
                     make_question_from_contactmoment(self.contactmoment_new),
                     make_question_from_contactmoment(self.contactmoment_balie),
-                    MockOpenKlant2Service().retrieve_question()[0],
```
swrichards: Does this mean we're no longer testing questions retrieved by OK2?

pi-sigma: Removed this when debugging the tests, but forgot to add back in.

src/eherkenning/tests/test_mock_views.py
```diff
@@ -88,6 +89,10 @@ def test_get_returns_valid_response(self):
 @override_settings(**OVERRIDE_SETTINGS)
 @modify_settings(**MODIFY_SETTINGS)
 class PasswordLoginViewTests(eHerkenningMockTestCase):
+    @classmethod
+    def setUpTestData(cls):
+        signals.user_logged_in.disconnect(update_user_from_klant_on_login)
+
```
swrichards: Just out of idle curiosity: why was this one the only require disconnect?

swrichards (approved):

src/eherkenning/tests/test_mock_views.py
pi-sigma: Argh, missed this one, sorry!

swrichards (approved):

alextreme (approved):

PR #1532, opened 2024-12-18
[#1468] Add configuration step for OpenKlant2 connectivity

### LINKED ISSUE #1468: Add django-setup-configuration to configure open-inwoner to openklant
Add django-setup-configuration to configure open-inwoner to openklant

swrichards:


src/open_inwoner/configurations/bootstrap/openklant.py
```diff
@@ -88,3 +106,39 @@ def execute(self, model: KlantenApiConfigurationModel):
...
                 "Unable to validate and save configuration"
             ) from exc
+
+
+class OpenKlant2ConfigurationStep(BaseConfigurationStep[OpenKlant2Configuration]):
+    """
+    Configure the KIC settings and set any feature flags or other options if specified
+    """
+
+    verbose_name = "OpenKlant2 APIs configuration"
```
pi-sigma, line 116: Are we expecting to support multiple?

pi-sigma (approved):

src/open_inwoner/configurations/bootstrap/openklant.py
swrichards: As discussed, the direction of travel is singleton: we'll keep the setup step on this assumption for now (and the nice thing about that is we can refactor the underlying model without too much pain later).

PR #1533, opened 2024-12-19
[#2931] Update company name from KVK API on login

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2931

### DIFF
--- a/src/open_inwoner/accounts/tests/test_auth.py
+++ b/src/open_inwoner/accounts/tests/test_auth.py
@@ -14,7 +14,7 @@ from furl import furl
 from pyquery import PyQuery as PQ
 
 from open_inwoner.accounts.choices import NotificationChannelChoice
-from open_inwoner.accounts.signals import update_user_from_klant_on_login
+from open_inwoner.accounts.signals import KvKClient, update_user_on_login
 from open_inwoner.configurations.models import SiteConfiguration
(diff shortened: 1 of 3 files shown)


src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -1896,7 +1896,7 @@ def test_anonymous_user_is_redirected_to_login_page_if_password_change_is_access
 
 
 @requests_mock.Mocker()
-class TestUpdateUserFromKlantUponLoginTests(TestCase):
+class UpdateUserOnLoginTest(TestCase):
```
swrichards, line 1899: PR LGTM, though technically it looks like we're not actually verifying that the name is _saved_ I think? Happy to go with this for now, we can treat it as a stretch goal.

swrichards (approved):
Just one question about a test, but treat it as a nitpick.

src/open_inwoner/accounts/tests/test_auth.py
pi-sigma: Yes, we're only verifying that the functions are called, not that the update works. I was going for the minimal solution, but will add this

swrichards (approved):

alextreme (approved):

PR #1535, opened 2024-12-20
[#2946] Fix login for eHerkenning user with single vestiging

pi-sigma:
The KVK branch selection view should be displayed, and the vestigingsnummer of the company should be stored in the session, for companies that have only a single vestiging (i.e. a hoofdvestiging but no nevenvestiging).

We currently skip the KVK branch selection view for companies with a single vestiging. The reason for doing this was a problem with endless redirects caused by our middleware (see [https://taiga.maykinmedia.nl/project/open-inwoner/task/2000](https://taiga.maykinmedia.nl/project/open-inwoner/task/2000.)).

The issue now is that if a client logs in with a KVK nummer that has only one vestiging, the vestigingsnummer is not stored in the session. Consequently, any calls to APIS that make use of the KVK + vestigingsnummer will give the wrong results for this KVK number.

Reproducible with KVK number 90002490, which has only a hoofdvestiging.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2946

### DIFF
--- a/src/open_inwoner/kvk/views.py
+++ b/src/open_inwoner/kvk/views.py
@@ -7,13 +7,14 @@ from django.views.generic import FormView
 from furl import furl
 
 from open_inwoner.kvk.branches import KVK_BRANCH_SESSION_VARIABLE
+from open_inwoner.utils.views import LogMixin
 
 from ..utils.url import get_next_url_from
 from .client import KvKClient
 from .forms import CompanyBranchChoiceForm
 
 
-class CompanyBranchChoiceView(FormView):
+class CompanyBranchChoiceView(LogMixin, FormView):
     """Choose the branch ("vestiging") of a company"""
 
     template_name = "pages/kvk/branches.html"
@@ -32,12 +33,14 @@ class CompanyBranchChoiceView(FormView):
         company_branches = kvk_client.get_all_company_branches(
             kvk=self.request.user.kvk
         )
-        # create pseudo-branch representing the company as a whole
-        master_branch = {
+        # create pseudo-branch representing the company as a whole (the "rechtspersoon").
+        # technically, the compnay as a legal entity is represented as "rechtspersoon",
+        # but this is not always included in query results
+        rechtspersoon_entry = {
             "vestigingsnummer": "",
             "naam": company_branches[0].get("naam", "") if company_branches else "",
         }
-        company_branches.insert(0, master_branch)
+        company_branches.insert(0, rechtspersoon_entry)
 
         kwargs["company_branches"] = company_branches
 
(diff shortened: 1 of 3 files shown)


swrichards:
One note about the copy:

Screenshot from 2024-12-20 13-00-58

> Log in namens alle vestigingen

Is a bit misleading: what you're actually doing is "Log in namens rechtspersoon" which specifically _excludes all vestigingen_, including the hoofdvestiging.

src/open_inwoner/kvk/views.py
```diff
@@ -32,7 +32,9 @@ def get_form_kwargs(self):
         company_branches = kvk_client.get_all_company_branches(
             kvk=self.request.user.kvk
         )
-        # create pseudo-branch representing the company as a whole
+        # create pseudo-branch representing the company as a whole.
+        # technically, the compnay as a legal entity is represented as "rechtspersoon",
+        # but this is not always included in query results
```
swrichards: As discussed, I think `pseudo-branch` is confusing here because what we're after is the rechtspersoon which is definitely not a branch. I would rename this to: `rechtspersoon_entry` to avoid confusion.

src/open_inwoner/kvk/views.py
```diff
@@ -63,11 +65,9 @@ def get(self, request, *args, **kwargs):
...
         form = context["form"]
 
-        company_branches_with_master = form.company_branches
-        # Exclude the "master" branch from these checks, since we always insert this
-        company_branches = company_branches_with_master[1:]
-
-        if not company_branches or len(company_branches) == 1:
+        # check that there are company branches besides our artifical "master_branch"
+        vestigingen = form.company_branches[1:]
+        if not vestigingen or not any(v.get("vestigingsnummer") for v in vestigingen):
```
swrichards, line 71: My best guess is that this be incredibly rare, but it might warrant a `system_action` for diagnostics (including I think the selection of the branch when it's made, that would also help debugging)..

src/open_inwoner/kvk/views.py
```diff
@@ -32,7 +32,9 @@ def get_form_kwargs(self):
         company_branches = kvk_client.get_all_company_branches(
             kvk=self.request.user.kvk
         )
-        # create pseudo-branch representing the company as a whole
+        # create pseudo-branch representing the company as a whole.
+        # technically, the compnay as a legal entity is represented as "rechtspersoon",
+        # but this is not always included in query results
         master_branch = {
             "vestigingsnummer": "",
```
swrichards, line 40: Just checking: this means that the `POST` handler in will have an empty string in `request.POST["branch_number"]`, correct? And that would thus be empty and signify we're talking about the rechtspersoon rather than one of its vestigingen? If so, I would add a comment around `l.93` to make this explicit, so something like:

```
branch_number := request.POST["branch_number"])
request.session[KVK_BRANCH_SESSION_VARIABLE] = branch_number or None # an empty string is interpreted as "interact as the rechtsersoon, not as any specific branch"
```

swrichards (changes requested):
LGTM, just some contextual things, plus my comment about a change to the form copy, to avoid confusion about rechtspersoon versus "all branches".

I double checked this locally, and it looks like it's sending the correct values to both Mijn Aanvragen and Mijn Vragen, depending on the selection (either kvk or vestigingsnummer for Mijn Vragen, depending on selection, and either kvk or kvk + vestiging for Mijn Aanvragen).

src/open_inwoner/kvk/views.py
pi-sigma: That was the idea, yes

src/open_inwoner/kvk/views.py
pi-sigma: Included  a `system_action` with information about the vestigingen. Not sure what you mean by "the selection of the branch when it's made". Do you mean a timestamp? That should be nearly the same as the time of the log entry, no?

swrichards: Ah sorry, I meant logging the happy flow (user selected either rechtspersoon or branch) in addition to the unhappy flow you logged here, so something like: `system_action("User selected {rechtspersoon kvk123 | branch 456 for rechtspersoon kvk123")`. But I now feel that might be a bit overkill, feel free to ignore.

swrichards (approved):

PR #1538, opened 2024-12-27
[#2935] Mijn Afspraken: Don't show appoinments from the past

alextreme:


pi-sigma (changes requested):
- There is no test that past appointments are actually excluded
- Tests should use `freezegun.freezetime` instead of manually setting a future date

src/open_inwoner/cms/plugins/tests/test_appointments.py
```diff
@@ -19,12 +20,14 @@ def test_plugin(self, m):
...
-        )
+        with freeze_time("2020-01-01 00:00"):
+            html, context = cms_tools.render_plugin(
+                UserAppointmentsPlugin, plugin_data={}, user=data.user
+            )
 
         appointments = context["appointments"]
 
+        # past appointments should not show up
         self.assertEqual(len(appointments), 2)
```
swrichards, line 30: This might be more explicit (the `notIn` follows from the previous one, but code > comments). But it's a nitpick.

```suggestion
        appointment_titles = [a.title for a in appointments]
        self.assertEqual([data.appointment_idcard.title, data.appointment_passport.title], appointment_titles)
        self.assertNotIn(appointment_titles, data, appointment_old.title)
```

swrichards (approved):

src/open_inwoner/cms/plugins/tests/test_appointments.py
pi-sigma: You're right. I changed it, though using a slightly different approach.

swrichards (approved):

pi-sigma (approved):

PR #1544, opened 2025-01-08
[#2808] Use KVK API to check legal status of compay at login

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2808

### DIFF
--- a/src/open_inwoner/accounts/views/auth.py
+++ b/src/open_inwoner/accounts/views/auth.py
@@ -18,13 +18,13 @@ from digid_eherkenning.mock.views.digid import DigiDAssertionConsumerServiceMock
 from digid_eherkenning.views.base import get_redirect_url
 from digid_eherkenning.views.digid import DigiDAssertionConsumerServiceView
 from digid_eherkenning.views.eherkenning import eHerkenningAssertionConsumerServiceView
+from glom import glom
 
 from eherkenning.mock import eherkenning_conf
 from eherkenning.mock.views.eherkenning import (
     eHerkenningAssertionConsumerServiceMockView,
 )
-from open_inwoner.openklant.models import ESuiteKlantConfig
-from open_inwoner.openzaak.models import OpenZaakConfig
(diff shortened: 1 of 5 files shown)


src/open_inwoner/kvk/client.py
```diff
@@ -123,6 +125,10 @@ def get_all_company_branches(self, kvk: str, **kwargs) -> list[dict | None]:
 
         return branches
 
+    @cache_result("kvk:{kvk}")
+    def get_basisprofiel(self, kvk: str, **kwargs) -> dict:
+        return self._request(f"{self.basisprofielen_endpoint}/{kvk}", params=kwargs)
```
swrichards: We should be careful combining `**kwargs` with `@cache_result`: it means you'll get the same cached result even if `**kwargs` vary. I would prefer either writing out the kwargs and adding them to the caching key, or not accepting them until we need them.

src/open_inwoner/accounts/views/auth.py
```diff
@@ -136,16 +135,22 @@ class BlockEenmanszaakLoginMixin:
...
+        basisprofiel = client.get_basisprofiel(kvk=kvk)
+
+        if basisprofiel.get("fout"):
+            auth.logout(request)
+            message = basisprofiel["fout"][0]["omschrijving"]
+            messages.error(request, message)
+            failure_url = self.get_failure_url()
+            return HttpResponseRedirect(failure_url)
+
+        if basisprofiel["_embedded"]["eigenaar"]["rechtsvorm"] == "Eenmanszaak":
```
swrichards: Will the entire nested path be available for all responses? If not, we might want an extra guard in case e.g. `eigenaar` does not exist (perhaps `glom` can help).

swrichards (changes requested):

swrichards (approved):
LGTM, holding off on merge until after `v1.26.0`.

PR #1546, opened 2025-01-09
eSuite user syncing refactors

swrichards:
Some ground-clearing in preparation for a PR to actually sync new user data to eSuite upon user creation.


src/open_inwoner/openklant/api_models.py
```diff
@@ -6,13 +6,19 @@
 from zgw_consumers.api_models.base import ZGWModel
 
 
-class KlantCreateData(TypedDict):
-    bronorganisatie: str
-    voornaam: str
-    voorvoegselAchternaam: str
-    achternaam: str
+class KlantWritePayload(TypedDict, total=False):
+    """The writable fields for en eSuite klant.
```
pi-sigma:
```suggestion
    """The writable fields for an eSuite klant.
```

src/open_inwoner/openklant/api_models.py
```diff
@@ -6,13 +6,19 @@
 from zgw_consumers.api_models.base import ZGWModel
 
 
-class KlantCreateData(TypedDict):
-    bronorganisatie: str
-    voornaam: str
-    voorvoegselAchternaam: str
-    achternaam: str
+class KlantWritePayload(TypedDict, total=False):
```
pi-sigma, line 9: Perhaps the name itself (not just the comment) should make clear that this is intended for eSuite: `KlantWritePayloadESuite` or the like?

pi-sigma (approved):

alextreme:
@swrichards can be merged after resolving the conflict

src/open_inwoner/openklant/api_models.py
swrichards: Fair, but I think I'll leave it consistent with the rest of the file for now, because this is true for the whole submodule.

PR #1552, opened 2025-01-14
[#2966] Refactor logic for registering contactmomenten in eSuite

pi-sigma:
In preparation for adding support for registering questions with OpenKlant2, the existing logic for registering contactmomenten in the eSuite is refactored.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2966


swrichards (changes requested):
LGTM, just needs a rebase for the migration dependencies (it's a nitpick, but there's also some stash conflict markers in the first commit, not that it hurts much). I'd like to hold off on merging until `v1.26.0` is released.

src/open_inwoner/openklant/migrations/0019_klantensysteemconfig_register_contact_email_and_more.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.2.16 on 2025-01-14 13:17
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("openklant", "0018_rename_openklantconfig_esuiteklantconfig_and_more"),
```
swrichards: I think this needs to be dependent on `src/open_inwoner/openklant/migrations/0020_add_more_help_texts_to_openklant2_config.py`?

src/open_inwoner/openklant/migrations/0020_klantensysteemconfig_register_contact_email_and_more.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.2.16 on 2025-01-14 13:17
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("openklant", "0019_alter_klantensysteemconfig_primary_backend"),
+    ]
```
swrichards: @pi-sigma shouldn't this depend on https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/openklant/migrations/0020_add_more_help_texts_to_openklant2_config.py ?

pi-sigma: You're right, I mixed up the order  (`0020_add_more...` is already in `develop` but there's no logical dependency between the migrations).

swrichards (approved):

PR #1553, opened 2025-01-15
[#2959] Sync user to eSuite after completing registration

swrichards:
[Taiga](https://taiga.maykinmedia.nl/project/open-inwoner/issue/2959)

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -1,7 +1,6 @@
 import logging
 
 from django.contrib.auth.signals import user_logged_in, user_logged_out
-from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.urls import reverse
 from django.utils.translation import gettext as _
@@ -37,6 +36,7 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
     if user.login_type not in [LoginTypeChoices.digid, LoginTypeChoices.eherkenning]:
         return
 
+    # KvK API
     if user.login_type is LoginTypeChoices.eherkenning:
         _update_eherkenning_user_from_kvk_api(user=user)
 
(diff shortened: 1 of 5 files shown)


src/open_inwoner/accounts/signals.py
```diff
@@ -89,38 +89,6 @@ def _update_eherkenning_user_from_kvk_api(user: User):
         user.save()
```
swrichards, line 91: @pi-sigma Note here that this hook did not add much: all DigiD/eHerkenning users would _also_ be logged in upon creation, which would trigger the handler for the `user_logged_in` flow, so I removed this as it just causes confusion.

src/open_inwoner/accounts/views/registration.py
```diff
@@ -186,18 +187,31 @@ def get_initial(self):
...
-                }
-            )
+        user = form.instance
+
+        klant, _ = service.get_or_create_klant(
+            user=user, fetch_params=service.get_fetch_parameters(user=user)
+        )
+
+        if not klant:
+            logger.error("Unable to create klant during post-registration sync")
```
pi-sigma: Perhaps add `user` and/or fetch_parameters to the log?

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -507,6 +509,72 @@ def test_existing_user_digid_login_fails_brp_update_when_brp_http_404(self, m):
 
         self.assertNotEqual(user.first_name, "UpdatedName")
 
+    @requests_mock.Mocker()
+    def test_user_without_klant_is_created_and_updated(self, m):
+        """
+        Assert that if not klant exists during the filling the necessary fields form,
```
pi-sigma:
```suggestion
        Assert that if no klant exists during the filling of the necessary fields form,
```

pi-sigma (approved):

alextreme:
Will merge to ensure we can test this today before the esuite is closed off end-of-day... but I'll only trust this after I see it working

PR #1554, opened 2025-01-15
Guard against indefinite caching

swrichards:
Because the timeout value for the ZGW caching is taken from the Django settings, it is not unthinkable that, due to some misconfiguration, the timeout value could end up being None. If this is passed as-is to the low-level Django caching backend, it will be interpreted as "cache forever", because the default kwarg of 60 would be overridden. This can lead to unexpected behavior, e.g. ZGW API objects appearing to be stale because the cache never expires.

This commit makes this process more explicit by setting a default only if no timeout kwarg is provided, and also interpreting a timeout of "None" as a request for the default timeout, to fully prohibit "cache forever" scenarios.

### DIFF
--- a/src/open_inwoner/utils/decorators.py
+++ b/src/open_inwoner/utils/decorators.py
@@ -56,24 +56,33 @@ def _map_cache_key_instance_attrs_to_placeholders(key):
     return mapped_key, identifiers
 
 
+_DEFAULT_CACHE_TIMEOUT = 60
+_CACHE_MISS = object()
+
+
 def cache(
     key: str,
     alias: str = "default",
     *,
-    timeout: int = 60,
+    timeout: int = _DEFAULT_CACHE_TIMEOUT,
 ):
     """
     Decorator factory for updating the django low-level cache.
 
-    It determines if the key exists in cache and skips it by calling the decorated function
-    or creates it if doesn't exist.
+    It determines if the key exists in cache and skips it by calling the decorated
+    function or creates it if doesn't exist.
 
-    :param key: the caching key to use. Can contain any named positional and keyword argument
-    of the wrapped function as interpolation placeholders. When deocarating a method,
-    you can also include instance attributes using the `"cache:{self.attr}"` syntax.
+    :param key: the caching key to use. Can contain any named positional and keyword
+    argument of the wrapped function as interpolation placeholders. When deocarating a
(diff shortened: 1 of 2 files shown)


src/open_inwoner/utils/decorators.py
```diff
@@ -72,9 +76,14 @@ def cache(
...
     you can also include instance attributes using the `"cache:{self.attr}"` syntax.
     :param alias: the Django cache to use, defaults to "default"
-    :param timeout: the timeout for the cache in seconds. Defaults to 60
+    :param timeout: the timeout for the cache in seconds. Defaults to 60. Note that you
+    cannot cache forever (`None` will be interpreted as a request to use the default),
+    but you can set this to `0` to avoid caching, as per the Django low-level cache
+    docs.
     """
 
+    _timeout = timeout if timeout is not None else _DEFAULT_CACHE_TIMEOUT
```
pi-sigma: I agree that we should disallow caching forever (though I'm not sure why anyone would want to try that), but the approach to `None` seems needlessly complicated and potentially confusing. Instead of re-interpreting some expected behavior in a different way because that is found more desirable (`timeout=None` doesn't mean that `timeout` is `None`, as one would expect, but rather `timeout` is `default` because that is found more desirable), I would prefer to keep the meaning of the original behavior and simply disallow it. That is, `if not isinstance(timeout, int): raise ValueError` or something like it.

What is the advantage of this approach to `None`? Perhaps I'm missing something...

pi-sigma (changes requested):

src/open_inwoner/utils/decorators.py
swrichards: That's a fair question.

To be clear: I don't think anybody _intends_ to cache forever, my concern was that this may _accidentally_ happen if a setting is missing/misconfigured, thus defaulting to `None` and infinite caching.

So your proposal would be to 1) keep `60` (or whatever) as the kwarg default but 2) raising if some other value is passed that is not an integer?

The advantage of this approach is basically that it uses a sane default if `None` was not intended, but I agree that's not preferable if the user meant to say something like "no caching" (which should be `0`), in which case a runtime error is probably preferable to implicit behavior.

TLDR: I'll go with your suggestion.

pi-sigma (approved):

PR #1563, opened 2025-01-16
:chart_with_upwards_trend: [#2965] Add 'Mijn aanvragen' Siteimprove tracking (Dynamic errors)

jiromaykin:
Separate issue for more complex dynamic tracking,
"..._Ik wil weten hoeveel mensen een verkeerd bestand uploaden en daardoor een error krijgen. Dit uitgesplitst in een error van een te groot bestand en een error van een verkeerd bestand_..."
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2965

+This code also corrects an old bug from 3 years ago "Unresolved variable or type selects" in the Mijn Samenwerkingen dropdown Select-filter.


jiromaykin:
@stefrado Kun je alvast uittesten of deze eigenlijk wel goed werkt? (dus: files uploaden met fouten als "te groot / verkeerd type/ beide")? En dan files verwijderen/nieuwe uploaden etc. De code staat nog helemaal vol met console.logs die er in de definitieve versie uit moeten + ook de "_sz" mock moet worden uitgecomment zodra deze goed is voor productie.

src/open_inwoner/js/components/autosumbit/index.js
swrichards, line 1: Mostly unrelated, but: I assume that's a typo in the (pre-existing) filename?

src/open_inwoner/js/components/siteimprove/error-tracking.js
```diff
@@ -0,0 +1,133 @@
...
+  handleChanges(event) {
+    if (!this.#fileInputElement || event.target !== this.#fileInputElement)
+      return
+
+    const currentErrors = Object.entries(this.#occurringErrors).reduce(
+      (acc, [key, nodes]) => {
+        nodes.forEach((node) => {
+          const fileName =
+            node.querySelector('.file__name').textContent ??
+            [...this.#fileListElement.children].indexOf(node)
```
swrichards, line 56: I am not sure I follow the logic behind the fallback value: this would simply return an empty ID no? Perhaps something like `unnamed-file-{index}` would be better? I suppose most browsers won't let you select unnamed files, so this is really just a guard, but still.

Also: isn't the spread operator redundant here?

```suggestion
            node.querySelector('.file__name').textContent ??
            this.#fileListElement.children.indexOf(node)
```

src/open_inwoner/js/components/siteimprove/error-tracking.js
```diff
@@ -0,0 +1,133 @@
...
+          acc.set(fileName, this.#ERROR_MAP[key])
+        })
+        return acc
+      },
+      new Map()
+    )
+
+    currentErrors.forEach((message, id) => {
+      if (!this.previousErrorState.has(id) && typeof _sz !== 'undefined')
```
swrichards, line 65: Is there a case here where the number of errors changes between runs and `this.previousErrorState.has(id)` gives false positives?

swrichards (approved):
LGTM, just a few questions (mainly for my own understanding).

pi-sigma (approved):

src/open_inwoner/js/components/autosumbit/index.js
jiromaykin: Yeah, I wanted to keep the filename so we can see the delta/changes - but the developer who made this may have set this name to avoid conflicts with other JS files? But prob a typo. I don't know...

src/open_inwoner/js/components/siteimprove/error-tracking.js
jiromaykin: Not in real life reality. But if the user has very fast hands and can first _delete_ a faulty file AND upload a similar one that is also faulty _within 50 milliseconds_ or less, then there is an edge case, where the error tracking may misinterpret a previously removed error as a "new" one, because it replaces the Map. But I really don't think that's humanly possible.

swrichards: Clear, thanks for the explanation!

src/open_inwoner/js/components/siteimprove/error-tracking.js
jiromaykin: This could better be answered by @stefrado but here is my take:
in almost all cases, the .file__name element should contain some text in reality. However, the fallback (indexOf(node)) is a safeguard for unexpected edge cases:
-If there's a rendering delay and the .file__name text isn't populated yet.
-If the .file__name selector is somehow missing due to a DOM update issue (like flakey HTMX).
-If the system processes files asynchronously and some file metadata is not yet available.
While unlikely, having a fallback (here the fallback is an index number) ensures that error tracking doesn't break due to an empty identifier.
Just the ID is indescriptive so yeah, unnamed-file-{index} could be better but humans won't actually see that, it's code only.

The spread operator isn't quite redundant. Since "this.#fileListElement.children" is an **HTMLCollection**, but the .indexOf() method only works on arrays, so the code is converting the collection into an array before calling .indexOf(node). The spread operator creates a shallow copy.
A more readable way to do this, would be `const index = Array.from(this.#fileListElement.children).indexOf(node)` but you'll still need it.

All together it could be something like this:
`const currentErrors = Object.entries(this.#occurringErrors).reduce(
  (acc, [key, nodes]) => {
    nodes.forEach((node) => {
      const index = Array.from(this.#fileListElement.children).indexOf(node)
      const fileName = node.querySelector('.file__name')?.textContent || `unnamed-file-${index}`
      acc.set(fileName, this.#ERROR_MAP[key])
    })
    return acc
  },
  new Map()
)
`

src/open_inwoner/js/components/siteimprove/error-tracking.js
jiromaykin: Also: keep in mind that this is not 'functional' code for making the components work so no need to be 100% perfect. 
It's only meant for analytics, where we just want to track IF users are actually uploading faulty files; we don't care much about all of the individual files forever and ever. Groningen just wants to check if 'enough' individual users pay attention to the paragraph that tells them the required types/sizes.

src/open_inwoner/js/components/siteimprove/error-tracking.js
swrichards: That's a very useful explanation, thanks. I take your point that the fallback is really more of a theoretical option, so feel free to ignore.

I didn't realize that `HTMLCollection` didn't implement the full array interface, good to know :+1: .

jiromaykin: That may be one of the reasons why React code is chockfull of spread operators.

PR #1573, opened 2025-01-22
Make case title description explicitly configurable from the source ZGW fields

swrichards:
[Taiga 2950](https://taiga.maykinmedia.nl/project/open-inwoner/us/2950?milestone=521)

Now that there's three options, rather than two, to configure how a case title is displayed, the `use_zaak_omschrijving_as_title` flag has been deprecated in favor of an explicit enum:

Screenshot from 2025-01-22 15-55-12

As @alextreme pointed out earlier, this should really be a flag configurable on each individual API group rather than globally, but TLDR, that's a much larger refactor than it looks like, so we'll have to do that another day.


src/open_inwoner/openzaak/migrations/0060_remove_openzaakconfig_use_zaak_omschrijving_as_title_and_more.py
```diff
@@ -0,0 +1,87 @@
...
+            logger.info(
+                "Setting OpenZaakConfig.derive_zaak_titel_from to %s because "
+                "OpenZaakConfig.use_zaak_omschrijving_as_title = True",
+                ZaakTitleDisplayChoices.zaak_titel,
+            )
+            config.derive_zaak_titel_from = ZaakTitleDisplayChoices.zaak_titel
+        else:
+            logger.info(
+                "Setting OpenZaakConfig.derive_zaak_titel_from to %s because "
+                "OpenZaakConfig.use_zaak_omschrijving_as_title = True",
```
pi-sigma:
```suggestion
                "OpenZaakConfig.use_zaak_omschrijving_as_title = False",
```

src/open_inwoner/openzaak/migrations/0060_remove_openzaakconfig_use_zaak_omschrijving_as_title_and_more.py
```diff
@@ -0,0 +1,87 @@
...
+                ZaakTitleDisplayChoices.zaak_titel,
```
pi-sigma:
```suggestion
                ZaakTitleDisplayChoices.zaaktype_onderwerp,
```

src/open_inwoner/openzaak/constants.py
```diff
@@ -7,3 +7,13 @@ class StatusIndicators(models.TextChoices):
     warning = "warning", _("Warning")
     failure = "failure", _("Failure")
     success = "success", _("Success")
+
+
+class ZaakTitleDisplayChoices(models.TextChoices):
+    zaak_titel = "zaaktype_title", _("The title of the case (`zaak.omschrijving`)")
```
pi-sigma:
```suggestion
    zaak_omschrijving = "zaak_omschrijving", _("The description of the case (`zaak.omschrijving`)")
```

or even:
```
    zaak_omschrijving = "zaak_omschrijving", _("The description of the particular case (`zaak.omschrijving`)")
``` 
to bring out the contrast with the zaak type.

The option is to use the `omschrijving` of the particular `zaak` as `title` for the `zaak`, right?

pi-sigma (changes requested):
I think the label for the first option in the new Enum is misleading and should be changed, unless I'm misunderstanding the current title/description situation.

src/open_inwoner/openzaak/constants.py
swrichards: :facepalm: yup.

pi-sigma (approved):

alextreme (approved):

PR #1574, opened 2025-01-22
[#2969] Create questions with OpenKlant2

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2969


src/open_inwoner/openklant/models.py
```diff
@@ -117,17 +118,24 @@ class ContactFormSubject(OrderedModel):
...
+    esuite_config = models.ForeignKey(
         "ESuiteKlantConfig",
+        null=True,
+        on_delete=models.CASCADE,
+    )
+    openklant_config = models.ForeignKey(
+        "OpenKlant2Config",
+        null=True,
         on_delete=models.CASCADE,
     )
```
swrichards, line 136: I don't quite follow why this model needs references to the esuite/openklant config.

swrichards: (I guess for the inline admin?)

src/open_inwoner/openklant/models.py
```diff
@@ -213,6 +221,11 @@ class OpenKlant2Config(SingletonModel):
         ),
     )
 
+    register_api_required_fields = ("service",)
+
+    def has_api_configuration(self):
+        return all(getattr(self, f, "") for f in self.register_api_required_fields)
```
swrichards: nitpick, but we can make the intent clearer:
```suggestion
        return all(getattr(self, f, False) for f in self.register_api_required_fields)
```

src/open_inwoner/openklant/models.py
```diff
@@ -285,24 +298,23 @@ def __str__(self):
...
         if self.primary_backend == KlantenServiceType.ESUITE.value:
-            esuite_config = ESuiteKlantConfig.get_solo()
-            return esuite_config.has_api_configuration()
-
-        # TODO: support `has_api_configuration` check for OK2?
+            config = ESuiteKlantConfig.get_solo()
+            return config.has_api_configuration()
+        if self.primary_backend == KlantenServiceType.OPENKLANT2.value:
+            config = OpenKlant2Config.get_solo()
+            return config.has_api_configuration()
```
swrichards: I am somewhat inclined to either use a `match` statement, or to raise a `ValueError` after the conditionals to guard against us ever expanding the enum but not handling it.

src/open_inwoner/openklant/models.py
```diff
@@ -285,24 +298,23 @@ def __str__(self):
...
         if not self.contact_registration_enabled:
             return False
-
         if self.primary_backend == KlantenServiceType.ESUITE.value:
             esuite_config = ESuiteKlantConfig.get_solo()
             return esuite_config.contactformsubject_set.exists()
-
-        # TODO: check conditions for OK2
+        if self.primary_backend == KlantenServiceType.OPENKLANT2.value:
+            return user.is_authenticated
```
swrichards: I don't quite follow the logic here, it feels like it's mixing concerns: should this level of abstraction care about the authentication status? Feel like a view/controller level concern.

swrichards: I suppose this is because OK2 does not support anonymous questions, so I agree this is tricky. It means the feature sets between the backends differ, and the question is where to do the "does this backend support this functionality [anonymous questions]?" check. My gut feeling is that this would belong on the `QuestionService` protocol: `supports_anonymous_questions() -> bool`, and we could perform this check in the view, which I think would be closer to doing the explicit checks in the right place.

src/open_inwoner/openklant/tests/test_contactform.py
```diff
@@ -332,9 +368,50 @@ def test_register_contactmoment_for_anon_user_via_api(
...
+
+        config = KlantenSysteemConfig.get_solo()
+        config.primary_backend = KlantenServiceType.OPENKLANT2.value
+        config.register_contact_via_api = True
+        config.send_email_confirmation = True
+        config.save()
+
+        subject = ContactFormSubjectFactory(subject="Aanvraag")
+
+        response = self.app.get(self.url, user=data.digid_user)
```
swrichards, line 412: This does not seem to be about an anon user as the test name suggests: `user=data.digid_user`

swrichards (changes requested):
Just a question about the anonymous user/ok2 configuration, otherwise just nitpicks.

src/open_inwoner/openklant/models.py
pi-sigma: It already has a ref to `esuite_config`, I just renamed the ref from `config` to `esuite_config` and added the ref to `OpenKlant2Config`. It is used in `has_contactform_configuration` to check if a set of subjects exists. After double-checking, however, I see this seems to be the only reference to the esuite configuration. So we could remove the reference and do only `ContactFormSubject.objects.exists()`?

Edit: Forget it, we need the references anyway to inline the models in the admin

src/open_inwoner/openklant/models.py
```diff
@@ -309,25 +323,24 @@ def __str__(self):
...
-
-        # TODO: support `has_api_configuration` check for OK2?
-        return False
+        match self.primary_backend:
+            case KlantenServiceType.ESUITE.value:
+                config = ESuiteKlantConfig.get_solo()
+            case KlantenServiceType.OPENKLANT2.value:
+                config = OpenKlant2Config.get_solo()
+            case _:
+                config = None
```
swrichards, line 332: I still feel this should be a `raise`: we don't want to silently allow unexpected values in the enum without this code being refactored.

src/open_inwoner/cms/footer/tests/test_plugin.py
```diff
@@ -12,14 +13,16 @@
 class ContactFormTestCase(ClearCachesMixin, TestCase):
     def setUp(self):
         super().setUp()
+
+        self.user = UserFactory()
```
swrichards: Is this guy used for anything?

src/open_inwoner/openklant/services.py
```diff
@@ -136,9 +136,17 @@ def get_fetch_parameters(
 
         return None
 
+    @property
+    def supports_anonymous_questions(self):
+        if isinstance(self.config, ESuiteKlantConfig):
+            return True
+        if isinstance(self.config, OpenKlant2Config):
+            return False
```
swrichards: Here the "match and raise" pattern would also be appropriate I think: don't want to silently fail (in that case `None` would be return, and if we ever used `is True` we'd be in trouble).

src/open_inwoner/openklant/views/contactform.py
```diff
@@ -36,12 +29,23 @@
...
+        self.klanten_client = None
+        self.vragen_service = None
+
+        self.klanten_config = KlantenSysteemConfig.get_solo()
+        match self.klanten_config.primary_backend:
+            case KlantenServiceType.ESUITE.value:
+                self.klanten_client = build_klanten_client()
+                self.vragen_service = eSuiteVragenService()
+            case KlantenServiceType.OPENKLANT2.value:
+                self.vragen_service = OpenKlant2Service()
```
swrichards, line 52: `case _:
   raise RuntimeError(Unknown service type)
`

src/open_inwoner/openklant/models.py
swrichards: Ah but of course this could not be set... my bad, ignore.

swrichards (approved):
Just some minor nitpicks now, all to do with handling the (far-fetched) case in which we add something to the service enum but forget to handle it in all our getters, otherwise LGTM.

src/open_inwoner/cms/footer/tests/test_plugin.py
pi-sigma: Nope, deleted.

src/open_inwoner/openklant/services.py
pi-sigma: I agree about `raise`, but I found it difficult to work with `match` in this situation. If you do `match type(foo)...case(bar):...do stuff` you'll get `SyntaxError: name capture 'bar' makes remaining patterns unreachable`. I stuck with the `if...else` therefore. Though perhaps I'm missing an obvious way to do this with `match`.

swrichards: The syntax for instance checks would be:

```python
match self.config:
    case OpenKlant2Config():  # equivalent to isinstance(self.config, OpenKlant2Config)
         ...
    case _:
        raise RuntimeError("Unhandled case")
```

src/open_inwoner/openklant/views/contactform.py
pi-sigma: If we `raise` in the view, then every access to the contactformulier page will result in a `500` due to `RuntimeError`. It makes sense to `raise` in the service, but I wonder if it's not better, in the view, to add a log about the missing service and then pass on.

swrichards: Indeed that's not desirable. Just double checking: having no backend defined here is a valid use-case right? Then we register it by email, but not by API. If that is the case then I agree, log and move on.

pi-sigma: That's right. In the service we `raise` if there is no valid configuration because we don't support a service without config. But in the view we don't `raise` because we do support a contactformulier without API service.

swrichards: Check, then no `raise` in the view :+1: , thanks.

swrichards (approved):

PR #1575, opened 2025-01-23
[#2975] Consolidate feature flags for OpenZaak config

pi-sigma:
- [x] Move option to fetch eHerkenning zaken with RSIN from global OpenZaak config to ZGW API group config
- [x] Add enum to ZGW API group config for selecting backend for klantgegevens

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2975


src/open_inwoner/openzaak/tests/test_notification_zaak_infoobject.py
```diff
@@ -117,10 +117,7 @@ def test_ziohandle_zaken_notification_niet_natuurlijk_persoon_initiator(
                 mock_handle.reset_mock()
                 self.clearTimelineLogs()
 
-                config.fetch_eherkenning_zaken_with_rsin = (
-                    fetch_eherkenning_zaken_with_rsin
-                )
-                config.save()
+                #
```
swrichards: Gremlin?

src/open_inwoner/openzaak/tests/factories.py
```diff
@@ -42,6 +42,7 @@ class ZGWApiGroupConfigFactory(factory.django.DjangoModelFactory):
     drc_service = factory.SubFactory(ServiceFactory, api_type=APITypes.drc)
     ztc_service = factory.SubFactory(ServiceFactory, api_type=APITypes.ztc)
     form_service = factory.SubFactory(ServiceFactory, api_type=APITypes.orc)
+    fetch_eherkenning_zaken_with_rsin = factory.Faker("boolean")
```
swrichards: I think maybe just `False` is more predictable here.

src/open_inwoner/openzaak/admin.py
```diff
@@ -77,7 +77,6 @@ class OpenZaakConfigAdmin(SingletonModelAdmin):
                 "fields": [
                     "skip_notification_statustype_informeren",
                     "reformat_esuite_zaak_identificatie",
-                    "fetch_eherkenning_zaken_with_rsin",
```
swrichards, line 80: I think we should re-add this to the ZGW API Group admin as well.

src/open_inwoner/configurations/bootstrap/zgw.py
```diff
@@ -28,7 +28,6 @@ class Meta:
                 "max_upload_size",
                 "skip_notification_statustype_informeren",
                 "reformat_esuite_zaak_identificatie",
-                "fetch_eherkenning_zaken_with_rsin",
```
swrichards, line 31: This should be added to the `ZGWAPIGroup` config model.

src/open_inwoner/openklant/migrations/0019_alter_klantensysteemconfig_primary_backend.py
```diff
@@ -1,6 +1,7 @@
 # Generated by Django 4.2.16 on 2025-01-16 14:41
 
 from django.db import migrations, models
+
 import open_inwoner.openklant.models
```
swrichards, line 5: Nitpick, but we might want to reduce the diff-noise here.

src/open_inwoner/openklant/models.py
```diff
@@ -217,26 +217,50 @@ class Meta:
         verbose_name = _("OpenKlant2 configuration")
 
 
+# Deprecated in favor of `validate_backend_choice`, which has more general use
```
swrichards, line 220: Maybe also add an explicit `TODO: Can be removed after migration openklant.00xx_yyyy has been squashed`

src/open_inwoner/openzaak/migrations/0060_zgwapigroupconfig_fetch_eherkenning_zaken_with_rsin.py
```diff
@@ -0,0 +1,62 @@
...
+            "No zgw_api_group_config found. "
+            "Skipping migration of fetch_eherkenning_zaken_with_rsin"
+        )
+        return
+    except MultipleObjectsReturned:
+        logger.info(
+            "Multiple zgw_api_group_config instances found. "
+            "Skipping migration of fetch_eherkenning_zaken_with_rsin"
+        )
+        return
```
swrichards: I don't think it's a problem to have the same value set on all group configs if they're set -- that is the status quo we're trying to preserve, that all groups show the same behavior.

src/open_inwoner/openzaak/migrations/0061_merge_20250124_1149.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        (
+            "openzaak",
+            "0060_remove_openzaakconfig_use_zaak_omschrijving_as_title_and_more",
+        ),
+        ("openzaak", "0060_zgwapigroupconfig_fetch_eherkenning_zaken_with_rsin"),
+    ]
```
swrichards, line 14: For my understanding, why was this necessary, can't we tweak the dependencies so that we have a linear history again?

src/open_inwoner/openzaak/models.py
```diff
@@ -252,6 +254,28 @@ def forms_client(self):
...
+        ),
+        help_text=_(
+            "If enabled, Zaken for eHerkenning users are fetched using the company RSIN (Open Zaak). "
+            "If not enabled, Zaken are fetched using the KvK number (eSuite)."
+        ),
+        default=False,
+    )
+
+    # flags related to other services
+    klant_backend = models.CharField(
```
swrichards, line 270: As discussed, I think this should be nullable: not all gemeentes actually use a klantensysteem (but perhaps that's for a subsequent PR if it requires too much logic changes).

swrichards (changes requested):

src/open_inwoner/openzaak/migrations/0061_merge_20250124_1149.py
pi-sigma: Short answer: lazyness. Will modify the migrations manually.

swrichards (approved):

PR #1583, opened 2025-01-28
Check for missing OpenKlant2 Service model in ok2 vragen service

swrichards:


src/open_inwoner/openklant/services.py
```diff
@@ -846,6 +846,9 @@ class OpenKlant2Service(KlantenService):
 
     def __init__(self, config: OpenKlant2Config | None = None):
         self.config = config or OpenKlant2Config.get_solo()
+        if not self.config.service:
```
pi-sigma, line 849:
```suggestion
        if not getattr(self.config, "service", None):
```
since `service` is nullable?

pi-sigma (changes requested):

src/open_inwoner/openklant/services.py
swrichards: I think the semantics of `get_solo()` means it's always `get_or_create` under the hood, no?

pi-sigma: You're right, didn't realize this applies to all fields

pi-sigma (approved):

PR #1585, opened 2025-01-28
Upgrade openklant2 client for 2.5.0

swrichards:


src/openklant2/types/resources/partij_identificator.py
```diff
@@ -7,12 +7,33 @@
...
 #
 
+CodeObjecttype = Literal[
+    "natuurlijk_persoon",
+    "niet_natuurlijk_persoon",
+    "overig",
+    "vestiging",
+]
+
+SodeSoortObjectId = Literal[
```
pi-sigma:
```suggestion
CodeSoortObjectId = Literal[
```

pi-sigma (changes requested):

src/openklant2/types/resources/partij_identificator.py
swrichards: Good catch!

pi-sigma (approved):

PR #1586, opened 2025-01-29
[#2993] Use custom page title for zaak detail page

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/2993

### DIFF
--- a/src/open_inwoner/templates/pages/cases/status_inner.html
+++ b/src/open_inwoner/templates/pages/cases/status_inner.html
@@ -1,5 +1,9 @@
 {% load i18n utils anchor_menu_tags card_tags file_tags form_tags grid_tags solo_tags link_tags button_tags icon_tags notification_tags %}
 
+{% block title %}
+    <title>{% if page_title %}{{ page_title }}{% else %}{% trans "Aanvraag" %}{% endif %}</title>
+{% endblock title %}
+
 {# Messages #}
 <div class="container container--no-margin" id="cases-status-messages-oob" hx-swap-oob="true">
     {% notifications messages closable=True %}
@@ -9,6 +13,13 @@
 
 <div class="case__grid">
 {% if case %}
+    {# Dynamically update page title after content is loaded via HTMX #}
+    <script>
+        document.body.addEventListener('htmx:afterSwap', function(event) {
+            document.title = "{{ case.description|default:'Aanvraag' }}";
+        });
+    </script>
+
     {% render_grid %}
         {% render_column span=12 %}
             {# Title/dashboard #}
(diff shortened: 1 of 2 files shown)


src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -1,5 +1,9 @@
 {% load i18n utils anchor_menu_tags card_tags file_tags form_tags grid_tags solo_tags link_tags button_tags icon_tags notification_tags %}
 
+{% block title %}
+    <title>{% if page_title %}{{ page_title }}{% else %}{% trans "Aanvraag" %}{% endif %}</title>
+{% endblock title %}
```
swrichards, line 5: I just wonder whether there's issues of having this in the inner template, for two reasons:

1. The `<title>` tag will only appear in the DOM tree after the HTMX has finished loading (which can be a few seconds). I am not sure this change will be properly picked up by screenreaders and announced, but perhaps @jiromaykin knows more here.
2. I think `<title>` has to be a child of `<head>`, having it nested under `<body>` is probably not valid spec-compliant (not sure that's a show stopper, but again, it might have a11y knock-on effects).

I don't really have a better option, either then setting the title via some kind of HTMX post-load handler (doing it in the main template would defeat the purpose of the async load). Happy to merge this and address it later if it turns out to be a problem.

swrichards (approved):

jiromaykin (changes requested):
@pi-sigma  You are misunderstanding/mixing-up a couple of things:
1. [Story nr.2397](https://taiga.maykinmedia.nl/project/open-inwoner/us/2397) is about the META title tags and how multiple pages are missing a proper <title> there, so they can be differentiated in the browse History, for example: all of the Flat pages.
So yeah: that is a TITLE META tag that indeed needs to be part of the HEAD tag, and you will have to resarch first, which pages have Titles missing, And a <title> tag will neverbe functional inside the Body tag.
Also: the unique title needs to come first, then a hyphen after which the Sitename can appear.
So this is a bigger/different issue.
2. We are only using HTMX to swap and load **_partial_** components in this project, so the DOM and it's META tags are already loaded first, so no worries for screenreaders there.
3. The cases status page already has a 'unique' meta title ("status") so you can either delete the stuff you made in this PR or add the status-name to the actual meta tag (not the content of the Body...).
3. Is this the only PR you are making for this issue? If so: this does not tackle that issue at all. 
4. Oh dear, we do need to make a separate issue for the Heading-level-1 titles that do not load immediately on pageload. Which indeed is bad for accessibility, but that's completely separet from this one. So I made [a new one here you do not need to worry about](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3000) right now.

PR #1597, opened 2025-02-04
Move eherkenning eenmanszaak flag to more natural SiteConfiguration

swrichards:
This flag didn't really make sense in the `OpenZaakConfig`, it's site-wide: I've moved it together with the other login related flags in the general `SiteConfiguration`.


src/open_inwoner/configurations/models.py
```diff
@@ -128,6 +128,16 @@ class SiteConfiguration(SingletonModel):
         verbose_name=_("Login tekst"),
         help_text=_("Deze tekst wordt getoond op de login pagina."),
     )
+    enable_eherkenning_for_eenmanszaak = models.BooleanField(
+        verbose_name=_("Allow eenmanszaken to authenticate using eHerkenning"),
+        help_text=_(
+            "If enabled, eenmanszaken may authenticate using eHerkenning and "
+            "subsequently their kvk nummer will be used to interface with the zaken and"
+            " klanten backends). If not, an eenmanszaak will be forced to use DigiD, "
```
pi-sigma:
```suggestion
            " klanten backends. If not, an eenmanszaak will be forced to use DigiD, "
```

pi-sigma (approved):

src/open_inwoner/configurations/models.py
swrichards: Good catch, thanks.

pi-sigma (approved):

PR #1599, opened 2025-02-05
Update uWSGI settings for Docker

swrichards:
This commit limits the number of uWSGI settings that
are externally configurable. We are opinionated with
regards to certain settings (mainly to do with process
management), for others we provide sane defaults but
allow overrides if they're mainly to do with the
surrounding environment and resource constraints
(threads, processes, timeouts, etc.).

---
Charts counterpart: https://github.com/maykinmedia/charts/pull/179

### DIFF
--- a/bin/docker_start.sh
+++ b/bin/docker_start.sh
@@ -26,46 +26,38 @@ export UWSGI_CHDIR="src"
 export UWSGI_STATIC_MAP="/static=/app/static /media=/app/media"
 
 # --- Process Management
-export UWSGI_MASTER=true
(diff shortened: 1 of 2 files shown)


swrichards:
Discussed with @SilviaAmAm -- we agreed that these settings are not a regression on the current configuration, and _probably_ an improvement, but there are still things we don't quite understand about uWSGI's inner workings. We agreed to approve this approach for now, but to iterate on this in the next sprint with a smaller lab Django project where we can reproduce some production issues (timeouts, thread deadlocks) to validate our assumptions.

alextreme (approved):

PR #1600, opened 2025-02-05
[#2967] Fix verification email not sent

pi-sigma:
The verification email flow was previously changed to account for duplicate emails sent during registration or profile edits. This had the effect that emails were *not* automatically sent at all when the user changed their email again.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2967

### DIFF
--- a/src/open_inwoner/accounts/views/registration.py
+++ b/src/open_inwoner/accounts/views/registration.py
@@ -245,12 +245,12 @@ class EmailVerificationUserView(LogMixin, LoginRequiredMixin, TemplateView):
                 get_next_url_from(self.request, default=reverse("pages-root"))
             )
 
-        # send verification email immediately on requesting page, but only once
-        if not request.session.get("verification_email_sent"):
+        # send verification email immediately on requesting page
+        if request.session.get("verification_email_sent_to") != user.email:
             send_user_email_verification_mail(
                 user, next_url=get_next_url_from(self.request, default="")
             )
-            request.session["verification_email_sent"] = True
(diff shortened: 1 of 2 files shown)


src/open_inwoner/mail/tests/test_verification.py
```diff
@@ -281,6 +282,8 @@ def test_unverified_user_redirects(self, mock_send):
         self.assertRedirects(response, verify_url + "?sent=1")
         response.follow(status=200)
 
+        # breakpoint()
+
```
swrichards:
```suggestion
```

This is the kind of thing where it would be [really nice to have Ruff](https://docs.astral.sh/ruff/rules/debugger/), going to try and make a point of that this sprint.

src/open_inwoner/accounts/views/registration.py
```diff
@@ -245,12 +245,13 @@ def get(self, request, *args, **kwargs):
                 get_next_url_from(self.request, default=reverse("pages-root"))
             )
 
-        # send verification email immediately on requesting page, but only once
-        if not request.session.get("verification_email_sent"):
+        # send verification email immediately on requesting page
+        verify_email = request.session.get("verification_email_sent_to")
+        if not verify_email or verify_email != user.email:
```
swrichards:
```suggestion
        if verify_email != user.email:
```

swrichards (changes requested):
LGTM, nitpicks only.

swrichards (approved):

PR #1601, opened 2025-02-05
Fix broken zgw config step

swrichards:


docker/setup_configuration/data.yaml
```diff
@@ -94,8 +94,7 @@ openzaak_config:
   max_upload_size: 50
   skip_notification_statustype_informeren: false
   reformat_esuite_zaak_identificatie: true
-  fetch_eherkenning_zaken_with_rsin: false
-  use_zaak_omschrijving_as_title: 'true'
+  derive_zaak_titel_from: zaaktype_omschrijving
```
pi-sigma, line 97: I don't see an assertion about this field in `test_configure_full_sets_the_correct_fields`. On the other hand, there are other fields missing from the test (e.g. there are no asserts for `zaken_filter_enabled` and `max_upload_size` either). Shouldn't these be added?

pi-sigma (changes requested):

docker/setup_configuration/data.yaml
swrichards: Good call: this does need all the asserts, easy to get drift here. Added.

pi-sigma (approved):

PR #1604, opened 2025-02-06
[#3007] Redirect from zaak to vraag created with OpenKlant

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3007

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -276,6 +276,43 @@ class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
     """
 
     def get(self, request, *args, **kwargs):
+        api_service = kwargs.get("api_service")
+
+        match api_service:
+            case KlantenServiceType.OPENKLANT2.value:
+                question_uuid = self._get_klantcontact_openklant(
+                    request, *args, **kwargs
+                )
+            case KlantenServiceType.ESUITE.value:
+                question_uuid = self._get_klantcontactmoment_esuite(
+                    request, *args, **kwargs
+                )
+            case _:
+                raise RuntimeError("Unsupported klanten backend")
+
+        return HttpResponseRedirect(
+            reverse(
+                "cases:contactmoment_detail",
+                kwargs={
+                    "api_service": api_service,
+                    "kcm_uuid": question_uuid,
+                },
+            )
+        )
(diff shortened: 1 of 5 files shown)


src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -276,6 +276,43 @@ class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
...
+
+        return HttpResponseRedirect(
+            reverse(
+                "cases:contactmoment_detail",
+                kwargs={
+                    "api_service": klant_config.primary_backend,
+                    "kcm_uuid": question_uuid,
+                },
+            )
+        )
```
swrichards: For my understanding: why can we assume it is the primary backend we need to redirect to, if both are possibilities?

src/open_inwoner/openzaak/tests/test_case_detail.py
```diff
@@ -2597,6 +2613,135 @@ def test_kcm_redirect(self, m):
...
+        #
+        # extra configs + mocks
+        #
+        self.klanten_service = ServiceFactory(
+            api_root=KLANTEN_ROOT, api_type=APITypes.kc
+        )
+        self.openklant_config.service = self.klanten_service
+        self.openklant_config.save()
+
+        # OpenKlant mocks
```
swrichards, line 2659: Let's validate these to guard against changes in the expected shapes.

swrichards: (Using the validator in our client module).

src/open_inwoner/accounts/views/contactmoments.py
swrichards: (My memory is a bit hazy on this point, but I recall we said the linkage between a zaak and klant systeem would be record on the API group, which I would then expect to be used here).

swrichards (changes requested):

src/open_inwoner/accounts/views/contactmoments.py
pi-sigma: You're right, we need to pass the `api_service` to the view via the url instead of checking the `KlantSysteemConfig`

pi-sigma:
Tackled in https://github.com/maykinmedia/open-inwoner/pull/1612

PR #1610, opened 2025-02-06
Fix uWSGI port config for Docker

swrichards:
The Docker uWSGI config had two issues:

- It didn't specfy an address to listen on, so it would only listen on localhost by default
- It used UWSGI_PORT to customize the port, which is not a recognized uWSGI option, but because it follows the naming scheme, was evaluated by uWSGI which then failed because of strict mode.

This commit changes this to always bind on all addresses, and to use the more conventional $PORT environment variable.


alextreme (approved):

alextreme:
Tagged with v1.27.0-rc2, thanks for the late-night fix!

PR #1612, opened 2025-02-10
[#3002] Support clickthrough from vraag to zaak

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3002

Depends on https://github.com/maykinmedia/open-inwoner/pull/1604

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -276,6 +276,43 @@ class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
     """
 
     def get(self, request, *args, **kwargs):
+        api_service = kwargs.get("api_service")
+
+        match api_service:
+            case KlantenServiceType.OPENKLANT2.value:
+                question_uuid = self._get_klantcontact_openklant(
+                    request, *args, **kwargs
+                )
+            case KlantenServiceType.ESUITE.value:
+                question_uuid = self._get_klantcontactmoment_esuite(
+                    request, *args, **kwargs
+                )
+            case _:
+                raise RuntimeError("Unsupported klanten backend")
+
+        return HttpResponseRedirect(
+            reverse(
+                "cases:contactmoment_detail",
+                kwargs={
+                    "api_service": api_service,
+                    "kcm_uuid": question_uuid,
+                },
+            )
+        )
+
+    def _get_klantcontact_openklant(self, request, *args, **kwargs):
+        service: OpenKlant2Service = OpenKlant2Service()
+        fetch_params = service.get_fetch_parameters(user=request.user)
+
+        question_dto, _ = service.retrieve_question(
+            fetch_params=fetch_params,
+            question_uuid=str(kwargs.get("uuid")),
+            user=request.user,
+        )
+
+        return question_dto.get("api_source_uuid")
+
+    def _get_klantcontactmoment_esuite(self, request, *args, **kwargs):
         vragen_service: VragenService = eSuiteVragenService()
         klanten_service: KlantenService = eSuiteKlantenService()
         fetch_params = klanten_service.get_fetch_parameters(self.request)
@@ -294,12 +331,4 @@ class KlantContactMomentRedirectView(KlantContactMomentAccessMixin, View):
         if not kcm:
             raise Http404
 
-        return HttpResponseRedirect(
-            reverse(
-                "cases:contactmoment_detail",
-                kwargs={
-                    "api_service": KlantenServiceType.ESUITE.value,
-                    "kcm_uuid": kcm.uuid,
-                },
-            )
-        )
+        return kcm.uuid
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -201,8 +201,11 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
         elif KlantenServiceType.OPENKLANT2.value in self.request.path:
             service = self.get_service(service_type=KlantenServiceType.OPENKLANT2)
 
+        origin = self.request.headers.get("Referer")
         question, zaak = service.retrieve_question(
-            self.get_fetch_params(service), kwargs["kcm_uuid"], user=self.request.user
+            self.get_fetch_params(service),
+            question_uuid=kwargs["kcm_uuid"],
+            user=self.request.user,
         )
         if not question:
             raise Http404()
@@ -247,7 +250,6 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
                 "value": question["channel"].capitalize(),
             },
         ]
-        origin = self.request.headers.get("Referer")
         if origin and reverse("cases:contactmoment_list") in origin:
             ctx["origin"] = {
                 "label": _("Terug naar overzicht"),
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1326,7 +1326,6 @@ class OpenKlant2Service(LogMixin, KlantenService):
         answers_for_klantcontact_uuid = {}
         question_uuids = []
         klantcontact_uuid_to_klantcontact_object = {}
-
         for klantcontact in self.klantcontacten_for_partij(
             partij_uuid, kanaal=self.config.mijn_vragen_kanaal
         ):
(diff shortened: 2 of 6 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
-        return self._build_question_dto(question_ok2=question, user=user), None
+        zaak_with_api_group = None
+
+        # fetch onderwerp_object linked to klantcontact
+        onderwerp_objecten = self.client.onderwerp_object.list()["results"]
+        onderwerp_object = next(
+            o
+            for o in onderwerp_objecten
+            if o["klantcontact"]["uuid"] == question.question_kcm_uuid
+        )
```
swrichards: Just checking our assumption here: one question can only be connected to 1 Zaak, correct? (That's the current behaviour, but it might be worth seeing if there are `>1` and logging an unexpected state error).

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
+            return self._build_question_dto(question_ok2=question, user=user), None
+
+        # find zaak_with_api_group which is linked to the question
+        for response in truthy_responses:
+            zaken, client = response.result, response.client
+            for zaak in zaken:
+                if (
+                    not zaak.identificatie
+                    == onderwerp_object["onderwerpobjectidentificator"]["objectId"]
+                ):
```
swrichards: Taking an advance on #1614 :
```suggestion
                if (
                    zaak.identificatie
                    != onderwerp_object["onderwerpobjectidentificator"]["objectId"]
                ):
```

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
+            for o in onderwerp_objecten
+            if o["klantcontact"]["uuid"] == question.question_kcm_uuid
+        )
+
+        # fetch zaken for user
+        groups = list(
+            ZGWApiGroupConfig.objects.filter(
+                klant_backend=KlantenServiceType.OPENKLANT2.value
+            )
+        )
```
swrichards: nitpick, but the `list()` is not needed here (I think the original snippet needed to index the group, but it doesn't appear we have to do so here).

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
+        proxy = MultiZgwClientProxy([group.zaken_client for group in groups])
+        proxy_response = proxy.fetch_cases(user_bsn=user.bsn)
+        truthy_responses = proxy_response.truthy_responses
+
+        if (case_count := len(truthy_responses)) == 0:
+            logger.error(
+                "Unable to find matched contactmomenten zaak with OpenKlant2 backend"
+            )
```
swrichards: I am not sure this is an _error_ per se, I think we'll want to avoid sending this to Sentry as it's possibly a user simply has no cases. Let's make it an `info` instead.

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
+
+        # find zaak_with_api_group which is linked to the question
+        for response in truthy_responses:
+            zaken, client = response.result, response.client
+            for zaak in zaken:
+                if (
+                    not zaak.identificatie
+                    == onderwerp_object["onderwerpobjectidentificator"]["objectId"]
+                ):
+                    continue
```
```suggestion
        for response in truthy_responses:
            zaken, client = response.result, response.client
            for zaak in filter(lambda z: z.identificatie == onderwerp_object["onderwerpobjectidentificator"]["objectId"], zaken):

```

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1407,55 @@ def retrieve_question(
...
 
-        # TODO
-        # should return (Question, zaak_with_api_group); the latter is left out until a
-        # standard for linking klantcontact + zaak is agreed upon
-        # https://github.com/Klantinteractie-Servicesysteem/KISS-frontend/issues/808#issuecomment-2357637675
-        return self._build_question_dto(question_ok2=question, user=user), None
+        zaak_with_api_group = None
+
+        # fetch onderwerp_object linked to klantcontact
+        onderwerp_objecten = self.client.onderwerp_object.list()["results"]
```
swrichards: Note that `list()` excepts `params` which could avoid some of the filtering below:

```
class OnderwerpobjectIdentificatorListParams(TypedDict, total=False):
    onderwerpobjectidentificatorCodeObjecttype: str
    onderwerpobjectidentificatorCodeRegister: str
    onderwerpobjectidentificatorCodeSoortObjectId: str
    onderwerpobjectidentificatorObjectId: str
```

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -247,7 +250,7 @@ def get_context_data(self, **kwargs):
                 "value": question["channel"].capitalize(),
             },
         ]
-        origin = self.request.headers.get("Referer")
+        # TODO: review
```
swrichards: What's the concern here?

swrichards (changes requested):

src/open_inwoner/accounts/views/contactmoments.py
pi-sigma: I was having issues with the redirect; works now.

src/open_inwoner/openklant/services.py
pi-sigma: We need to filter by checking the `uuid` of the associated `klantcontact`, not any of the features that identify the `onderwerp_object` itself (like object id), hence I don't think we can avoid the filtering.

swrichards: Right, clear.

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1405,61 @@ def retrieve_question(
...
+            ):
+                group = ZGWApiGroupConfig.objects.resolve_group_from_hints(
+                    client=client
+                )
+                zaak_with_api_group = ZaakWithApiGroup(zaak=zaak, api_group=group)
+                if case_count > 1:
+                    # In principle this should not happen, a zaak should be stored in
+                    # exactly one backend. But: https://www.xkcd.com/2200/
+                    # We should at least receive a ping if this happens.
+                    logger.error("Found one zaak in multiple backends")
```
swrichards: One more issue here: the `case_count` is the wrong variable to check here, because we only care about cases that _match the `objectId`. I think we can just log the error if `zaak_with_api_group is not None`: then we'll default to taking the most recent rather than the first value, but that's equally legitimate to taking the first I think.

swrichards (changes requested):

src/open_inwoner/openklant/services.py
pi-sigma: I was indeed confused about `case_count` (copy/paste from above, where we deal with a single case). But I don't think we can just check `zaak_with_api_group is not None`, because we do need at least one. We need to check separately (a) that we have a unique zaak, (b) the zaak is found in a unique backend. It's tricky because we're dealing with multiple cases instead of just one.

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1405,80 @@ def retrieve_question(
...
+        if len(zaken_for_question) > 1:
+            logger.error(
+                "More than one zaak found for question %s",
+                question.question_kcm_uuid,
+            )
+            return self._build_question_dto(question_ok2=question, user=user), None
+
+        zaak_with_api_group = ZaakWithApiGroup(
+            zaak=zaken_for_question.pop(),
+            api_group=api_groups.pop(),
```
swrichards: I am not sure this will work: I think `pop` on a `set` just returns a random element, and even then, `api_groups` and `zaken_for_questions` are not guaranteed to be the same size as far as I can tell, so we might match the wrong group to the zaak, or am I missing something?

src/open_inwoner/openklant/services.py
```diff
@@ -1404,11 +1405,66 @@ def retrieve_question(
...
+            logger.info(
+                "Unable to find matched contactmomenten zaak with OpenKlant2 backend"
+            )
+            return self._build_question_dto(question_ok2=question, user=user), None
+
+        # find the unique zaak for the question
+        zaken_for_question = []
+        for response in truthy_responses:
+            # discard the client for determining the api_group; we only need the zaak
+            zaken, _ = response.result, response.client
```
```suggestion
            zaken = response.result
```

swrichards (approved):
One nitpick, but otherwise LGTM

swrichards (approved):

PR #1613, opened 2025-02-10
Add ruff to ci

swrichards:


ruff.toml
```diff
@@ -0,0 +1,73 @@
...
+    # Line too long
+    "E501",
+    # Do not assign a lambda expression
+    "E731",
+    # Name may be undefined from '*' import
+    "F405",
+    # Blank line contains whitespace
+    "W293",
+    # Trailing whitespace
+    "W291",
```
pi-sigma: This seems like a useful check. Why ignore it? Have to admint I'm biased on this one. I've configured my editor to automatically delete trailing whitespace on save, so whenever I come across a file that contains trailing whitespace somewhere this is a bit of a nuisance.

pi-sigma (changes requested):
- Why was `src/openklant2/tests/test_partij.py` modified?
- Minor suggestion about the check concerning trailing whitespace

ruff.toml
swrichards: Removed, good call. (The others are a bit trickier, too many hits to fix in one quick pass).

PR #1616, opened 2025-02-12
Apply ruff pep8-naming rule

pi-sigma:
https://docs.astral.sh/ruff/rules/#pep8-naming-n

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -52,6 +52,9 @@ select = [
     "B006",
     "B010",
     "B904",
+    # https://docs.astral.sh/ruff/rules/#pep8-naming-n
+    "N",
+>>>>>>> 1fccb49c (Enable ruff pep8-naming rule)
 ]
 ignore = [
     # Whitespace before ':' (conflicts with Black)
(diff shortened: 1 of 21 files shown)


ruff.toml
```diff
@@ -63,13 +65,28 @@ ignore = [
     # Use a single `with` statement with multiple contexts instead of nested `with` statements
     "SIM117",
     # Checks for nested if statements that can be collapsed into a single if statement.
-    "SIM102"
+    "SIM102",
+    # Variables in function scope should be lower-case
+    "N806",
+    # Cariables in class scope should not be mixed-case
```
swrichards:
```suggestion
    # Variables in class scope should not be mixed-case
```

src/open_inwoner/qmatic/client.py
```diff
@@ -85,7 +85,7 @@ class NoServiceConfigured(RuntimeError):
 # API CLIENT IMPLEMENTATIONS, per major version of the API
 
 
-def QmaticClient() -> "Client":
+def make_qmatic_client() -> "Client":
```
swrichards: I think this is more in line with what we use elsewhere (e.g. for zgw clients):

```suggestion
def qmatic_client_factory() -> "Client":
```

swrichards (approved):
LGTM, just a few small suggestions

PR #1617, opened 2025-02-12
[#venray-144] Expose case description (zaak.omschrijving) in email templates

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/gemeente-venray-open-formulier/issue/144

### DIFF
--- a/src/open_inwoner/openzaak/notifications.py
+++ b/src/open_inwoner/openzaak/notifications.py
@@ -153,6 +153,7 @@ def send_case_update_email(
     template = find_template(template_name)
     context = {
         "identification": case.identification,
+        "case_description": case.omschrijving,
         "type_description": case.zaaktype.omschrijving,
         "start_date": case.startdatum,
         "end_date": date.today() + timedelta(days=config.action_required_deadline_days),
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/notifications.py
```diff
@@ -153,6 +153,7 @@ def send_case_update_email(
     template = find_template(template_name)
     context = {
         "identification": case.identification,
+        "case_description": case.omschrijving,
```
swrichards, line 156: If I understand the ticket correctly, they want the same title displayed as on the case detail page, which currently would be `case.description` (which contains logic to observe the flag) rather than `omschrijving` which is a direct ZGW field. Did you read it differently?

swrichards (changes requested):

src/open_inwoner/openzaak/notifications.py
pi-sigma: Yes, I thought they explicitly want the `omschrijving` of the particular `zaak` in addition to the `omschrijving` of the `zaaktype`: "Daar staat standaard de omschrijving van het zaaktype. Maar dat zegt de aanvrager niets." (see also Alex's comment). If `case.description` resolves to the `omschrijving` of the `zaaktype` because they've made that choice, we would end up displaying the `omschrijving` of the `zaaktype` twice.

Incidentally, if it's generally true that the description of the `zaaktype`  "zegt de aanvrager niets", one might wonder why that's an option for `case.description` to begin with - but that's a separate issue.

swrichards: Check, thanks for the explanation.

swrichards (approved):

PR #1620, opened 2025-02-13
[#3033] Support multiple phone numbers

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3033


src/open_inwoner/openklant/services.py
```diff
@@ -283,6 +283,12 @@ def update_user_from_klant(self, klant: Klant, user: User):
         if klant.telefoonnummer and klant.telefoonnummer != user.phonenumber:
             update_data["phonenumber"] = klant.telefoonnummer
 
+        if (
+            klant.telefoonnummerAlternatief
+            and klant.telefoonnummerAlternatief != user.phonenumber_alternative
+        ):
+            update_data["phonenumber_alternative"] = klant.telefoonnummerAlternatief
+
```
swrichards, line 298: I think we'd also want to implement this in the OK2 version, where we would add it as another digital address.

The difficulty might be that the primary phone number needs to have the `isStandaardAdres` flag explicitly set to true, and the alternative phonenumber must have it explicitly set to false.

src/open_inwoner/accounts/forms.py
```diff
@@ -111,6 +111,7 @@ class Meta:
             "infix",
             "last_name",
             "phonenumber",
+            "phonenumber_alternative",
```
swrichards, line 114: There's one more caveat here (apologies, I should've added this to the card): eSuite requires the primary to be set before an alternative can be added, and we should really enforce the same behavior at the form validation level (at least) and arguably also the database level. That is: you cannot have a value for the alternative, but not for the primary. Should've mentioned this earlier, my bad.

swrichards (changes requested):

src/open_inwoner/accounts/forms.py
```diff
@@ -119,12 +120,21 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        # make phonenumber required when 2fa-sms login is enabled
         config = SiteConfiguration.get_solo()
-        if not config.login_2fa_sms:
-            del self.fields["phonenumber"]
-        else:
-            self.fields["phonenumber"].required = True
+        self.fields["phonenumber"].required = bool(config.login_2fa_sms)
```
swrichards: `config.login_2fa_sms` is already a non-nullable bool, so should be able to just do:

```suggestion
        self.fields["phonenumber"].required = config.login_2fa_sms
```

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -295,14 +307,19 @@ def test_update_partij_from_user(self):
...
-            ],
-            [("email", "user@bar.com"), ("telefoonnummer", "0644938475")],
+            set([(adres["soortDigitaalAdres"], adres["adres"]) for adres in adressen]),
+            set(
+                [
+                    ("email", "user@bar.com"),
+                    ("telefoonnummer", "0644938475"),
+                    ("telefoonnummer", "0687654321"),
+                ]
+            ),
```
swrichards:
```suggestion
            set([(adres["soortDigitaalAdres"], adres["adres"], adres["isStandaardAddress"]) for adres in adressen]),
            {
                    ("email", "user@bar.com", False),
                    ("telefoonnummer", "0644938475", True),
                    ("telefoonnummer", "0687654321", False),
            },
```

src/open_inwoner/accounts/models.py
```diff
@@ -324,31 +331,10 @@ class Meta:
...
-            #     name="unique_bsn_when_digid",
-            # ),
-            # CheckConstraint(
-            #     # maybe this is not correct?
-            #     check=(Q(bsn="") | Q(bsn__isnull=True))
-            #     | ~Q(login_type=LoginTypeChoices.digid),
-            #     name="check_digid_bsn_required_when_digid",
-            # ),
+            models.CheckConstraint(
+                check=~Q(phonenumber__exact="") | Q(phonenumber_alternative__exact=""),
```
swrichards, line 335: I was expecting:

```suggestion
                check=Q(phonenumber__exact="") & ~Q(phonenumber_alternative__exact=""),
```

That is: we should avoid a primary being unset _and_ a secondary being set.

src/open_inwoner/openklant/services.py
```diff
@@ -1096,19 +1105,24 @@ def get_or_create_digitaal_adres(
         partij_uuid: str,
         soortAdres: Literal["email", "telefoonnummer"],
         adres: str,
+        isStandaardAdres: bool = False,
     ) -> tuple[DigitaalAdres, bool]:
         digitale_adressen = self.filter_digitale_addressen_for_partij(
```
swrichards, line 1110: Filed an issue to add `isStandaardAdres` to the list filters (https://github.com/maykinmedia/open-klant/issues/338).

src/open_inwoner/openklant/services.py
```diff
@@ -1134,7 +1148,11 @@ def update_user_from_partij(self, partij_uuid: str, user: User):
         if phone_adressen := self.filter_digitale_addressen_for_partij(
             partij_uuid, soortDigitaalAdres="telefoonnummer", adressen=adressen
         ):
-            update_data["phonenumber"] = phone_adressen[0]["adres"]
+            for adres in phone_adressen:
+                if adres["isStandaardAdres"]:
+                    update_data["phonenumber"] = adres["adres"]
+                else:
+                    update_data["phonenumber_alternative"] = adres["adres"]
```
swrichards, line 1155: There's a slight edge case here in that there may be `>1` address for which `isStandardAdresFalse`. This is another one of those "not much we can do" cases as this is not enforced, but it might be worth explicating in a log call that there were more addresses and that we picked the last one in the last (I would also update the test to this effect).

src/open_inwoner/openklant/services.py
```diff
@@ -1096,19 +1105,24 @@ def get_or_create_digitaal_adres(
...
+        isStandaardAdres: bool = False,
     ) -> tuple[DigitaalAdres, bool]:
         digitale_adressen = self.filter_digitale_addressen_for_partij(
             partij_uuid, soortDigitaalAdres=soortAdres
         )
         for digitaal_adres in digitale_adressen:
-            if digitaal_adres["adres"] == adres:
+            if (
+                digitaal_adres["adres"] == adres
+                and digitaal_adres["isStandaardAdres"] is isStandaardAdres
```
swrichards: Because we're not directly comparing to `True`/`False` I feel that `==` might be a bit more readable than `is`, but that's a nitpick.

```suggestion
                and digitaal_adres["isStandaardAdres"] == isStandaardAdres
```

src/openklant2/types/resources/digitaal_adres.py
```diff
@@ -36,6 +36,7 @@ class DigitaalAdres(TypedDict):
     adres: str
     omschrijving: str
     soortDigitaalAdres: SoortDigitaalAdres
+    isStandaardAdres: NotRequired[bool]
```
swrichards: I think in the _output_ this field is required, but not in the create/patch data.

swrichards (changes requested):

src/openklant2/types/resources/digitaal_adres.py
pi-sigma: The docs don't specify it as required in either case. I did add it to the create data though.

src/open_inwoner/accounts/models.py
pi-sigma: The check passes if the condition evaluates to `True` (or `Null`, which is not relevant though since the phone number fields are non-nullable). The condition here describes the allowed case: primary number is not empty or secondary is empty.

src/openklant2/types/resources/digitaal_adres.py
swrichards: I am a bit confused by this myself, but my understanding is that the schema of the _read_ endpoints always contains _all_ fields, but a field might be nullable. That is, I don't think fields are ever omitted from a response, though they might be for a write endpoint (because there the API can resort to a default). Hence I am not sure if `NotRequired` ever applies to read endpoints, it would imply the shape of the response changes, which is not impossible but is usually reserved for sub-parts of an object (e.g. an attribute `details` can differ based on an `objectType` field). But the OpenAPI docs are peculiar: because they do in fact list some fields as required and `isStandaardAdres` as not, but I doubt the field is ever omitted and that this is a shortcoming of the OpenAPI Spec (or how we generate the spec).

pi-sigma: Thanks for the explanation, I wasn't aware of this peculiarity of the OpenAPI spec (or our use of it)! In that case I will update the dicts so that they reflect the actual input/output requirements.

src/open_inwoner/accounts/models.py
swrichards: Got it, confusion from my end, thanks.

swrichards (approved):

PR #1623, opened 2025-02-14
[#3045] Add option to enable/disable indexing by search engines

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3011?milestone=523


swrichards (changes requested):
LGTM, but two points:

1. We also have a `base_pdf.html` which might also avail itself of the flag.
2. We also have a bunch of error templates (`403.html`, etc). While we're doing this, it would make sense to me to simply add a static noindex to those (search engines should never pick those up, regardless of the environment).

src/open_inwoner/templates/master.html
```diff
@@ -2,6 +2,7 @@
 <html lang="nl" class="view utrecht-document openinwoner-theme {% block view_class %}view--{{ request.resolver_match.namespaces|join:'-' }}-{{ request.resolver_match.url_name }}{% endblock %}">
     <head>
         <meta charset="utf-8">
+        <meta name="robots" {% if enable_indexing %}content="all"{% else %}content="noindex"{% endif %}>
```
swrichards: I think it might be good to have a single unit test for this (perhaps just loading the index page and ensuring the flag is observed).

swrichards:
```suggestion
        <meta name="robots" content="{% if enable_indexing %}all{% else %}noindex{% endif %}">
```

src/open_inwoner/templates/403.html
```diff
@@ -1,6 +1,8 @@
 {% extends 'master.html' %}
 
-{% block menu %}{% endblock %}
+{% block extra_head %}
+    <meta name="robots" content="noindex">
+{% endblock extra_head %}
```
swrichards: Sorry I didn't notice this earlier, but I didn't realize they all inherit from `master.html`, which means we'll have the tag twice, and I am not sure what that entails for the behavior or validity of the page. But if we try to fix this at the flag level, it becomes more complicated, because then we have to also branch on whether this is an error page or not. So perhaps it makes more sense to not add this to the error pages for now, to avoid the danger the DOM tree becomes invalid.

pi-sigma: Double-checked: overriding the tag via template inheritance simply replaces the original (works like regular inheritance), so no danger of inconsistency.

swrichards: That's strange though, because the first `<meta name="robots" ...` tag is not in the `extra_head` block.

When you say "replace", do you mean you only see it once in the node inspector? (It might still be duplicated at the raw HTML level).

pi-sigma: You were right about the template inheritance. Strange that it seemed to work previously but I now see the original tag from `master.html`. Can be solved by placing the `meta` tag in `master` in its own block. However, there is one template for which this doesn't work: the error template for the case pages: `templates/pages/cases/403.html`. I suspect the problem here is the same as with updating the title of the case detail page (content is loaded via HTMX).

swrichards (approved):

PR #1626, opened 2025-02-18
[#3052] Fix clickthrough zaak -> vraag

pi-sigma:
Clicking through from a zaak to an associated vraag results in `AttributeError` for vragen fetched with OpenKlant.

Cause: our use of the `MultiZgwClientProxy` results is inconsistent: sometimes we use it to return a zaak, sometimes a tuple (zaak, zgw_api_group).

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3052

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -202,7 +202,7 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             service = self.get_service(service_type=KlantenServiceType.OPENKLANT2)
 
         origin = self.request.headers.get("Referer")
-        question, zaak = service.retrieve_question(
+        question, zaak_with_api_group = service.retrieve_question(
             self.get_fetch_params(service),
             question_uuid=kwargs["kcm_uuid"],
             user=self.request.user,
@@ -220,16 +220,16 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             local_kcm.save()
 
         ctx["question"] = question
-        ctx["zaak"] = zaak.zaak if zaak else None
+        ctx["zaak"] = getattr(zaak_with_api_group, "zaak", None)
         case_url = (
             reverse(
                 "cases:case_detail",
                 kwargs={
-                    "object_id": str(zaak.zaak.uuid),
-                    "api_group_id": zaak.api_group.id,
+                    "object_id": str(zaak_with_api_group.zaak.uuid),
+                    "api_group_id": zaak_with_api_group.api_group.id,
                 },
             )
-            if zaak
+            if zaak_with_api_group
             else None
         )
         ctx["metrics"] = [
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1464,10 +1467,18 @@ def retrieve_question(
                 question.question_kcm_uuid,
             )
             return self._build_question_dto(question_ok2=question, user=user), None
+        if len(clients) > 1:
+            logger.error("Found one zaak in multiple backends")
+            return self._build_question_dto(question_ok2=question, user=user), None
```
swrichards, line 1511: I wonder if it wouldn't make more sense to still return the first but error loudly to Sentry as we do here?

swrichards: So just error but otherwise continue as below:

```suggestion
        if len(clients) > 1:
            logger.error("Found one zaak in multiple backends")
```

The reasoning here would be that this is an invariant violation that still permits a sane default (show the first), and indeed that's the behavior the user would've experienced before adding the unwarranted second zaak. It would be undesirable to break that behavior.

On the other hand, the concept of "first" here is not deterministic, it depends on which API responds first. So that could lead to arbitrary behavior. We could perhaps fetch all and always return the oldest, and still error, so we have consistent behavior.

What do you think?

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -220,16 +220,16 @@ def get_context_data(self, **kwargs):
             local_kcm.save()
 
         ctx["question"] = question
-        ctx["zaak"] = zaak.zaak if zaak else None
+        ctx["zaak"] = zaak_with_api_group.zaak if zaak_with_api_group else None
```
swrichards: Might be more readable:

```suggestion
        ctx["zaak"] = getattr(zaak_with_api_group, 'zaak', None)
```

swrichards (changes requested):

src/open_inwoner/openklant/services.py
pi-sigma: I agree that keeping the behavior consistent would be good, but I'm not sure the first or oldest zaak is always a sane default. Consider this scenario: a single question happens to be about z1 and z2 such that z1 < z2 in the order of time (i.e. `start_date`), but z1 has been closed already while z2 is still open. In that case we would return z1 as the topic of the question while it would make more sense to return z2, because arguably z2 is more relevant to the client. Or the other way around. A sane default would be the zaak that is most likely the most relevant for the client. But just because one zaak is older/newer doesn't mean that it's more relevant.

swrichards: That's a good point. I am still uncomfortable with not showing _any_ linked case here, I'd like to add it as a sub-task to https://taiga.maykinmedia.nl/project/open-inwoner/us/3039?no-milestone=1 -- because this is a case where a user expectation could be violated (they might know about a zaak and see it disappear later). Then we can revisit it more properly when we get around to that story. Merging this now.

pi-sigma: Sounds good. I'll create the task.

swrichards (approved):

PR #1627, opened 2025-02-19
[#3048] Fix meervoud begeleider niet correct

linssen814:
Changed "Mijn begeleider" to its plural form (i.e. "Mijn begeleiders") to reflect the fact that a citizen can have multiple mentor contacts.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3048

### DIFF
--- a/src/open_inwoner/templates/pages/profile/me.html
+++ b/src/open_inwoner/templates/pages/profile/me.html
@@ -206,12 +206,20 @@
                 {% endif %}
                         <div class="card__body">
                             <div class="ellipsis">
-                                <p class="card__heading-4"><span class="link link__text">{% trans "Mijn begeleider" %}</span></p>
+                                <p class="card__heading-4">
+                                    <span class="link link__text">
+                                        {% if mentor_contacts|length < 2 %}
+                                            {% trans "Mijn begeleider" %}
+                                        {% else %}
+                                            {% trans "Mijn begeleiders" %}
+                                        {% endif %}
+                                    </span>
+                                </p>
                                 {% render_list %}
                                     {% for name in mentor_contact_names %}
                                         {% list_item text=name compact=True strong=False %}
                                     {% empty %}
-                                        <li class="card__text--small">{% trans "U heeft geen gemeentelijke begeleider." %}</li>
+                                        <li class="card__text--small">{% trans "U heeft geen gemeentelijke begeleiders." %}</li>
                                     {% endfor %}
                                 {% endrender_list %}
 


src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -198,12 +198,12 @@ <h2 class="utrecht-heading-2" id="overview">{% trans "Overzicht" %}</h2>
                 {% endif %}
                         <div class="card__body">
                             <div class="ellipsis">
-                                <p class="card__heading-4"><span class="link link__text">{% trans "Mijn begeleider" %}</span></p>
```
jiromaykin, line 209: This task is about what users see in reality, so you will have to account for multiple situations.

1. Make front-end logic so the display shows the single form when there is only 1 mentor, and let it show the plural if there are more
2. Make sure you do a replacement of the entire word begeleider/begeleiders so not just the letter 'S' since it is a translated block (plurals in different languages work differently)
3. Check in the back-end to see if you can get the length of the mentor_contact_names from a proper list (or dictionary) so it acts properly on arguments

So in the front-end you will need to check if this works correctly for **zero and 1** mentor,  and more than 1 mentors (just an example):

```
{% if mentor_contact_names|length < 2 %}
    {% trans "Mijn begeleider" %}
{% else %}
    {% trans "Mijn begeleiders" %}
{% endif %}
```

Or if it's a dictionary/object: something like 

```
{% with mentor_count=mentor_contact_names|length %}
...bla
{% if mentor_count < 2 %}
...bla
```

jiromaykin (changes requested):

src/open_inwoner/templates/pages/profile/me.html
linssen814: I actually considered this, but thought that it could be potentially misleading to display the plural/singular based on the amount of mentors you have, since when you have 0 or 1 mentor, "Mijn begeleider" seems to imply that you can only have 1 mentor even though you can actually have multiple. I guess it does make more sense visually though.

jiromaykin: I am not sure if the multiple is the most common situation; if so, then your solution is indeed correct. 
In these cases it is better to ask in Taiga if the municipality can clarify what they think is better. I can approve either way but would prefer Daan or Bobby to specify the task.

linssen814: I asked the municipality and they do actually prefer your solution so I'll stick to that.

jiromaykin (approved):
Seems to work!
Now all you have to do is:

1. Add the PR link to Github in the Taiga issue/comment as well  (like this for example: https://taiga.maykinmedia.nl/project/open-inwoner/task/3017)
2. For the future: keep the first old commit intact, and only add 1 new one that is named something like "processed codereview for 'meervoud begeleider') after which you can do rebasing etc.

PR #1628, opened 2025-02-19
[#3047] Add option to send confirmation email for creation of klantcontact

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3047


src/open_inwoner/openklant/models.py
```diff
@@ -95,6 +95,15 @@ class ESuiteKlantConfig(SingletonModel):
         default=list,
         validators=[validate_array_contents_non_empty],
     )
+    send_klantcontact_confirmation_email = models.BooleanField(
+        verbose_name=_("Send confirmation email for registration of klantcontact"),
+        default=False,
+        help_text=_(
+            "If enabled, a confirmation email will be sent to the email address specified in the "
+            "contactmoment. If no email is given, the email will be sent to the address of "
+            "the klant."
```
swrichards: With this help text the flag naming seems slightly off to me: `send_klantcontact_confirmation_email` suggests this is about whether to send the email at all, but the help texts suggests it's about which address to use. 

The docs suggest the eSuite flag is about sending the email at all:

```
	
boolean (Verzend bevestigingsmail)
Default: false

Geeft aan of een bevestigingsmail verzonden moet worden naar de klant. De mail wordt verzonden naar het emailadres in de contactgegevens van het contactmoment. Indien deze leeg is naar het emailadres van de Klant. Indien deze ook leeg is, wordt de mail niet verzonden.
```

I forget, is this to do with the fact that will send our own if eSuite does not?

swrichards (approved):
LGTM, but take a look at my comment re naming.

src/open_inwoner/openklant/models.py
pi-sigma: This is indeed about whether the the email is sent at all. At second glance, I agree that my help text isn't the most helpful. I'll rephrase it.

swrichards (approved):

PR #1632, opened 2025-02-21
💄 [#3016] Desktop search filters - new design

stefrado:
- ✨ Hide filters if there are no choices available.
- 💄 Implement new search filter UI.
- 🎨 Open always open the first filter on page load.

Combine 2 issues: 
https://taiga.maykinmedia.nl/project/open-inwoner/task/3016
https://taiga.maykinmedia.nl/project/open-inwoner/task/3018


src/open_inwoner/utils/templatetags/utils.py
```diff
@@ -106,3 +106,13 @@ def cookies_accepted(request: HttpRequest) -> bool:
 @stringfilter
 def markdown(value):
     return md.markdown(value, extensions=["markdown.extensions.fenced_code"])
+
+
+@register.simple_tag
```
jiromaykin: We want less inbetween-layer template tags, not more
(see https://taiga.maykinmedia.nl/project/open-inwoner/issue/2062)
if this is just for a checked or 'toggle opened' state (?), this is not necessary. and should be solved with JS/scss.

(Also it's not quite a util, so would preferably be located inside `src/open_inwoner/components/templatetags/form_tags.py` )

src/open_inwoner/scss/components/SearchResults/SearchResults.scss
```diff
@@ -4,6 +4,10 @@
     flex-direction: column;
     gap: var(--spacing-large);
 
+    @media (min-width: 768px) {
+      margin-top: 3.375rem; // 54px
```
jiromaykin: I agree REMs are better but the project is pixelbased. Also: we need to be able to adjust spaces in future with design-tokens. So municipalities can adjust spacing themselves (if they use different font etc.) 
Using `calc(2 * var(--spacing-giant) - var(--spacing-small))` could be another solution

src/open_inwoner/scss/components/Filter/Filter.scss
```diff
@@ -1,22 +1,42 @@
...
+  }
+
+  &__reset {
+    display: flex;
+    justify-content: flex-end;
+
+    .button {
+      height: fit-content;
+      padding: 0;
+      margin-bottom: 0.75rem; // 12px
```
jiromaykin: Again not sure if REMs are the way to go when municipalities will need to change margin-values with design-tokens.
This one has a tricky top-margin. In the design the distance between "Wis alle?" and the text with number of Resultaten (search-results__title-small) is exactly 60 pixels, but will need to be adjust due to inner padding etc.

src/open_inwoner/scss/components/Form/Checkbox.scss
```diff
@@ -1,4 +1,42 @@
...
+    &:has(input:checked) {
+      color: var(--color-font-primary);
+      background: var(--color-primary);
+      border-color: var(--color-primary);
+
+      .checkbox__label:after {
+        content: '';
+        display: block;
+        background-image: url("data:image/svg+xml,%3Csvg width='20' height='21' viewBox='0 0 20 21' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cg clip-path='url(%23clip0_2541_1098)'%3E%3Cpath d='M15.1263 5.84163L10.8215 10.1464L10.468 10.5L10.8215 10.8535L15.1263 15.1583L14.6584 15.6262L10.3536 11.3214L10.0001 10.9679L9.64653 11.3214L5.34175 15.6262L4.87386 15.1583L9.17863 10.8535L9.53219 10.5L9.17863 10.1464L4.87385 5.84163L5.34175 5.37373L9.64653 9.67851L10.0001 10.0321L10.3536 9.67851L14.6584 5.37373L15.1263 5.84163Z' fill='white' stroke='white'/%3E%3C/g%3E%3Cdefs%3E%3CclipPath id='clip0_2541_1098'%3E%3Crect width='20' height='20' fill='white' transform='translate(0 0.5)'/%3E%3C/clipPath%3E%3C/defs%3E%3C/svg%3E%0A");
+        width: 1.25rem;
```
jiromaykin: Width and height could perhaps better be `var(--font-size-body-large)`, since we usually use the Material-Icons font for these, but I know there is no Close icon in UTF 😞

src/open_inwoner/scss/components/Form/Checkbox.scss
```diff
@@ -1,4 +1,42 @@
 .checkbox {
+  &--tag {
+    display: flex;
+    max-height: 33px;
+    padding: var(--spacing-small) var(--spacing-medium);
+    border-radius: var(--border-radius);
+    color: var(--color-gray-dark);
+    background: var(--color-gray-lightest);
+    border: 1px solid var(--color-gray-light);
+    width: fit-content;
```
jiromaykin, line 10: Note for future: not sure yet if we need to set max-width to 320px and use ellipsis overflow/white-space on the Label text only, (but you may need to separate the close-icon and just add it as an initially invisible material-icon instead). Since I am not sure if ellipsis on interactive elements are allowed (users need to be able to see the entire text).

src/open_inwoner/search/tests/test_page.py
```diff
@@ -319,6 +333,8 @@ def _test_search(checkbox_name, expected_text):
                 expected_text
             )
 
+        print("Tag 1", self.product1.name)
```
jiromaykin: The 'name' should be coming from `self.tag1 = TagFactory.create(name="Tag 1")` I think? 
I'd really ask @pi-sigma if he can help out with this failing test. 
Since all the proper elements seem to be there, so I am unsure how to solve this 😞

jiromaykin (changes requested):
Even though tests are not passing yet, I'm adding a review so we can rework it before final review.

src/open_inwoner/scss/components/Form/Checkbox.scss
stefrado: Will update the value to the `var()`!

src/open_inwoner/scss/components/Form/Checkbox.scss
stefrado: Tbh I don't think the overflow ellipsis on an is what we want without giving the users another option to see te full text. But we need to find a durable solution that also solves the text size issues occuring with the accessibility options (text-size and dyslexia font).

src/open_inwoner/utils/templatetags/utils.py
stefrado: 👍 I'll come back on this later today.

src/open_inwoner/templates/pages/search.html
```diff
@@ -13,35 +13,10 @@ <h1 class="utrecht-heading-1">{% trans "Zoekresultaten voor " %} "{% firstof sea
...
-                    {% include "components/Filter/Filter.html" with field=search_form.categories form_id="search-form" initial_open=open_index|is_same:0 only  %}
-                {% endif %}
-                {% if search_filter_tags and search_form.tags.field.choices|length != 0  %}
-                    TAGS
-                    {% include "components/Filter/Filter.html" with field=search_form.tags form_id="search-form" initial_open=open_index|is_same:1 only %}
-                {% endif %}
-                {% if search_filter_organizations and search_form.organizations.field.choices|length != 0 %}
-                    {% include "components/Filter/Filter.html" with field=search_form.organizations form_id="search-form" initial_open=open_index|is_same:2 only %}
-                {% endif %}
+                {% search_filters search_form form_id="search-form" search_filter_categories search_filter_tags search_filter_organizations %}
```
pi-sigma: Keyword arguments (`form_id="search-form"`) must come after positional arguments (`search_filter_categories` etc.). This causes a `TemplateSyntaxError` which in turn causes `expect(page.locator(".search-results__item")).to_have_count(2)` to fail.

src/open_inwoner/search/tests/test_page.py
```diff
@@ -312,7 +312,6 @@ def _click_checkbox_for_name(page=page, name=''):
 
             # Open het <details>-element
             # details_element.click()
```
pi-sigma: You need to query the `detail_element` with `get_by_text` and pass in the `name` to get the correct locator:

```
details_element = page.locator(".filter").filter(has=page.get_by_text(name))
details_element.click()
```

This will make the relevant `details_element` (a different one for each call to `_click_checkbox_for_name` visible

src/open_inwoner/scss/components/Filter/FilterDropdown.scss
```diff
@@ -0,0 +1,106 @@
+// .filter-dropdown {
```
jiromaykin: If we don't need this, it can be removed :-)

src/open_inwoner/scss/components/_index.scss
```diff
@@ -26,6 +26,8 @@
 @import './File/File.scss';
 @import './File/FileList.scss';
 @import './Filter/Filter.scss';
+@import './Filter/FilterDropdown';
```
jiromaykin: Since all code is commented out, FilterDropdown can be removed.

src/open_inwoner/templates/pages/search.html
```diff
@@ -12,23 +12,13 @@ <h1 class="utrecht-heading-1">{% trans "Zoekresultaten voor " %} "{% firstof sea
 
     {#  facets and results section #}
     <div class="grid">
-    {% if paginator.count %}
```
jiromaykin, line 15: If there are no query results, filters should not be visible. Not sure if you may need the {% if paginator.count %}...{% else %} block from PR https://github.com/maykinmedia/open-inwoner/pull/1638 or can solve it in this one already. Here the problem is, it is pushing the column to the side on Desktop.

Desktop:

Screenshot 2025-03-18 at 11 12 31

Mobile:

Screenshot 2025-03-18 at 11 17 05

src/open_inwoner/components/templates/components/Filter/SearchFilters.html
```diff
@@ -0,0 +1,36 @@
+{% load i18n icon_tags button_tags form_tags filter_tags utils %}
+
+
+{# Mobile search filters - inside modal #}
+<div class="filters filters--mobile">
+    {% render_filter_modal %}
+        {% if search_filter_categories and search_form.categories.field.choices|length != 0 %}
+            {% filter field=search_form.categories form_id=form_id open_filter_index=open_filter_index index=0 append_to_checkbox_id='mobile' %}
+        {% endif %}
+        {% if search_filter_tags and search_form.tags.field.choices|length != 0  %}
```
jiromaykin: Unnecessary extra space before search_form.tags closing tag "`!= 0  %}`"

src/open_inwoner/components/templates/components/Filter/SearchFilters.html
```diff
@@ -0,0 +1,36 @@
...
+<div class="filters filters--desktop">
+    <div class="filters__reset">
+        {% button transparent=True type='button' inline=True text=_('Wis alle?') extra_classes='filters__reset-btn filter__reset' %}
+    </div>
+
+    {% if search_filter_categories and search_form.categories.field.choices|length != 0 %}
+        {% filter field=search_form.categories form_id=form_id open_filter_index=open_filter_index index=0 checkbox_as_tag=True %}
+    {% endif %}
+
+    {% if search_filter_tags and search_form.tags.field.choices|length != 0  %}
```
jiromaykin: Also unnecessary extra space before search_form.tags closing tag "`!= 0  %}`"

src/open_inwoner/scss/components/Filter/Filter.scss
```diff
@@ -1,105 +1,171 @@
...
+      padding: 0;
+      margin-bottom: var(--spacing-medium);
+      margin-top: calc(var(--spacing-small) + var(--spacing-medium));
+    }
+  }
+}
+
 .filter {
-  border: 0 solid transparent;
+  user-select: none;
```
jiromaykin, line 32: Neat! 👍

src/open_inwoner/search/tests/test_page.py
```diff
@@ -294,26 +302,60 @@ def test_search_with_filters(self):
...
+            # Check if filter is not open yet.
+            if filter_button.get_attribute("aria-expanded") != "true":
+                filter_button.click()
+
+            # Filter should be open.
+            expect(filter_button).to_have_attribute("aria-expanded", "true")
+
+        def _click_checkbox_for_name(page=page, name=""):
+            filter = page.locator(".filters--desktop").locator(".filter", has_text=name)
+            # first we need to open the list so that all the checkboxes are visible.
```
jiromaykin: Meaning of the comment: do "we" need to open it? Or Playwright needs to check for opened list?

src/open_inwoner/search/tests/test_page.py
```diff
@@ -326,6 +368,39 @@ def _test_search(checkbox_name, expected_text):
         _test_search("Category 1", self.product1.name)
         _test_search("Category 2", self.product2.name)
 
+    # Find out how playwright can view this test on a different viewport.
```
jiromaykin: Clarify meaning of the comment: Since the code is already setting a custom viewport size, a more meaningful comment would be one that explains the purpose of this test and why the specific viewport size was chosen. Also since we mostly have only 1 mobile breaking point (768px), so this looked a bit arbitrary.

Maybe something like this?
#Test mobile search functionality with a small viewport (500x480)
#to verify the search modal dialog opens and closes correctly on mobile devices

src/open_inwoner/components/templatetags/filter_tags.py
```diff
@@ -0,0 +1,149 @@
...
+    kwargs["initial_open"] = kwargs["open_filter_index"] is kwargs["index"]
+    checkbox_as_tag = kwargs.get("checkbox_as_tag", False)
+
+    return {"field": field, "checkbox_as_tag": checkbox_as_tag, **kwargs}
+
+
+@register.tag
+def render_filter_modal(parser, token):
+    """
+    This is used to render in a grid.
```
jiromaykin: Copy/paste mistake :-) its not for a grid.

src/open_inwoner/components/templatetags/filter_tags.py
```diff
@@ -0,0 +1,149 @@
...
+@register.inclusion_tag("components/Filter/SearchFiltersMobile.html")
+def search_filters_mobile(
+    search_form,
+    search_filter_categories,
+    search_filter_tags,
+    search_filter_organizations,
+    **kwargs
+):
+    """
+    Renders the actions in a filterable table.
```
jiromaykin: Needs its own description ;-)

src/open_inwoner/components/templatetags/filter_tags.py
```diff
@@ -0,0 +1,149 @@
...
+@register.inclusion_tag("components/Filter/SearchFilters.html")
+def search_filters(
+    search_form,
+    search_filter_categories,
+    search_filter_tags,
+    search_filter_organizations,
+    **kwargs
+):
+    """
+    Renders the actions in a filterable table.
```
jiromaykin: Needs its own description ;-)

src/open_inwoner/components/templatetags/filter_tags.py
```diff
@@ -0,0 +1,149 @@
...
+    kwargs["search_filter_organizations"] = search_filter_organizations
+    kwargs["open_filter_index"] = open_filter_index
+
+    return {**kwargs}
+
+
+@register.inclusion_tag("components/Filter/Filter.html")
+def filter(field, **kwargs):
+    """
+    Renders the actions in a filterable table.
```
jiromaykin, line 61: Needs its own description ;-)

jiromaykin (changes requested):

jiromaykin (approved):
Approving this so we can change the sidebar widths later.

PR #1631, opened 2025-02-20
[#3056] Open links to open formulieren and acties in new tabs

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3056

PR #1633, opened 2025-02-21
[#1432] Opening links from FAQ

linssen814:
Edited `utils/ckeditor.py` so that the links on the FAQ answers don't open in a new tab anymore. Also added link icons to them. 

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/1432

image

### DIFF
--- a/src/open_inwoner/utils/ckeditor.py
+++ b/src/open_inwoner/utils/ckeditor.py
@@ -58,7 +58,7 @@ def get_rendered_content(content: str) -> str:
                 icon.attrs.update(
                     {
                         "aria-hidden": "true",
-                        "class": "material-icons text-decoration-none",
+                        "class": "material-icons",
                     }
                 )
                 icon.append("open_in_new")
--- a/src/open_inwoner/utils/ckeditor.py
+++ b/src/open_inwoner/utils/ckeditor.py
@@ -53,7 +53,21 @@ def get_rendered_content(content: str) -> str:
         for element in soup.find_all(tag):
             element.attrs["class"] = class_name
             if element.name == "a" and element.attrs.get("href", "").startswith("http"):
-                element.attrs["target"] = "_blank"
+                # icon & screenreader support
+                icon = soup.new_tag("span")
+                icon.attrs.update({"aria-hidden": "true", "class": "material-icons"})
+                icon.append("open_in_new")
+
+                screen_reader_only_text = soup.new_tag("span")
+                screen_reader_only_text.attrs.update(
+                    {
+                        "class": "sr-only",
+                    }
+                )
+                screen_reader_only_text.append(_("Opens external website"))
+
+                element.append(icon)
+                element.append(screen_reader_only_text)
 
     return str(soup)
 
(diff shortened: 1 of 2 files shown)


jiromaykin (changes requested):
Requesting changes:
1. Change the underline/text-decoration. External links with an icon should only have their text itself be underlined, there should not be a line under the icon itself. For an example, look at the product pages, like this one https://open-inwoner-test.maykin.nl/onderwerpen/zorg-en-ondersteuning/producten/huiselijk-geweld/ that contains external links with proper icons and proper text-decoration: 
2. Slightly better process when creating a new PR: first switch back to Develop branch, then do a `git pull`, and then create a new branch from Develop. This way you will have a correct commit history. When a task takes longer you can just rebase to Develop, before doing your final commit and only then ask for review.
3. Add a link to your PR in the Taiga issue as well.

pi-sigma:
@Jiro @linssen814 Note that this PR is based on an older user story and seems to be in tension with the recent https://taiga.maykinmedia.nl/project/open-inwoner/us/3014 (PR: https://github.com/maykinmedia/open-inwoner/pull/1631), which requires that links to formulieren and acties be opened in a new tab. The rationale is that otherwise users get the wrong impression that closing the tab will destroy their OIP session (which is not the case because they will log in via an external service to create/resume a form, thus closing the tab will destroy *that* session but not their OIP session).

src/open_inwoner/utils/ckeditor.py
```diff
@@ -53,7 +53,26 @@ def get_rendered_content(content: str) -> str:
         for element in soup.find_all(tag):
             element.attrs["class"] = class_name
             if element.name == "a" and element.attrs.get("href", "").startswith("http"):
-                element.attrs["target"] = "_blank"
+                # icon & screenreader support
+                icon = soup.new_tag("span")
+                icon.attrs.update(
+                    {
+                        "aria-hidden": "true",
+                        "class": "material-icons text-decoration-none",
```
jiromaykin: We do not have any class called "`.text-decoration-none`" so that will not do anything at all.
The underline is still there.
Remove that non-existing class.

All of our CSS is custom made and organized in its own component.
So what I would do: 
- Keep `npm run watch` running so you can immediately see all changes in the browser
- Look for the _Faq.scss file
- look for the section that starts with `&__answer--open { ... ...`
- now the thing that is the problem here is the .link styling: we need to set that particular type of link to display as 'inline' on the surrounding parent element (right now it is inheriting from the wrong .link parent, and in the browser inspector you can see it is inheriting the display=inline-flex" styling but that should only go to the FAQ 'dropdowns' - also in order to remove the icon 'child' underline style there is a trick: give it a display 'inline-block'! 🌈

So that whole part you should just replace like this:

```
&__answer--open {
    margin-top: var(--spacing-large) !important;

    .utrecht-paragraph .link {
      display: inline;

      [class*='icons'] {
        display: inline-block;
        text-decoration: none;
        transform: scale(.75);
        vertical-align: bottom;
      }
    }
  }
```

jiromaykin (changes requested):
Check to see if suggested CSS will work and remove non-existing 'text-decoration-none' class.

src/open_inwoner/utils/ckeditor.py
linssen814: Makes sense. Thanks for the feedback!

jiromaykin (approved):
Approved!
Note: we will never use Bootstrap classes for custom styling in the public front-end.

EDIT: you may need to run "`isort src/`" and add a migration, since there's failing tests.

PR #1634, opened 2025-02-21
[#3043] Reduce Sentry noise by downgrading expected exceptions as warnings

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3043?no-milestone=1

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -47,14 +47,14 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
         try:
             service = OpenKlant2Service()
         except Exception:
-            logger.error("OpenKlant2 service failed to build")
+            logger.warning("OpenKlant2 service failed to build")
         else:
             _update_user_from_openklant2(user=user, service=service, request=request)
     if config.primary_backend == KlantenServiceType.ESUITE.value:
         try:
             service = eSuiteKlantenService()
         except Exception:
-            logger.error("eSuiteKlantenService failed to build")
+            logger.warning("eSuiteKlantenService failed to build")
         else:
             _update_user_from_esuite(user=user, service=service, request=request)
 
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -47,14 +47,14 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
         try:
             service = OpenKlant2Service()
         except Exception:
-            logger.warning("OpenKlant2 service failed to build")
+            logger.warning("OpenKlant2 service failed to build", exc_info=True)
         else:
             _update_user_from_openklant2(user=user, service=service, request=request)
     if config.primary_backend == KlantenServiceType.ESUITE.value:
         try:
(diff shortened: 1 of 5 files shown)


src/open_inwoner/accounts/signals.py
```diff
@@ -47,14 +47,14 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
         try:
             service = OpenKlant2Service()
         except Exception:
-            logger.error("OpenKlant2 service failed to build")
+            logger.warning("OpenKlant2 service failed to build")
```
swrichards: Let's also log the exception info, so if we are reading the logs we can see what the error in lieu of finer-grained exceptions):

```suggestion
            logger.warning("OpenKlant2 service failed to build", exc_info=True)
```

src/open_inwoner/accounts/signals.py
```diff
@@ -47,14 +47,14 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
...
         else:
             _update_user_from_openklant2(user=user, service=service, request=request)
     if config.primary_backend == KlantenServiceType.ESUITE.value:
         try:
             service = eSuiteKlantenService()
         except Exception:
-            logger.error("eSuiteKlantenService failed to build")
+            logger.warning("eSuiteKlantenService failed to build")
```
```suggestion
            logger.warning("eSuiteKlantenService failed to build", exc_info=True)
```

src/open_inwoner/openzaak/clients.py
```diff
@@ -828,7 +828,7 @@ def __init__(self, clients: list[TClient]):
         self.clients = clients
 
         if len(clients) == 0:
-            raise ValueError("You must specify at least one client")
+            logger.warning("No client specified for MultiZgwClientProxy")
```
swrichards: I am not sure this can safely be changed to a log without raising: it will now simply continue execution. Woud at the very least need a `return`.

However: I think it's best to leave this as-is, this is not spamming Sentry much, and this is a genuine case we shouldn't reach.

src/open_inwoner/openzaak/notifications.py
```diff
@@ -531,6 +531,7 @@ def _handle_status_notification(
     status = zaken_client.fetch_single_status(case.status)
     if not status:
         logger.error("Unable to fetch status for %s", case.status)
+        return
```
swrichards: I would leave this: I can't oversee the behavioral changes here, and this PR should really just be changed to how we log issues, not the control flow.

swrichards (changes requested):
Reviewed this pro-actively because of https://taiga.maykinmedia.nl/project/open-inwoner/issue/3150 -- would like to get this in for the release.

src/open_inwoner/openzaak/notifications.py
pi-sigma: Fair enough, but I suspect the behavior is wrong. I'll add a TODO so we can check later

swrichards: Indeed, very much up for fixing this, but preferably in its own focused PR.

swrichards (approved):

PR #1635, opened 2025-02-21
[2833] Fixed white screen after switching to desktop from mobile nav

linssen814:
When the mobile navigation menu is opened, and the user switches to the desktop view without first clicking the 'close navigation' button, the main elements of the page do not reappear and the user sees an empty/white screen. This is because the `body--open` class, which hides the main elements on the page, does not get removed unless explicitly clicking the 'close navigation' button and so it stays behind still hiding the main content even though the mobile navigation menu is closed.

To fix it, I added a media query around the CSS controlling `.body_open` in `_body.scss` that limits the effects of `body--open` to smaller screens only. 

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2833

### DIFF
--- a/src/open_inwoner/scss/views/_body.scss
+++ b/src/open_inwoner/scss/views/_body.scss
@@ -1,16 +1,18 @@
-.body--open {
-  overflow: hidden;
-  position: fixed;
-  width: 100%;
+@media (max-width: 767px) {
+  .body--open {
+    overflow: hidden;
+    position: fixed;
+    width: 100%;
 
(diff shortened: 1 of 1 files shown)


jiromaykin (approved):
Beautiful.

jiromaykin:
Nice, but don't forget to rebase to Develop before final commit.
You can set your environment as:
`$ git config --global pull.rebase true`
`$ git config --global pull.rebase`

also: add the Github link to this PR to the Taiga issue in the comments-section.

PR #1636, opened 2025-02-21
[#2731] Fix whitespace in footer

linssen814:
Gets rid of the whitespace in a portion of the footer in the mobile view.

Not sure if my solution is the most syntactically consistent/pleasing way but this does work.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2731

### DIFF
--- a/src/open_inwoner/scss/components/Footer/Footer.scss
+++ b/src/open_inwoner/scss/components/Footer/Footer.scss
@@ -68,3 +68,7 @@
     padding: var(--spacing-small) 0;
   }
 }
+
+.footer__right :first-child {
+  padding-left: 0;
+}
--- a/src/open_inwoner/scss/components/Footer/Footer.scss
+++ b/src/open_inwoner/scss/components/Footer/Footer.scss
@@ -67,8 +67,10 @@
   &__list-item {
     padding: var(--spacing-small) 0;
   }
-}
 
-.footer__right :first-child {
-  padding-left: 0;
+  &__links {
+    .footer__list {
+      padding-left: 0;
+    }
+  }
 }


src/open_inwoner/scss/components/Footer/Footer.scss
```diff
@@ -68,3 +68,7 @@
     padding: var(--spacing-small) 0;
   }
 }
+
+.footer__right :first-child {
```
jiromaykin: We already have a selector for this inside the CSS .footer class; so you could have added your padding style to the one that already exists on line 44/45:

```
&__right {
    :first-child {
```

Or better...
_Actually_ target this specific unique `.footer__links` element for both mobile/desktop and its UL child:

```
&__links {
    .footer__list {
      padding-left: 0;
      background-color: yellow;
    }
  }
```

So inside the .footer class within the SCSS syntax, before the last bracket. The ampersand will inherit from the footer parent.

Adjacent topic: 
 
Actually we may need to override the UL styling coming from the CKEDITOR specific styling here `src/open_inwoner/scss/components/Typography/wysiwyg.scss` so in future we'll need to remove that. But keep it like this for now.

Another thing to keep in mind: this list will not always be any-of-the-first-children. And we will loose vertical spacing if a municipality decides to add a custom list above the `nav.footer__links` element and change the order of things. But let's not add that in this PR, if none of the municipalities hasn't had that situation yet.

So I mean this possibly rare situation: 
In the CMS editor they can change the order by dragging, and add different stuff (in CSS we would need multiple 'adjacent' `~` Tilde selectors for that):

Screenshot 2025-03-17 at 15 33 01

jiromaykin (changes requested):

src/open_inwoner/scss/components/Footer/Footer.scss
linssen814: Makes sense. I've used your `.footer__links` solution for now.

jiromaykin (approved):

PR #1638, opened 2025-02-25
:recycle: [#3059] Refactoring of searchpage Sidebar/grids

jiromaykin:
Rearranged grid, in order to make space for filters on desktop
https://taiga.maykinmedia.nl/project/open-inwoner/task/3059


jiromaykin:
@stefrado I tried a CSS only solution first, but that won't work, since it's grids-inside-grids-inside grids... 
+the Grids need to respond to having filters set on/off by municipalities, so this refactoring into columns might be the better solution. Other pages won't have this problem with multiple grids, but feel free to suggest better!

stefrado:
### Feedback

#### Some incorrect styled elements (see screenshot)
1. The pagination alignment is not correct.
2. The feedback form is not aligned correct.

image

#### Code formatting
1. The html of `search.html` isn't formatted properly
    - Some inconsistent blank lines.
    - Added left whitespace on some places where it shouldn't be.

stefrado (approved):

PR #1645, opened 2025-02-28
Initial pass at applying ruff bugbear

swrichards:
[B006](https://docs.astral.sh/ruff/rules/mutable-argument-default/)
[B010](https://docs.astral.sh/ruff/rules/set-attr-with-constant/)
[B904](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/)

### DIFF
--- a/src/open_inwoner/accounts/gateways.py
+++ b/src/open_inwoner/accounts/gateways.py
@@ -64,12 +64,12 @@ class MessageBird(Gateway):
                 to,
                 self.get_message(token),
             )
-        except messagebird.client.ErrorException as e:
-            for error in e.errors:
+        except messagebird.client.ErrorException as exc:
+            for error in exc.errors:
                 logger.critical(
                     ("Could not send SMS to {to}:\n{error}").format(to=to, error=error)
                 )
-            raise GatewayError()
+            raise GatewayError() from exc
         else:
             logging.debug('Sent SMS to %s: "%s"', to, self.get_message(token))
             return True
--- a/src/open_inwoner/components/tests/abstract.py
+++ b/src/open_inwoner/components/tests/abstract.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from django.template import Context, Template
 from django.test import RequestFactory
 
@@ -55,7 +57,7 @@ class InclusionTagWebTest(WebTest):
     # The tag name, a value of "list_item" would be the equivalent of "{% list_item %}".
     tag = ""
 
-    def assertContext(self, args: dict = {}) -> any:
+    def assertContext(self, args: dict | None = None) -> Any:
         """
         Asserts that a context is returned by the inclusion tag.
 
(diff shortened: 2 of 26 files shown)


src/open_inwoner/accounts/gateways.py
```diff
@@ -69,7 +69,7 @@ def send(self, to, token, **kwargs):
                 logger.critical(
                     ("Could not send SMS to {to}:\n{error}").format(to=to, error=error)
                 )
-            raise GatewayError()
+            raise GatewayError() from e
```
pi-sigma:
```suggestion
            raise GatewayError() from exc
```
for consistency with the rest of our codeboase (requires change above of course)

src/open_inwoner/components/tests/abstract.py
```diff
@@ -79,12 +83,16 @@ def assertRender(self, args: dict = {}, data={}) -> str:
 
         Returns: The str rendered by the tag.
         """
+        if data is None:
+            data = {}
+        if args is None:
+            args = {}
```
pi-sigma: Never-nester:
```
data = data or {}
args = args or {}
```
Same for the other cases

src/open_inwoner/configurations/bootstrap/utils.py
```diff
@@ -29,4 +29,4 @@ def get_service(slug: str) -> Service:
     try:
         return Service.objects.get(slug=slug)
     except Service.DoesNotExist as e:
-        raise Service.DoesNotExist(f"{str(e)} (identifier = {slug})")
+        raise Service.DoesNotExist(f"{str(e)} (identifier = {slug})") from e
```
pi-sigma:
```suggestion
        raise Service.DoesNotExist(f"{str(e)} (identifier = {slug})") from exc
```

src/open_inwoner/openzaak/clients.py
```diff
@@ -895,7 +895,7 @@ def build_zgw_client_from_service(service: Service) -> ZgwClientFactoryReturn:
     except KeyError:
         raise ValueError(
             f"No client defined for API type {service.api_type} on service {service}"
-        )
+        ) from None
```
pi-sigma, line 898: Is this for hiding the original traceback? Is it necessary?

pi-sigma (changes requested):
Exceptions should be `exc` instead of `e`. Couple of other comments/questions. Nice that we're getting checks for mutable defaults!

src/open_inwoner/accounts/gateways.py
swrichards: Check.

src/open_inwoner/components/tests/abstract.py
swrichards: Yes!

src/open_inwoner/openzaak/clients.py
swrichards: My thinking here is that the original exception does not contain useful context, but we might as well be permissive. Added the original exception.

pi-sigma (approved):

PR #1648, opened 2025-03-03
Improve klanten service typing

swrichards:


src/open_inwoner/openklant/services.py
```diff
@@ -780,13 +793,13 @@ def list_questions_for_zaak(self, zaak: Zaak, user: User) -> list[Question]:
...
         kcm_answer_mapping = get_kcm_answer_mapping(contactmomenten, user)
-        questions = []
+        questions: list[Question] = []
         for contactmoment in contactmomenten:
             new_answer_available = contactmoment_has_new_answer(
                 contactmoment, kcm_answer_mapping
             )
             questions.append(
-                Question(
+                dict(
```
pi-sigma, line 802: Why change from `Question` to `dict` if the type is `list[Question]`?

swrichards: I have a (vague) sense that using a `TypedDict` as an actual object to be instantiated rather than a type reference is a bit of an anti-pattern, though I Don't think it has any runtime effects (unlike the Pydantic validation we do a few lines below). More generally, we want the list to meet the spec even if we change the construction of the object or get it from elsewhere, so this seems neater to me, but happy to revert if you feel strongly otherwise.

pi-sigma: No, this makes sense, lgtm then.

pi-sigma (approved):

PR #1649, opened 2025-03-03
Reduce configurations.0075 migration logging noise

swrichards:
@pi-sigma I was getting a bit annoyed by the logging noise for this migration. This only changes the logging, not the migration behavior (and practically you only notice it while running tests when bootstrapping a fresh db).


pi-sigma (approved):
Nice!

PR #1653, opened 2025-03-04
[#3086] Add defaults for ContactFormSubject klant configs

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3086


src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def set_default_for_klant_config(apps, _):
+    ContactFormSubject = apps.get_model("openklant", "ContactFormSubject")
+    ESuiteKlantConfig = apps.get_model("openklant", "ESuiteKlantConfig")
+    OpenKlant2Config = apps.get_model("openklant", "OpenKlant2Config")
+
+    for contact_form_subject in ContactFormSubject.objects.all():
+        contact_form_subject.esuite_config = ESuiteKlantConfig.objects.first()
+        contact_form_subject.openklant_config = OpenKlant2Config.objects.first()
```
swrichards:
```suggestion
    for contact_form_subject in ContactFormSubject.objects.all():
        contact_form_subject.esuite_config = ESuiteKlantConfig.objects.first()
        contact_form_subject.openklant_config = OpenKlant2Config.objects.first()
        contact_form_subject.save()
```

src/open_inwoner/openklant/tests/test_migrations.py
```diff
@@ -0,0 +1,34 @@
...
+            contact_form_subject.esuite_config = None
+            contact_form_subject.openklant_config = None
+
+    def test_set_default_config(self):
+        ContactFormSubject = self.apps.get_model("openklant", "ContactFormSubject")
+        ESuiteKlantConfig = self.apps.get_model("openklant", "ESuiteKlantConfig")
+        OpenKlant2Config = self.apps.get_model("openklant", "OpenKlant2Config")
+
+        for contact_form_subject in ContactFormSubject.objects.all():
+            self.assertEqual(
```
swrichards: See above: if the defaults are not saved, I believe these will both evaluate to `None`, meaning the tests pass but we haven't actually set the configs. Best would be to store the created config objects and assert that the config now points to those.

swrichards (changes requested):

src/open_inwoner/openklant/models.py
```diff
@@ -120,6 +120,14 @@ def has_api_configuration(self):
         return all(getattr(self, f, "") for f in self.register_api_required_fields)
 
 
+def get_first_esuite_config():
+    return ESuiteKlantConfig.objects.first()
+
+
+def get_first_openklant_config():
+    return OpenKlant2Config.objects.first()
```
swrichards: I feel slightly uneasy about this, though I can't quite figure out why -- I suspect it's because this might be run in (various) migrations, the path will become fixed in the migration, and generally it feels very tightly coupled. I'd say if we want a default like this, it might be better to set it at the admin form level.

src/open_inwoner/openklant/admin.py
```diff
@@ -22,6 +22,16 @@ def __init__(self, *args, **kwargs):
...
 
+        # set defaults for klant configs if they're missing
+        if not self._meta.model.esuite_config:
+            self._meta.model.esuite_config = (
+                self._meta.model.esuite_config.field.related_model.objects.first()
+            )
+        if not self._meta.model.openklant_config:
+            self._meta.model.openklant_config = (
+                self._meta.model.openklant_config.field.related_model.objects.first()
+            )
```
swrichards: I am a bit unfamiliar with this approach: this seems to amend the model _spec_, no? Is this needed because the fields are not exposed in the admin?

Otherwise I'd expect something like:

```
self.fields['esuite_config'].initial = ESuiteConfig.objects.first()
# ...
```

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db import migrations
+
+
+def set_default_for_klant_config(apps, _):
+    ContactFormSubject = apps.get_model("openklant", "ContactFormSubject")
+    ESuiteKlantConfig = apps.get_model("openklant", "ESuiteKlantConfig")
+    OpenKlant2Config = apps.get_model("openklant", "OpenKlant2Config")
+
+    for contact_form_subject in ContactFormSubject.objects.all():
```
swrichards:
```suggestion
    for contact_form_subject in ContactFormSubject.objects.filter(Q(esuite_config__is_null=False) | Q(openklant_config__is_null=False)):
```

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
```diff
@@ -0,0 +1,37 @@
...
+        contact_form_subject.esuite_config = ESuiteKlantConfig.objects.first()
+        contact_form_subject.openklant_config = OpenKlant2Config.objects.first()
+        contact_form_subject.save()
```
swrichards: Let's check that the values are not already set, to avoid overriding (shouldn't matter, but better safe than sorry).

src/open_inwoner/openklant/tests/test_migrations.py
```diff
@@ -0,0 +1,32 @@
...
+        OpenKlant2Config = app.get_model("openklant", "OpenKlant2Config")
+
+        self.esuite_config_1 = ESuiteKlantConfig.objects.create()
+        self.esuite_config_2 = ESuiteKlantConfig.objects.create()
+        self.openklant_config_1 = OpenKlant2Config.objects.create()
+        self.openklant_config_2 = OpenKlant2Config.objects.create()
+
+        for contact_form_subject in ContactFormSubject.objects.all():
+            contact_form_subject.esuite_config = None
+            contact_form_subject.openklant_config = None
```
swrichards, line 21:
```suggestion
            contact_form_subject.openklant_config = None
            contact_form_subject.save()
```

src/open_inwoner/openklant/admin.py
swrichards: Note I am not sure we really need this if the migration is run. I'd be ok leaving it out.

swrichards (changes requested):

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
pi-sigma: Do you mean:

```
ContactFormSubject.objects.filter(Q(esuite_config__is_null=True) | Q(openklant_config__is_null=True))
```

to ensure we only set the default for subjects that have no config yet?

src/open_inwoner/openklant/admin.py
pi-sigma: That was indeed the reason. But I think you're right this won't be necessary after the data migration, so I'll remove it

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
swrichards: My bad, you're right, that's what I meant.

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
pi-sigma: I thought that was the point of filtering? But then I thought you meant to filter on `isnull=True`. Did I misunderstand the intention behind the filtering?

swrichards: That was the intent, but because it's an `or`, you don't know which of the two configs is set.

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
```diff
@@ -0,0 +1,42 @@
...
+def set_default_for_klant_config(apps, _):
+    ContactFormSubject = apps.get_model("openklant", "ContactFormSubject")
+    ESuiteKlantConfig = apps.get_model("openklant", "ESuiteKlantConfig")
+    OpenKlant2Config = apps.get_model("openklant", "OpenKlant2Config")
+
+    for contact_form_subject in ContactFormSubject.objects.filter(
+        Q(esuite_config__isnull=True) | Q(openklant_config__isnull=True)
+    ):
+        contact_form_subject.esuite_config = ESuiteKlantConfig.objects.first()
+        contact_form_subject.openklant_config = OpenKlant2Config.objects.first()
```
swrichards: I would still add this, to be on the safe side:

```suggestion
        contact_form_subject.esuite_config = contact_form_subject.esuite_config or ESuiteKlantConfig.objects.first()
        contact_form_subject.openklant_config = contact_form_subject.openklant_config or OpenKlant2Config.objects.first()
```

swrichards (changes requested):

src/open_inwoner/openklant/migrations/0026_contactform_subject_config_default.py
pi-sigma: Ah, I misunderstood you then.

PR #1657, opened 2025-03-06
[#3081] New designs for Plans / checkboxes / modal / screens

jiromaykin:
Features 3 tasks in Story:
https://taiga.maykinmedia.nl/project/open-inwoner/us/2990
Multiple (new) designs:
https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?node-id=4446-10133&t=WWOTjCAtHjwBwzan-1

Idea here is that the choosing of a Plan based on Template should become much more prominent.
Creating a custom Plan without any Template should be less important.

So the original '`create`' classes should probably all be renamed to 'create-without-template' or something similar. Also the changes in URL flow will need new tests and rewriting of existing tests in future.

Navigation flow: 

1. Plans-list click 'Start nieuwe samenwerking'
http://localhost:8000/samenwerken/
2. Old: show the original 'create' plan template with scroll (#plan-templates) to templates - needs titles/goal/ enddate
http://localhost:8000/samenwerken/create/
3. New (non-functional): show only the selection view, new inbetween navigation step,  for creating from template first, with button to create custom, so no option for title/goal/date in this new step
http://localhost:8000/samenwerken/choose-template/
4. Old: immediately show new created plan with prefilled fields after submit
localhost:8000/samenwerken/1234567890-numbers...
5. New: show new form, inbetween navigation step, where goal/enddate still need to be filled in, title should come from the selected template
http://localhost:8000/samenwerken/create-from-template/
6. New: last step: show the filled and active plan
localhost:8000/samenwerken/1234567890-numbers...

Also new: selecting a template in the modal, should select the radio input for it.

Note: these are mostly designs only, not yet functional.
Like: currently it is required to give a 'name' to an uploaded Plan Document, but in the new design there is no option for that.

All new Dutch translations will be added after merge in a separate translations PR.


jiromaykin:
@pi-sigma @stefrado This PR has "old" and "new" simultaneously (so merging is made possible).
In the introductory 'conversation' on Github I've put up a bunch of explanations about the views you can actually review.

For municipalities the pages `/choose-template` and `/create-from-template` are not visible yet.

I've added the `TODO:` syntax in templates and tests where I think we need more back-end additions and moving about of things (due to the different order and a new 'select/choose template' step).

src/open_inwoner/cms/collaborate/urls.py
```diff
@@ -6,20 +6,32 @@
...
 )
 
 app_name = "collaborate"
 
 urlpatterns = [
     path("", PlanListView.as_view(), name="plan_list"),
-    path("create/", PlanCreateView.as_view(), name="plan_create"),
+    path(
+        "choose-template/",
+        PlanTemplateChooseView.as_view(),
```
pi-sigma:
```suggestion
        PlanTemplateChoiceView.as_view(),
```

src/open_inwoner/cms/collaborate/urls.py
```diff
@@ -6,20 +6,32 @@
     PlanActionEditStatusTagView,
     PlanActionEditView,
     PlanActionHistoryView,
-    PlanCreateView,
+    PlanCreateNoTemplateView,
```
pi-sigma:
```suggestion
    PlanCreateView,
```
```suggestion
    PlanCreateNoTemplateView,
```

src/open_inwoner/cms/collaborate/urls.py
```diff
@@ -6,20 +6,32 @@
...
+    PlanCreateWithTemplateView,
```
pi-sigma:
```suggestion
    PlanCreateFromTemplateView,
```

src/open_inwoner/plans/templates/plans/preview.html
```diff
@@ -1,39 +1,30 @@
...
-        <div>{{ plan_template.goal }}</div>
+        <p class='plan-preview__key'>{% trans "Doel" %}</p>
+        <div>{{ plan_template.goal|truncatewords:20 }}</div>
     {% endif %}
 
     {% if plan_template.description %}
-        <div class='preview__title'>{% trans "Description:" %}</div>
-        <div>{{ plan_template.description }}</div>
+        <div class='plan-preview__key'>{% trans "Description:" %}</div>
+        <div>{{ plan_template.description|truncatewords:10 }}</div>
```
pi-sigma: Why do we truncate at 10 words for the description while truncating at 20 words for the goal?

src/open_inwoner/plans/tests/test_logging.py
```diff
@@ -32,9 +32,9 @@ def setUp(self):
 
     def test_created_plan_is_logged(self):
         plan = PlanFactory.build(created_by=self.user)
-        form = self.app.get(reverse("collaborate:plan_create"), user=self.user).forms[
-            "plan-form"
-        ]
+        form = self.app.get(
+            reverse("collaborate:plan_create_no_template"), user=self.user
```
pi-sigma, line 36:
```suggestion
            reverse("collaborate:plan_create"), user=self.user
```
I would prefer keeping `plan_create` here and elsewhere for creation of plans without template, though perhaps @swrichards has a different view of the naming here.

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -44,7 +44,11 @@ def setUp(self) -> None:
 
         self.login_url = reverse("login")
         self.list_url = reverse("collaborate:plan_list")
-        self.create_url = reverse("collaborate:plan_create")
+        self.choose_template_url = reverse("collaborate:plan_choose_template")
+        # TODO: new tests for choosing template form
+        self.create_url = reverse("collaborate:plan_create_no_template")
+        self.create_with_template_url = reverse("collaborate:plan_create_with_template")
```
pi-sigma:
```suggestion
        self.create_from_template_url = reverse("collaborate:plan_create_with_template")
```
Same as above. Look at your comment: you say "creating plan from template", that's the natural way to express this, and should be reflected in the naming

src/open_inwoner/plans/views.py
```diff
@@ -233,23 +233,100 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class PlanCreateView(
+class PlanTemplateChooseView(
```
pi-sigma:
```suggestion
class PlanTemplateChoiceView(
```

src/open_inwoner/plans/views.py
```diff
@@ -233,23 +233,100 @@ def get_context_data(self, **kwargs):
...
     PlanActionsEnabledMixin,
     LogMixin,
     LoginRequiredMixin,
     CommonPageMixin,
     BaseBreadcrumbMixin,
     CreateView,
 ):
-    template_name = "pages/plans/create.html"
+    template_name = "pages/plans/choose-template.html"
```
pi-sigma:
```suggestion
    template_name = "pages/plans/template-choice.html"
```

src/open_inwoner/plans/views.py
```diff
@@ -233,23 +233,100 @@ def get_context_data(self, **kwargs):
...
+        self.object.plan_contacts.add(self.object.created_by)
+
+        self.log_addition(self.object, _("plan was created"))
+        return HttpResponseRedirect(self.get_success_url())
+
+    def get_success_url(self) -> str:
+        return self.object.get_absolute_url()
+
+
+class PlanCreateNoTemplateView(
```
pi-sigma:
```suggestion
class PlanCreateView(
```

src/open_inwoner/plans/views.py
```diff
@@ -233,23 +233,100 @@ def get_context_data(self, **kwargs):
...
+        self.object.plan_contacts.add(self.object.created_by)
+
+        self.log_addition(self.object, _("plan was created"))
+        return HttpResponseRedirect(self.get_success_url())
+
+    def get_success_url(self) -> str:
+        return self.object.get_absolute_url()
+
+
+class PlanCreateWithTemplateView(
```
pi-sigma:
```suggestion
class PlanCreateFromTemplateView(
```

src/open_inwoner/utils/tests/test_timeline_logger.py
```diff
@@ -72,7 +72,7 @@ def test_changed_object_is_logged(self):
         self.assertEqual(
             log_entry.extra_data,
             {
-                "message": "Titel, Doel and Einddatum gewijzigd.",
+                "message": "Titel, Doel and Expected end date gewijzigd.",
```
pi-sigma: Strange mix of languages. I would expect Dutch *or* English but not both. The original had only "and" instead of "en", which might have been an oversight? Can you check where the log is coming from and what format (including which language) we expect?

src/open_inwoner/templates/pages/plans/create-no-template.html
```diff
@@ -0,0 +1,99 @@
...
+
+                        <div class="">
+                            <div class="plan-template" id="plan-templates">
+                                <div class="label plan-template__row plan-template__table radio">
+                                    <div>
+                                        <input class="radio__input" type="radio" name="template" value="" id="id_template_0" {% if not form.data.template %}checked{% endif %}>
+                                        <label class="radio__label" for="id_template_0">Geen template</label>
+                                    </div>
+                                </div>
+                                {% for tpl_id, plan_template in form.template.field.choices %}
```
pi-sigma: Do we still want to display the template choices here? I think it would make more sense to add a link to the template choice page. This is relevant for how we want to deal with the forms to make this functional (see general comment).

src/open_inwoner/cms/collaborate/urls.py
```diff
@@ -6,20 +6,32 @@
...
+        "choose-template/",
+        PlanTemplateChooseView.as_view(),
+        name="plan_choose_template",
+    ),
+    path("create/", PlanCreateNoTemplateView.as_view(), name="plan_create_no_template"),
+    path(
+        "create-from-template/",
+        PlanCreateWithTemplateView.as_view(),
+        name="plan_create_with_template",
+    ),
```
pi-sigma, line 36: - `PlanCreateWithTemplateView` should be `PlanCreateFromTemplateView` imho
- This will likely need a parameter for the template, which is determined on the template choice page, hence:
```
path(
    "create-from-template/<int:template_id>/",
    PlanCreateFromTemplateView.as_view(),
    name="plan_create_with_template",
)
```
However, this suggestion depends on splitting up the form for creating plans into (a) a form for creating without templates and (b) a form for creating with templates. See general comment.

pi-sigma (changes requested):
Looking good! 

We'll need to think of course about how to handle the forms. We're probably best off splitting up the `PlanForm` into `PlanCreateForm` (getting rid of the checks for the template) + `PlanCreateFromTemplateForm`, and then create an additional `PlanTemplateChoiceForm` where the user selects a template, and this is passed to the `PlanCreateFromTemplateView`. The current form, which handles the with/without template cases at the same time, doesn't really fit our new situation. However, this would require changing the template for `create-no-template.html` (see comment ad loc). Let's discuss if this is in line with the design.

Since the PR is not the main focus, I think we don't have to do too much with the views/forms/tests here. Groundwork looks good and can be adapted later on.

src/open_inwoner/templates/pages/plans/create-no-template.html
jiromaykin: These will indeed need to be removed for the "no-template" option - I wanted to keep 1 functional page before merge with functional URL, but I guess I should just have kept `create.html`,
and should have added `create-no-template.html `without any link to it, then later give that new no-template page the same URL as the current `.../create` page.

src/open_inwoner/plans/templates/plans/preview.html
jiromaykin: Ah, wait, there shouldn't be any Description in the Preview Modal, only the Goal, 
so I'll remove it.

src/open_inwoner/utils/tests/test_timeline_logger.py
jiromaykin: I thought it was partially coming from our Translation file ...django.po, but this is outdated:

```
#: open_inwoner/plans/models.py:121
#: open_inwoner/templates/export/include/actions.html:10
#: open_inwoner/templates/export/plans/plan_export.html:22
#: open_inwoner/templates/export/profile/action_export.html:24
msgid "End date"
msgstr "Einddatum"
```

Where _Einddatum_ has to be replaced with '_Verwachte einddatum_'. But I did not yet want to update all of the translations in this PR.

The additional 'and' is a mystery to me. Let's see what happens if I update the translations partially.

src/open_inwoner/plans/tests/test_logging.py
jiromaykin: I wanted to give a very specific, more explicit meaning to the new situation where there is no such thing as just 'creating' a plan. 
Also we are sort of trying to hide the original 'create' default, and we wish to push the creating-plan-from-template situation.

I can of course reverse all of the `plan_create_no_template` to `plan_create` again if we all understand what it is for 😇 - just let me know.

pi-sigma (approved):

src/open_inwoner/js/components/plan-preview/index.js
```diff
@@ -13,19 +13,123 @@ export class PlanPreview {
...
+          // If no explicit label found, look for parent or ancestor label
+          radioLabel =
+            radioInput.closest('label') ||
+            radioInput.parentElement.querySelector('label')
+        }
+      }
+    }
+
+    const closeButtons = modalElement.querySelectorAll(
+      '.modal__close, .modal__close-title'
```
stefrado: Do we really want to select the viewed plan when we close the model with the cross?

src/open_inwoner/plans/templates/plans/preview.html
```diff
@@ -1,39 +1,25 @@
-{% load i18n link_tags %}
+{% load i18n link_tags file_tags %}
 
-<div class='preview'>
+<div class='plan-preview'>
+<p class='plan-preview__key'>{% trans "Titel" %}</p>
```
stefrado: Please add whitespace

src/open_inwoner/plans/templates/plans/preview.html
```diff
@@ -1,39 +1,25 @@
-{% load i18n link_tags %}
+{% load i18n link_tags file_tags %}
```
stefrado, line 1: Is this file still used? I can find a place where it is rendered.

src/open_inwoner/templates/pages/plans/create.html
```diff
@@ -3,79 +3,97 @@
...
 {% block content %}
     <div class="plan-create">
-        <h1 class="utrecht-heading-1">{% trans "Start nieuwe samenwerking" %}</h1>
-        {% optional_paragraph configurable_text.plans_page.plans_intro %}
+        <div class="plan-template__header">
+            <h1 class="utrecht-heading-1">{% trans "Start een nieuwe samenwerking" %}</h1>
+            <div class="plan-template__header-nav">
+                {% if request.user.get_active_contacts %}
+                    {% button href="#plan-templates" text=_("Gebruik sjabloon") primary=True transparent=True extra_classes="link--bold" icon="insert_drive_file" icon_outlined=True %}
+                    {# TODO: Change scrollbutton into "href="collaborate:plan_choose_template" #}
```
stefrado, line 11: So it is correct that i can't access `/samenwerkingen/choose-template` with the UI?

stefrado (changes requested):

stefrado:
No empty state?

Scherm­afbeelding 2025-03-25 om 12 48 34

stefrado:
Why is this button inside this modal. It gives the user an action that feels a bit misplaced.

Scherm­afbeelding 2025-03-25 om 13 00 36

src/open_inwoner/plans/templates/plans/preview.html
jiromaykin: You mean indent?

src/open_inwoner/js/components/plan-preview/index.js
jiromaykin: I don't know, maybe we don't. It does feel a bit irritating.

src/open_inwoner/templates/pages/plans/create.html
jiromaykin: Yes not currently. It's not functional. 
We need a lot of rearranging in the current forms/tests + add new stuff to the forms and also create a new form, due to 2 extra navigation steps.

src/open_inwoner/plans/templates/plans/preview.html
jiromaykin: Yes it is, scroll down and look at:

`{% if plan_template.file %}
        <div class='plan-preview__key'>{% trans 'Bestand' %}</div>
        <div>{% file file=plan_template.file extension=plan_template.file.extension size=plan_template.file.size show_download=False %}</div>
    {% endif %}`

jiromaykin:
> No empty state?

I'll add a fallback paragraph with the strong tag we also use in other list views.

jiromaykin:
> Why is this button inside this modal. It gives the user an action that feels a bit misplaced.

Oh awful.... I see @stefrado it's different in the design now. Seems it needs to be some kind of close button, since the text is "Terug naar voorstellen" but this won't make sense in the current version 
(it does need to be changed in the new views that are not public yet).

stefrado:
Some places it is called sjabloon other places voorstel. Shouldn't this be consistent?

Scherm­afbeelding 2025-03-28 om 10 31 06

stefrado (approved):

jiromaykin:
> Some places it is called sjabloon other places voorstel. Shouldn't this be consistent?

Maybe. It is like this in the design. I can ask in the Design meeting next Tuesday. For now keeping this same as the design.

PR #1662, opened 2025-03-11
[#3087] Add DB constraint that primary and alternative phonenumber differ

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3087


src/open_inwoner/accounts/migrations/0081_user_check_alternative_phonenumber_differs_from_primary_phonenumber.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.18 on 2025-03-11 15:35
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
swrichards, line 15: Can we add a migration to ensure that duplicate numbers are consolidated before applying the migration?

pi-sigma: What's the best option, setting the alternative to "" in case of duplication?

swrichards: Yes that makes sense, that would be the default otherwise if it weren't set.

src/open_inwoner/accounts/forms.py
```diff
@@ -194,6 +194,10 @@ def clean_phonenumber_alternative(self):
                     "A primary phone number is required for setting an alternative phone number"
                 )
             )
+        if phonenumber_alt and phonenumber_alt == phonenumber:
+            raise ValidationError(
+                _("Primary adn secondary phone numbers cannot be the same")
```
swrichards:
```suggestion
                _("Primary and secondary phone numbers cannot be the same")
```

swrichards (changes requested):

src/open_inwoner/accounts/tests/test_user.py
```diff
@@ -112,9 +112,16 @@ def test_plan_contact_new_count_methods(self):
...
 
-    def test_constraint_phonenumber_alternative(self):
+    def test_phonenumber_alternative_requires_primary_phonenumber(self):
         user = UserFactory(phonenumber="")
 
         with self.assertRaises(IntegrityError):
             user.phonenumber_alternative = "0612345678"
             user.save()
+
+    def test_phonenumber_alternative_differs_from_non_empty_primary_phonenumber(self):
```
swrichards, line 122: @pi-sigma sorry one more late thought: can we also add a test that confirms both phonenumbers can be empty?

pi-sigma: Good point, I'll add it

pi-sigma: I just added `phonenumber_alternative=""` to the `UserFactory` in one test to make it more explicit. I don't think this requires an extra test, since `UserFactory` will raise an `IntegrityError` upon creation if the constraint is violated.

swrichards: True, _but_ then we are counting on future refactorers (including ourselves) to be aware of that context, I'd still somewhat prefer having it as a separate test, even if small, to express that this is a specific invariant we care about.

pi-sigma: Fair enough, better explicit than implicit. I'll change it.

swrichards (approved):

PR #1665, opened 2025-03-11
[#3076] page title change with cms page title

linssen814:
Made it so that the browser tab title as well as the header of the 'Benefits', 'Collaboration' and 'Mijn Vragen' pages change with the page title configured in the cms.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3076

Note that for the 'Mijn Vragen' page, the way that the Page object is retrieved is kind of iffy, though it does seem to work. See the comment above it.

Edit: As per the discussion in the weekly, 'Mijn vragen' is kept hardcoded for this issue and will be addressed in a separate issue. See Taiga.

### DIFF
--- a/src/open_inwoner/plans/views.py
+++ b/src/open_inwoner/plans/views.py
@@ -99,8 +99,10 @@ class PlanListView(
 
     @cached_property
     def crumbs(self):
+        current_page = self.request.current_page
+        title = current_page.get_title() if current_page else _("Samenwerken")
         return [
-            (_("Samenwerken"), reverse("collaborate:plan_list")),
+            (title, reverse("collaborate:plan_list")),
         ]
 
     def get_queryset(self):
--- a/src/open_inwoner/ssd/views.py
+++ b/src/open_inwoner/ssd/views.py
@@ -6,23 +6,30 @@ from django.http import HttpResponse
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.functional import cached_property
-from django.utils.translation import gettext as _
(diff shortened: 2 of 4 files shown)


src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -136,7 +137,19 @@ def crumbs(self):
...
+        # solution is appreciated.
+        page = Page.objects.filter(
+            application_urls="", publisher_is_draft=True, is_home=False
+        ).first()
+
+        # Extra check to be safe
+        if page and page.get_absolute_url() == "/mijn-vragen/":
+            return page.get_title()
+        else:
+            return _("Mijn Vragen")
```
pi-sigma: Yeah this is annoying. The filtering and subsequent checks are not going to work: `page.get_absolute_url()` relies on the slug of the page, which in turn is determined by the page title (though can be manually edited) If you change the slug manually in the CMS menu for the vragen page, you'll notice that `page.get_absolute_url` will yield a different result. Since this is configurable by the user, we can't know in advance what it's going to be (currently we do have "mijn-vragen" as the slug and therefore as value of `page.get_absolute_url()`, but this could be different).

Unfortunately I don't know of any better way to do this. The only thing under our control regarding the CMS pages is the apphook (and its attributes). At this point, we might as well just stick with `return _("Mijn Vragen"`) as in the original code.

pi-sigma (changes requested):

pi-sigma (approved):

PR #1670, opened 2025-03-17
[#3096] Add a vestiging field to the User model

swrichards:
[Taiga 3096](https://taiga.maykinmedia.nl/project/open-inwoner/task/3096).

```
630a00032 - [#3096] Tag tests failing due to vestigingen attribute
This commit tags tests currently failing due to the tighter
constraints around the kvk and vestiging attribute on the User
model. We'll work our way through these tags and amend them to
pass as we move through the feature.

8d8758af6 - [#3096] Exclude tagged vestiging tests while we complete the feature
Adding vestigingen is a big change. To make this manageable,
we'll merge these changes to a feature branch, and to facilitate
parallel PRs, we'll tag the currently failing tests and fix them
as we go along, aiming to remove all tags and fix all tests
prior to merging the feature branch into the trunk.

1bf318267 - [#3096] Add a vestiging field to the User model
This commit adds a `vestiging` attribute to the User model.
Vestigingen are for most use-cases, and in the eyes of most
clients, distinct users from the legal entity to which
the vestiging belongs. To handle this properly, we'll have
to make vestigingen first-class citizens, so we can e.g. store
profile information on a vestiging basis, and generally allow
vestigingen to interact with OIP on a full-fledged basis (e.g.
also having, for instance, different notification settings
then the main legal entity).

f4c1668e5 - [#3096] Make kvk attribute required for eherkenning users
The kvk attribute was practically already treated as a unique
identifier by the eHerkenning authentication backends, but this
was not yet enforced at the database level. In light of the upcoming
changes with regard to adding vestigingen to the user model, it makes
sense to tighten these constraints to make it less likely we'll run
into edge cases in narrowing down users to the correct kvk/vestiging
pair.
```


swrichards:
@pi-sigma probably best to go commit-by-commit here to separate out the refactors and test tagging from the main logic. In addition to adding the vestiging, I also made kvk required for eHerkenning users.

As discussed, we'll merge this into `feature/vestigingen-as-users` and fix the tests within the applicable stories.

src/open_inwoner/accounts/tests/test_user.py
```diff
@@ -137,3 +138,39 @@ def test_eherkenning_user_requires_kvk(self):
...
+                    user.full_clean()
+
+    def test_vestiging_validation_must_be_numeric(self):
+        for invalid_vestiging in tuple(str("a" * i) for i in range(1, 11)) + (
+            "a1111111111",
+        ):
+            with self.subTest(invalid_vestiging):
+                user = eHerkenningVestigingUserFactory(vestiging=invalid_vestiging)
+                with self.assertRaises(ValidationError):
+                    user.full_clean()
```
pi-sigma, line 153: The second test is explicitly about ruling out non-numeric chars, but would also fail if non-numeric chars were allowed but had to be minimally 11 tokens long. In other words,  `tuple(str("a" * i) for i in range(1, 11)` doesn't do useful work I think; it would be better to keep the length of the test strings constant at 11 and only check for "numericity" (?)

src/open_inwoner/accounts/tests/test_user_manager.py
```diff
@@ -35,3 +37,127 @@ def test_having_usable_email(self):
...
+    def setUp(self):
+        self.user_kvk_without_vestiging = eHerkenningUserFactory(
+            kvk="12345678",
+            vestiging="",
+        )
+        self.user_kvk_with_vestiging = eHerkenningUserFactory(
+            kvk="12345678",
+            vestiging="00123456789",
+        )
+        self.another_user_with_kvk_vestiging = eHerkenningUserFactory(
```
pi-sigma: `self.another_user_without_kvk_vestiging = eHerkenningUserFactory`

or add vestiging

src/open_inwoner/accounts/tests/test_user_manager.py
```diff
@@ -35,3 +37,127 @@ def test_having_usable_email(self):
...
+                kvk="12345678", vestiging="98765432100"
+            )
+
+    def test_filter_by_kvk_and_vestiging(self):
+        """Test filtering users by KVK and vestiging"""
+        # Test with KVK only, empty vestiging
+        queryset = self.manager.filter_by_kvk_and_vestiging(
+            kvk="12345678", vestiging=""
+        )
+        self.assertEqual(list(queryset), [self.user_kvk_without_vestiging])
```
pi-sigma, line 119: May have have to be changed depending on how you handle the previous comment concerning `self.another_user_with_kvk_vestiging`

pi-sigma (changes requested):

pi-sigma (approved):

PR #1671, opened 2025-03-17
Use uv to compile dependencies

pi-sigma:


.github/workflows/dependencies.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         with:
           python-version: '3.11'
 
+      - name: Install uv
```
swrichards, line 30: @pi-sigma You might be able to use [a dedicated plugin for this](https://docs.astral.sh/uv/guides/integration/github/#using-uv-in-github-actions).

pi-sigma: Nice, will give it a try!

requirements/dev.txt
```diff
@@ -1268,12 +1262,11 @@ svglib==1.5.1
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
     #   easy-thumbnails
-tablib[html,ods,xls,xlsx,yaml]==3.1.0
+tablib==3.1.0
```
swrichards: I am wondering why `uv` no longer includes the optional dependencies: is that a feature/intentional?

swrichards (changes requested):
Looks good, but have a look at the optional dependencies issue. It seems the `uv` does support this, but they've disappeared from the requirements files. Maybe it needs another flag?

requirements/dev.txt
swrichards: @pi-sigma [you might need](https://docs.astral.sh/uv/pip/compatibility/#pip-compile-defaults) `--no-strip-extras`.

pi-sigma: Good catch. Apparently this is uv's default behavior; needs `--no-strip-extras` to prevent this.

requirements/dev.txt
```diff
@@ -1433,7 +1430,3 @@ zopfli==0.1.9
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt
     #   fonttools
-
-# The following packages are considered to be unsafe in a requirements file:
-# pip
-# setuptools
```
swrichards: One more thing here: this is another deviation from default pip-tools behavior. This thread (https://github.com/astral-sh/uv/issues/1353) suggests pip-tools will change this default behavior (similar to the strip extras command), so this might be fine. However, to lower mental load, perhaps we also do not allow unsafe just so we have strict parity with existing behavior. I would suggest we then make a tech debt ticket to review both flags (extras/unsafe) for future review.

swrichards (approved):

PR #1672, opened 2025-03-17
💄 [#3083] New search feedback design

jiromaykin:
Feature: https://taiga.maykinmedia.nl/project/open-inwoner/task/3083

Figma: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs-(new)?node-id=3052-15300&t=dMCqsMx7LklAegER-1 

Also incorporates new mobile design for the blue 'no results' box.

### DIFF
--- a/src/open_inwoner/js/components/search/feedback.js
+++ b/src/open_inwoner/js/components/search/feedback.js
@@ -17,6 +17,10 @@ export class FeedbackForm {
     this.radioButtons.forEach((radio) => {
       radio.addEventListener('click', this.handleRadioClick.bind(this))
     })
+
+    this.radioLabels.forEach((label) => {
+      label.addEventListener('click', this.handleLabelClick.bind(this))
+    })
   }
 
   /**
@@ -24,7 +28,7 @@ export class FeedbackForm {
    * @param {Event} event
    */
   handleRadioClick(event) {
-    this.feedbackContainer.classList.add('feedback__remark--show')
+    this.feedbackRemarkContainer.classList.add('feedback__remark--show')
 
     if (event.target.value === 'true') {
       this.feedbackLabelText.textContent =
@@ -36,13 +40,28 @@ export class FeedbackForm {
   }
 
   /**
+   * Ensure the input gets focus when the label is clicked in Safari
+   * @param {Event} event
+   */
+  handleLabelClick(event) {
+    const input = event.currentTarget.querySelector('.button-radio__input')
+    if (input) {
+      input.focus()
+    }
+  }
+
+  /**
    * Getters for various elements within the form
    */
   get radioButtons() {
     return this.node.querySelectorAll('.feedback__options .button-radio__input')
   }
 
-  get feedbackContainer() {
+  get radioLabels() {
+    return this.node.querySelectorAll('.feedback__options .button-radio')
+  }
+
+  get feedbackRemarkContainer() {
     return this.node.querySelector('.feedback__remark')
   }
 
--- a/src/open_inwoner/js/components/search/feedback.js
+++ b/src/open_inwoner/js/components/search/feedback.js
@@ -1,0 +1,59 @@
+const FEEDBACK_FORMS = document.querySelectorAll('.feedback')
+
+export class FeedbackForm {
+  /**
+   * Construct the FeedbackForm class.
+   * @param {HTMLElement} node
+   */
+  constructor(node) {
+    this.node = node
+    this.bindEvents()
+  }
+
+  /**
+   * Bind all event handlers.
+   */
+  bindEvents() {
+    this.radioButtons.forEach((radio) => {
+      radio.addEventListener('click', this.handleRadioClick.bind(this))
+    })
+  }
+
+  /**
+   * Handle radio button clicks
+   * @param {Event} event
+   */
+  handleRadioClick(event) {
+    this.feedbackContainer.classList.add('feedback__remark--show')
+
+    if (event.target.value === 'true') {
+      this.feedbackLabelText.textContent =
+        this.translations.dataset.positiveLabel
+    } else {
+      this.feedbackLabelText.textContent =
+        this.translations.dataset.negativeLabel
+    }
+  }
+
+  /**
+   * Getters for various elements within the form
+   */
+  get radioButtons() {
+    return this.node.querySelectorAll('.feedback__options .button-radio__input')
+  }
+
+  get feedbackContainer() {
+    return this.node.querySelector('.feedback__remark')
+  }
+
+  get feedbackLabelText() {
+    return this.node.querySelector('.feedback__label-text')
+  }
+
+  get translations() {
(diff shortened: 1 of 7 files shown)


pi-sigma (approved):

src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -5,34 +5,45 @@
     z-index: -1;
 
     &:focus ~ .button-radio__button {
-      outline: 2px solid var(--color-info);
+      outline: var(--focus-border-width) var(--focus-border-style)
+        var(--color-primary-lighter);
```
stefrado, line 10: This outline results in a blue outline on a red button. Is this really what we want?

src/open_inwoner/templates/pages/search.html
```diff
@@ -48,32 +48,52 @@ <h3 class="utrecht-heading-3 search-results__item-title">{{ hit.name }}</h3>
...
+                                     data-negative-label="{% trans 'Waarom heeft u niet gevonden wat u zocht?*' %}">
+                                </div>
+
+                                {# Translatable label and textarea #}
+                                <div class="feedback__remark">
+                                    <div class="form__control">
+                                        <label class="label">
+                                            <span class="feedback__label-text utrecht-paragraph"></span>
+                                            {{ feedback_form.remark|addclass:"textarea" }}
+                                            {% if feedback_form.remark.errors %}
```
stefrado: Somehow these errors are not shown and i can submit an empty textarea.

src/open_inwoner/js/components/search/feedback.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Getters for various elements within the form
+   */
+  get radioButtons() {
+    return this.node.querySelectorAll('.feedback__options .button-radio__input')
+  }
+
+  get feedbackContainer() {
```
stefrado: Is this the feedback container or the feedback remark container?

I can see it is the remark container but i do think the name could fit better.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -5,34 +5,45 @@
...
     }
+
     &#negative {
       &:checked ~ .button-radio__button {
-        color: #cf1c08;
+        background-color: var(--color-error-lightest);
+        border: var(--border-width-thin) solid var(--color-error-darker);
+        color: var(--color-error-darker);
         cursor: default;
+        outline: 1px solid var(--color-primary-lighter);
```
stefrado: Outline always visible on purspose?

src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -5,34 +5,45 @@
...
     }
+
     &#positive {
       &:checked ~ .button-radio__button {
-        color: var(--color-green-dark);
+        background-color: var(--color-success-lighter);
+        border: var(--border-width-thin) solid var(--color-success);
+        color: var(--color-success);
         cursor: default;
+        outline: 1px solid var(--color-primary-lighter);
```
stefrado: Same as above

src/open_inwoner/scss/components/Form/ButtonRadio.scss
```diff
@@ -46,9 +57,8 @@
...
 
-  &:first-child {
-    .button-radio__button {
-      border-left: 1px solid #d2d2d2;
-    }
+  *[class*='icons'],
+  *[class*='Icons'] {
+    font-size: var(--font-size-body);
   }
 }
```
stefrado, line 73: No hover state like al other buttons?

src/open_inwoner/js/components/search/feedback.js
```diff
@@ -0,0 +1,59 @@
...
+    return this.node.querySelector('.feedback__label-text')
+  }
+
+  get translations() {
+    return this.node.querySelector('.feedback__translations')
+  }
+}
+
+// Start!
+;[...FEEDBACK_FORMS].forEach((node) => new FeedbackForm(node))
```
stefrado, line 78: I must say i do like the way you wrote this class! Looks very nice, clean and readable 👍

stefrado (changes requested):

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: Ah, good idea, all the others have a 'bump' animation. I 'll check if it's not on it. We don't really have nice hover-styles in the entire project.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: The focus/active border is a hard requirement - it won't always be blue (depends on municipalities' Primary color)

What color should we choose for this? A lighter hint of the Error/Success color will not stand out enough I think.

src/open_inwoner/templates/pages/search.html
jiromaykin: Ah yeah. The text is actually not required so users are allowed to just click the Thumbs up/down buttons, without typing text.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: Yes, this was a demand from an older Accessibility report 
(active/focus can be moved with arrowkeys but must remain visible). 

This style could be a generic color (on both buttons, higher up in the SCSS) or here we can set an individual outline for each Ja/Nee button respectively.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: same: This style could be a generic color (on both buttons, higher up in the SCSS) or here we can set an individual outline for each Ja/Nee button respectively with their own color.

src/open_inwoner/scss/components/Form/ButtonRadio.scss
jiromaykin: In order to comply with the maximum color value outline I'm making them look similar to this element in Mijn Zaken - detail page:

Screenshot 2025-03-25 at 14 19 09

stefrado:
One last point - is it intentional that i can submit the form without adding a feedback remark.
If this is intentional - let's merge. Otherwise please make sure the field is required.

jiromaykin:
> One last point - is it intentional that i can submit the form without adding a feedback remark. If this is intentional - let's merge. Otherwise please make sure the field is required.

I asked that same question last week in Figma but did not get a reply yet.

stefrado (approved):

jiromaykin:
Update: it would be very annoying if the text-field would be required if the answer is just "Yes'. So keeping it as is.

PR #1676, opened 2025-03-19
[#3136] Add option to fetch only zaken where user has relevant role

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3136


src/open_inwoner/openzaak/models.py
```diff
@@ -448,6 +448,15 @@ def form_service(self, service):
         ),
         default=False,
     )
+    relevant_role = models.CharField(
```
swrichards: The naming feels a bit opaque, especially when you're out of context (relevant for what?). What about:

```suggestion
    limit_user_visible_cases_to_role = models.CharField(
```
Bit more verbose, but perhaps more readable and explicit in the long run.

src/open_inwoner/openzaak/tests/test_cases.py
```diff
@@ -951,6 +963,74 @@ def test_list_cases_for_eherkenning_user_missing_rsin(self, m):
...
+            self.assertEqual(
+                list_zaken_req.qs,
+                {
+                    "rol__betrokkeneidentificatie__natuurlijkpersoon__inpbsn": [
+                        self.user.bsn
+                    ],
+                    "maximalevertrouwelijkheidaanduiding": [
+                        VertrouwelijkheidsAanduidingen.beperkt_openbaar
+                    ],
+                    "rol__omschrijvinggeneriek": [ZaakBetrokkeneRol.initiator],
```
swrichards, line 1030: Just double-checking: this is now all lower-case, is that an artifact of the mocking library (we pass `rol__omschrijvingGeneriek`).

swrichards (changes requested):

src/open_inwoner/openzaak/tests/test_cases.py
pi-sigma: I think so, yes, the discrepancy is quite annoying (it's the same for the other params).

src/open_inwoner/openzaak/models.py
```diff
@@ -448,6 +448,15 @@ def form_service(self, service):
         ),
         default=False,
     )
+    limit_user_visible_cases_to_role = models.CharField(
+        verbose_name=_("Betrokkene rol"),
```
swrichards: Sorry for another nitpick (I'd leave it for a separate PR, but because this has to be cherry-picked for a backport it's nice if it's in a single commit/migration): I'd like to have the verbose name reflect the function a bit better

```suggestion
        verbose_name=_("Limit cases to role"),
```

swrichards (changes requested):

swrichards (approved):

PR #1687, opened 2025-03-24
Feature branch: explicitly model vestigingen as part of the User model

swrichards:
This feature branch models vestigingen as part of the User model (each vestiging as a user, in addition to the legal entity itself, which has only a kvk nummer but not vestigingnummer). It involves changing the user model, OIDC/SAML login flows for eHerkenning, branch selection, and the parameters sent to the various backend APIs.


src/open_inwoner/kvk/branches.py
```diff
@@ -3,7 +3,3 @@
 
 def kvk_branch_selected_done(session) -> bool:
     return KVK_BRANCH_SESSION_VARIABLE in session
```
pi-sigma, line 5: Do we still  need this? I can't find any references to `kvk_branch_selected_done`, and the only refs to `KVK_BRANCH_SESSION_VARIABLE` are in tests.

pi-sigma (changes requested):

pi-sigma (approved):

src/open_inwoner/kvk/branches.py
swrichards: Discussed, @pi-sigma will do this in a followup PR post-merge.

pi-sigma: Discussed: we'll merge this one because other PR's need the changes. Cleanup will be done on develop

PR #1688, opened 2025-03-25
Bump django-csp-reports to 1.9.1

pi-sigma:
`description-file` (with a dash) in `setup.cfg` of older versions of `django-csp-reports` is deprecated as it doesn't conform to PEP naming conventions, causing CI to tail. It is replaced with `description_file` in newer versions.


swrichards:
Closed in favour of #1690 (needed a tweak to the script as well).

PR #1691, opened 2025-03-25
Integrate vestiging-users into eHerkenning OIDC/SAML and branch selection

swrichards:
This is the core of the vestiging-user change and breaks down into three parts:

- Updating the OIDC auth backend to handle vestiging-users
- Ibid, for the SAML (and SAML mock!) backends
- Managing branch selection and switching between vestiging-users

You'll want to go commit-by-commit, hopefully it's easy enough to follow. I created a helper (`EHerkenningSessionContext`) which does a lot of the heavy lifting. Specifically, it enforces that:

1. _If_ a vestiging was provided via OIDC/SAML claims, the user _cannot_ switch to either another vestiging _or_ the rechtsperson itself. This might ultimately not be the desired behavior, but let's be restrictive until we hear otherwise.
2. It tries to maintain parity with the existing branch selection behavior, basically: you are only redirected to the selection screen if your eHerkenning does not indicate a specific branch (meaning you _also_ shouldn't see it when you switch vestiging-users!).

The functionality whereby a vestiging is provided via OIDC/SAML is hard to test locally (our mocks don't support passing a vestiging right now), but you can still login with one of the KvK Test users and select one of the vestigingen. On your profile page, you'll see the currently selected vestiging (or not if you are the rechtspersoon itself).

### DIFF
--- a/src/eherkenning/backends.py
+++ b/src/eherkenning/backends.py
@@ -4,7 +4,7 @@ from digid_eherkenning.backends import eHerkenningBackend as _eHerkenningBackend
 from digid_eherkenning.exceptions import eHerkenningError
 from digid_eherkenning.utils import get_client_ip
 
-from open_inwoner.kvk.branches import KVK_BRANCH_SESSION_VARIABLE
+from open_inwoner.accounts.eherkenning_session import EHerkenningSessionContext
 
 UserModel = get_user_model()
 
@@ -13,6 +13,14 @@ class eHerkenningBackend(_eHerkenningBackend):
     """
     Custom backend to identify users based on the KvK number instead of RSIN
     """
+
+    def _persist_eherkenning_params_to_session(self, request, user):
+        session_context = EHerkenningSessionContext(request)
+        session_context.persist_eherkenning_state_for_user(
+            user=user,
+            is_branch_restricted=bool(user.vestiging),
+            initial_branch_selection_done=False,
+        )
 
     def get_company_branch_number(self, attributes):
         company_branch_number = attributes.get(
@@ -27,16 +35,20 @@ class eHerkenningBackend(_eHerkenningBackend):
                 "Login failed due to no KvK being returned by eHerkenning."
             )
 
+        vestigingsnummer = self.get_company_branch_number(saml_attributes)
+
         created = False
         try:
-            user = UserModel.eherkenning_objects.get_by_kvk(kvk)
+            user = UserModel.eherkenning_objects.get_by_kvk_and_vestiging(
+                kvk=kvk, vestiging=vestigingsnummer
+            )
         except UserModel.DoesNotExist:
-            user = UserModel.eherkenning_objects.create(kvk=kvk)
+            user = UserModel.eherkenning_objects.create(
+                kvk=kvk, vestiging=vestigingsnummer
+            )
             created = True
 
-        if vestigingsnummer := self.get_company_branch_number(saml_attributes):
-            self.request.session[KVK_BRANCH_SESSION_VARIABLE] = vestigingsnummer
-            self.request.session.save()
+        self._persist_eherkenning_params_to_session(request, user)
 
         success_message = self.error_messages["login_success"] % {
             "user": str(user),
--- a/src/open_inwoner/accounts/backends.py
+++ b/src/open_inwoner/accounts/backends.py
@@ -1,27 +1,29 @@
 import logging
-from typing import Literal
+from typing import Literal, NamedTuple
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.hashers import check_password
 from django.contrib.auth.models import AbstractUser
+from django.core.exceptions import SuspiciousOperation
 from django.urls import reverse, reverse_lazy
 
 from axes.backends import AxesBackend
 from digid_eherkenning.oidc.backends import BaseBackend
+from glom import Path, glom
 from mozilla_django_oidc_db.backends import OIDCAuthenticationBackend
 from mozilla_django_oidc_db.config import dynamic_setting
 from mozilla_django_oidc_db.typing import JSONObject
 from oath import accept_totp
 
+from open_inwoner.accounts.eherkenning_session import EHerkenningSessionContext
 from open_inwoner.configurations.models import SiteConfiguration
-from open_inwoner.kvk.branches import KVK_BRANCH_SESSION_VARIABLE
 from open_inwoner.utils.hash import generate_email_from_string
 from open_inwoner.utils.views import LogMixin
 
 from .choices import LoginTypeChoices
-from .models import OpenIDDigiDConfig, OpenIDEHerkenningConfig
+from .models import OpenIDDigiDConfig, OpenIDEHerkenningConfig, User
 
 logger = logging.getLogger(__name__)
 
@@ -191,6 +193,11 @@ class DigiDOIDCBackend(LogMixin, BaseBackend):
         return user
 
 
+class KvkEntityInformation(NamedTuple):
+    kvk: str
+    vestigingsnummer: str | None
+
+
 class EHerkenningOIDCBackend(LogMixin, BaseBackend):
     OIP_UNIQUE_ID_USER_FIELDNAME = dynamic_setting[Literal["kvk"]]()
     OIP_LOGIN_TYPE = dynamic_setting[LoginTypeChoices]()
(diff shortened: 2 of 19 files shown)


swrichards:
~@pi-sigma have a look at #1707 first to get that refactor out of the way, will make this easier to review~.

src/open_inwoner/accounts/backends.py
```diff
@@ -199,34 +206,74 @@ def _check_candidate_backend(self) -> bool:
...
+        # But there is no guarantee this mapping will be used, so it's difficult to
+        # raise confidently. Instead, for now, we log this so we can detect it early.
+        if identifier_type:
+            if identifier_type == "urn:etoegang:1.9:EntityConcernedID:RSIN":
+                logger.error(
+                    "eHerkenning backend configured to use RSIN whilst KvK was expected."
+                )
+            elif "rsin" in identifier_type.lower():
+                # TODO: We don't really support this
+                logger.warning("Possible RSIN configured")
```
pi-sigma: Perhaps something like this for easier debugging if this happens:

`logger.warning("eHerkenning backend configured with unsupported claim for identifier type: %s", identifyer_type)` 

Also: do we expect to change something about this because it's a `TODO`?

src/open_inwoner/accounts/backends.py
```diff
@@ -199,34 +206,74 @@ def _check_candidate_backend(self) -> bool:
...
+        if not entity_info:
             return self.UserModel.objects.none()
-        return self.UserModel.objects.filter(
-            **{f"{self.OIP_UNIQUE_ID_USER_FIELDNAME}__iexact": unique_id}
+
+        return self.UserModel.eherkenning_objects.filter_by_kvk_and_vestiging(
+            kvk=entity_info.kvk, vestiging=entity_info.vestigingsnummer
+        )
+
+    def _sync_eherkenning_session_context(self, user):
```
pi-sigma: "sync" makes it sound like this works in both directions, when in fact it only works in one (update eherkenning session context from user)

src/open_inwoner/accounts/backends.py
```diff
@@ -199,34 +206,74 @@ def _check_candidate_backend(self) -> bool:
...
+        session_context = EHerkenningSessionContext(self.request)
+        session_context.persist_eherkenning_state_for_user(
+            user=user,
+            # Restricted if no branch claim was provided
```
pi-sigma: Should be `Restricted if branch claim was provided`, no?

src/open_inwoner/accounts/backends.py
```diff
@@ -199,34 +206,74 @@ def _check_candidate_backend(self) -> bool:
...
+        # TODO: The user may be using a broker which provides RSIN rather than KvK,
+        # a scenario for which we are not equipped. Our usual model just assumes that
+        # the identifier claim is KvK, even if it's actually an RSIN. This is hard to
+        # check for, because there is no fixed convention as to the naming of the OIDC
+        # claims. In theory, they should match the SAML claims verbatim:
+        #
+        #   urn:etoegang:1.9:EntityConcernedID:KvKnr
+        #   urn:etoegang:1.9:EntityConcernedID:RSIN
+        #
+        # But there is no guarantee this mapping will be used, so it's difficult to
```
pi-sigma, line 239: Can we provide this information to the user in the admin to make sure they use the correct claim?

src/open_inwoner/accounts/eherkenning_session.py
```diff
@@ -0,0 +1,146 @@
...
+            request.session.get("_auth_user_backend", None)
+            not in EHerkenningSessionContext._expected_auth_backends()
+        ):
+            raise ValueError(
+                "The user in this session was not authenticated via the expected "
+                f"EHerkenning backends: {', '.join(EHerkenningSessionContext._expected_auth_backends())}"
+            )
+
+    def _set_session_flag(self, key: str, flag: bool):
+        self._request.session[key] = bool(flag)
```
pi-sigma: `bool(flag)` should not be necessary if the type of the argument is as expected (namely `bool`). If this is meant to make the method more resilient, I wonder if the result might be the opposite: if we unexpectedly get an argument of the wrong type because something went wrong and now convert it to `bool`...

src/open_inwoner/accounts/eherkenning_session.py
```diff
@@ -0,0 +1,146 @@
...
+        self._request.session.save()
+
+    def _get_session_flag(self, key: str):
```
pi-sigma: Type annotation for return value missing:

`def _get_session_flag(self, key: str) -> bool:`

Similarly for the other methods.

src/open_inwoner/accounts/eherkenning_session.py
```diff
@@ -0,0 +1,146 @@
...
+        *,
+        kvk: str,
+        vestiging: str | None,
+    ):
+        self.assert_valid_eherkenning_request(self._request)
+        self.assert_valid_eherkenning_user(self._request.user)
+
+        if self._request.user.kvk != kvk:
+            raise ValueError(
+                "You cannot change the authenticated user to the kvk for a different type"
```
pi-sigma: Does this mean: `You cannot change the authenticated user to a different kvk`?

src/open_inwoner/accounts/eherkenning_session.py
```diff
@@ -0,0 +1,146 @@
...
+                kvk=kvk, vestiging=vestiging
+            )
+        except User.DoesNotExist:
+            target_user = User.eherkenning_objects.create(kvk=kvk, vestiging=vestiging)
+
+        if target_user == self._request.user:
+            return  # Nothing to do
+
+        # Persist these values before the session is cleared
+        previous_is_branch_restricted = self.is_branch_restricted()
```
pi-sigma: If the user can change from one vestiging to another, how can the branch be restricted? If it were, shouldn't the user be prevented from changing to a different branch?

src/open_inwoner/accounts/tests/test_eherkenning_session.py
```diff
@@ -0,0 +1,204 @@
...
+            str(ctx.exception),
+            "The user in this session was not authenticated via the expected "
+            "EHerkenning backends: open_inwoner.accounts.backends."
+            "EHerkenningOIDCBackend, eherkenning.backends.eHerkenningBackend, "
+            "eherkenning.mock.backends.eHerkenningBackend",
+        )
+
+    def test_change_authenticated_user_raises_on_incorrect_backend(self):
+        request = self.make_request_with_session()
+        request.session["_auth_user_backend"] = "not.the.expected.backend"
```
pi-sigma: Perhaps use a real but incorrect backend (e.g. the digid backend)

src/open_inwoner/accounts/tests/test_eherkenning_session.py
```diff
@@ -0,0 +1,204 @@
...
+        mock_logout.assert_not_called()
+        self.assertEqual(request.user, self.user)
+        self.assertEqual(
+            request.session["_auth_user_backend"],
+            EHerkenningSessionContext._expected_auth_backends()[0],
+        )
+        self.assertFalse(context.is_branch_restricted())
+        self.assertFalse(context.is_initial_branch_selection_done())
+
+    def test_change_authenticated_user_maintains_branch_restriction_and_backend(self):
```
pi-sigma, line 213: I see a test that checks that a user cannot switch to a different kvk. Do we test that a vestiging user with branch restriction cannot switch to a different branch?

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -1060,6 +1062,59 @@ def setUpClass(cls):
...
+        mock_get_solo,
+        mock_get_token,
+        mock_verify_token,
+        mock_store_tokens,
+        mock_get_userinfo,
+        mock_retrieve_rsin_with_kvk,
+        mock_kvk,
+        mock_get_basisprofiel,
+    ):
+        """Test that missing claims configurations raises an exception."""
```
pi-sigma: The test name is self-explanatory and sufficient here

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -1073,6 +1128,58 @@ def setUpClass(cls):
...
+        mock_verify_token,
+        mock_store_tokens,
+        mock_get_userinfo,
+        mock_retrieve_rsin_with_kvk,
+        mock_kvk,
+        mock_get_basisprofiel,
+    ):
+        mock_get_userinfo.return_value = {
+            "email": "existing_user@example.com",
+            # "sub": "12345678",
```
pi-sigma: Left-over comment?

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -1201,6 +1315,173 @@ def test_new_user_is_created_when_new_kvk(
...
+        )
+
+        self.assertRedirects(
+            callback_response, reverse("pages-root"), fetch_redirect_response=False
+        )
+        db_user = User.objects.get()
+
+        mock_retrieve_rsin_with_kvk.assert_not_called()
+        self.assertEqual(db_user.email, existing_user.email)
+        self.assertEqual(db_user.kvk, "12345678")
```
pi-sigma: `self.assertEqual(db_user.kvk, existing_user.kvk)`

Similarly for the lines below

src/eherkenning/backends.py
```diff
@@ -14,6 +14,14 @@ class eHerkenningBackend(_eHerkenningBackend):
     Custom backend to identify users based on the KvK number instead of RSIN
     """
 
+    def _sync_eherkenning_session_context(self, request, user):
+        session_context = EHerkenningSessionContext(request)
+        session_context.persist_eherkenning_state_for_user(
+            user=user,
+            is_branch_restricted=bool(user.vestiging),
```
pi-sigma, line 21: For the `EHerkenningOIDCBackend` you have `is_branch_restricted=user.vestiging != ""`. Should be consistent.

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -962,18 +966,27 @@ def test_eherkenning_user_is_redirected_to_necessary_registration(
 
         user = eHerkenningUserFactory.create(kvk="12345678", email="example@localhost")
 
-        response = self.app.get(reverse("pages-root"), user=user)
+        self.client.force_login(
+            user, backend=EHerkenningSessionContext._expected_auth_backends()[0]
+        )
+        response = self.client.get(reverse("pages-root"), user=user)
```
pi-sigma: Do we need to switch from `app` to `client` for this test (we're using the `WebTest`'s `app` in the other tests)?

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -101,13 +157,18 @@ def test_post_branches_page_with_incorrect_vestigingsnummer(
...
+        )
         response = self.client.post(self.url, data={"branch_number": "4321"})
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(kvk_branch_selected_done(self.client.session), False)
-        self.assertNotIn(KVK_BRANCH_SESSION_VARIABLE, self.client.session)
+        self.assertEqual(
+            response.wsgi_request.user,
+            self.user,
+            msg="No branch change effectuated",
```
pi-sigma: `No branch change effectuated` is a bit cryptic/elaborate. Perhaps `KVK branch was not changed`

src/open_inwoner/kvk/views.py
```diff
@@ -57,10 +60,34 @@ def get_redirect(self):
...
+
+        # Only eHerkenning users can potentially switch to a vestiging
+        try:
+            context.assert_valid_eherkenning_user(user)
+        except ValueError:
             return HttpResponse(_("Unauthorized"), status=401)
 
+        if context.is_branch_restricted():
+            return HttpResponse(
+                _("Your eHerkenning account cannot access other branches."),
```
pi-sigma, line 75: Do we have a test for this?

pi-sigma (changes requested):
Double-checking: we discussed to make the branch selection page accessible via a link on "Mijn Profiel". This is for a separate PR?

src/eherkenning/backends.py
swrichards: Check!

src/open_inwoner/accounts/backends.py
swrichards: Good idea, I'll add a note to the `help_text`.

src/open_inwoner/accounts/backends.py
swrichards: The `TODO` would be to support RSIN. Changed the comment to clarify that, and also simplified the condition to only check if `kvk` is missing from the identifier type.

src/open_inwoner/accounts/eherkenning_session.py
swrichards: Fair point, this does neither resilience nor validation. I'll raise a `ValueError`, #Yolo

src/open_inwoner/accounts/eherkenning_session.py
swrichards: Good point, I'll add a guard for that and change this to the literal.

src/open_inwoner/accounts/tests/test_auth.py
swrichards: Good catch, it was previously needed to be able to access the session directly, but that need dissipated in a refactor. Will change back.

src/open_inwoner/accounts/tests/test_eherkenning_session.py
swrichards: :facepalm:

src/open_inwoner/kvk/views.py
swrichards: We do!

```
def test_post_as_branch_restricted_user_throws_401(self):
        self.client.force_login(eHerkenningVestigingUserFactory())

        # Note: we have to persist the session in a separate variable because every
        # property access on self.client insantiates a new session
        session = self.client.session
        session.update(
            {EHerkenningSessionContext.KVK_BRANCH_RESTRICTION_SESSION_KEY: True}
        )
        session.save()

        response = self.client.post(self.url)

        self.assertEqual(response.status_code, 401)
```

src/open_inwoner/accounts/backends.py
swrichards: Actually, I am making a TODO for this: it would have to happen in the library, or need some system check, etc. I'll get back to it, but would be good to not hold this up for that.

src/open_inwoner/accounts/tests/test_eherkenning_session.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+
+        self.assertEqual(
+            str(ctx.exception),
+            "You cannot change the authenticated user to the rechtspersoon or "
+            "vestiging of a different kvk number",
+        )
+        self.assertEqual(request.user, self.user)
+
+    def test_change_authenticated_user_with_branch_restriction_raises(
```
pi-sigma: Duplicate of l.107 ff. I like the formatting here better (separation of setup, test, assert with newlines), otherwise they're they same.

src/eherkenning/backends.py
```diff
@@ -14,6 +14,14 @@ class eHerkenningBackend(_eHerkenningBackend):
     Custom backend to identify users based on the KvK number instead of RSIN
     """
 
+    def _sync_eherkenning_session_context(self, request, user):
```
pi-sigma: For the OIDC backend you changed it to `_persist_eherkenning_params_to_session`. Should be the same (I should have marked this in the first review as well)

pi-sigma (changes requested):

src/open_inwoner/accounts/tests/test_eherkenning_session.py
swrichards: Good catch, must've messed up my rebase, thanks.

pi-sigma (approved):

PR #1692, opened 2025-03-25
[#3126] Make use of user.vestiging to fetch klanten and contactmomenten

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3126

### DIFF
--- a/src/open_inwoner/openklant/tests/test_fetch.py
+++ b/src/open_inwoner/openklant/tests/test_fetch.py
@@ -2,6 +2,11 @@ from django.test import TestCase
 
 import requests_mock
 
+from open_inwoner.accounts.tests.factories import (
+    DigidUserFactory,
+    eHerkenningUserFactory,
+    eHerkenningVestigingUserFactory,
+)
 from open_inwoner.openklant.api_models import KlantContactMoment
 from open_inwoner.openklant.tests.data import MockAPIReadData
 from open_inwoner.openklant.wrap import (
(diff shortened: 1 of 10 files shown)


src/open_inwoner/openklant/tests/test_fetch.py
```diff
@@ -58,3 +66,43 @@ def test_fetch_klantcontactmoment_for_kvk(self, m):
         self.assertIsNotNone(kcm)
         self.assertIsInstance(kcm, KlantContactMoment)
         self.assertEqual(str(kcm.uuid), data.klant_contactmoment2["uuid"])
+
+
+class KlantFetchParametersTest(TestCase):
```
swrichards, line 71: I think it couldn't hurt in this case to also assert that the fetch parameters are actually used by the OK2/eSuite fetchers respectively: we don't need full on mocks with actual responses, we just need to assert against the request history that the vestigingsnummer/kvk are properly bassed on to the target URL.

swrichards (changes requested):

src/open_inwoner/openklant/tests/test_fetch.py
pi-sigma: What's the best approach for this while avoiding to mock out all the requests involved and providing mocks to avoid errors from the services/clients? Is it acceptable to `try/except` the fetch request and then assert against the request_mock history (making a request will produce and error without mocks, but the request_history is still there)?

swrichards: Ah so I meant we do add a mock entry for the _endpoint_, but we don't have to return much of a sane response. But I guess we then run into validation errors and such?

pi-sigma: Yeah, there are intermediate requests which need to be mocked, and the response data must have a certain shape to avoid validation errors.

swrichards: Then let's skip that for now, but as a stretch goal this week see if we can at least add one test/change per category of test to ensure the vestiging actually trickles through to the backends.

swrichards (approved):

swrichards:
@pi-sigma this needs a rebase it seems due to a conflict.

PR #1695, opened 2025-03-26
[#3143] Make use of user.vestiging for updates from KVK API

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3143

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -84,7 +84,10 @@ def _update_user_from_esuite(
 def _update_eherkenning_user_from_kvk_api(user: User):
     kvk_client = KvKClient()
 
-    vestiging = kvk_client.get_company_headquarters(kvk=user.kvk)
+    if user.vestiging:
+        vestiging = kvk_client.get_vestiging(vestiging=user.vestiging)
+    else:
+        vestiging = kvk_client.get_company_headquarters(kvk=user.kvk)
 
     if company_name := vestiging.get("naam"):
         user.company_name = company_name
--- a/src/open_inwoner/kvk/client.py
+++ b/src/open_inwoner/kvk/client.py
@@ -105,6 +105,15 @@ class KvKClient:
 
         return headquarters[0]
 
+    def get_vestiging(self, vestiging: str) -> dict | None:
+        vestigingen = self.search(vestigingsnummer=vestiging).get("resultaten", [])
+
+        if not vestigingen:
+            logger.info("No vestiging found for vestigingsnummer %s", vestiging)
+            return None
+
+        return vestigingen[0]
+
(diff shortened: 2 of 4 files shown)


src/open_inwoner/kvk/client.py
```diff
@@ -105,6 +105,15 @@ def get_company_headquarters(self, kvk: str, **kwargs) -> dict:
 
         return headquarters[0]
 
+    def get_vestiting(self, vestiging: str) -> dict:
```
swrichards:
```suggestion
    def get_vestiging(self, vestiging: str) -> dict:
```

src/open_inwoner/kvk/client.py
```diff
@@ -105,6 +105,15 @@ def get_company_headquarters(self, kvk: str, **kwargs) -> dict:
...
+        vestigingen = self.search(vestigingsnummer=vestiging).get("resultaten", [])
+
+        if not vestigingen:
+            logger.info("No vestiging found for vestigingsnummer %s", vestiging)
+            return {}
```
swrichards: I think `None` is more explicit here.

src/open_inwoner/accounts/signals.py
```diff
@@ -84,7 +84,10 @@ def _update_user_from_esuite(
 def _update_eherkenning_user_from_kvk_api(user: User):
     kvk_client = KvKClient()
 
-    vestiging = kvk_client.get_company_headquarters(kvk=user.kvk)
+    if user.vestiging:
+        kvk_client.get_vestiting(vestiging=user.vestiging)
```
swrichards: Missed one:
```suggestion
        kvk_client.get_vestiging(vestiging=user.vestiging)
```

swrichards: @pi-sigma typo still appears to be there

swrichards: (Maybe also a sign that we don't have decent test coverage here, maybe a simple unit test where we mock `KvkClient` and simply assert the correct arguments are passed?

src/open_inwoner/accounts/signals.py
```diff
@@ -84,7 +84,10 @@ def _update_user_from_esuite(
 def _update_eherkenning_user_from_kvk_api(user: User):
     kvk_client = KvKClient()
 
-    vestiging = kvk_client.get_company_headquarters(kvk=user.kvk)
+    if user.vestiging:
+        kvk_client.get_vestiging(vestiging=user.vestiging)
+    else:
+        vestiging = kvk_client.get_company_headquarters(kvk=user.kvk)
```
```suggestion
        vestiging = kvk_client.get_vestiging(vestiging=user.vestiging) if user.vestiging else kvk_client.get_company_headquarters(kvk=user.kvk)
```

(Both paths need to set `vestiging`, otherwise it's potentially undefined on `l.92`).

swrichards (approved):

PR #1698, opened 2025-03-26
[#3127] Update case service/client to work with user.vestiging

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3127

### DIFF
--- a/src/open_inwoner/pdc/managers.py
+++ b/src/open_inwoner/pdc/managers.py
@@ -12,6 +12,7 @@ from open_inwoner.openzaak.api_models import Zaak
 from open_inwoner.openzaak.clients import MultiZgwClientProxy, build_zaken_clients
 from open_inwoner.openzaak.models import ZaakTypeConfig
 from open_inwoner.openzaak.utils import get_user_fetch_parameters
+from open_inwoner.utils.views import LogMixin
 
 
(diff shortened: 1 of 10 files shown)


src/open_inwoner/pdc/managers.py
```diff
@@ -12,6 +12,7 @@
 from open_inwoner.openzaak.clients import MultiZgwClientProxy, build_zaken_clients
 from open_inwoner.openzaak.models import ZaakTypeConfig
 from open_inwoner.openzaak.utils import get_user_fetch_parameters
+from open_inwoner.utils.views import LogMixin
```
swrichards, line 15: Why where these needed for this PR?

swrichards (changes requested):

src/open_inwoner/pdc/managers.py
pi-sigma: Strictly speaking not needed, but a test was failing because the class tried to log something when the `LogMixin` was not present. I could do this in a separate PR though.

swrichards: All good, just checking :+1: .

swrichards (approved):

PR #1699, opened 2025-03-27
Preliminary refactors and cleanups for vestigingen users

swrichards:
Split off from #1691 -- just some small tweaks and refactors to avoid cluttering the main PR.

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -249,8 +249,16 @@ class eSuiteKlantenService(
         *,
         data: KlantWritePayload | None = None,
     ) -> Klant | None:
-        if sum(bool(arg) for arg in (user_bsn, user_kvk_or_rsin, vestigingsnummer)) > 1:
-            raise ValueError("Only one argument can be specified")
+        if user_bsn and user_kvk_or_rsin:
+            raise ValueError(
+                "Specify either `user_bsn` or `user_kvk_or_rsin` with optional "
+                "`vestigingsnummer`"
+            )
+
+        if vestigingsnummer and not user_kvk_or_rsin:
+            raise ValueError(
+                "You must specify `user_kvk_or_rsin` if `vestigingsnummer` is set"
+            )
 
         payload = {}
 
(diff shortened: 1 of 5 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -249,8 +249,13 @@ def create_klant(
         *,
         data: KlantWritePayload | None = None,
     ) -> Klant | None:
-        if sum(bool(arg) for arg in (user_bsn, user_kvk_or_rsin, vestigingsnummer)) > 1:
-            raise ValueError("Only one argument can be specified")
+        if user_bsn and user_kvk_or_rsin:
+            raise ValueError("Specify either a natural person or a legal person")
```
pi-sigma: I'm curious about the form of the error message. This is an instruction, but for whom? The user who performs some action that would trigger the error won't see this message, or will they? It depends on how `ValueError` for `service.create_klant` is handled by the caller, but as far as I can see, this is an unexpected error condition that the user can't do anything about. Perhaps `Failed to create klant because both user_bsn and user_kvk_or_rsin supplied, expected one` or something like it? Similarly for the following one.

src/open_inwoner/openklant/tests/test_esuite_service.py
```diff
@@ -22,24 +19,27 @@ def setUp(self):
...
+            {
+                "user_bsn": "123",
+                "user_kvk_or_rsin": "123",
+                "vestigingsnummer": "123",
+            },
+            # Needs kvk or rsin
+            {
+                "vestigingsnummer": "123",
+            },
+        ):
```
pi-sigma, line 43: Maybe add a case that includes both `user_bsn` and `vestigingsnummer` to exclude the possibility that `user_bsn` is treated by the code as equivalent to `kvk_or_rsin`.

pi-sigma (approved):
Looking good! Two suggestions, but nothing big.

src/open_inwoner/openklant/services.py
swrichards: I'll rewrite it to name the specific args.

PR #1700, opened 2025-03-27
⬆️  [#3145] Upgrade HTMX and dependencies

jiromaykin:
In HTMX v1, htmx was globally available via window.htmx.

In HTMX v2, you must import it explicitly in modern module-based builds.

issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3145 

⚠️  Note:

### Run `npm ci --legacy-peer-deps` and do an `npm run build` first!
 

1. @babel/polyfill is now deprecated with new HTMX.
2. It was deprecated in Babel 7.4.0. It pollutes the global scope, which can cause conflicts.
4. Instead, Babel recommends using core-js and regenerator-runtime explicitly.
6.  "0,25% not dead" means: "Transpile JavaScript only for modern browsers that have more than 0.25% market share, and are still actively maintained."
8. +Update for this 3rd party: https://django-htmx.readthedocs.io/en/latest/


jiromaykin:
@stefrado @swrichards Deze hoeft echt nog niet in de huidige nieuwe release 1.29 - Laten we eerst maar testen of het allemaal werkt en goed gaat...

src/open_inwoner/js/components/cases/spinner.js
```diff
@@ -1,37 +1,39 @@
-const htmx = (window.htmx = require('htmx.org'))
+document.addEventListener('DOMContentLoaded', function () {
+  const htmx = window.htmx
```
stefrado, line 2: Please fix indentation in this file

stefrado (changes requested):
Small indentation comment.

stefrado (changes requested):
Somehow i can't interact with any of the buttons.

- fixed with `npm ci --legacy-peer-deps` and a new build `npm run build`

src/open_inwoner/js/components/cases/spinner.js
jiromaykin: No, Prettier did it :-D 
because of new higher level function-wrapper, due HTMX v2:  "...import htmx explicitly in modern module-based builds..."

I can't find another way to make the Spinner work - suggestions welcome.

swrichards (approved):
Works on my machine (tm). Let's get this in now `v1.29.0` is out and deploy it to test soon so we can start putting it through its paces.

stefrado (approved):

PR #1701, opened 2025-03-27
Consolidate eherkenning post login handling

swrichards:
Another cleanup for #1691 -- this time, it ensures the eHerkenning user is fully synced before pushing data to the klanten systeem. This surfaced in the other PR because, if the config stipulates that RSIN is needed to identify users, we would get failures when syncing to the klanten API because the RSIN could not be inferred in the OIDC/SAML backends.

This PR also took the opportunity to clean this logic up a bit and refactor the tests.

```
    Merge on_kvk_change into account post_login sync
    
    This commit consolidates the RSIN updating logic for the
    post_login signal in the accounts module. Apart from being
    cleaner (keeping user data synchronized is an accounts module
    concern, not a kvk module concern), it also ensures all the
    various eHerkenning user updates happen _before_ any pushes
    to the klanten APIs take place. This was not guaranteed to
    happen when the sync functions were split up, causing issues
    in the klanten API syncs which, depending on the config,
    might require the RSIN to be used to identify klanten.

    Handle exceptions in the eHerkenning user_logged_in signal
    
    This updater runs as part of a signal, therefore it
    should be resilient to common failure modes, most
    notably an error when communicating with the API.
```


src/open_inwoner/accounts/signals.py
```diff
@@ -82,16 +84,49 @@ def _update_user_from_esuite(
...
+        # only have to fetch it if it's not set.
+        if not user.rsin:
+            rsin = kvk_client.retrieve_rsin_with_kvk(user.kvk)
+
+            if rsin:
+                user.rsin = rsin
+                user.save(update_fields=["rsin"])
+                updated_fields.append("rsin")
+            else:
+                logger.error("Unable to connect to KvK API to retrieve rsin")
```
pi-sigma: It's not 
```suggestion
                logger.error("Unable to retrieve RSIN from KvK API")
```
It's not necessarily a connection error (perhaps also log the KVK number so that we know for which KVK the attempt failed)

pi-sigma (approved):

PR #1705, opened 2025-03-31
Move BRP user sync to main user_logged_in handler

swrichards:
Small refactor to round off #1701 -- the function is still run on the same signal, just in a different handler.

### DIFF
--- a/src/open_inwoner/accounts/signals.py
+++ b/src/open_inwoner/accounts/signals.py
@@ -44,6 +44,10 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
     if user.is_eherkenning_user:
         _update_eherkenning_user_from_kvk_api(user=user)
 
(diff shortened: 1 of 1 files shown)


pi-sigma (approved):

pi-sigma:
@swrichards Somehow my comment disappeared. I suggested moving `brp_config = HaalCentraalConfig.get_solo()` into the consquent of the conditional as another check `if HaalCentraalConfig.get_solo().service`, as we don't need the `brp_config` elsewhere.

swrichards:
> @swrichards Somehow my comment disappeared. I suggested moving `brp_config = HaalCentraalConfig.get_solo()` into the consquent of the conditional as another check `if HaalCentraalConfig.get_solo().service`, as we don't need the `brp_config` elsewhere.

Check, fixed.

PR #1707, opened 2025-03-31
[#3101] Split DigiD and eHerkenning OIDC backends

swrichards:
Another precursor for #1691.

### DIFF
--- a/src/open_inwoner/accounts/backends.py
+++ b/src/open_inwoner/accounts/backends.py
@@ -151,16 +151,53 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
         return self.UserModel.objects.filter(**{"oidc_id__iexact": unique_id})
 
 
-class DigiDEHerkenningOIDCBackend(LogMixin, BaseBackend):
-    OIP_UNIQUE_ID_USER_FIELDNAME = dynamic_setting[Literal["bsn", "kvk"]]()
+class DigiDOIDCBackend(LogMixin, BaseBackend):
+    OIP_UNIQUE_ID_USER_FIELDNAME = dynamic_setting[Literal["bsn"]]()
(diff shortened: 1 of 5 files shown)


src/open_inwoner/accounts/backends.py
```diff
@@ -151,17 +151,54 @@ def filter_users_by_claims(self, claims):
...
+
+class EHerkenningOIDCBackend(LogMixin, BaseBackend):
+    OIP_UNIQUE_ID_USER_FIELDNAME = dynamic_setting[Literal["kvk"]]()
+    OIP_LOGIN_TYPE = dynamic_setting[LoginTypeChoices]()
+
+    def _check_candidate_backend(self) -> bool:
+        parent = super()._check_candidate_backend()
+        return parent and self.config_class is OpenIDEHerkenningConfig
+
     def _store_vestigingsnummer_in_session(self, claims: JSONObject):
```
pi-sigma, line 202: Do we still use the vestigingsnummer in the session now that we have `user.vestiging`?

pi-sigma (commented):
Do we still need to store the vestiging in the session (see comment ad loc)? If not, the method could be deleted from the eHerkennning backend. In that case, I wonder if it makes sense to lump the identical functions in an intermediate class (could be `BaseBackend` if the corresponding name of the import of `BaseBackend` is changed), as there is not much that's specific to the two backends.

src/open_inwoner/accounts/backends.py
swrichards: @pi-sigma We indeed don't, but to be clear: this PR is mainly meant to maintain parity with the existing functionality, the actual changes are upcoming in #1691 (I split this one off to make reviewing the other PR a bit easier).

pi-sigma (approved):

PR #1712, opened 2025-04-02
Persist full session on eHerkenning user switch

swrichards:
When switching users for a different vestiging, the user will initially have authenticated through SAML or OIDC. In the latter case especially, there will be a number of session variables on which the OIDC flows depend. For instance, the ID token may be persisted and this will be used for the logout flow.

When a user switches to a different vestiging-user, we do not want to break any of these flows. Essentially, it should be as if the session was not changed _except_ for the user (and possible the branch selection flag).

To address this, we copy over all session variables, apart from the Django auth session keys which obviously will change, because we now have a new user.

The alternative was to copy over the specific variables which the OIDC backend expects, but that leaves us vulnerable to breaking interface changes in these packages. More generally, other modules might depend on the existence of certain session variables. In that light, simply copying over as much as we can seems like the safest path forward.

### DIFF
--- a/src/open_inwoner/accounts/eherkenning_session.py
+++ b/src/open_inwoner/accounts/eherkenning_session.py
@@ -1,5 +1,11 @@
 from django.conf import settings
-from django.contrib.auth import login as django_login, logout as django_logout
+from django.contrib.auth import (
+    BACKEND_SESSION_KEY,
+    HASH_SESSION_KEY,
+    SESSION_KEY,
+    login as django_login,
+    logout as django_logout,
+)
 from django.http import HttpRequest
 
 from open_inwoner.accounts.models import User
@@ -138,6 +144,7 @@ class EHerkenningSessionContext:
         # Persist these values before the session is cleared
         previous_backend = self._request.session["_auth_user_backend"]
         previous_initial_branch_selection_done = self.is_initial_branch_selection_done()
+        previous_session = dict(self._request.session)
 
         # Clear the session
         django_logout(self._request)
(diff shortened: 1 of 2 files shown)


src/open_inwoner/accounts/tests/test_eherkenning_session.py
```diff
@@ -232,3 +232,31 @@ def test_change_authenticated_user_maintains_branch_restriction_and_backend(self
...
+
+        self.assertEqual(request.user, self.vestiging_user)
+        self.assertEqual(
+            request.session["_auth_user_backend"],
+            EHerkenningSessionContext._expected_auth_backends()[0],
+        )
+        self.assertEqual(context.is_branch_restricted(), False)
+        self.assertFalse(context.is_initial_branch_selection_done())
+        self.assertEqual(context._request.session["foo"], "bar")
+        self.assertEqual(context._request.session["baz"], 42)
```
pi-sigma, line 271: We could also check that the `HASH_SESSION_KEY` is indeed changed since we have a new user. Set `session["_auth_user_hash"] = "1234"` and then `self.assertNotEqual(context._request.session["_auth_user_hash"], session["_auth_user_hash"])`.

src/open_inwoner/accounts/eherkenning_session.py
```diff
@@ -148,6 +155,16 @@ def change_authenticated_user(
             user=target_user,
             backend=previous_backend,
         )
+
+        # We want to preserve as much as possible from the previous session values
+        # as possible. This is especially important because digi
```
pi-sigma: Comment incomplete

pi-sigma (changes requested):

src/open_inwoner/accounts/eherkenning_session.py
swrichards: Careless: finished the thought in the commit message but not here :)

src/open_inwoner/accounts/tests/test_eherkenning_session.py
swrichards: Quite right

pi-sigma (approved):

PR #1717, opened 2025-04-03
[#3153] Add rich text-editor for zero results box

jiromaykin:
User story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3152?milestone=530
Task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3153


jiromaykin:
@pi-sigma @stefrado I am not sure if in Search.html I should add **Markdown** to `{{ configurable_text.search_page.search_zero_results_text|safe }}`.
We do use that for `<div class="wysiwyg">{{ login_text|markdown|safe }}</div>` but not sure if that is a requirement (probably won't hurt?).

pi-sigma:
@jiromaykin The content is stored as HTML by the editor, so `markdown` has no effect. The same holds for `login_text`.

pi-sigma (approved):

src/open_inwoner/templates/pages/search.html
```diff
@@ -97,13 +97,16 @@ <h2 class="utrecht-heading-2">{% trans "Heeft u gevonden wat u zocht?" %}</h2>
                         <h2 class="utrecht-heading-2 search-results__title">
                             {% trans "We konden geen zoekresultaten vinden" %}
```
stefrado, line 98: Is this title left here on purpose?
Why can't i change the whole no result text and am restricted to only the content and not the title?

src/open_inwoner/templates/pages/search.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n form_tags utils icon_tags grid_tags pagination_tags filter_tags %}
+{% load i18n form_tags string_tags icon_tags grid_tags pagination_tags filter_tags %}
```
stefrado:
```suggestion
{% load i18n form_tags icon_tags grid_tags pagination_tags filter_tags %}
```

String_tags is included, but `safe` is part of the built-in filters and can be used without this inclusion.

stefrado (changes requested):

src/open_inwoner/templates/pages/search.html
jiromaykin: This was indeed left here on purpose - look at the description in the user story in Taiga:
"De heading blijft hetzelfde"

src/open_inwoner/templates/pages/search.html
jiromaykin: Ah yes - in my first version I made use of our `optional_paragraph` tag which is part of `string_tags`. I'll remove it here.

stefrado (approved):

PR #1720, opened 2025-04-07
[#3128] Distinguish between company_name and branch_name

pi-sigma:


src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -43,6 +43,10 @@ <h2 class="utrecht-heading-2" id="business-overview">{% trans "Bedrijfsgegevens"
                 <div class="tabled__item tabled__key">{% trans "Vestigingsnummer" %}</div>
                 <div class="tabled__item tabled__value">{{ user.vestiging }}</div>
             </div>
+            <div class="tabled__row tabled__row--blank">
+                <div class="tabled__item tabled__key">{% trans "Trading name" %}</div>
```
swrichards: Let's use `Branch name` for the label too, to avoid confusion if people actually have trading names distinct from the branch names.

src/open_inwoner/kvk/client.py
```diff
@@ -136,10 +140,23 @@ def get_all_company_branches(self, kvk: str, **kwargs) -> list[dict | None]:
...
+                f"{self.vestigingsprofielen_endpoint}/{vestiging}"
+            )
+        ):
+            logger.error("Unable to retrieve vestigingsprofiel for %s", vestiging)
+        return vestigingsprofiel
+
+    def retrieve_rsin_with_kvk(
+        self, kvk: str, basisprofiel: dict | None = None
+    ) -> str | None:
+        if not (basisprofiel := basisprofiel or self.get_basisprofiel(kvk)):
```
swrichards: This injection of `basisprofiel` feels a bit messy to me: note that we already added a cache decorator to `get_basisprofiel`, so if this is mainly to avoid duplicate requests, that should be covered by the cache, given that the requests will be done in short succession.

src/open_inwoner/accounts/signals.py
```diff
@@ -95,27 +95,31 @@ def _update_eherkenning_user_from_kvk_api(user: User):
...
-                user.save(update_fields=["company_name"])
-                updated_fields.append("company_name")
+        basisprofiel = kvk_client.get_basisprofiel(kvk=user.kvk)
+        company_name = basisprofiel.get("naam")
+        if company_name and user.company_name != company_name:
+            user.company_name = company_name
+            user.save(update_fields=["company_name"])
+            updated_fields.append("company_name")
+
+        # Update branch name (use hoofdvestiging for rechtspersoon)
```
swrichards: We don't appear to use the hoofdvestiging anymore: that's fine I think, because the hoofdvestiging is still a vestiging and would be its own user, so the rechtspersoon should be something different (as here the first handelsnaam, but see comment below)

```suggestion
        # Update branch name (use first handelsnaam for rechtspersoon)
```

src/open_inwoner/accounts/signals.py
```diff
@@ -95,27 +95,31 @@ def _update_eherkenning_user_from_kvk_api(user: User):
...
-            logger.error(
-                "Unable to sync company_name from KvK API",
-                extra=extra_exc_params,
-            )
+            company_names = basisprofiel.get("handelsnamen")
+            branch_name = company_names[0].get("naam") if company_names else ""
+        if branch_name and user.branch_name != branch_name:
+            user.branch_name = branch_name
+            user.save()
+            updated_fields.append("branch_name")
```
swrichards: This remains tricky.

The name feels off to me because of the usual reason: a rechtspersoon is not a branch, and thus can't have a branch name.

How would you feel about simply ignoring this option? The rechtspersoon has no branch name, and we hide it in the template if no vestiging is set. That way there's no ambiguity about what the branch name refers to.

swrichards (changes requested):

src/open_inwoner/accounts/signals.py
pi-sigma: Yeah I was second-guessing myself about this one. I'll change it

src/open_inwoner/kvk/client.py
pi-sigma: You're right about the motivation, I didn't think of the caching.

src/open_inwoner/accounts/signals.py
```diff
@@ -96,27 +96,26 @@ def _update_eherkenning_user_from_kvk_api(user: User):
...
-                extra=extra_exc_params,
-            )
+        basisprofiel = kvk_client.get_basisprofiel(kvk=user.kvk)
+        company_name = basisprofiel.get("naam")
+        if company_name and user.company_name != company_name:
+            user.company_name = company_name
+            user.save(update_fields=["company_name"])
+            updated_fields.append("company_name")
+
+        # Update branch name (use first handelsnaam for rechtspersoon)
```
```suggestion
        # Update branch name (leave empty for rechtspersoon)
```

src/open_inwoner/kvk/client.py
```diff
@@ -136,11 +140,23 @@ def get_all_company_branches(self, kvk: str, **kwargs) -> list[dict | None]:
 
     @cache_result("kvk:{kvk}")
     def get_basisprofiel(self, kvk: str) -> dict:
-        return self._request(f"{self.basisprofielen_endpoint}/{kvk}")
-
-    def retrieve_rsin_with_kvk(self, kvk: str) -> str | None:
-        if not (basisprofiel := self.get_basisprofiel(kvk)):
-            return None
+        if not (basisprofiel := self._request(f"{self.basisprofielen_endpoint}/{kvk}")):
+            logger.error("Unable to retrieve basisprofiel for KVK %s", kvk)
```
swrichards: Note we'll already get a `logger.exception` from the `self._request` logic. This might be overkill (we should be able to get the location from the stacktrace).

swrichards (approved):
Just a nitpick

src/open_inwoner/kvk/client.py
pi-sigma: Right, I'll make the same change for `get_vestigingsprofiel`.

PR #1722, opened 2025-04-07
:lipstick: [#3156] Add new design for Samenwerkings detail page (top half)

jiromaykin:
Feature with screenshot of current changes: https://taiga.maykinmedia.nl/project/open-inwoner/task/3156 but do note that the gap width is now 40px instead of 80px

- [x] - remove PDF dropdown 
- [x] - change Gumshoe color to accent 
- [x] - make dashboard border grow with content 
- [x] - move PDF export button into anchor Gumshoe menu
- [x] - add bewerk button 
- [x] - add different create action button
- [x] - add sections with border and headergroups 
- [x] - add file component
- [x] add reusable section styling for sections with or without borders and headings
- [x] correct "forced right" buttons for mobile in tabled sections


src/open_inwoner/components/templates/components/Action/Actions.html
```diff
@@ -92,4 +92,10 @@
             <div class="table__item" colspan="4"><p class="utrecht-paragraph">{% trans "er zijn geen acties gevonden met de huidige filters of er zijn geen acties" %}</p></div>
         {% endfor %}
     </div>
+    {# TODO: conditional rendering cap of button to be determined later #}
+    {% if actions|length > 3 %}
```
stefrado, line 96: I think > 3 is a good default and works well.

src/open_inwoner/scss/components/Plan/PlanContact.scss
```diff
@@ -23,7 +13,16 @@
...
 
   .plan-contact__info {
+    display: inline-block;
     margin: 0;
+    word-wrap: break-word;
+    word-break: break-word;
+    white-space: normal;
+  }
+
+  .plan-contact__info {
```
stefrado: Duplicate className

src/open_inwoner/scss/components/Plan/PlanContact.scss
```diff
@@ -1,15 +1,5 @@
...
-  gap: var(--spacing-medium);
-  margin-bottom: 56px;
-  color: var(--font-color-body);
-  font-family: var(--font-family-body);
-  font-size: var(--font-size-body);
-
-  @media (min-width: 768px) {
-    grid-template-columns: 1fr 1fr 2fr;
-  }
+  //margin-bottom: 0;
```
stefrado: Forgotten comment?

src/open_inwoner/scss/components/Dashboard/Dashboard.scss
```diff
@@ -20,28 +20,29 @@
...
-      @media (min-width: 900px) {
-        content: none;
-      }
-    }
+    //&:last-child p::after {
+    //  @media (min-width: 900px) {
+    //    content: none;
+    //    border: none;
+    //  }
+    //}
```
stefrado: Forgotten comments?

src/open_inwoner/templates/pages/plans/detail.html
```diff
@@ -2,82 +2,104 @@
...
+    <section class="tabled tabled--flexible oip-section">
+        <div class="tabled__section">
+            <header class="oip-header-group title">
+                <h2 class="utrecht-heading-2" id="files">{% trans "Samenwerkingsplan" %}</h2>
+            </header>
+        </div>
+        <div class="plan__file">
+            {% if object.get_latest_file %}
+                {% url "profile:documents_download" uuid=object.get_latest_file.uuid as download_url %}
+                {% file file=object.get_latest_file.file created=file.created_on extension=plan_template.file.extension size=plan_template.file.size show_download=True %}
```
stefrado: When I try to download this file i see the url 'private_files/filename.ext' the file is not found on the server.
After inspecting my media folder i find that is is named 'private_media', is there something wrong with my configuration or is there something wrong?

stefrado (changes requested):
Besides the code feedback i noticed that there is a horizontal scroll on the 'samenwerken detail page'.

- It occurs from screen width 900px until ~1200px -> I saw you comment in Figma
    - This causes a strange behavior when I click in the sidebar on the first item.

src/open_inwoner/components/templates/components/Action/Actions.html
jiromaykin: Oops, I forgot to add `download_url=download_url`

src/open_inwoner/scss/components/Dashboard/Dashboard.scss
jiromaykin: oops... if all margins and paddings look correct then i can remove these, yes.

src/open_inwoner/templates/pages/plans/detail.html
jiromaykin: Oops, I forgot to add `download_url=download_url`

jiromaykin:
> Besides the code feedback i noticed that there is a horizontal scroll on the 'samenwerken detail page'.
> * It occurs from screen width 900px until ~1200px -> I saw you comment in Figma
>   * This causes a strange behavior when I click in the sidebar on the first item.

I know - @stefrado  suggestions welcome.
The scroll occurs because of the 3-columns-wide 'dashboard' component, for which we do not have an intermediate stage.
it is either 3 columns and fullwidth, or immediately drops to mobile view (rows).
Not sure what to do with this. Could perhaps use a 'container' query but that is badly supported I think.
~~I could set unique mediaquery for this page only, but it's annoying.~~
In the design there is actually no scroll-anchor to the title of the Plan, so I'll remove that in order to not have the weird side-scroll 'jump'.

This Dashboard is also used on Zaken detailpage and Contactmoment, which do not have a sidebar, but are not fullwidth either. Which makes it a design question as well.

stefrado (approved):

stefrado:
> The scroll occurs because of the 3-columns-wide 'dashboard' component, for which we do not have an intermediate stage. it is either 3 columns and fullwidth, or immediately drops to mobile view (rows).

A solution could be that for now we show a 3-columns width dashboard from 1200px screen width and all size less that 1200px we keep showing the rows. This would be a short-term solution until there is a design decision.

PR #1724, opened 2025-04-08
Update Openklant2 client to target 2.7.0

swrichards:


pi-sigma (approved):

src/open_inwoner/openklant/services.py
```diff
@@ -952,37 +953,79 @@ def _kvk_list_param(kvk: str) -> PartijListParams:
...
+            if result["subIdentificatorVan"]
+            and result["subIdentificatorVan"]["uuid"] == kvk_pi["uuid"]
+        )
+
+        if not vestiging_pi:
+            return None
+
+        if not vestiging_pi["identificeerdePartij"]:
+            logger.warning(
+                "Found a Partij Identificator vor vestiging %s without a linked Partij",
```
pi-sigma: vor > for

src/open_inwoner/openklant/services.py
```diff
@@ -952,37 +953,79 @@ def _kvk_list_param(kvk: str) -> PartijListParams:
...
+            result
+            for result in self.client.partij_identificator.list_iter(
+                params={
+                    "partijIdentificatorCodeSoortObjectId": "vestigingsnummer",
+                    "partijIdentificatorCodeRegister": "hr",
+                    "partijIdentificatorCodeObjecttype": "vestiging",
+                    "partijIdentificatorObjectId": vestigingsnummer,
+                }
+            )
+            if result["subIdentificatorVan"]
```
pi-sigma, line 1014: Is this guaranteed to exist, so no `KeyError` can result? Otherwise `result.get("subIdentificatorVan")`

src/open_inwoner/openklant/services.py
```diff
@@ -1051,29 +1084,68 @@ def get_or_create_partij_for_user(
                 )
                 created = True
 
-                for object_type, soort_object_id, object_id in (
-                    ("niet_natuurlijk_persoon", "kvk_nummer", kvk),
-                    ("vestiging", "vestigingsnummer", vestigingsnummer),
-                ):
-                    if object_id:
+                try:
+                    # Check for an existing entity-only PI (e.g. without)
```
pi-sigma: Incomplete comment?

src/open_inwoner/openklant/services.py
```diff
@@ -952,37 +953,79 @@ def _kvk_list_param(kvk: str) -> PartijListParams:
...
 
-    def get_or_create_partij_for_user(
-        self, fetch_params: FetchParameters, user: User
-    ) -> tuple[Partij | None, bool]:
+        kvk_pi: PartijIdentificator
+        match len(parent_kvk_results):
+            case 0:
+                return None
+            case x if x > 1:
+                raise RuntimeError("Unexpectedly found multiple PIs for a single kvk")
```
pi-sigma: We normally do `logger.error` and then return `None` or similar. Is `RuntimeError` the right choice (especially since the error is not handled by the caller)?

pi-sigma (changes requested):

src/open_inwoner/openklant/services.py
swrichards: The key should be mandatory (and I'd thus prefer the `KeyError` so it's caught in testing).

src/open_inwoner/openklant/services.py
swrichards: The problem with returning `None` is that if you then try to actually create the PI (because you think there's nothing there), you'll run into an API error.

I went for defaulting to the first PI found and logging an error for now.

pi-sigma (approved):

PR #1729, opened 2025-04-14
[#3092] Remove loon zorgverzekeringswet value from SSD jaaropgaaf report

swrichards:
[Taiga 3092](https://taiga.maykinmedia.nl/project/open-inwoner/us/3092?milestone=530).

New look:

Screenshot from 2025-04-14 14-42-45

### DIFF
--- a/src/open_inwoner/ssd/xml.py
+++ b/src/open_inwoner/ssd/xml.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, TypedDict, cast
 
 import requests
 from lxml import etree  # nosec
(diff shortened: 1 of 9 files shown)


src/open_inwoner/ssd/xml.py
```diff
@@ -56,7 +62,13 @@ def _get_report_info(
     return info
 
 
-def get_jaaropgaven(response: requests.Response) -> list[dict] | None:
+class JaaropgaveReturn(TypedDict):
```
pi-sigma, line 65: Might add the equivalent for the monthly uitkeringen as well. It's orthogonal to the PR though so nice to have but not strictly necessary

pi-sigma (approved):

src/open_inwoner/ssd/xml.py
swrichards: I might leave it as-is if it's all the same to you: I was winging this a bit, not quite clear what fields should be optional, etc. and wanted to limit the refactor given the story size.

PR #1731, opened 2025-04-15
Deprecate black and flake8

swrichards:


swrichards:
@pi-sigma huge PR obviously, I'd focus on the docs and just eye-ball the diffs. The only rule changed was for isort, which should be safe fixes, as should applying ruff format.

src/open_inwoner/configurations/tests/bootstrap/_test_setup_site_config.py
```diff
@@ -122,7 +122,7 @@ def test_site_configure(self):
         self.assertEqual(config.home_questionnaire_intro, "home questionnaire intro")
         self.assertEqual(config.home_product_finder_title, "home product finder title")
         self.assertEqual(config.home_product_finder_intro, "home product finder intro")
-        self.assertEqual(config.search_zero_results_text, "zero search results text"),
+        (self.assertEqual(config.search_zero_results_text, "zero search results text"),)
```
pi-sigma:
```suggestion
        self.assertEqual(config.search_zero_results_text, "zero search results text")
```
?

pi-sigma (changes requested):

src/open_inwoner/configurations/tests/bootstrap/_test_setup_site_config.py
swrichards: Argh missed this! I tried to fix these earlier, that was one of the nice upshots of doing this, we had a whole bunch of inadvertent tuple expressions due to trailing commas. Will fix this.

pi-sigma (approved):

PR #1735, opened 2025-04-16
[#3166] Use `case.description` for title in userfeed

pi-sigma:
The userfeed used `case.omschrijving` for the case title when updating the case status, ignoring the configurable option in the Zaak configuration. Changed to use `case.description` which looks up the field in the configuration.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3166

### DIFF
--- a/src/open_inwoner/userfeed/tests/hooks/test_case_status.py
+++ b/src/open_inwoner/userfeed/tests/hooks/test_case_status.py
@@ -8,7 +8,7 @@ from django.utils.translation import gettext as _, ngettext
 from zgw_consumers.api_models.base import factory
 
 from open_inwoner.openzaak.api_models import Status, StatusType, Zaak, ZaakType
-from open_inwoner.openzaak.constants import StatusIndicators
+from open_inwoner.openzaak.constants import StatusIndicators, ZaakTitleDisplayChoices
 from open_inwoner.openzaak.models import ZGWApiGroupConfig
 from open_inwoner.openzaak.tests.factories import (
     ZaakTypeConfigFactory,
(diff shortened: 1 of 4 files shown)


src/open_inwoner/userfeed/tests/hooks/test_case_status.py
swrichards, line 1: For one of these tests, it wouldn't be a bad idea to use a `subTest` and change the the configured field backing the description, to ensure it's picked up and not an accidental match.

src/open_inwoner/userfeed/tests/hooks/test_case_status.py
```diff
@@ -141,37 +141,51 @@ def test_status_update_with_status_type_config(
...
+        scenarios = [
+            (ZaakTitleDisplayChoices.zaak_omschrijving, case.omschrijving),
+            (ZaakTitleDisplayChoices.zaaktype_omschrijving, case.zaaktype.omschrijving),
+            (ZaakTitleDisplayChoices.zaaktype_onderwerp, case.zaaktype.onderwerp),
+        ]
+        for zaak_title_config, expected_zaak_title in scenarios:
+            with self.subTest(f"zaak_title_config: {zaak_title_config}"):
+                self.api_group.open_zaak_config.derive_zaak_titel_from = (
+                    zaak_title_config
+                )
```
swrichards: No `.save()` needed?
```suggestion
                self.api_group.open_zaak_config.derive_zaak_titel_from = (
                    zaak_title_config
                )
                self.api_group.open_zaak_config.save()
```

pi-sigma: Yep, already fixed

pi-sigma: It was indeed taking the title from `zaaktype.omschrijving` for all cases, so tests passed accidentally. Subtests were a good idea therefore.

swrichards (approved):

PR #1736, opened 2025-04-17
[#3161] Create/edit action redesign

stefrado:
----
📚 Documentation preview 📚: https://open-inwoner--1736.org.readthedocs.build/en/1736/


pi-sigma:
@stefrado Please squash the commits so that we can keep a meaningful commit history:
- There should be no independent commit like "run black formatter" (code pushed to GH should already be formatted with black. You can do this manually or use a git hook (see e.g. [here](https://black.readthedocs.io/en/stable/integrations/source_version_control.html) for instructions)
- The code in the "Fix CI" commits should either be part of the original commit that implements the feature, or at least squash the commits and give a more descriptive name to the resulting commit. "Fix CI" is too generic, it's not clear what it does.

As far as I can tell, there should be only one commit for this PR, possibly two: (1) implement feature for design, (2: possibly) Fix misaligned icon. Though the second could perhaps also be part of the first. Perhaps @swrichards can weigh in on this; we should all follow the same conventions here.

swrichards:
> As far as I can tell, there should be only one commit for this PR, possibly two: (1) implement feature for design, (2: possibly) Fix misaligned icon. Though the second could perhaps also be part of the first. Perhaps @swrichards can weigh in on this; we should all follow the same conventions here.

I agree with @pi-sigma on the general points -- I'd like to table this for the next weekly on Monday, but I generally I feel commit hygiene is an area where we can make some gains (which are several, but not the least is that it helps facilitating the release process to have a more readable and organized git history).

On the specific point: I think we can be pragmatic here and reduce to a single commit. The ideal of course is one commit, one logical unit of change, but we should discuss the meaning of that principle first. So for now I would propose squashing this into a single commit and including a meaningful commit message with ticket ref.

src/open_inwoner/accounts/migrations/0085_alter_action_status.py
swrichards, line 1: Following the commit discussion: would it be possible to squash the migrations into a single migration?

stefrado:
@pi-sigma @swrichards y'all are correct about the commits. I planned to squash all the commits into one. TBH i forgot to do it due to the CI that kept failing.

src/open_inwoner/accounts/forms.py
```diff
@@ -497,13 +497,15 @@ def has_changed(self, initial, data):
 
 class ActionForm(forms.ModelForm):
     is_for = UserField(
-        label=_("Is voor"),
+        label=_("Action for"),
```
jiromaykin, line 528: Why is this in English if the other model changes are in Dutch?

src/open_inwoner/accounts/forms.py
```diff
@@ -497,13 +497,15 @@ def has_changed(self, initial, data):
...
         queryset=User.objects.all(),
         empty_label=_("Myself"),
```
jiromaykin, line 530: English?

src/open_inwoner/accounts/forms.py
```diff
@@ -497,13 +497,15 @@ def has_changed(self, initial, data):
...
         required=False,
     )
     file = LimitedUploadFileField(
-        required=False, widget=PrivateFileWidget(url_name="profile:action_download")
+        label=_("Add a file"),
```
jiromaykin, line 534: Maybe this one did not need to be changed yet, due to future documentupload challenge. But I might have done this in Dutch.

src/open_inwoner/js/components/form/submitOnce.js
```diff
@@ -0,0 +1,57 @@
+const SUBMIT_ONCE_BUTTONS = document.querySelectorAll('.submit-once-button')
+
+/**
+ * SubmitOnceButton Class
+ * This class handles the functionality for buttons that should only be submitted once.
```
jiromaykin: Not for 'buttons' but for 'forms' that should only be submitted once :-)
But: was this actually really happening? Do you mean this situation occurs after an error situation?

src/open_inwoner/scss/components/Form/Form.scss
```diff
@@ -273,6 +273,12 @@
       transform: translateY(-50%);
     }
 
+    // Icon alignment fix if there is no help_text.
+    &--no-help *[class*='icons'],
+    &--no-help *[class*='Icons'] {
+      transform: translateY(15%) !important;
+    }
```
jiromaykin, line 280: Nice, seems to solve it for almost all other icons as well, except for the Select here: http://localhost:8000/mijn-aanvragen/contactmomenten/ but let's make that a separate issue.

src/open_inwoner/accounts/migrations/0085_alter_action_status.py
jiromaykin: That would be nice, more importantly the squashing of commits is high priority.
Squashing of commits is one of the first articles to read when onboarding, so it's a maykin-wide practice: https://www.maykinmedia.nl/blog/2017/nov/20/git-rebasing-make-your-commit-history-pretty/

src/open_inwoner/templates/pages/profile/actions/edit.html
```diff
@@ -1,31 +1,50 @@
...
+                    </div>
+                    
+                    <div class="form__actions form__actions--fullwidth">
+                        {% if action %}
+                            {% button text=_("Actie opslaan") primary=True type="submit" extra_classes='submit-once-button' %}
+                        {% else %}
+                            {% button text=_("Maak actie aan") primary=True type="submit" extra_classes='submit-once-button' %}
+                        {% endif %}
+                        {% if plan %}
+                            {% button href=plan.get_absolute_url icon="west" text=_("Terug naar ")|add:plan.title icon_outlined=True transparent=True %}
```
jiromaykin, line 40: Excellent!

jiromaykin (changes requested):
Perhaps ignore my comments about Dutch/English, but please do not put up PR's up for review if CI is still failing.
And squash commits.
Also: you will need to continuously rebase your PR to Develop branch, due to incoming changes, so the last person to merge is responsible for solving merge conflicts.

src/open_inwoner/accounts/forms.py
stefrado: Everything that was already there was written in english. Now it is more consistent.

src/open_inwoner/accounts/forms.py
stefrado: Same comment as above.

src/open_inwoner/accounts/forms.py
stefrado: I changed this because the label of the file input was 'File'. Now it is configured to 'Add a file', which is in line with the design.

src/open_inwoner/scss/components/Form/Form.scss
stefrado: This select can't be fixed with this solution, due to an extra margin below the field. 
I agree we should make a separate issue that tackles the icon alignment.

src/open_inwoner/js/components/form/submitOnce.js
stefrado: Yes when i clicked the create action form submit button multiple times the same action was created more than once.

src/open_inwoner/accounts/migrations/0085_alter_action_end_date_alter_action_file_and_more.py
```diff
@@ -0,0 +1,76 @@
+# Generated by Django 4.2.18 on 2025-04-24 16:17
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+import privates.storages
+
+
+class Migration(migrations.Migration):
```
jiromaykin, line 9: Looking good.

jiromaykin (approved):
In Figma we still have some 'gaps' of 80px (like the distance between the H1 and the content), but as far as I remember these all needed to be changed to 40px, so this PR is correct for now.

jiromaykin (approved):

PR #1737, opened 2025-04-22
[#3122] Amend plans creation flows to match new designs

swrichards:
[Taiga 3122](https://taiga.maykinmedia.nl/project/open-inwoner/us/3122?milestone=530).

For reference, [the Figma designs illustrating the flows](https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=3108-14182&p=f).


src/open_inwoner/accounts/views/contacts.py
```diff
@@ -74,18 +76,25 @@ def crumbs(self):
...
+    def form_valid(self, form: ContactCreateForm):
+        user = cast(User, self.request.user)
         email = form.cleaned_data["email"]
 
-        added_contacts = form.cleaned_data.get("added_contacts")
-        if added_contacts:
+        # If the provided email matched any existing users, send them an approval
+        # request.
+        if added_contacts := form.cleaned_data.get("added_contacts"):
             for contact_user in added_contacts:
```
pi-sigma, line 87: You could just do: `for contact_user in form.cleaned_data.get("added_contacts")`. The value of `added_contacts` is guaranteed to be an iterable, not `None`, so you can loop directly.

src/open_inwoner/accounts/forms.py
```diff
@@ -444,39 +458,53 @@ def clean(self):
...
+                already_added_users.add(contact_user)
             else:
+                # User is a valid candidate to be a contact
                 added_contacts.add(contact_user)
 
-        # remember the contacts and let the view add records, logs and the emails
+        if inactive_users:
+            logger.warning(
+                "User attempted to add inactive users as contacts",
+                extra={"users": inactive_users},
```
pi-sigma:
```suggestion
                extra={"inactive_users": inactive_users},
```

src/open_inwoner/accounts/forms.py
```diff
@@ -444,39 +458,53 @@ def clean(self):
...
+            )
+
+        if already_added_users:
+            logger.warning(
+                "User attempted to add contacts that are already added",
+                extra={"users": already_added_users},
```
pi-sigma:
```suggestion
                extra={"already_added_users": already_added_users},
```

src/open_inwoner/plans/forms.py
```diff
@@ -13,6 +14,104 @@
...
+        data = self.cleaned_data["plan_contacts"]
+        if self.instance.pk:
+            data |= User.objects.filter(pk=self.user.pk)
+        return data.distinct()
+
+    def clean(self) -> dict[str, Any]:
+        cleaned_data = super().clean()
+        end_date = cleaned_data.get("end_date")
+        if end_date:
+            actionTemplates = self.template.actiontemplates.all()
```
pi-sigma:
```suggestion
            action_templates = self.template.actiontemplates.all()
```

src/open_inwoner/plans/forms.py
```diff
@@ -13,6 +14,104 @@
...
+
+            if actionTemplates:
+                latest_end_in_days = max([a.end_in_days for a in actionTemplates])
```
pi-sigma:
```suggestion
                latest_end_in_days = max([a.end_in_days for a in action_templates])
```

src/open_inwoner/plans/forms.py
```diff
@@ -13,6 +14,104 @@
...
+        # Make sure current user exists in plan_contacts when editing form
+        data = self.cleaned_data["plan_contacts"]
+        if self.instance.pk:
+            data |= User.objects.filter(pk=self.user.pk)
+        return data.distinct()
+
+    def clean(self) -> dict[str, Any]:
+        cleaned_data = super().clean()
+        end_date = cleaned_data.get("end_date")
+        if end_date:
```
pi-sigma: You could do:

```
        if (
            (end_date := cleaned_data.get("end_date"))
            and (action_templates := self.template.action_template.all())
        ):
            latest_end_in_days = max([a.end_in_days for a in actionTemplates])
```
to reduce nesting, though I'm  not sure if this is cleaner in this case

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -477,21 +529,27 @@ def test_plan_create_no_template_plan_validation_error_reselects_template_and_co
...
 
         # nothing was created
         self.assertEqual(Plan.objects.count(), 1)
 
-        # check if we reselected the template and contact
-        elem = response.pyquery(f"#id_template_{plan_template.id}")[0]
-        self.assertEqual(elem.attrib.get("checked"), "checked")
+        # # check if we reselected the template and contact
+        # elem = response.pyquery(f"#id_template_{plan_template.id}")[0]
+        # self.assertEqual(elem.attrib.get("checked"), "checked")
```
pi-sigma: Do we still need these checks? Uncomment or delete.

src/open_inwoner/plans/views.py
```diff
@@ -278,7 +286,7 @@ class PlanCreateView(
     BaseBreadcrumbMixin,
     CreateView,
 ):
-    template_name = "pages/plans/create.html"
+    template_name = "pages/plans/create-no-template.html"
```
pi-sigma: We don't have the `-no-template` for the view or url names, why here? Generally i think it would be nice to have the naming of the templates go along with the naming of the urls, so `collaborate:plan_create` <-> `pages/plans/plan-create.html`. Probably outside the scope of this PR though.

pi-sigma (changes requested):

src/open_inwoner/plans/forms.py
swrichards: I think the problem with the underscore is that it conflicts with the naming of the relation: I'll lower-case it but stay consistent.

src/open_inwoner/plans/views.py
swrichards: I'll remove the `no-template`, but the more consistent naming you suggest indeed seems like a bigger undertaking separate from this PR.

pi-sigma (approved):

PR #1742, opened 2025-04-23
docs: update the admin link shortcut to point to correct docs

swrichards:
Note we'll have to update the ref when we do the release commit.

### DIFF
--- a/src/open_inwoner/conf/fixtures/django-admin-index.json
+++ b/src/open_inwoner/conf/fixtures/django-admin-index.json
@@ -470,7 +470,7 @@
             "overige-diverse"
         ],
         "name": "Link naar Documentatie",
-        "link": "https://taiga.maykinmedia.nl/project/open-inwoner/wiki/home"
+        "link": "https://docs.openinwoner.nl/en/v1.29.0/"
     }
 },
 {


src/open_inwoner/conf/fixtures/django-admin-index.json
```diff
@@ -470,7 +470,7 @@
             "overige-diverse"
         ],
         "name": "Link naar Documentatie",
-        "link": "https://taiga.maykinmedia.nl/project/open-inwoner/wiki/home"
+        "link": "https://docs.openinwoner.nl/en/v1.29.0/"
```
pi-sigma, line 473: Why not:
```suggestion
        "link": "https://docs.openinwoner.nl/en/latest/"
```
Then we don't have to update the version for every release.

pi-sigma (changes requested):

src/open_inwoner/conf/fixtures/django-admin-index.json
swrichards: Because if users click the link, we want them to arrive in the correct docs for their currently active version. The RTD interface is subtle and it might not be immediately obvious to the median user that the docs are versioned.

pi-sigma (approved):

PR #1743, opened 2025-04-23
feat: make SECRET_KEY_FALLBACKS environment configurable

swrichards:
Apart from allowing key rotation as a general best practice,
making the SECRET_KEY_FALLBACKS configurable can facilitate
instance migrations which keep existing sessions intact (using
the fallback) whilst progressively phasing these ciphertexts
out on the basis of the new primary SECRET_KEY.

We only allow a single value to avoid commas intended to be
part of a high-entropy SECRET_KEY being interpreted as a
list (and thereby giving several low-entropy secret keys).

This is in anticipation of some planned OIP migrations, to
give us the option to migrate crypto-verified state if we
need to.

### DIFF
--- a/src/open_inwoner/conf/base.py
+++ b/src/open_inwoner/conf/base.py
@@ -29,6 +29,11 @@ DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = config("SECRET_KEY")
 
+# To facilitate key rotation and migrations. We currently only allow a single value
+# (hence the singular `SECRET_KEY_FALLBACK` config attribute).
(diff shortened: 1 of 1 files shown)


pi-sigma (approved):
Are we going to support multiple fallbacks at some point? Otherwise we might as well just use the single fallback instead of a list containing one...

swrichards:
> Are we going to support multiple fallbacks at some point? Otherwise we might as well just use the single fallback instead of a list containing one...

Django [expects the setting to be a list](https://docs.djangoproject.com/en/4.2/ref/settings/#secret-key-fallbacks). I just allowed the one to avoid ambiguity of comma-separation through environment variables, as this value is quite likely to have commas in it as part of the key, rather than a separator.

PR #1744, opened 2025-04-24
fix: use correct navigation urls in plan creation views

swrichards:
[Taiga 3122](https://taiga.maykinmedia.nl/project/open-inwoner/us/3122?milestone=530)

### DIFF
--- a/src/open_inwoner/templates/pages/plans/create-with-template.html
+++ b/src/open_inwoner/templates/pages/plans/create-with-template.html
@@ -7,7 +7,7 @@
             <h1>{{ plan_template.name }}</h1>
             <div class="plan-template__header-nav">
                 {% if request.user.get_active_contacts %}
-                    {% button href="collaborate:plan_create_no_template" text=_("Zelf in te vullen samenwerking") primary=True transparent=True extra_classes="link--bold" icon="insert_drive_file" icon_outlined=True %}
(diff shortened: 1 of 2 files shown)


src/open_inwoner/templates/pages/plans/create-with-template.html
```diff
@@ -7,7 +7,7 @@
             <h1>{{ plan_template.name }}</h1>
             <div class="plan-template__header-nav">
                 {% if request.user.get_active_contacts %}
-                    {% button href="collaborate:plan_create_no_template" text=_("Zelf in te vullen samenwerking") primary=True transparent=True extra_classes="link--bold" icon="insert_drive_file" icon_outlined=True %}
+                    {% button href="collaborate:plan_create" text=_("Zelf in te vullen samenwerking") primary=True transparent=True extra_classes="link--bold" icon="insert_drive_file" icon_outlined=True %}
```
jiromaykin, line 10: This works but other link is still broken (so go to `src/open_inwoner/templates/pages/plans/create.html `)
There is one more TODO there that needs to be removed. 
The old link used to be a scroll link, but should now be replaced with the URL pointing to the default `http://localhost:8000/samenwerken/choose-template/` which will give that option back again to the user.

jiromaykin (changes requested):

jiromaykin (approved):
Ready for merge :-)

PR #1747, opened 2025-04-28
fix: improve error message for invalid plan end date

pi-sigma:
The end date for a plan cannot precede the earliest end date of the actions in the template. The current error message is not informative/actionable, since it doesn't tell the user when the earliest end date for the plan is.

### DIFF
--- a/src/open_inwoner/plans/forms.py
+++ b/src/open_inwoner/plans/forms.py
@@ -5,7 +5,7 @@ from typing import Any, cast
 from django import forms
 from django.core.exceptions import ValidationError
 from django.core.files.base import File
-from django.utils import timezone
+from django.utils import formats, timezone
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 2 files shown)


src/open_inwoner/plans/forms.py
```diff
@@ -66,8 +66,9 @@ def clean(self) -> dict[str, Any]:
                     "end_date",
                     _(
                         "The end date of the plan cannot precede the end dates of the "
-                        "actions in the selected template."
-                    ),
+                        "actions in the selected template. Earliest possible end date is %s"
```
swrichards:
```suggestion
                        "actions in the selected template. The earliest possible end date for this type of plan is %s."
```

swrichards: (The trailing period was for consistency, and thought it couldn't hurt to be a bit more prosaic)

swrichards (approved):

PR #1748, opened 2025-04-28
:bug: [#3223] Faulty alignment for input-icons with error-notifications

jiromaykin:
Issue screenshots: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3223


pi-sigma (approved):

stefrado (approved):

stefrado:
For now this works fine, but i think for the long term we need to refactor the inputs so that the icon is positioned relative to the `input` instead of the `.form__control`

PR #1749, opened 2025-04-28
[#3218] Use correct anchors and titles in profile navigation menu

pi-sigma:
The PR fixes two issues with the navigation menu in the profile:

- the navigation anchor for eherkenning users uses the wrong tile (is: 'Persoonlijke gegevens', should be: 'Bedrijfsgegevens')
- the navigation anchor 'Overzicht' should only be displayed if there are selected categories to be displayed (configurable via the CMS profile app)

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3218


jiromaykin (approved):
Works well.

PR #1750, opened 2025-04-28
fix: validate collaborators are selected for plan create from template

swrichards:


src/open_inwoner/plans/forms.py
```diff
@@ -72,6 +72,13 @@ def clean(self) -> dict[str, Any]:
                     % formats.date_format(actions_end_date, use_l10n=True),
                 )
 
+        if not (plan_contacts := cleaned_data["plan_contacts"]) or (
+            plan_contacts and not plan_contacts.exclude(pk=self.user.pk)
+        ):
```
pi-sigma:
```
if not (plan_contacts := cleaned_data["plan_contacts"]) or not plan_contacts.exclude(pk=self.user.pk):
```
Checking for `plan_contacts` in the second disjunct should not be necessary if the first disjunct evaluates to `False` (because this means that `plan_contacts` is truthy)

pi-sigma (approved):

src/open_inwoner/plans/forms.py
swrichards: Check, changed it below as well.

PR #1751, opened 2025-04-28
[#3225] Add HTML autocomplete attribute to profile edit inputs

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3225

### DIFF
--- a/src/open_inwoner/components/templatetags/form_tags.py
+++ b/src/open_inwoner/components/templatetags/form_tags.py
@@ -18,6 +18,17 @@ WIDGET_TEMPLATES = {
     "TEXTAREA": "components/Form/Textarea.html",
 }
 
+HTML_NAME_TO_AUTOCOMPLETE = {
+    "first_name": "given-name",
+    "last_name": "family-name",
+    "email": "email",
+    "phonenumber": "tel-national",
+    "phonenumber_alternative": "tel-national",
+    "street": "address-line1",
+    "city": "address-level2",
+    "postcode": "postal-code",
+}
+
 
 def get_form_classes(**kwargs):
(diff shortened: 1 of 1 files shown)


src/open_inwoner/components/templatetags/form_tags.py
```diff
@@ -18,6 +18,17 @@
     "TEXTAREA": "components/Form/Textarea.html",
 }
 
+HTML_NAME_TO_AUTOCOMPLETE = {
+    "first_name": "given-name",
+    "last_name": "family-name",
+    "email": "email",
+    "phonenumber": "tel-national",
+    "phonenumber_alternative": "tel-national",
+    "street": "street-address",
```
jiromaykin: The value for street is actually not correct. There is no proper dedicated street-name attribute.
`street-address` only exists as an unstructured combination of **both** streetname **and** housenumber (and more).
So this means you'll have to get 'lucky' and hope that other developers made the same mistake, so that users who have visited such sites, have already used this before.

This has been an ongoing problem for many years now (with many differences between countries): "[autocomplete attribute for street-address W3C issue 2019](https://github.com/w3c/i18n-activity/issues/779)"

This post suggests the attribute should be `address-line1` instead.
[https://www.200ok.nl/tips/autocomplete/](https://www.200ok.nl/tips/autocomplete/)

There you can also see `address-level2` is indeed correct for 'plaats'.

jiromaykin (changes requested):

jiromaykin (approved):
This will likely come up again for other forms, but the issue was only about this individual page. 
We need to keep this in mind for new forms somehow...

PR #1754, opened 2025-04-29
[#3231] Add sitemap to footer

pi-sigma:
To improve navigability of the site, a sitemap with internal links to select pages on OIP is added to the footer. To start with, the sitemap contains links to the categories (onderwerpen).

Taiga: 
- https://taiga.maykinmedia.nl/project/open-inwoner/us/2398 
- https://taiga.maykinmedia.nl/project/open-inwoner/task/3231

### DIFF
--- a/src/open_inwoner/scss/components/Sitemap/sitemap.scss
+++ b/src/open_inwoner/scss/components/Sitemap/sitemap.scss
@@ -1,0 +1,5 @@
+.sitemap {
+  &__section {
+    padding-top: var(--spacing-small);
+  }
+}
--- a/src/open_inwoner/templates/pages/sitemap/sitemap.html
+++ b/src/open_inwoner/templates/pages/sitemap/sitemap.html
@@ -1,0 +1,60 @@
+{% extends "master.html" %}
+{% load i18n list_tags %}
+
+{% block content %}
+    <h1 class="utrecht-heading-1">{% trans "Links naar pagina's op het Open Inwoner Platform" %}</h1>
+
+    <div class="grid sitemap">
+        {# pages with information about OIP #}
+        <section class="sitemap__section">
+            <span class="utrecht-heading-2">{% trans "Platform" %}</span>
+            <ul class="sitemap__list">
+                {% for page in platform_pages %}
+                    {% with link=page.url_name text=page.link_text %}
+                        <li class="sitemap__list-item"><a href="{% url link %}">{% trans text %}</a></li>
+                    {% endwith %}
+                {% endfor %}
+                {% for page in flatpages %}
+                    {% with text=page.link_text %}
+                        <li class="sitemap__list-item"><a href={{ page.url }}>{% trans text %}</a></li>
+                    {% endwith %}
+                {% endfor %}
+            </ul>
+        </section>
+
+        {# onderwerpen #}
+        <section class="sitemap__section">
+            <a class="utrecht-heading-2" href="{% url 'products:category_list' %}">{% trans "Onderwerpen" %}</a>
+            <ul class="sitemap__list">
+                {% for node in category_nodes %}
+                    {% include "pages/sitemap/category_item.html" with node=node %}
+                {% endfor %}
+            </ul>
+        </section>
+
+        {% if request.user.is_authenticated %}
+            <section class="sitemap__section">
+                {# profile pages heading #}
+                {% for page in cms_profile_pages|slice:":1" %}
+                    {% with link=page.url_name text=page.link_text %}
+                        <a class="utrecht-heading-2" href="{% url link %}">{% trans text %}</a>
+                    {% endwith %}
+                {% endfor %}
+                <ul class="sitemap__list">
+                    {# profile pages #}
+                    {% for page in cms_profile_pages|slice:"1:" %}
+                        {% with link=page.url_name text=page.link_text %}
+                            <li class="sitemap__list-item"><a href="{% url link %}">{% trans text %}</a></li>
+                        {% endwith %}
+                    {% endfor %}
+                    {# other CMS pages #}
+                    {% for page in cms_pages %}
+                        {% with link=page.url_name text=page.link_text %}
+                            <li class="sitemap__list-item"><a href="{% url link %}">{% trans text %}</a></li>
+                        {% endwith %}
+                    {% endfor %}
+                </ul>
+            </section>
+        {% endif %}
+    </div>
+{% endblock content %}
--- a/src/open_inwoner/accounts/templates/registration/login.html
+++ b/src/open_inwoner/accounts/templates/registration/login.html
@@ -6,6 +6,7 @@
     {% static_placeholder 'login_banner' %}
 {% endblock header_image %}
 
+{# {% block title %}{% with site_name=site_name page_title=_("Log in") %}{{ block.super }}{% endwith %}{% endblock %} #}
 
 {% block content %}
     <h1 class="utrecht-heading-1">{% trans 'Welkom' %}</h1>
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -176,7 +176,7 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
         return [(_("Mijn vragen"), reverse("cases:contactmoment_list"))]
 
     def page_title(self):
-        return _("Mijn vraag")
+        return f"{_('Mijn vraag')} {self.question['identification']}"
 
     def get_anchors(self) -> list:
         return [
@@ -204,7 +204,7 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
         )
         if not question:
             raise Http404()
-
+        self.question = question
         QuestionValidator.validate_python(question)
 
         local_kcm, created = KlantContactMomentAnswer.objects.get_or_create(  # noqa
--- a/src/open_inwoner/accounts/views/login.py
+++ b/src/open_inwoner/accounts/views/login.py
@@ -90,6 +90,15 @@ class CustomLoginView(LogMixin, LoginView):
 
         return redirect(furl(reverse("verify_token")).add(params).url)
 
+    # Manually overwrite the context site_name and page_title for the LoginView.
+    # A wrong default is set in django.contrib.auth.LoginView.
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        config = SiteConfiguration.get_solo()
+        context["site_name"] = config.name
+        context["page_title"] = _("Log in")
+        return context
+
 
 class VerifyTokenView(ThrottleMixin, FormView):
     throttle_visits = 3
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -87,19 +87,39 @@ class VragenService(Protocol):
 
 
 class OuterCaseDetailView(
-    OuterCaseAccessMixin, CommonPageMixin, BaseBreadcrumbMixin, TemplateView
+    OuterCaseAccessMixin,
+    CommonPageMixin,
+    BaseBreadcrumbMixin,
+    CaseAccessMixin,
+    TemplateView,
 ):
     template_name = "pages/cases/status_outer.html"
+    case: Zaak | None = None
 
     @cached_property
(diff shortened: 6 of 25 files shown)


jiromaykin (changes requested):
Spelling mistake in `pagina's`,
but more importantly:

This is not what a sitemap is for: linked pages should not be limited to this small a selection.
A sitemap is a file that lists ALL OF the 'reachable' pages on a website, to make it easier for users to navigate the site, and to help search engines find and index the content.
Look at this example: https://www.rijksoverheid.nl/sitemap 

So it's quite a bit more work. You will have to think about perhaps making content linked correctly for users that are logged in as well (which differs for different municipalities).
But it would be better to decide on what to include/exclude here with @swrichards .

src/open_inwoner/templates/pages/sitemap.html
swrichards, line 1: Is one of those redundant?

src/open_inwoner/templates/pages/sitemap.html
```diff
@@ -0,0 +1,38 @@
...
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn contacten" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn acties" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn uitkeringen" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn aanvragen" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn vragen" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn afspraken" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn vragen" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Samenwerken" %}</a></li>
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Zelftest" %}</a></li>
```
swrichards: Incorrect links?

src/open_inwoner/templates/pages/sitemap.html
```diff
@@ -0,0 +1,38 @@
...
+                    <li class="list-item"><a class="link" href={{ node.category.get_absolute_url }}>{% trans node.category.name %}</a></li>
+                {% endfor %}
+            </ul>
+        </section>
+
+        {% if request.user.is_authenticated %}
+            <section>
+                <a class="link" href="{% url 'profile:detail' %}"><h2 class="utrecht-heading-2">{% trans "Mijn Profiel" %}</h2></a>
+                <ul class="list">
+                    <li class="list-item"><a class="link" href="{% url 'profile:categories' %}">{% trans "Mijn Interessegebieden" %}</a></li>
```
swrichards: I think the presence of some of these links is also conditional on other factors (e.g. whether a ZGW api is enabled). Is there a way we can re-use the CMS menu logic here?

swrichards (changes requested):
There might be some low-hanging fruit here regarding the visuals, would be good for @jiromaykin to weigh in there. Otherwise I think the main concern is how far we want to go with making the authenticated links fully match the logic we use elsewhere to determine what links to show (e.g. on the profile page). My sense is we shouldn't spend too much time on that given the ticket size, but perhaps there is some re-usable logic there.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,37 @@
+{% extends "master.html" %}
+{% load i18n list_tags %}
+
```
jiromaykin, line 3: For now this looks pretty good for styling.
I wonder if there is an easy way to "crawl" the OIP site so you can generate your own sitemap locally easily and then add the missing pages to this template.
We still need links to 'home' and the login-page + registration page and maybe I am forgetting some that are publicly visible, like the flat pages. So yeah: even if they already are in the footer, they need to be in the Sitemap as well.

src/open_inwoner/scss/components/Sitemap/sitemap.scss
```diff
@@ -0,0 +1,17 @@
+section {
```
jiromaykin: These section tag styles are messing up the login-page Tabs and all of the other sections in the site... 
At least just add a `.sitemap` class to the whole page and add `.sitemap__section` class to each section, else: why add a file called 'sitemap' if we don't have a sitemap class?

It's okay if it's all vertically placed for now.

Also these global tag-styles for the lists are messing up the list-styles inside notifications etc.

Please don't ever do this - no tag styles ever. Use specific classes or just add the classes we already have, so you don't have to style anything.

Tag-styling: no, unless you're targeting unknown elements inside a rich text-editor field (like we do in `src/open_inwoner/scss/components/Typography/wysiwyg.scss`)

So just surround the whole section with `.section` and then add the components with the usual .list and .list-item classes,
so we can add better styling later that shows the discs for the items, etc. Right now all we need is just the correct List elements and List-items.

src/open_inwoner/scss/components/Sitemap/sitemap.scss
```diff
@@ -0,0 +1,17 @@
...
+  padding-top: 1em;
+}
+
+ul {
+  list-style-type: disc;
```
jiromaykin: This also messes up some entries for the FAQ's 😄

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+{% block content %}
+    <h1 class="utrecht-heading-1">{% trans "Links naar pagina's op het Open Inwoner Platform" %}</h1>
+
+    <div class="grid">
```
jiromaykin: Add sitemap class here as well.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+        <section>
```
jiromaykin: this section would like a `class="sitemap__section"` class.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+            <a class="link" href="{% url 'products:category_list' %}"><h2>{% trans "Onderwerpen" %}</h2></a>
```
jiromaykin: Many of the </h2> closing tags match anything = they are missing their starting tag. You can use H2 `utrecht-heading-2` classes for the starting tags.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+                {% endfor %}
+            </ul>
+        </section>
+
+        {% if request.user.is_authenticated %}
+            <section>
+                {# profile pages heading #}
+                {% for page in cms_profile_pages|slice:":1" %}
+                    {% with link=page.url_name text=page.link_text %}
+                        <a class="link" href="{% url link %}"><h2>{% trans text %}</h2></a>
```
jiromaykin: Could have nicer `<h2 class="utrecht-heading-2">` class.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+                    {% endwith %}
+                {% endfor %}
+                <ul>
+                    {# profile pages #}
+                    {% for page in cms_profile_pages|slice:"1:" %}
+                        {% with link=page.url_name text=page.link_text %}
+                            <li><a class="link" href="{% url link %}">{% trans text %}</h2></a></li>
```
jiromaykin: The links look nicer with a "link--secondary" class but that could be a future improvement, as a design question.

jiromaykin: Also this link has a closing H2 tag, but it's missing a starting tag; perhaps no H2 needed at all here?

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+            </section>
+        {% endif %}
+
+        {# pages with information about OIP #}
+        <section>
+            <h2>{% trans "Platform" %}</h2>
+            <ul>
+                {% for page in platform_pages %}
+                    {% with link=page.url_name text=page.link_text %}
+                        <li><a class="link" href="{% url link %}">{% trans text %}</h2></a></li>
```
jiromaykin: An orphan (?) H2 tag.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,59 @@
...
+                    {% endwith %}
+                {% endfor %}
+                {% for page in flatpages %}
+                    {% with text=page.link_text %}
+                        <li><a class="link" href={{ page.url }}>{% trans text %}</h2></a></li>
```
jiromaykin: Another orphan (?) H2 tag.

jiromaykin (changes requested):
A good start, but needs some style love later 
(and multiple template improvements now regarding HTML tags, making sure it's not interfering with the rest of the site).

swrichards (approved):
Listing looks correct to me, I'll defer to @jiromaykin on the styling issues.

src/open_inwoner/scss/components/Sitemap/sitemap.scss
```diff
@@ -0,0 +1,5 @@
+.sitemap {
+  section {
```
jiromaykin: You actually made a correct `sitemap__section` this time, so why not target that...? (using `&__section`) No styles on tags please, but i can create a PR later and clean up these things myself if you promise to never do it again :-)

src/open_inwoner/scss/components/Sitemap/sitemap.scss
```diff
@@ -0,0 +1,5 @@
...
+    padding-top: 1em;
```
jiromaykin: And yeah, we should not use static margins, but the variables instead, which you need to know, but I can clean it up.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,60 @@
...
+
+        {% if request.user.is_authenticated %}
+            <section class="sitemap__section">
+                {# profile pages heading #}
+                {% for page in cms_profile_pages|slice:":1" %}
+                    {% with link=page.url_name text=page.link_text %}
+                        <a class="utrecht-heading-2" href="{% url link %}">{% trans text %}</a>
+                    {% endwith %}
+                {% endfor %}
+                <ul class="sitemap__list">
```
jiromaykin, line 43: Since we need to start using NL-Design_System components, I don't think we should create new ones. But just keep these for now.

I'm thinking of directly using the nice LinkList https://nl-design-system.github.io/utrecht/storybook/?path=/docs/css_css-link-list--docs

Of for simplicity just these: https://nl-design-system.github.io/utrecht/storybook/?path=/story/css_css-unordered-list--nested

So that's what I'll be doing later.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -0,0 +1,60 @@
...
+                            <li class="sitemap__list-item"><a href="{% url link %}">{% trans text %}</a></li>
+                        {% endwith %}
+                    {% endfor %}
+                </ul>
+            </section>
+        {% endif %}
+
+        {# pages with information about OIP #}
+        <section class="sitemap__section">
+            <span class="utrecht-heading-2">{% trans "Platform" %}</span>
```
jiromaykin: I would actually move the Platform all the way up since that's most common?
That's the only thing that I'd like to put up as a "**request changes**" for this review - all the rest I can pick up myself after approve and merge.

jiromaykin (changes requested):
Not sure about the order of the sections, all else I can pick up later for styling in a separate PR.

jiromaykin (approved):

PR #1757, opened 2025-05-01
More resilient thread management in CaseListService

swrichards:
[Taiga 2999](https://taiga.maykinmedia.nl/project/open-inwoner/us/2999?milestone=532)

The main changes here are to:

- De-nest threaded calls, to make it easier to apply timeouts and worker limits
- Ensure we have timeouts for all concurrent operations
- Better exception handling to ensure we show what we can, even with failures
- Make sure that asynchronous code has as few side-effects as possible (no state sharing, minimal DB requests)

Not done in this PR, but which will follow separately: keeping track of when we timeout/have failures resolving, and reporting this to the user in a "You may not see everything, try again later" error.

### DIFF
--- a/src/open_inwoner/cms/cases/views/services.py
+++ b/src/open_inwoner/cms/cases/views/services.py
@@ -11,7 +11,6 @@ from django.utils.translation import gettext_lazy as _
 from zgw_consumers.concurrent import parallel
 
 from open_inwoner.openzaak.api_models import OpenSubmission, Zaak
-from open_inwoner.openzaak.clients import CatalogiClient, ZakenClient
 from open_inwoner.openzaak.models import (
     ZaakTypeConfig,
     ZaakTypeStatusTypeConfig,
@@ -250,37 +249,33 @@ class CaseListService:
         functions = [
             functools.partial(
                 self._resolve_resultaat_and_resultaat_type,
-                zaken_client=group.zaken_client,
-                catalogi_client=group.catalogi_client,
+                group=group,
             ),
             functools.partial(
                 self._resolve_status_and_status_type,
-                zaken_client=group.zaken_client,
-                catalogi_client=group.catalogi_client,
+                group=group,
             ),
             functools.partial(
                 self._resolve_zaak_type,
-                client=group.catalogi_client,
+                group=group,
             ),
         ]
 
-        # use contextmanager to ensure the `requests.Session` is reused
-        with group.catalogi_client, group.zaken_client:
-            with parallel(
-                max_workers=self._thread_limits["resolve_case_instance"]
-            ) as executor:
-                futures = [executor.submit(func, case) for func in functions]
+        with parallel(
+            max_workers=self._thread_limits["resolve_case_instance"]
+        ) as executor:
+            futures = [executor.submit(func, case) for func in functions]
 
-            for task in concurrent.futures.as_completed(
-                futures,
-                timeout=self._thread_timeouts["resolve_case_instance"],
-            ):
-                try:
-                    update_case = task.result()
-                    if hasattr(update_case, "__call__"):
-                        update_case(case)
-                except BaseException:
-                    logger.exception("Error in resolving case", stack_info=True)
+        for task in concurrent.futures.as_completed(
+            futures,
+            timeout=self._thread_timeouts["resolve_case_instance"],
+        ):
+            try:
+                update_case = task.result()
+                if hasattr(update_case, "__call__"):
+                    update_case(case)
+            except BaseException:
+                logger.exception("Error in resolving case", stack_info=True)
 
         try:
             zaaktype_config = ZaakTypeConfig.objects.filter_case_type(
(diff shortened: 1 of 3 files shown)


src/open_inwoner/cms/cases/views/services.py
```diff
@@ -112,26 +115,36 @@ def get_submissions_for_api_group(
 
     def get_submissions(self) -> list[SubmissionWithApiGroup]:
         all_api_groups = list(
-            ZGWApiGroupConfig.objects.exclude(form_service__isnull=True)
+            ZGWApiGroupConfig.objects.select_related(
+                "form_service",
+            ).exclude(form_service__isnull=True)
```
pi-sigma: I would expect:

```
all_api_groups = list(
            ZGWApiGroupConfig.objects.filter(form_service__isnull=False).select_related(
                "form_service",
            )
        )
  ```
  although Django might create the same SQL for both so it's just a cosmetic issue (?)

File: src/open_inwoner/cms/cases/views/services.py [hunk 2]
```diff
@@ -112,26 +115,36 @@ def get_submissions_for_api_group(
...
+                                submission=row, api_group=group_for_task
+                            )
+                            for row in task.result()
+                        )
 
-                except BaseException:
-                    logger.exception("Error fetching and pre-processing cases")
+                    except BaseException:
+                        logger.exception("Error fetching and pre-processing cases")
+            except TimeoutError:
```
pi-sigma: Are you using the builtin `TimeoutError`? I thought that one is for OS-related errors (e.g. the opening of a file taking too long) rather than I/O bound operations. There is a `concurrent.futures.TimeoutError`. Why not use this?

src/open_inwoner/cms/cases/views/services.py
```diff
@@ -160,225 +173,243 @@ def get_case_status_frequencies(self) -> dict[CaseFilterFormOption, int]:
...
-                client=group.catalogi_client,
-            ),
-        ]
+                for task in concurrent.futures.as_completed(
+                    fetch_raw_cases_futures,
+                    timeout=self._timeouts["fetch_raw_cases"],
+                ):
+                    raw_cases_for_group = task.result()
+                    fetched_cases.extend(raw_cases_for_group)
+            except TimeoutError:
```
pi-sigma: See previous question about `TimeoutError`

src/open_inwoner/cms/cases/views/services.py
```diff
@@ -53,52 +61,49 @@ def identification(self) -> str:
...
     def __init__(self, request: HttpRequest):
         self.request = request
-        self._thread_timeouts = {
-            "zgw_api_groups": 60,
-            "resolve_case_instance": 15,
-            "resolve_case_list": 15,
+        self._timeouts = {
+            "fetch_raw_cases": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.3,
+            "resolve_cases": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.5,
+            "fetch_submissions": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.2,
```
pi-sigma, line 84: What's the rationale for multiplying the values with a constant instead of taking them as they are? No objection, just curious.

src/open_inwoner/cms/cases/views/services.py
```diff
@@ -112,17 +117,23 @@ def get_submissions_for_api_group(
 
     def get_submissions(self) -> list[SubmissionWithApiGroup]:
```
pi-sigma, line 121: I notice that for `get_cases` you construct `ZaakWithApiGroup` inside the helper function `_get_raw_cases_for_api_group` which is submitted to the futures. OTOH, for `get_submissions`, the helper function `_get_submissions_for_api_group` only returns the submissions for the relevant api group; the `SubmissionWithApiGroup` is then constructed in the main function `get_submissions`. It would be nice if the fetching of zaken and submissions was uniform as it would make the code easier to read - unless there's a reason for the discrepancy which I've missed.

pi-sigma (approved):
The fetching of zaken and submissions could be made more uniform (potentially, see comment), otherwise looking good.

src/open_inwoner/cms/cases/views/services.py
swrichards: The setting is for the _aggregate_ timeout, which means we have to distribute the total timeout proportionally over the timeouts for sub-tasks.

Two alternatives:
- Make the timeouts for individual subtasks configurable (but what criterion would you use to change them?)
- Record our own timings after each block and subtract those from the total timeout, so that e.g. if one task is faster, that leaves more time for the other tasks. But that seemed like a fair bit of complexity.

So this is the low-hanging fruit. The `resolve_cases` task involves the highest number of requests, hence it gets the biggest slice of the timeout.

src/open_inwoner/cms/cases/views/services.py
swrichards: Good point.

PR #1760, opened 2025-05-08
✨  [#3248] Add rich note field to plan - hide CSP warning

jiromaykin:
Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3244 

Test with styles in WYSIWYG:

Screenshot 2025-05-15 at 15 19 44


jiromaykin:
@pi-sigma @stefrado I am assuming there is no need to add this field to "src/open_inwoner/conf/app/setup_configuration.py"

Please take extra time to review the tests, since I am not sure if they are either too much / or not enough.

I've also limited/changed the number of style-options for the CKeditor fields (I've removed the color options for example, but added new options to add UL's and OL's).

src/open_inwoner/conf/base.py
```diff
@@ -476,14 +476,12 @@
     },
 }
 
-
```
pi-sigma: The removal of these newlines is unrelated to the PR. This should be done (if we want this) in a separate dedicated clean-up PR that adds the commits to our [.git-blame-ignore-revs](https://github.com/maykinmedia/open-inwoner/blob/develop/.git-blame-ignore-revs)

src/open_inwoner/configurations/models.py
```diff
@@ -252,7 +252,6 @@ class SiteConfiguration(SingletonModel):
         verbose_name=_("Edit goal message"),
         help_text=_("The message when a user edits a goal."),
     )
-
```
pi-sigma: See other comment about empty lines

src/open_inwoner/js/components/rich-text-editors/ckeditor-hide-notification.js
stefrado, line 1: I think the logic of this file needs to be changed. This observer is now present on every page looking for this warning. A better solution is to use this observer only when the warning element is present. Besides that an even better solution is to hide this warning with just CSS. This can be done by adding this code to `scss/views/_view.scss`
```scss
&--collaborate-plan_edit_note .cke_notification_warning {
  display: none;
}
```

src/open_inwoner/scss/components/Plan/Plan.scss
```diff
@@ -11,9 +11,71 @@
...
+
+  &__note {
+    .form__actions {
+      margin-top: var(--spacing-large);
+    }
+
+    @media (max-width: 768px) {
+      #cke_id_note {
+        // TODO: Remove style when switching to different rich text-editor
+        width: 100% !important;
```
stefrado, line 75: I think this width should be applied until screen width laptop. Now after 768px this editor is overflowing the grid columns.

src/open_inwoner/templates/pages/plans/detail.html
```diff
@@ -89,6 +89,27 @@ <h3 class="utrecht-heading-3" id="files">{% trans "Upload een nieuwe versie" %}<
...
+                <h2 class="utrecht-heading-2" id="notes">{% trans "Notities" %}</h2>
+                <div class="tabled__item--mobile-big">
+                    {% button href="collaborate:plan_edit_note" uuid=object.uuid icon="edit" icon_outlined=True text=_("Bewerk") title=_("Notitie aanvullen") ariaLabel=_("Notitie aanvullen") transparent=True %}
+                </div>
+            </header>
+        </div>
+        <div class="tabled__inner">
+        {% if object.note %}
+            <p class="utrecht-paragraph">{{ object.note|safe }}</p>
+            {% else %}
```
stefrado:
```suggestion
          {% else %}
```

src/open_inwoner/templates/pages/plans/note_edit.html
```diff
@@ -0,0 +1,28 @@
...
+  <p class="utrecht-paragraph">
+    {% trans "Voeg een notitie toe en/of vul een bestaande notitie aan met nieuwe tekst." %}
+  </p>
+
+  {% render_grid %}
+    {% render_column start=4 span=7 %}
+      <div class="plan__note">
+        {% render_form id="note-edit" form=form method="POST" %}
+          {% csrf_token %}
+          {% input form.note %} {# TODO: Remove the javascript that hides the CKeditor notification when switching to different text-editor #}
```
stefrado: If you use the JS solution and switch over to CSS.
```suggestion
          {% input form.note %}
```

src/open_inwoner/js/components/rich-text-editors/ckeditor-hide-notification.js
stefrado: See other comment for another solution.

src/open_inwoner/conf/base.py
```diff
@@ -1018,6 +1008,18 @@
...
+            ["Format"],  # Headings
+            ["Bold", "Italic", "Underline"],
+            ["NumberedList", "BulletedList"],
+            ["Link", "Unlink"],
+            ["Table", "HorizontalRule"],
+            ["RemoveFormat", "Source"],
+            ["Undo", "Redo"],
+        ],
+        "removeButtons": "TextColor,BGColor",  # Remove color-styling
+        "width": 600,
```
stefrado, line 1033: Apparently, by adding `"versionCheck": False,` the warning is completely disabled — both in the frontend and the admin panel.

stefrado (changes requested):
Besides these changes, I noticed that the output of the notes editor is not styled. This may be scheduled for another time, but is something that should not be forgotten.

src/open_inwoner/plans/models.py
```diff
@@ -115,6 +116,15 @@ class Plan(models.Model):
         blank=True,
         help_text=_("How do you intend to achieve this goal?"),
     )
+    note = ckeditor_fields.RichTextField(
+        max_length=1000,
```
pi-sigma: Leave the `max_length` out. Since this is a single note composed of several logically distinct notes, it's not unlikely that we're going to hit the 1000 mark very soon.

src/open_inwoner/plans/models.py
```diff
@@ -115,6 +116,15 @@ class Plan(models.Model):
...
+        default=_(""),
```
pi-sigma: No default

src/open_inwoner/plans/models.py
```diff
@@ -115,6 +116,15 @@ class Plan(models.Model):
...
+        blank=True,
+        verbose_name=_("Edit note message"),
```
pi-sigma: For the admin the `verbose_name` can be useful; for the citizen I'm not so sure. We already have the title from the page and the help text below the input field, also from the HTML. Suggestions:

- Change the text for the beheerder: `"Edit note"` instead of `"Edit note message"`
- Hide this from the editor for the citizen. To this end, update the `PlanNoteForm` as follows:

```
class PlanNoteForm(forms.ModelForm):
        self.fields["note"].label = ""

    class Meta:
        [your code...]

```

Of course, if there's a design/style reason to keep the label, then keep it.

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -329,6 +332,55 @@ def test_plan_file_add_contact_can_access(self):
...
+        # Simulate appending to existing note
+        form["note"] = self.plan.note + "<p>Second note content.</p>"
+        response = form.submit(user=self.user)
+        self.assertEqual(response.status_code, 302)
+
+        self.plan.refresh_from_db()
+        self.assertIn("First note content", self.plan.note)
+        self.assertIn("Second note content", self.plan.note)
+
+    def test_plan_note_edit_contact_can_edit(self):
```
pi-sigma, line 372: The test should include a note by a citizen, then a note by the contact, and check that both contents are present.

pi-sigma (changes requested):
Looking good for the most part. However, the editor behaves differently for staff and regular users on my end. For staff users the editor works correctly:

form-admin

For regular users the editor doesn't work:

form-citizen

Something prevents the editor from being initialized, leading to the "deformed" editor from the image and a type error visible in the console. Are you getting the same result or does this work correctly on your end?

src/open_inwoner/conf/base.py
jiromaykin: @pi-sigma ~~I did not remove these lines - so it must have been Ruff.~~

It may have been PyCharm! I will check again.

src/open_inwoner/conf/base.py
jiromaykin: @stefrado But we kind of decided not to do this during refinement, since we don't want to bite licensing? We decided to only hide it on the front-end for now.

src/open_inwoner/plans/models.py
jiromaykin: We have no design, and no reason.

src/open_inwoner/plans/models.py
jiromaykin: Yeah, adding translation to that was silly... (happened because I had something in there first). But I do see we have multiple models in the project that have `default="",` but i will just completely remove the default if that works.

src/open_inwoner/js/components/rich-text-editors/ckeditor-hide-notification.js
jiromaykin: The CSS solution was of course the first one I tried but that will never work because the CKeditor warning is 'timed' and does not appear on pageload, hence why i am looping Javascript to look for it.

Also: it is actually _**not**_ looking for this element on every page; only on the page with the HTML class `.view--collaborate-plan_edit_note`
In OIP each page has its own .view.... class added to the HTML element.

Since we decided to only hide the warning on the 'citizen' front-end, javascript is still the go-to for now.

src/open_inwoner/configurations/models.py
jiromaykin: ~~I needed Ruff to properly sort and format the files, so I was unable to avoid these removals (i think).~~

It was likely a Black setting in Pycharm.

src/open_inwoner/plans/models.py
jiromaykin: Also: we do not actually have any Admin side for notes, since notes will never be prefilled. 
Notes do not need to be prefilled in the templates for example.
But adding the empty label is a good idea, so we can change it later.

So I think the verbose_name is actually not needed at all?
We only need to think of the mentors and the citizens.

jiromaykin:
> Looking good for the most part. However, the editor behaves differently for staff and regular users on my end. For staff users the editor works correctly:
> [....]
> For regular users the editor doesn't work: [....]
> Something prevents the editor from being initialized, leading to the "deformed" editor from the image and a type error visible in the console. Are you getting the same result or does this work correctly on your end?

@pi-sigma  ~~I do not see this happening at all. All my users see a properly working editor,
I've checked by logging in as a mentor, and as a DigiD citizen, also as an Admin and with a KVK user.~~

~~Are you sure you did a proper Migration and rebuilt NPM and refreshed the cache of your browsers?
In what kind of console do you see a Type error? What kind of error?~~

EDIT/UPDATE: yes, this is happening because of `CSP` blocking (a version request to https://cke4.ckeditor.com/) so we may need to revert this whole story or add an unsafe(?) exemption for this view.

jiromaykin:
> Besides these changes, I noticed that the output of the notes editor is not styled. This may be scheduled for another time, but is something that should not be forgotten.

@stefrado  What do you mean with "the output is not styled"?
I _think_ you mean that we do not add classes to unpredictable rich-text editor content. For now we usually solve this by adding the `.wysiwyg` class around those areas, like in the footer.
Only on the Product pages we have a system to actually add OIP/NLDS classes to the elements, but that's very closely linked to CKeditor only.

Open-Forms i think uses a different solution, with the wysiwyg **mixin** (auto tag based styling) from **microscope sass** with auto margin for typographic elements. Which is useful for styling uncontrolled html like the output from any (?) wysiwyg editor.

src/open_inwoner/templates/pages/plans/note_edit.html
jiromaykin: No.

src/open_inwoner/plans/tests/test_views.py
jiromaykin: ~~You mean mentor?~~ Oh i guess a new test that shows a difference between the user and the contact, like some kind of test_plan_note_edit_appends_content_by_different_users.

src/open_inwoner/scss/components/Plan/Plan.scss
jiromaykin: I'll set it to our Container width (1024px).

src/open_inwoner/js/components/rich-text-editors/ckeditor-hide-notification.js
```diff
@@ -0,0 +1,44 @@
...
+    // Just in case it's already rendered by the time this runs.
+    this.hideWarning()
+  }
+
+  /**
+   * Hides the CKEditor version warning if it exists.
+   */
+  hideWarning() {
+    const warning = this.context.querySelector(
+      '.view--collaborate-plan_edit_note .cke_notification_warning'
```
jiromaykin: @stefrado This `.view--collaborate-plan_edit_note class` on the HTML element makes sure this script is only executed on this 1 unique page in the project.

jiromaykin:
Putting this up for review again because I have a bunch of questions and remarks we need to discuss before finishing this completely @pi-sigma and @stefrado. So if you've seen the remarks and agree with them you can say so here as well. Also if you disagree of course :-)

src/open_inwoner/conf/base.py
jiromaykin: I'll implement this, if this conforms to CSP handling.

jiromaykin:
Closing this version in wait on new rich-text editor.

PR #1761, opened 2025-05-08
fix(pdc): use case list service in filtering categories by zaken

swrichards:
When filtering categories on the frontpage by a users' cases, we used the ZGW client directly, not the CaseList service, which apart from being inconsistent, in this case meant we would always use the RSIN for eHerkenning users due to a hard-coded default in get_user_fetch_params (which had a use_rsin=True default kwarg, which this code did not override).

Fixes: #3249

### DIFF
--- a/src/open_inwoner/pdc/managers.py
+++ b/src/open_inwoner/pdc/managers.py
@@ -7,11 +7,10 @@ from ordered_model.models import OrderedModelQuerySet
 from treebeard.mp_tree import MP_NodeQuerySet
 
 from open_inwoner.accounts.models import User
+from open_inwoner.cms.cases.views.services import CaseListService
 from open_inwoner.configurations.models import SiteConfiguration
-from open_inwoner.openzaak.api_models import Zaak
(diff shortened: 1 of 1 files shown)


pi-sigma (approved):
Looks good. One question: we also use the `get_user_fetch_parameters` as a standalone function (with `use_rsin` default of `True`) in the `SearchView`. Should this be changed as well (because the same reasoning applies) or is it harmless there?

swrichards:
> Looks good. One question: we also use the get_user_fetch_parameters as a standalone function (with use_rsin default of True) in the SearchView. Should this be changed as well (because the same reasoning applies) or is it harmless there?

It's harmless in the sense that nothing breaks, but eHerkenning users with a zaken backend expecting RSIN will not be able to use the search feature successfully. I had a quick look at fixing this but it's non-trivial for a few reasons mostly because the API groups are needed, and it would require a fair bit of test/mock changing and moving things around, unfortunately we're out of time for this. Will pick it up in the next sprint if there's some slack.

PR #1765, opened 2025-05-12
[#3253] Display different metrics for completed/un-completed cases

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3253

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -330,11 +330,26 @@ class InnerCaseDetailView(
                     "label": _("Zaak ingediend op:"),
                     "value": context["case"].get("start_date"),
                 },
-                {
-                    "label": _("Beslissing op zijn laatst:"),
-                    "value": context["case"].get("end_date_legal"),
-                },
             ]
+            if self.case.einddatum:
+                context["metrics"].append(
+                    {
+                        "label": _("Besluit genomen op:"),
+                        "value": self.case.einddatum,
+                    },
+                )
+            else:
+                end_date = context["case"].get("end_date_legal") or context["case"].get(
+                    "end_date_planned"
+                )
+                context["metrics"].append(
+                    {
+                        "label": _("U ontvangt een besluit vóór:"),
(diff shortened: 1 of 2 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -301,11 +301,26 @@ def get_context_data(self, **kwargs):
...
-                    "value": context["case"].get("end_date_legal"),
-                },
             ]
+            if self.case.einddatum:
+                context["metrics"].append(
+                    {
+                        "label": _("Besluit genomen op:"),
+                        "value": self.case.einddatum,
+                    },
+                )
```
swrichards, line 340: will this localize the date (ie does it happen in the template?)

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -301,11 +301,26 @@ def get_context_data(self, **kwargs):
...
+            else:
+                end_date = context["case"].get("end_date_legal") or context["case"].get(
+                    "end_date_planned"
+                )
+                context["metrics"].append(
+                    {
+                        "label": _("U ontvangt een besluit voor:"),
```
swrichards:
```suggestion
                        "label": _("U ontvangt een besluit vóór:"),
```

src/open_inwoner/cms/cases/views/status.py
swrichards, line 1: Any tests we could add an assert to for some coverage?

swrichards (changes requested):

src/open_inwoner/cms/cases/views/status.py
pi-sigma: No, has to be done in the view indeed.

src/open_inwoner/cms/cases/views/status.py
pi-sigma: Yes, but requires new mock. Not sure it's worth it for this one assert, but we might want to use the new mock for testing other variations between different individual zaken.

swrichards (approved):

PR #1766, opened 2025-05-12
[#3254] Display `esuite_compat_naam` for resultaattype in admin

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3254

### DIFF
--- a/src/open_inwoner/openzaak/admin.py
+++ b/src/open_inwoner/openzaak/admin.py
@@ -17,6 +17,7 @@ from privates.storages import PrivateMediaFileSystemStorage
 from solo.admin import SingletonModelAdmin
 
 from open_inwoner.ckeditor5.widgets import CKEditorWidget
+from open_inwoner.openzaak.clients import build_zgw_client_from_service
 from open_inwoner.openzaak.import_export import ZGWConfigExport, ZGWConfigImport
 from open_inwoner.utils.forms import LimitedUploadFileField
 
@@ -377,11 +378,13 @@ class ZaakTypeResultaattypeConfigInline(admin.StackedInline):
         "omschrijving",
         "resultaattype_url",
         "zaaktype_uuids",
+        "esuite_compat_naam",
         "description",
     ]
     readonly_fields = [
         "omschrijving",
         "resultaattype_url",
+        "esuite_compat_naam",
         "zaaktype_uuids",
     ]
     ordering = (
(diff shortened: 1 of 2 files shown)


swrichards (commented):
Discussed: no new model field, instead add a new admin read only field that fetches `esuite_compat_naam` from the configured ZTC api (via the `Service` connected to the `ResultaatType` model).

src/open_inwoner/openzaak/admin.py
```diff
@@ -395,6 +398,15 @@ def has_add_permission(self, request, obj):
...
         return request.user.is_superuser
 
+    def esuite_compat_naam(self, instance):
+        catalogi_client = build_zgw_client_from_service(
+            service=instance.zaaktype_config.catalogus.service
+        )
+        result_type = catalogi_client.fetch_single_resultaat_type(
+            resultaat_type_url=instance.resultaattype_url
+        )
+        return getattr(result_type, "esuite_compat_naam", "")
```
swrichards: I think it might be a good idea to wrap this in a `try/except` to guard against unknown issues which would have the effect of rendering the whole admin page unreachable if there's an issue with the connection. Best to be cautious here I think.

swrichards: (I would then add a `logger.exception`, as it would be an actual issue, not one of our abuse-Sentry-as-a-logger things :).

swrichards (changes requested):
LGTM, just one final comment.

src/open_inwoner/openzaak/admin.py
```diff
@@ -395,6 +398,21 @@ def has_add_permission(self, request, obj):
...
+        try:
+            catalogi_client = build_zgw_client_from_service(
+                service=instance.zaaktype_config.catalogus.service
+            )
+            result_type = catalogi_client.fetch_single_resultaat_type(
+                resultaat_type_url=instance.resultaattype_url
+            )
+        except Exception as exc:
+            logger.exception(
+                "Error fetching resultaattype for zaaktype_config admin: %s", exc
```
swrichards: The `exc` argument is not necessary I think because the exception info will be included.

pi-sigma: Need to remember this, I have this habit of automatically including the `exc` myself :)

swrichards (approved):

PR #1767, opened 2025-05-13
[#3039] Remove error suppression in synchronous API based flows

pi-sigma:
A couple of improvements to our error handling in API calls.

~~- [x] Warn users on "Mijn Aanvragen" if one or more ZGW backends failed to respond~~
- [x] Display a user error if we are unable to sync a user's saved profile fields to the klanten API
- [x] Error if digid-based eenmanszaak login cannot reach KvK
- [x] Warn user if more than one zaak was found for a question
- [x] Gracefully handle misconfigured SSD clients
- [x] Display an error message if we were unable to create a Question via API

In addition, error handling for the KVK and SSD clients was improved by introducing custom exceptions which are handled closer to the views.
> [!NOTE]  
>Tthe change to the `BlockEenmanszaakLoginMixin` is a change in behavior, not merely a refactor: if there is an error when fetching the basisprofiel from the KVK API, eherkenning login is down for everyone, and an error message is displayed to the user. The potential inconvenience for eherkenning users who are not eenmanszaken is preferable to the risk of allowing eenmanszaken to login via eherkenning (instead of digid).

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3039

### DIFF
--- a/src/open_inwoner/accounts/views/contactmoments.py
+++ b/src/open_inwoner/accounts/views/contactmoments.py
@@ -1,13 +1,14 @@
 import logging
 from typing import Iterable, Protocol
 
+from django.contrib import messages
 from django.contrib.auth.mixins import AccessMixin
 from django.core.exceptions import ImproperlyConfigured
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.functional import cached_property
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext as _
 from django.views import View
 from django.views.generic import TemplateView
 
@@ -197,14 +198,27 @@ class KlantContactMomentDetailView(KlantContactMomentBaseView):
             raise ImproperlyConfigured("Unknown KlantenServiceType")
 
         origin = self.request.headers.get("Referer")
-        question, zaak_with_api_group = service.retrieve_question(
+        question, zaken_with_api_group = service.retrieve_question(
             self.get_fetch_params(service),
             question_uuid=kwargs["kcm_uuid"],
             user=self.request.user,
         )
         if not question:
             raise Http404()
+        if zaken_with_api_group and len(zaken_with_api_group) > 1:
+            # In principle this should not happen, a zaak should be stored in
+            # exactly one backend. But: https://www.xkcd.com/2200/
+            # We should at least receive a ping if this happens and warn the user.
+            logger.error("Found one zaak in multiple backends")
+            messages.warning(
+                self.request,
+                _(
+                    "Your question was connected to multiple cases, we are showing you the first."
+                ),
+            )
+
         self.question = question
+        zaak_with_api_group = zaken_with_api_group[0] if zaken_with_api_group else None
         QuestionValidator.validate_python(question)
 
         local_kcm, created = KlantContactMomentAnswer.objects.get_or_create(  # noqa
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -28,6 +28,7 @@ from open_inwoner.configurations.models import SiteConfiguration
 from open_inwoner.haalcentraal.utils import fetch_brp
 from open_inwoner.laposta.forms import NewsletterSubscriptionForm
 from open_inwoner.laposta.models import LapostaConfig
+from open_inwoner.openklant.api_models import Klant
 from open_inwoner.openklant.services import eSuiteKlantenService
 from open_inwoner.plans.models import Plan
 from open_inwoner.qmatic.client import NoServiceConfigured, qmatic_client_factory
@@ -230,18 +231,35 @@ class EditProfileView(
         return self.request.user
 
     def form_valid(self, form):
-        form.save()
         user: User = self.get_object()
 
-        self.update_esuite_klant(
+        klant = self.update_esuite_klant(
             {k: form.cleaned_data[k] for k in form.changed_data}, user
         )
 
-        messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
-        self.log_change(self.get_object(), _("profile was modified"))
+        # only save the form if changes have been written to Klanten API or
+        # changes do not require writing to API
+        if klant or not any(
+            key in form.changed_data
+            for key in ("email", "phonenumber", "phonenumber_alternative")
+        ):
+            form.save()
+            messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
+            self.log_change(self.get_object(), _("profile was modified"))
+        else:
+            messages.error(
+                request=self.request,
+                message=_(
+                    "Your changes could not be saved due to technical problems. "
+                    "Please try again later"
+                ),
+            )
+            self.log_change(
+                self.get_object(), _("profile changes not saved due to Klant API error")
+            )
         return HttpResponseRedirect(self.get_success_url())
 
-    def update_esuite_klant(self, user_form_data: dict, user: User):
+    def update_esuite_klant(self, user_form_data: dict, user: User) -> Klant | None:
         field_mapping = {
             "emailadres": "email",
             "telefoonnummer": "phonenumber",
@@ -266,7 +284,7 @@ class EditProfileView(
                 fetch_params=fetch_params, user=user
             )
             if klant and not created:
-                service.update_klant_from_user(
+                return service.update_klant_from_user(
                     klant, user, update_fields=list(update_data.keys())
                 )
 
--- a/src/open_inwoner/kvk/client.py
+++ b/src/open_inwoner/kvk/client.py
@@ -3,11 +3,12 @@ from functools import cached_property
 from urllib.parse import urlencode
 
 import requests
-from requests.exceptions import JSONDecodeError
+from requests.exceptions import InvalidJSONError, JSONDecodeError
 
 from open_inwoner.utils.decorators import cache as cache_result
 
 from .constants import CompanyType
+from .exceptions import KVKAPIException
 from .models import KvKConfig
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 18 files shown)


pi-sigma:
@swrichards For https://taiga.maykinmedia.nl/project/open-inwoner/task/3062 (the edge case concerning a question related to several zaken from different backends) I've stuck with current behavior of displaying the first question if we have many. Any ideas on how we could determine which zaak is the most relevant for a user? I'm skeptical this can be done in a general fashion and therefore went with current behavior, but of course I'm open to suggestions.

swrichards:
> @swrichards For https://taiga.maykinmedia.nl/project/open-inwoner/task/3062 (the edge case concerning a question related to several zaken from different backends) I've stuck with current behavior of displaying the first question if we have many. Any ideas on how we could determine which zaak is the most relevant for a user? I'm skeptical this can be done in a general fashion and therefore went with current behavior, but of course I'm open to suggestions.

I agree there's no real good way to do this. My concern was not directly to change the behavior, but warn the user there _might_ be something wrong. I am aware that's questionable UX: I was thinking of the case where a user knows they've submitted a question, but don't see it when they log in (so something like "You might not see all questions, please try again later", similar to the case for when one but not all of the ZGW backends fails).

src/open_inwoner/kvk/exceptions.py
```diff
@@ -0,0 +1,15 @@
...
+        super().__init__(message)
+        self.original_exception = original_exception
+        self.custom_message = message
+        self.default_message = _(
+            "We're experiencing technical difficulties. Please try again later."
+        )
+
+    @property
+    def message(self):
+        return self.custom_message or self.default_message
```
swrichards: A few points here.

I find `custom` and `default` a bit confusing here. The common pattern for exceptions is that you can provide a message to override the default message, which can just be a default on the kwarg.

More broadly: the pattern here is that you intend to use the default message to display directly to the user. That feels wrong to me: generally, exception messages should be considered internal implementation details, and in the exception you don't know where or how the message will be used. I am happy to _have_ a default message, but I would define it elsewhere to make sure it's intended as a user-facing message.

I am not sure the `original_exception` adds much given we're using (and having `ruff` enforce) the `... raise from` syntax. _However_, what I think would be useful is to check for e.g. the `fout` or whatever key holds the error, extract it, and store that in a `api_error_message` string. I would make a separate constructor for this, e.g. `@classmethod` and `def from_error_response(cls, response: Response)`, so that we can keep the common pattern of `message` being the primary positional arg on `__init__` for _developer facing_ messages.

src/open_inwoner/ssd/exceptions.py
```diff
@@ -0,0 +1,15 @@
...
+        super().__init__(message)
+        self.original_exception = original_exception
+        self.custom_message = message
+        self.default_message = _(
+            "We're experiencing technical difficulties. Please try again later."
+        )
+
+    @property
+    def message(self):
+        return self.custom_message or self.default_message
```
swrichards: Similar comments as above with the KVK exception.

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -197,14 +198,28 @@ def get_context_data(self, **kwargs):
...
+            # exactly one backend. But: https://www.xkcd.com/2200/
+            # We should at least receive a ping if this happens and warn the user.
+            logger.error("Found one zaak in multiple backends")
+            messages.warning(
+                self.request,
+                _(
+                    "The question was found to be related to several zaken, but we cannot "
+                    "determine which zaak is the most relevant"
+                ),
+            )
```
swrichards: Perhaps:

```suggestion
            messages.warning(
                self.request,
                _(
                    "Your question was connected to multiple cases, we are showing you the first."
                ),
            )
```

I think the point of the message is to let the user know the page might not meet their prior expectations, rather than letting them know what _our_ problem is.

swrichards (changes requested):

src/open_inwoner/kvk/exceptions.py
pi-sigma: Agree, that's a better pattern

src/open_inwoner/accounts/views/profile.py
```diff
@@ -230,18 +231,31 @@ def get_object(self):
...
 
-        self.update_esuite_klant(
+        klant = self.update_esuite_klant(
             {k: form.cleaned_data[k] for k in form.changed_data}, user
         )
 
-        messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
-        self.log_change(self.get_object(), _("profile was modified"))
+        # only save the form if changes have been written to Klanten API
+        if klant:
```
swrichards: `klant` will also be `None` if there's no update-able fields (it looks like we only update email and such, because eSuite does not allow updating names, so if you change your `first_name`, you still see the error and the changes are not saved).

src/open_inwoner/openklant/views/contactform.py
```diff
@@ -330,4 +330,10 @@ def _create_contactmoment(
                 identificatie=employee_id
             )
 
-        return self.vragen_service.create_contactmoment(data, klant=klant)
+        contactmoment = self.vragen_service.create_contactmoment(data, klant=klant)
+        if not contactmoment:
+            messages.warning(
+                request=self.request,
+                message=_("Your question could not be saved. Please try again later."),
+            )
```
swrichards: Double-check that this is not redundant with `self.set_result_message`

src/open_inwoner/kvk/exceptions.py
```diff
@@ -0,0 +1,21 @@
...
+    def from_error_response(cls, response: Response) -> Self:
+        try:
+            data = response.json()
+        except Exception:
+            api_error_message = ""
+        else:
+            api_error_message = data.get("fout", [{}])[0].get("omschrijving", "")
+        cls.message = api_error_message
+
+        return cls(message=api_error_message)
```
swrichards: I think we can simplify this, and keep the response around for traceability:

```suggestion
class KVKAPIException(Exception):
    response: Response | None
    def __init__(self, message=None, *, response: Response | None = None):
        super().__init__(message or "An error occurred while communicating with the KvK API")
        self.response = response

    @classmethod
    def from_error_response(cls, response: Response) -> Self:
        api_error_message: str | None = None
        with contextlib.suppress(BaseException):
            data = response.json()
            api_error_message = data.get("fout", [{}])[0].get("omschrijving", "")
  
        return cls(message=api_error_message, response=response)
```

src/open_inwoner/kvk/views.py
```diff
@@ -33,9 +35,14 @@ def get_form_kwargs(self):
...
 
-        company_branches = kvk_client.get_all_company_branches(
-            kvk=self.request.user.kvk
-        )
+        try:
+            company_branches = kvk_client.get_all_company_branches(
+                kvk=self.request.user.kvk
+            )
+        except KVKAPIException as exc:
+            messages.error(request=self.request, message=exc.default_message)
```
swrichards: The exception message is internal:

```suggestion
            messages.error(request=self.request, message="We are temporarily unable to show your branches, please try again at a later point.")
```

swrichards: (I also think we deprecated `default_message` in an earlier iteration).

src/open_inwoner/openklant/services.py
```diff
@@ -765,18 +766,19 @@ def retrieve_question(
...
+                groups = [
+                    ZGWApiGroupConfig.objects.resolve_group_from_hints(
+                        client=case.client
+                    )
+                    for case in cases_found
+                ]
+                zaken_with_api_group = [
+                    ZaakWithApiGroup(zaak=zaak, api_group=group)
+                    for zaak, group in zip(zaken, groups)
+                ]
```
```suggestion
                zaken_with_api_group = [
                    ZaakWithApiGroup(zaak=case.result, api_group=ZGWApiGroupConfig.objects.resolve_group_from_hints(case.client))
                    for zaak, group in zip(zaken, groups)
                ]
```

src/open_inwoner/ssd/exceptions.py
```diff
@@ -0,0 +1,27 @@
...
+        # fout, waarshuwing, informatie
+        fwi = xml_response.fwi
+
+        exceptions = fwi.fout or fwi.waarshuwing or fwi.informatie
+        if not exceptions:
+            return cls("SSD client error")
+
+        melding = exceptions[0]
+
+        return cls(message=melding.tekst)
```
swrichards:
```suggestion
class SSDClientException(Exception):
    response: JaarOpgaveInfoResponse | UitkeringInfoResponse | None
    def __init__(self, message: str | None=None):
        super().__init__(message or "SSD client error")

    @classmethod
    def from_xml_response(
        cls, xml_response: JaarOpgaveInfoResponse | UitkeringInfoResponse
    ) -> Self:       
        # fout, waarshuwing, informatie
        fwi = xml_response.fwi
        error_messages = ", ".join(e for e in (fwi.fout or fwi.waarshuwing or fwi.informatie,) if e)
        
        return cls(message=error_messages)
```

src/open_inwoner/kvk/client.py
```diff
@@ -56,15 +57,19 @@ def _request(self, endpoint: str, params: dict | None = None) -> dict:
...
-        except requests.RequestException as ex:
-            logger.exception("Unable to retrieve information from the KVK API: %s", ex)
-            return {}
+            response.raise_for_status()
+        except requests.HTTPError as exc:
+            logger.exception("Error while making request to KVK API")
+            raise KVKAPIException.from_error_response(exc.response) from exc
+        except requests.RequestException as exc:
+            logger.exception("Error while making request to KVK API")
+            raise KVKAPIException from exc
```
swrichards: I know I said earlier that differentiating between error types is not relevant if the user cannot change their behaviour because of it, but I think this reveals that _we_ might want to change, specifically, what we log. I think if all of these are `exception` that would be quite loud.

I don't think we have to differentiate now, but let's use `logger.warning(..., exc_info=True)` here. I'd like us to move towards only sending _unhandled_ errors to Sentry, and to (eventually) pick up the warnings through log aggregation.

swrichards (changes requested):
Almost there. just a few more issues re exception design and a double check of the flow in profile/contactmomenten.

swrichards (approved):

PR #1769, opened 2025-05-19
:sparkles: Add Plain text notes field to plan

jiromaykin:
Reverted richt text, added plain text instead
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3248


jiromaykin:
@pi-sigma @stefrado I made a completely new PR for this issue, because we're now making it **plain** text, instead of rich.
My other PR #1760  does work but we are not using that, since we'll either be 1. ditching CKeditor, or 2. solving CSP differences with 'front' front-end and the admin safely.

src/open_inwoner/conf/base.py
```diff
@@ -1019,6 +1019,19 @@
...
+            ["Format"],  # Headings
+            ["Bold", "Italic", "Underline"],
+            ["NumberedList", "BulletedList"],
+            ["Link", "Unlink"],
+            ["Table", "HorizontalRule"],
+            ["RemoveFormat", "Source"],
+            ["Undo", "Redo"],
+        ],
+        "removeButtons": "TextColor,BGColor",  # Remove color-styling
+        "versionCheck": False,
```
jiromaykin, line 1033: @pi-sigma @stefrado Even though this PR is no longer about using the rich text-editors I thought it would be nice to improve the options of the existing ones, but I'm not sure if I should keep the `versionCheck` here. 
It does look nicer without the warnings.

pi-sigma (approved):

src/open_inwoner/conf/base.py
pi-sigma: I think it's nice to disable the warning for the time we're stuck with the outdated version of CKEditor.

stefrado: I agree with @pi-sigma - it is better to disable the warning for now.

src/open_inwoner/scss/components/Plan/Plan.scss
```diff
@@ -11,9 +11,68 @@
...
     }
+
     .button {
       margin-bottom: var(--row-height);
       margin-top: var(--spacing-medium);
     }
   }
+
+  // Deviant breakpoints due to sidebar
+  .dashboard {
```
stefrado, line 22: I think it would be better to move the new dashboard styling to a separate PR to keep this one concise.

src/open_inwoner/templates/pages/plans/detail.html
```diff
@@ -89,6 +89,29 @@ <h3 class="utrecht-heading-3" id="files">{% trans "Upload een nieuwe versie" %}<
...
+                </div>
+            </header>
+        </div>
+        <div class="tabled__inner">
+        {% if object.note %}
+            <div class="wysiwyg">
+                <p class="utrecht-paragraph">{{ object.note|linebreaksbr }}</p>
+            </div>
+        {% else %}
+            <p class="utrecht-paragraph"><strong>{% trans "Er zijn nog geen notities toegevoegd." %}</strong></p>
```
stefrado, line 107: Shouldn't this be a `h4` instead of a p & strong?

src/open_inwoner/templates/pages/plans/goal_edit.html
stefrado, line 1: Same for this file -> I think it would be better to move the new goal edit page styling to a separate PR to keep this one concise.

src/open_inwoner/templates/pages/plans/note_edit.html
```diff
@@ -0,0 +1,28 @@
+{% extends 'master.html' %}
+{% load i18n button_tags card_tags map_tags form_tags grid_tags %}
+
+{% block extra_head %}
+    {{ form.media }} {# Loads necessary JS for CKEditor #}
```
stefrado: Is this still needed?

stefrado (changes requested):

src/open_inwoner/templates/pages/plans/detail.html
jiromaykin: Ah yes. Maybe. A heading should only be used if it actually describes the paragraph underneath, but in this case that could work.

jiromaykin: Oh no wait: actually another heading is NOT allowed here. Since there is already a Heading 2 above it. 
So the content underneath is what is described by that H2, which should not be directly followed by an H4 (or even an H3) anyway.
So: no.

src/open_inwoner/templates/pages/plans/note_edit.html
jiromaykin: Oops, no, will remove.

src/open_inwoner/scss/components/Plan/Plan.scss
jiromaykin: In this case this style is unique for this 1 page that is already being changed in this PR, so I'd rather keep this change in the same PR.

src/open_inwoner/templates/pages/plans/goal_edit.html
jiromaykin: Usually I would agree, but this is such a small change, and I am expecting more people to now see this eye sore, since this is the first time in over 2 years we're building new things for the whole Plans section.

stefrado (approved):

PR #1770, opened 2025-05-19
✨  [#3257] Improve Sitemap styling

jiromaykin:
Correcting HTML and adding design-tokens: https://taiga.maykinmedia.nl/project/open-inwoner/task/3257 

Making the design more similar to https://www.rijksoverheid.nl/sitemap with our own icon handling.
Style looks a bit similar to our Card-link-lists with Sub-Categories, but not bold: https://open-inwoner-test.maykin.nl/onderwerpen/werk-en-inkomen/


src/open_inwoner/scss/components/Typography/H2.scss
```diff
@@ -15,6 +14,10 @@
     overflow-wrap: break-word;
   }
 
+  .link {
+    font-size: var(--utrecht-heading-2-font-size);
+  }
+
```
pi-sigma, line 20: For regular links we have `text-decoration: none;` and then `text-decoration: underline;` on `hover`. On the other hand: `h2` headings which are also links are always underlined. Do we want to treat the headings differently or should we make them consistent with regular links?

pi-sigma (approved):
Looking nice! Small question about the styling of heading links compared to other links, but nothing blocking.

src/open_inwoner/scss/components/Typography/H2.scss
jiromaykin: Good question. We don't have a design for this, but it's actually better for accessibility to make them underlined first, which is why I had to make links in the FAQ and product pages underlined. Surprisingly none of the accessibility audits asked us to change/reverse all of our anchor links this way. :astonished:
I'd say: let's keep headings conform to the overall style first, so it looks better on this page. If in future I'll need to adjust all link styles everywhere, that'd be a new issue.

stefrado (approved):
Looks good to me. One remark: do we really want to show the 'login or create an account' link to authenticated users?

jiromaykin:
> Looks good to me. One remark: do we really want to show the 'login or create an account' link to authenticated users?

Good point, no we shouldn't have that. Also I feel like some translations are missing in the view, so I will add them.

src/open_inwoner/core/views.py
```diff
@@ -44,42 +45,51 @@ def sitemap(request):
...
-            {"url_name": "profile:categories", "link_text": "Mijn interessegebieden"},
-            {"url_name": "profile:contact_list", "link_text": "Mijn begeleiders"},
-            {"url_name": "profile:action_list", "link_text": "Mijn contacten"},
-            {"url_name": "profile:appointments", "link_text": "Mijn afspraken"},
+            {"url_name": "profile:detail", "link_text": _("Mijn profiel")},
+            {
+                "url_name": "profile:categories",
+                "link_text": _("Mijn interessegebieden"),
+            },
+            {"url_name": "profile:contact_list", "link_text": _("Mijn begeleiders")},
```
jiromaykin: @stefrado @pi-sigma  Maybe we should omit this one (?), since the actual correct URL for this would be `{% url 'profile:contact_list' %}?type=begeleider` but many users will not actually have a mentor, so then there would be no results.

src/open_inwoner/core/views.py
```diff
@@ -44,42 +45,51 @@ def sitemap(request):
...
+            {"url_name": "profile:action_list", "link_text": _("Mijn contacten")},
+            {"url_name": "profile:appointments", "link_text": _("Mijn afspraken")},
         ]
 
     # add "platform pages" (login/create account, cookie info etc.)
```
stefrado: Forgot to update the comment :)

src/open_inwoner/scss/components/Sitemap/sitemap.scss
```diff
@@ -1,4 +1,30 @@
...
+  display: grid;
+  gap: var(--spacing-giant) var(--spacing-mega);
+  grid-template-columns: 1fr;
+  grid-template-areas:
+    'platform'
+    'onderwerpen'
+    'mijn_profiel';
+
+  @media (min-width: 768px) {
+    grid-template-columns: 1fr 1fr;
```
stefrado, line 11: Also add `grid-template-rows: auto 1fr;` to make the first row fit it's content.

src/open_inwoner/core/views.py
```diff
@@ -44,42 +45,51 @@ def sitemap(request):
...
-            {"url_name": "profile:contact_list", "link_text": "Mijn begeleiders"},
-            {"url_name": "profile:action_list", "link_text": "Mijn contacten"},
-            {"url_name": "profile:appointments", "link_text": "Mijn afspraken"},
+            {"url_name": "profile:detail", "link_text": _("Mijn profiel")},
+            {
+                "url_name": "profile:categories",
+                "link_text": _("Mijn interessegebieden"),
+            },
+            {"url_name": "profile:contact_list", "link_text": _("Mijn begeleiders")},
+            {"url_name": "profile:action_list", "link_text": _("Mijn contacten")},
```
stefrado: This is the wrong url. `"profile:action_list"` -> `"profile:contact_list"`

src/open_inwoner/core/views.py
stefrado: I think we can leave it like this - in this case, the url should be changed.

src/open_inwoner/core/views.py
```diff
@@ -44,42 +45,51 @@ def sitemap(request):
...
+            {"url_name": "collaborate:plan_list", "link_text": _("Samenwerken")}
         )
     if inbox_page_is_published():
         context["cms_pages"].append(
-            {"url_name": "inbox:index", "link_text": "Mijn berichten"}
+            {"url_name": "inbox:index", "link_text": _("Mijn berichten")}
         )
     if products_page_is_published():
         context["cms_pages"].append(
             {"url_name": "products:questionnaire_list", "link_text": "Zelftest"}
```
stefrado, line 69: Wrap with `_("")`?

stefrado (changes requested):

src/open_inwoner/core/views.py
jiromaykin: @stefrado ? You are contradicting your own comment here? Leave it or change it? Note we already have a link to My Contacts.

src/open_inwoner/scss/components/Sitemap/sitemap.scss
jiromaykin: I did that at first but looks terribly hideous in case of very long page titles.

src/open_inwoner/core/views.py
stefrado: @jiromaykin I meant changing by adding the query `?type=begeleiger` to this url.

src/open_inwoner/core/views.py
jiromaykin: Is this a review-comment? What do you mean?

stefrado: I meant with this comment that you forgot to remove `login/create account` from this comment, but it is still there only now conditional. (so i'll resolve this comment ;) )

jiromaykin: I see, that was a code-comment by Paul - the meaning is a bit odd now yes.

src/open_inwoner/core/views.py
```diff
@@ -1,4 +1,8 @@
+from urllib.parse import urlencode
+
 from django.shortcuts import render
+from django.urls import reverse
```
jiromaykin, line 4: @pi-sigma Perhaps you could look at this part again to see if I'm doing this correctly - I'm doing a bunch of reverseURL things here in order to try and get the Begeleiders URL.
It seems to work...

stefrado (approved):

PR #1758, opened 2025-05-06
[#3242] Replace 'welcome' with icon and add firstname/businessname in authenticated menu

jiromaykin:
Taiga story:
https://taiga.maykinmedia.nl/project/open-inwoner/us/3222

separated issue:
https://taiga.maykinmedia.nl/project/open-inwoner/task/3242

- [x] remove Welkom texts in menu's
- [x] also add vestigingen or business names in desktop
- [x] add how_to_reg icon before names on desktop
- [x] due to new spacing because of extra icon: correct all spacing (including for Dyslexia styles etc.)
- [x] check tablet styles
- [ ] ~~change how_to_reg icon on mobile into Account_Circle~~ Keep as is
- [x] make mobile icon only link to Profiel page
- [x] keep/add tracking with Siteimprove
- [x] change icon color on Desktop into secondary color when dropdown is opened (? not sure if that is needed)

PR #1772, opened 2025-05-20
[#3246] Hide branch data from Siteimprove tracking

jiromaykin:
Fixing issues: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3146

Tracking personal data from the Registration page was already solved in PR #1758 

So from now on:
- the Profile edit page should not send names/mailadresses etc. to Siteimprove
- nor should the Profiel Registration page
- also clicking on the KVK branches should not send any of the business names or other data to Siteimprove and should be overwritten with different labels or the Siteimprove Dashboard
- the `_sz` mock object should only work in Dev environments, so inspection of the Console is possible for Devs, but this should not be visible in Production (and perhaps Acceptance)

### DIFF
--- a/src/open_inwoner/templates/master.html
+++ b/src/open_inwoner/templates/master.html
@@ -28,7 +28,7 @@
 
         {% if favicon %}<link rel="icon" href="{{ favicon }}" />
         <link rel="apple-touch-icon" href="{{ favicon }}">{% endif %}
-
+    
         <link nonce="{{ request.csp_nonce }}" href="{% static 'bundles/open_inwoner-css.css' %}" media="all" rel="stylesheet" />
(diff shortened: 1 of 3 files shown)


src/open_inwoner/templates/master.html
```diff
@@ -174,6 +174,13 @@
 
         {% render_block "js" %}
         {% block extra_js %}{% endblock %}
+        
+        {% if debug %}
```
stefrado, line 178: There is a double check to see if debug is true.
If debug is true we can do `window.IS_DEV = true`, otherwise the script is not rendered.

stefrado (approved):

PR #1774, opened 2025-05-22
Integrate CMS pages into the search results

swrichards:
[Taiga 3220](https://taiga.maykinmedia.nl/project/open-inwoner/us/3220?milestone=532)


src/open_inwoner/search/searches.py
```diff
@@ -67,14 +80,76 @@ def query(self, search, query):
...
+
+
+def build_cms_search(query_str: str) -> Search:
+    """Build a search object targeting the CMS Page index."""
+    search = CMSPageDocument.search()
+    search = search.query(
+        Q(
+            "multi_match",
+            query=query_str,
+            fields=["title", "description", "rendered_content_placeholder"],
```
pi-sigma: The definition of `CMSPageDocument` has `rendered_placeholders`, not `rendered_content_placeholder`. Is this a typo or am I misunderstanding the relationship between document and query?

src/open_inwoner/search/searches.py
```diff
@@ -67,14 +80,76 @@ def query(self, search, query):
...
+    search_query = search_query.add(page_search)
+
+    logger.debug("built_search_query: %s", search_query.to_dict())
+    products_response, pages_response = search_query.execute()
+
+    return ProductSearchResult.build_from_response(
+        response=product_search.faceted_response_from_multi_search_response(
+            products_response,
+        )
+    ), CMSPageSearchResult.build_from_response(pages_response)
```
pi-sigma: An error in one of the searches will bring down the the entire `multi_search`. Could use error handling to ensure that `multi_search` will display the results of one search of the other one fails.

src/open_inwoner/search/tests/test_search_pages.py
```diff
@@ -0,0 +1,42 @@
...
+        ):
+            with self.subTest(term):
+                _, pages_result = multi_search(term)
+                self.assertEqual(
+                    [r.meta.id for r in pages_result.results], [str(expected_page.id)]
+                )
+                self.assertEqual(pages_result.get_generic_hits(), [expected_hit])
+
+    def test_unpublished_pages_are_not_indexed(self):
+        pass
```
pi-sigma: #yolo?

src/open_inwoner/search/documents.py
```diff
@@ -72,3 +78,39 @@ def get_queryset(self):
...
+    rendered_placeholders = fields.TextField(analyzer=html_strip)
+    description = fields.TextField()
+    url = fields.TextField()
+
+    def prepare_description(self, instance: Page):
+        return instance.get_meta_description(fallback=False, language="nl")
+
+    def prepare_rendered_html(self, instance: Page):
+        content = render_full_page(instance)
+        return content
```
pi-sigma, line 98: Perhaps consider caching. CMS pages could be re-indexed several times if e.g. the admin edits and saves in quick succession, or if the `Page` model has related models or placeholders tied to other models. In every case we're rendering the full page, which is wasteful. Though this might also be per-mature optimization

src/open_inwoner/search/documents.py
```diff
@@ -72,3 +78,39 @@ def get_queryset(self):
...
+
+    def prepare_rendered_placeholders(self, instance: Page):
```
pi-sigma, line 100: Same as previous comment

pi-sigma (changes requested):
The definition of query search fields for the CMS page search has a mistake (?), the other comments are suggestions.

src/open_inwoner/search/searches.py
swrichards: You have a sharp eye, that was an earlier incarnation. Will fix, good catch.

src/open_inwoner/search/documents.py
swrichards: In principle we shouldn't be indexing often: one a day, sometimes more if triggered, but always asynchronously, so I don't think we have to be too worried about performance yet (and generally we'd want to be cautious using caches when indexing, we want the freshest wares given that the search index is itself a kind of cache!).

src/open_inwoner/search/tests/test_search_pages.py
swrichards: Yea :/ I'll remove for now.

src/open_inwoner/search/documents.py
swrichards: (Note that I did turn off the auto-sync on saves -- I think the indexing does not need to be as frequent as that, users can always run the Celery task to explicitly re-index before the periodic job, and it helps the performance concerns).

src/open_inwoner/search/searches.py
swrichards: Currently I struggle to think what would cause one request to fail but another to succeed (given that they happen in the same HTTP request). The library is sketchy on how to deal with errors. I'd like to see how this evolves, as the exceptions here are caught in the view. Then we can see how to do more fine-grained handling once we see what cases we're talking about in practice.

pi-sigma (approved):
Good to go, two stylistic recommendations only.

swrichards:
> Good to go, two stylistic recommendations only.

I can't see them, are they pre-resolved?

src/open_inwoner/cms/tests/cms_tools.py
```diff
@@ -142,3 +216,20 @@ def create_apphook_page(
         raise Exception("failed to publish page")
 
     return p
+
+
+def create_cms_page_with_content(*, title: str, content: str, language: str = "nl"):
```
pi-sigma:
```suggestion
def create_cms_page_with_content(*, title: str, content: str, language: str = "nl") -> Page:
```

src/open_inwoner/search/searches.py
```diff
@@ -67,12 +80,68 @@ def query(self, search, query):
...
+) -> tuple[ProductSearchResult, CMSPageSearchResult]:
+    """Execute a search across the Products and CMS Page indexes."""
+    product_search = ProductSearch(query_str, filters=filters or {})[
+        : settings.ES_MAX_SIZE
+    ]
+    page_search = build_cms_search(query_str)
+
+    search_query = MultiSearch()
+    search_query = search_query.add(search=product_search._s)
+    search_query = search_query.add(page_search)
```
pi-sigma:
```suggestion
    search_query = search_query.add(search=product_search._s)
    search_query = search_query.add(search=page_search)
```
for uniformity

PR #1775, opened 2025-05-26
:sparkles: [#3291] Add single file-upload into Plan detailpage

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3291

Notes: Form used for Cases: `#document-upload`, 
Form used for Plans: was `DocumentForm` but that had an unnecessary name field, so creating new form `PlanFileUploadForm`

- [x] make upload FileInput component more generic and not just linked to Cases
- [x] move Disable-submit-button functionality out of cases into Form JS
- [x] create new uploadform for Plan-detailpage
- [x] change file_input tags so that it can be used for multiple/single file uploads and show proper plural/single button texts
- [x] remove `pages/plans/file.html`
- [x] expand PlanFileUploadView with HTMX/HTML template partial
- [x] make page-refresh update the file-list when file is uploaded
- [x] make errors/notifications appear correctly
- [x] move upload CSS from Cases to generic `FileInputForm` CSS
- [x] add Spinner 
- [x] change `src/open_inwoner/js/components/form/FileInput.js` so that it works properly for a Single fileinput (no adding new file list-items, only replacing selected file)
- [x] change Case Tests
- [x] add new Plans file-upload tests, add HTMX test (like `src/open_inwoner/cms/cases/tests/test_htmx.py`)


src/open_inwoner/plans/tests/test_htmx.py
```diff
@@ -0,0 +1,80 @@
...
+
+def live_reverse(name, kwargs=None, star=False):
+    """
+    Optionally append wildcard to Playwright's `wait_for_url` function to work with plan.uuid's.
+    """
+    url = reverse(name, kwargs=kwargs or {})
+    return f"{url}{'*' if star else ''}"
+
+
+class TestPlansHTMX:
```
pi-sigma: Couple of things:
- (minor but please follow the convention)
    ```suggestion
    class PlansHTMXTest:
    ```
    or
    ```
    class PlansHTMXTestCase
    ```
    (see https://docs.djangoproject.com/en/5.2/topics/testing/overview/ for the naming convention). After double-checking I saw that the documentation for `unittest` actually recommends the pattern you followed (i.e. `TestClassName`, cf. https://docs.python.org/2/library/unittest.html#basic-example). That's unfortunate; the Django convention makes more sense semantically and should be adopted for uniformity.

- (serious) This is not a test class at all because it doesn't inherit from either `unittest.TestCase` or `django.test.TestCase` or (this is what we need here) `open_inwoner.utils.tests.playwright.PlaywrightSyncLiveServerTestCase` (see https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/utils/tests/playwright.py). As a result, none of the tests are discovered and run. You get no error in CI because the class is simply ignored. If you try to run the test manually on your machine (`python src/manage.py test open_inwoner.plans.tests.test_htmx.TestPlansHTMX`) you will get an error. Inheriting from one of the aforementioned classes will mark the class as a test class and make it discoverable for the Django test runner.

src/open_inwoner/plans/tests/test_htmx.py
```diff
@@ -0,0 +1,80 @@
...
+    uploaded_bytes = b"fake pdf"
+    uploaded_filename = "upload_test.pdf"
+
+    def test_plan_file_upload_e2e(
+        self, live_server, browser, plan_factory, user, user_login_state
+    ):
```
pi-sigma: `browser` comes from from `PlaywrightSyncLiveServerTestCase` if you inherit from it. However, these are not values for test method params, but attributes on the test instance. Hence remove these params and inside the test: `self.browser.do_stuff()`. Similarly, the `user` should be created in the setup for the test class and accessed with `self.user`. See https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/accounts/tests/test_action_views.py for an example

src/open_inwoner/plans/tests/test_views.py
```diff
@@ -324,14 +325,22 @@ def test_plan_file_cannot_be_downloaded_by_non_contact(self):
...
         self.assertEqual(self.plan.documents.count(), 1)
 
+    def test_htmx_get_plan_file_add_form_renders_upload_form(self):
+        response = self.app.get(
+            self.file_add_url, user=self.user, extra_environ={"HTTP_HX_REQUEST": "true"}
+        )
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('id="document-upload"', response.text)
+        self.assertIn("Upload document", response.text)
```
pi-sigma: The last assert is true by accident. The actual text (as well as title) on the button is "Upload documenten". Better to target a specific element with `PyQuery` and then do `self.assertEqual(py_query_element.text, "Upload documenten")`. This assert will fail for "Upload document" as it should.

src/open_inwoner/templates/pages/plans/file_upload_form.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n solo_tags form_tags button_tags icon_tags %}
+
+{% get_solo 'openzaak.OpenZaakConfig' as openzaak_config %}
+
+<div id="form_upload">
+{% render_form id="document-upload" form=form method="POST" form_action=form_action enctype="multipart/form-data" submit_text=_("Bestand uploaden") extra_classes="file-input-form" %}
+    {% csrf_token %}
+    {# Name of the file is a hidden input #}
+    {% file_input form.file multiple=False primary_text=_("Upload document") max_upload_size=openzaak_config.max_upload_size allowed_file_extensions=openzaak_config.allowed_file_extensions %}
```
pi-sigma, line 9: `primary_text=_("Upload document")` is the default on the form tag, so not necessary here (removing it here means we won't have to stop by here if we decide to change the text)

pi-sigma (changes requested):

src/open_inwoner/plans/tests/test_htmx.py
jiromaykin: In `src/open_inwoner/cms/cases/tests/test_htmx.py` the class is called `CasesPlaywrightTests` and now I wonder if that follows the convention.

pi-sigma: I think it should always be `FooBarTest`, in the singular (since it's the name of a class, a *kind* of thing, and kind names are singular). But yeah we're not consistent here.

src/open_inwoner/templates/pages/plans/file_upload_form.html
jiromaykin: I will have to change the default then, because for a single upload it needs to be singular 'Document' and not the plural.

src/open_inwoner/components/templatetags/form_tags.py
```diff
@@ -338,19 +338,43 @@ def input(field, **kwargs):
...
+        or (_("Upload documenten") if multiple else _("Upload document")),
+        "empty_label": _("Sleep of selecteer bestanden")
+        if multiple
+        else _("Sleep of selecteer bestand"),
+        "selected_label": _("Selecteer meer bestanden")
+        if multiple
+        else _("Selecteer ander bestand"),
+        "selected_heading": _("Geselecteerde bestanden")
+        if multiple
+        else _("Geselecteerd bestand"),
```
stefrado: I think it is cleaner to use these statements into a variable
```python
empty_label = _("Sleep of selecteer bestanden") if multiple else _("Sleep of selecteer bestand")

return { "empty_label": empty_label }
```

src/open_inwoner/components/templates/components/Form/FileInput.html
```diff
@@ -4,22 +4,39 @@
...
+    {% endif %}
 
     <div class="file-list" aria-live="polite">
-        <div class="file-list-selection" hidden><h4 class="utrecht-heading-3 file-list-selection__heading">{% if multiple %}{% trans 'Geselecteerde bestanden' %}{% else %}{% trans 'Geselecteerd bestand' %}{% endif %}</h4></div>
+        <div class="file-list-selection" hidden>
+            <h4 class="utrecht-heading-3 file-list-selection__heading">
+                {{ selected_heading }}
+            </h4>
+        </div>
         <ul class="file-list__list" aria-live="polite" data-label-delete="{% trans 'Verwijderen' %}"></ul>
```
stefrado, line 40: When i add a file, on mobile breakpoints there is a horizontal scroll that is caused by this element.

src/open_inwoner/scss/components/Plan/PlanFile.scss
```diff
@@ -1,9 +1,7 @@
-#document-create {
-  /// Make required asterisk visible for this component
-  .label__label--required {
-    color: var(--color-red-notification);
-    padding-left: var(--spacing-tiny);
-    //make asterisks invisible by default, only make them visible if component has the caption
-    display: inline;
+.plan {
+  .file-input-form {
+    .form__actions--fullwidth {
```
stefrado: add these classes to one line, unnecessary nesting
```suggestion
    .plan .file-input-form .form__actions--fullwidth {
```

stefrado (changes requested):
Looking good! Some small css problems: 
- [x] There is horizontal scroll after i uploaded a file.
- [x] The container of the loading spinner is very big - this can be much smaller instead of this big white block.

src/open_inwoner/components/templates/components/Form/FileInput.html
jiromaykin: After some research: this issue was already there, and doen't happen in an ordinary file-list because those render an ellipsis, but _actually_... an ellipsis should not be there since this is text that should remain readable, for accessibility. 
So my proposal is I remove the ellipsis both in ordinary file lists as well as in the selected file list and set wrap to default breaking-to-new line, etc. For the file__data.

pi-sigma (approved):
Looks good to me. The playwright test is passing but has been flaky. I've increased the timeout for the upload element; let's see if this works consistently.

stefrado (approved):

PR #1779, opened 2025-05-27
feat: fix Service connected to CatalogusConfig on zgw_import_data

swrichards:
Previously, when syncing catalogus configs, we would match only on URLs but not check the connected Service. This means that certain CatalogusConfig objects may have gotten into a broken state, due to existing prior to introducing the service attribute on the CatalogusConfig model. In that case, there is a chance the field was never properly updated, or that the value was set to an incorrect service.

This is important becaus we use the Service to configure clients for all nested types (zaaktypes, resultaattypes, etc.), and we might get InvalidURL exceptions from ape_pie, because the api_root of the client does not match that of the resource being fetched.

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_zgw_imports.py
+++ b/src/open_inwoner/openzaak/tests/test_zgw_imports.py
@@ -5,6 +5,7 @@ import requests_mock
 from open_inwoner.openzaak.models import CatalogusConfig, OpenZaakConfig, ZaakTypeConfig
 from open_inwoner.openzaak.tests.factories import (
     CatalogusConfigFactory,
+    ServiceFactory,
     ZGWApiGroupConfigFactory,
 )
 from open_inwoner.openzaak.tests.helpers import generate_oas_component_cached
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openzaak/tests/test_zgw_imports.py
```diff
@@ -188,6 +190,23 @@ def test_import_catalogs(self, m):
...
+
+        import_catalog_configs()
+
+        existing_catalogus_with_same_url_but_different_service.refresh_from_db()
+        self.assertEqual(
+            existing_catalogus_with_same_url_but_different_service.service,
+            self.api_groups[0].ztc_service,
+            msg="Existing catalog with matching URL but differing Service should "
+            "be updated",
+        )
```
pi-sigma: Strictly speaking, we want to test that `existing_catalogus_with_same_url_but_different_service.service` is the same as the *original* `self.api_groups[0].ztc_service`. Your test is compatible with the scenario where `self.api_groups[0].ztc_service` was updated to equal `another_servicve`. The two would still be equal in this case. You could save the original `self.api_groups[0].ztc_service` in a variable before calling `import_catalog_configs` and use it in the assert to exclude this possibility. It's perhaps far-fetched since your code for updating the service is pretty straightforward...

pi-sigma (approved):

src/open_inwoner/openzaak/tests/test_zgw_imports.py
swrichards: I am not sure I understand the case: I specifically want to test here the case where an existing `CatalogusConfig` was created with Service A, and is now being synced with Service B (in which case `config.service` should be updated to Service B). 

I've simplified the test a bit to make this hopefully clearer, but let me know what I've missed.

pi-sigma: I understand, but the test does not distinguish between the case where `config.service` is updated to Service B (intended) and the case where Service B is updated to `config.service` (not intended) since the test just asserts that they're equal, which is true in both cases. As mentioned this is pretty far-fetched anyway.

swrichards: Right, I think those conditions would indeed be a bit far-fetched, and given that this is an internal job, we should be ok with just the basic coverage.

PR #1780, opened 2025-06-03
fix: only show complete CMS page search results

swrichards:
The description field can be None, in which case we should show an empty string, and we do not want to show hits without a title and url, because there'd be nothing to show the user.

### DIFF
--- a/src/open_inwoner/search/results.py
+++ b/src/open_inwoner/search/results.py
@@ -136,8 +136,9 @@ class CMSPageSearchResult(SearchResult):
 
     def get_generic_hits(self):
         return [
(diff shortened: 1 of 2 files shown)


pi-sigma (approved):
Could use an extra test scenario like:

```
(
    "baz",
    self.bar_page,
    GenericHit(title="", summary="", link="/bar-page/"),
),
```
which should not be included in the response of `multi_search`.

swrichards:
> Could use an extra test scenario like:
> 
> ```
> (
>     "baz",
>     self.bar_page,
>     GenericHit(title="", summary="", link="/bar-page/"),
> ),
> ```
> 
> which should not be included in the response of `multi_search`.

Good call! Added

PR #1786, opened 2025-06-06
[#3289] Display second-to-last login on profile page

linssen814:
Added a 'prev_login' field on User that gets updated to store the second-to-last login date/time. Used to show the previous login moment on the profile page.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3289


src/open_inwoner/accounts/models.py
```diff
@@ -604,6 +612,28 @@ def get_group_managed_categories(self):
 
         return Category.objects.filter(access_groups__user=self)
 
+    @staticmethod
+    def save_prev_login(sender, instance, update_fields, **kwargs):
```
swrichards: This is not a bad pattern (though I would make it a `@classmethod`, but it's a bit inconsistent with how we've organized the other signals. I would move this to a regular function in `accounts/signals.py`.

src/open_inwoner/accounts/tests/test_prev_login.py
```diff
@@ -0,0 +1,33 @@
...
+        # Should not be updated to a non-null value
+        # yet, since this was the first login
+        self.assertIsNone(self.user.prev_login)
+
+        old_last_login = self.user.last_login
+
+        self.user.last_login = datetime(2025, 8, 14, 7, 0, 0, tzinfo=pytz.UTC)
+        self.user.save(update_fields=["last_login"])
+
+        self.assertEqual(self.user.prev_login, old_last_login)
```
swrichards: You'll want to add a test that ensures the logic is applied over the course of a regular login: you're now directly calling `user.save()`, but we want to verify that this will also happen as part of a regular login flow. Otherwise we haven't validated that the built-in login flows actually use `update_fields` in the way intended.

src/open_inwoner/accounts/models.py
```diff
@@ -313,6 +314,13 @@ def has_verified_email(self):
             "Indicates if the user wants to receive notifications for updates concerning plans and actions."
         ),
     )
+    prev_login = models.DateTimeField(
```
swrichards: Nothing wrong with verbose names!

```suggestion
    previous_login = models.DateTimeField(
```

src/open_inwoner/accounts/models.py
```diff
@@ -313,6 +314,13 @@ def has_verified_email(self):
...
+        _("prev login"),
+        blank=True,
+        null=True,
+        help_text="The previous login time for the user. To replace 'last_login' for displaying the previous login.",
```
swrichards:
```suggestion
        help_text=_("The previous login time for the user, which will be shown upon login when `last_login` has been updated to the current timestamp."),
```

src/open_inwoner/accounts/models.py
```diff
@@ -604,6 +612,28 @@ def get_group_managed_categories(self):
...
+
+        if update_fields and "last_login" in update_fields:
+            old_instance = User.objects.get(id=instance.id)
+            instance.prev_login = old_instance.last_login
+
+    def save(
+        self, force_insert=False, force_update=False, using=None, update_fields=None
+    ):
+        if update_fields and "last_login" in update_fields:
+            update_fields.append("prev_login")
```
swrichards: Why is this needed?

swrichards (changes requested):

src/open_inwoner/accounts/models.py
jiromaykin: Yeah, texts need to be translatable.

src/open_inwoner/accounts/models.py
linssen814: Updating the `prev_login` property in the instance passed to `save_prev_login` doesn't trigger it to recognize prev_login as one of the fields that need to be updated, so I had to manually add prev_login to `update_fields` so that it doesn't overlook that field as it updates the object. It doesn't look very idiomatic but I'm not sure how else to do it.

jiromaykin (approved):
Works on my machine :-) and later: remember to 'squash' all commits before merging.

src/open_inwoner/accounts/migrations/0090_remove_user_prev_login_user_previous_login.py
pi-sigma, line 1: You add `prev_login` in the previous migration, then remove it here and add `previous_login`. That's one superfluous migration. Just add the `previous_migration` field.

src/open_inwoner/accounts/models.py
```diff
@@ -313,6 +313,15 @@ def has_verified_email(self):
             "Indicates if the user wants to receive notifications for updates concerning plans and actions."
         ),
     )
+    previous_login = models.DateTimeField(
+        _("previous login"),
```
```suggestion
        _("Previous Login"),
```
(for uniformity; see the other verbose fields)

src/open_inwoner/accounts/signals.py
```diff
@@ -170,3 +171,13 @@ def log_user_login(sender, user, request, *args, **kwargs):
 def log_user_logout(sender, user, request, *args, **kwargs):
     if user:
         user_action(request, user, MESSAGE_TYPE["logout"])
+
+
+@receiver(pre_save)
+def save_previous_login(sender, instance, update_fields, **kwargs):
+    "Saves previous 'last_login' value for use in"
+    "correctly displaying the last login datetime"
```
pi-sigma:
```suggestion
def save_previous_login(sender, instance, update_fields, **kwargs):
    """Save previous `last_login` value"""
```

pi-sigma (changes requested):

pi-sigma (dismissed):
One last thing: please squash the commits so that only one remains. Otherwise good to go!

src/open_inwoner/accounts/signals.py
```diff
@@ -170,3 +171,12 @@ def log_user_login(sender, user, request, *args, **kwargs):
...
     if user:
         user_action(request, user, MESSAGE_TYPE["logout"])
+
+
+@receiver(pre_save)
+def save_previous_login(sender, instance, update_fields, **kwargs):
+    """Save previous `last_login` value"""
+
+    if update_fields and "last_login" in update_fields:
+        old_instance = User.objects.get(id=instance.id)
```
swrichards, line 181: Could this ever be called on an unsaved `User` (in which case, there would be nice `.id` and we'd get an `AttributeError` or `NotFound` exception).

src/open_inwoner/accounts/models.py
```diff
@@ -604,6 +613,16 @@ def get_group_managed_categories(self):
 
         return Category.objects.filter(access_groups__user=self)
 
+    def save(
+        self, force_insert=False, force_update=False, using=None, update_fields=None
+    ):
+        if update_fields and "last_login" in update_fields:
+            update_fields.append("previous_login")
```
swrichards, line 620: I _think_ `update_fields` in the `pre_save` handler is mutable, meaning you could do this in `save_previous_login` and avoid having to override this at all (but you'd need to double check).

src/open_inwoner/accounts/models.py
```diff
@@ -313,6 +313,15 @@ def has_verified_email(self):
             "Indicates if the user wants to receive notifications for updates concerning plans and actions."
         ),
     )
+    previous_login = models.DateTimeField(
```
swrichards, line 316: Let's add this guy as one of the read_only fields on the admin.

src/open_inwoner/accounts/tests/test_previous_login.py
```diff
@@ -0,0 +1,58 @@
...
+        self.assertIsNone(self.user.previous_login)
+
+        old_last_login = self.user.last_login
+
+        self.user.last_login = datetime(2025, 8, 14, 7, 0, 0, tzinfo=pytz.UTC)
+        self.user.save(update_fields=["last_login"])
+
+        self.assertEqual(self.user.previous_login, old_last_login)
+
+    def test_previous_login_gets_updated_on_login(self):
```
swrichards, line 39: As per the discussion above: maybe add a test where you create an unsaved `User` instance (`pk=None1), `save()` it,. and ensure we get the behavior we want too, also validating it doesn't crash on an unsaved instance.

Similar to the tests you had earlier iirc, but it'd be good to test both paths (via `client.login` for the behavior, and via a `model.save()` to ensure it won't break anything else).

swrichards (changes requested):

src/open_inwoner/accounts/models.py
linssen814: It actually gets converted into a 'frozenset' before being passed to the pre_save so unfortunately it's not mutable. I do wish there was some other way around.

swrichards: Clear, thanks for checking.

src/open_inwoner/accounts/signals.py
linssen814: I actually did have a try/catch there before but removed it since 'update_fields' is really only populated on an 'update' and not upon 'creation'. But I will add a test like you suggested.

src/open_inwoner/accounts/tests/test_previous_login.py
```diff
@@ -0,0 +1,81 @@
...
+            new_user = User(first_name="Foo", last_name="Bar")
+            new_user.save()
+        except Exception as e:
+            self.fail(f"Exception raised on trying to save unsaved user: {e}")
+
+    def test_unsaved_user_does_not_break_previous_login_signal_via_account_creation(
+        self,
+    ):
+        try:
+            new_user = User.objects.create_user(
```
swrichards, line 71: This is not an unsaved user! I would just rename the test as per above.

src/open_inwoner/accounts/tests/test_previous_login.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual(self.user.previous_login, old_last_login)
+
+    def test_unsaved_user_does_not_break_previous_login_signal(self):
+        try:
+            new_user = User(first_name="Foo", last_name="Bar")
+            new_user.save()
+        except Exception as e:
+            self.fail(f"Exception raised on trying to save unsaved user: {e}")
+
+    def test_unsaved_user_does_not_break_previous_login_signal_via_account_creation(
```
swrichards:
```suggestion
    def test_programmatically_created_user_does_not_break_previous_login_signal(
```

swrichards (changes requested):

src/open_inwoner/accounts/tests/test_previous_login.py
linssen814: I actually assumed this is what you meant since you can't call 'client.login' on unsaved users, but I've adjusted it now.

swrichards (approved):

PR #1790, opened 2025-06-12
[#3296] E-herkenning switch between branches

jiromaykin:
Userstory: https://taiga.maykinmedia.nl/project/open-inwoner/us/3287

This PR is only about the switch functionality, not about the Search dropdown design-flow.
The backbutton should not be visible if a Vestiging has not been selected yet.
After a Vestiging is selected, it has to be checked in the selection screen so user can see which one is selected.
Switch button should be visible for entire company as well as individual branches (⚠️  _unless_ they are `is_branch_restricted`)

So only:
New button in Profile:  
https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6641-5628&t=FmveKziiPJO47Qd8-1 

Link in Nav menu:  
https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6908-12854&t=ytrmyhScRaLrhz8Y-1

A conditional backbutton on the KVK branch selection screen: 
https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6908-12899&t=FmveKziiPJO47Qd8-1


src/open_inwoner/accounts/context_processors.py
```diff
@@ -0,0 +1,18 @@
...
+
+from open_inwoner.accounts.eherkenning_session import EHerkenningSessionContext
+
+
+def set_branch_selection_state(request: HttpRequest):
+    context = {}
+    try:
+        ctx = EHerkenningSessionContext(request)
+        context["is_branch_restricted"] = ctx.is_branch_restricted
+        context["is_initial_Branch_selection_done"] = (
```
stefrado: Is the uppercase B in `is_initial_Branch_selection_done` on purpose?

src/open_inwoner/components/templates/components/Header/Header.html
```diff
@@ -82,9 +82,14 @@
...
                         <ul class="header__list">
+                            {% if is_eherkenning_session and not is_branch_restricted %}
+                            <li class="header__list-item">
+                                {% link text=_("Wissel van vestiging") href="kvk:branches" icon="autorenew" icon_position="before" primary=True icon_outlined=True %}
+                            </li>
+                            {% endif %}
                             <li class="header__list-item">
                                 {% trans "Logout" as logout %}
-                                {% link text=logout href=request.user.get_logout_url icon="login" icon_position="before" primary=True %}
+                                {% link text=logout href=request.user.get_logout_url icon="login" icon_position="before" primary=True icon_outlined=True %}
```
stefrado:
```suggestion
                                {% link text=_("Logout") href=request.user.get_logout_url icon="login" icon_position="before" primary=True icon_outlined=True %}
```

src/open_inwoner/components/templates/components/Header/NavigationAuthenticated.html
```diff
@@ -30,6 +30,11 @@
...
                     {% endif %}
 
+                    {% if is_eherkenning_session and not is_branch_restricted %}
+                    <li class="header__list-item">
+                        {% link text=_("Wissel van vestiging") href="kvk:branches" icon="autorenew" icon_position="before" primary=True icon_outlined=True %}
+                    </li>
+                    {% endif %}
                     <li class="header__list-item">
                         {% trans "Logout" as logout %}
                         {% link text=logout href=request.user.get_logout_url icon="logout" icon_position="before" primary=True %}
```
stefrado: Just to be consistent...

```suggestion
                        {% link text=_("Logout") href=request.user.get_logout_url icon="logout" icon_position="before" primary=True %}
```

stefrado (approved):

src/open_inwoner/accounts/context_processors.py
jiromaykin: Probably not (this is part of Sidney's commit).

src/open_inwoner/accounts/context_processors.py
```diff
@@ -0,0 +1,18 @@
...
+            ctx.is_initial_branch_selection_done
+        )
+        context["is_eherkenning_session"] = True
+    except ValueError:
```
pi-sigma: This is not raise in the `EHerkenningSessionContext.__init__`, thus if you log in as staff user via the admin and follow `Bekijk site`, you will see the option to change the vestiging in the dropdown (attempt to follow yiels `unauthorized`).

src/open_inwoner/accounts/context_processors.py
```diff
@@ -0,0 +1,18 @@
+from django.http import HttpRequest
+
+from open_inwoner.accounts.eherkenning_session import EHerkenningSessionContext
+
+
+def set_branch_selection_state(request: HttpRequest):
```
pi-sigma: In general, I'd like to avoid using context processors as these are injected into every request context, when it doesn't make sense to do this check for staff users.

This is a good use case for template tags. Create `/open_inwoner/core/templatetags/__init__.py` and `/open_inwoner/core/templatetags/custom_session_tags.py`. In the latter create a template tag:

```
from django import template

from open_inwoner.accounts.eherkenning_session import EHerkenningSessionContext

register = template.Library()


@register.simple_tag(takes_context=True)
def session_is_branch_restricted(context) -> bool:
    eherkenning_ctx = EHerkenningSessionContext(context["request"])
    return eherkenning_ctx.is_branch_restricted()
```

Load the tag in your templates and then you can do:

```
{% if user.is_eherkenning_user and not session_is_branch_restricted %}
   do_stuff()
{% endif %}
```

This will also prevent staff users from seeing the vestiging switch option.

src/open_inwoner/conf/base.py
```diff
@@ -312,6 +312,7 @@
                 "sekizai.context_processors.sekizai",
                 "cms.context_processors.cms_settings",
                 "django.template.context_processors.i18n",
+                "open_inwoner.accounts.context_processors.set_branch_selection_state",
```
pi-sigma: Better to avoid context processors if possible (see related comment)

pi-sigma (changes requested):

src/open_inwoner/conf/base.py
jiromaykin: I think Sidney added these because the switch must be visible in the navigation menu (on all pages, so basically all requests).

pi-sigma: That's strange. If you load the template tag in the navigation template (as I suggested), then it will be visible everywhere where it's needed. But it won't be visible for admin users. With the current implementation, if you log in as admin and the check the template context (with the debug toolbar), you'll see that `is_eherkenning_session = True`, which is obviously false for staff users.

Besides, in your context processor you have e.g. `context["is_branch_restricted"] = ctx.is_branch_restricted`, which puts the function itself into the context. You'd need: `context["is_branch_restricted"] = ctx.is_branch_restricted()`, which gives the return value of the function call.

If @swrichards suggested the approach with a context processor, then we should hear him out when he's back. Perhaps there's a reason for this approach that I'm missing.

jiromaykin: Sidney implemented this very quickly on the go, so I suspect your approach is better and probably more correct!

src/open_inwoner/accounts/context_processors.py
jiromaykin: I think you are right - and we definitely do not want this switch button to become visible for staff users.

pi-sigma (approved):

PR #1792, opened 2025-06-13
[#2497] Add config step that adds default staff user

linssen814:
Added a django-setup-configuration step that adds a default staff user.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/2497

Note that this is dependent on: https://github.com/maykinmedia/django-setup-configuration/pull/79

EDIT: On suggestion of @swrichards , I made this implementation separately from the one on django-setup-configuration.

### DIFF
--- a/src/open_inwoner/configurations/bootstrap/default_users.py
+++ b/src/open_inwoner/configurations/bootstrap/default_users.py
@@ -1,0 +1,62 @@
+import warnings
+
+from django_setup_configuration.configuration import BaseConfigurationStep
+from django_setup_configuration.models import ConfigurationModel
+
+from open_inwoner.accounts.models import User
+
+
+class UserConfigurationItem(ConfigurationModel):
+    """
+    Configuration model for a setting standard default users.
+    """
+
+    class Meta:
+        django_model_refs = {
+            User: (
+                "email",
+                "password",
+                "is_staff",
+                "is_superuser",
+            )
+        }
+
+
+class UserConfigurationModel(ConfigurationModel):
+    users: list[UserConfigurationItem]
+
+
+class UserConfigurationStep(BaseConfigurationStep):
+    """
+    Creates or updates a one or more default users based on
+    YAML settings. Note that a provided password
+    will only be used if the user does not exist yet.
+    """
+
+    verbose_name = "User Configuration Step"
+    enable_setting = "default_user_configuration_enable"
+    config_model = UserConfigurationModel
+    namespace = "default_user_configuration_config"
+
+    def execute(self, model):
+        for user_item in model.users:
+            user, created = User.objects.update_or_create(
+                email=user_item.email,
+                defaults={
+                    "email": user_item.email,
+                    "is_staff": user_item.is_staff,
+                    "is_superuser": user_item.is_superuser,
+                },
+            )
+
+            if created:
+                user.set_password(user_item.password)
+
+            user.save()
+
+            if user.check_password(user_item.password):
+                warnings.warn(
+                    "\nThe password for the automatically created "
+                    f"user '{user_item.email}' is currently set to a hardcoded default. "
+                    "Make sure to change the password in the admin panel.\n\n"
+                )
--- a/docker/setup_configuration/data.yaml
+++ b/docker/setup_configuration/data.yaml
@@ -140,3 +140,12 @@ sites_config:
   items:
     - domain: test-site
       name: The Django Sites default
+
+# Set default (staff) users
+default_user_configuration_enable: true
+default_user_configuration_config:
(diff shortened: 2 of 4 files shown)


src/open_inwoner/configurations/bootstrap/staff.py
```diff
@@ -0,0 +1,49 @@
...
+    enable_setting = "default_staff_user_enable"
+    config_model = StaffUserConfigurationModel
+    namespace = "default_staff_user_config"
+
+    def execute(self, model):
+        user, created = User.objects.update_or_create(
+            email=model.email, is_staff=model.is_staff, is_superuser=model.is_superuser
+        )
+
+        if created or not user.check_password(model.password):
```
pi-sigma: Why overwrite the password for existing staff users? If the staff changed their password from "change_me" (the default via the YAML file) to "foobar",  it would be surprising if the password was reset after the next run of `python src/manage.py setup_configuration`. This assumes that multiple configuration runs on an already existing instance of OIP (or other platforms/websites) make sense (e.g. for experimenting with different values), and that we'd want to keep already existing staff users in place. I would expect `if created: user.set_password(model.password)` and then an additional check further down (see below)

src/open_inwoner/configurations/bootstrap/staff.py
```diff
@@ -0,0 +1,49 @@
...
+            user.set_password(model.password)
+
+        user.save()
+        print("Updated/created default staff user account.")
```
pi-sigma: Superfluous (the success message emitted by the library is sufficient)

src/open_inwoner/configurations/bootstrap/staff.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if model.password == "change_me":  # nosec
+            print(
+                "[WARNING] The password is currently set to "
+                "a hardcoded default. Make sure to change the "
+                "password in the admin panel."
+            )
```
pi-sigma: Here you can check if the automatically created admin user has the default password and emit a warning. Instead of printing use `stdout` and the colors from `django.core.management.color` for consistency with the output of the library

```
# top of module obv
import sys
from django.core.management.color import make_style

if user.check_password(model.password):
    message = make_style().WARNING(
        "\n[WARNING] The password for the automatically created admin user is currently set 
        "to a hardcoded default. Make sure to change the password in the admin panel.\n\n")
    sys.stdout.write(message)
```

pi-sigma (changes requested):
Ideally this should be implemented in the library itself (https://github.com/maykinmedia/django-setup-configuration) so that it can be used for other projects as well.

src/open_inwoner/configurations/tests/test_default_staff_user.py
```diff
@@ -0,0 +1,47 @@
...
+        execute_single_step(UserConfigurationStep, object_source=config)
+
+        self.assertEqual(User.objects.count(), 1)
+
+        user = User.objects.get(email=email)
+        self.assertTrue(user.check_password(password))
+        self.assertTrue(user.is_staff)
+        self.assertTrue(user.is_superuser)
+
+    def test_verify_current_configuration(self):
```
swrichards: Let's add one more test to assert that existing users are updated, not  created

src/open_inwoner/configurations/bootstrap/default_users.py
```diff
@@ -0,0 +1,68 @@
...
+    config_model = UserConfigurationModel
+    namespace = "default_user_configuration_config"
+
+    def execute(self, model):
+        for user_item in model.users:
+            # user, created = User.objects.update_or_create(
+            #     email=user_item.email,
+            #     is_staff=user_item.is_staff,
+            #     is_superuser=user_item.is_superuser,
+            # )
```
swrichards:
```suggestion
```

src/open_inwoner/configurations/tests/test_default_staff_user.py
```diff
@@ -0,0 +1,85 @@
+from django.conf import settings
+from django.test import TestCase
+
+from django_setup_configuration.test_utils import execute_single_step
+
+from open_inwoner.accounts.models import User
+from open_inwoner.configurations.bootstrap.default_users import UserConfigurationStep
+
+
+class TestDefaultStaffUserStep(TestCase):
```
swrichards: Nitpick: no longer a `staff` user (also in the filename).

swrichards (approved):
Two nitpicks, but otherwise gtg.

swrichards (approved):

pi-sigma (approved):

PR #1793, opened 2025-06-17
[#3303] Disable performance transactions in Sentry

pi-sigma:
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3303


alextreme (approved):
Thanks!

PR #1795, opened 2025-06-18
[#3288] Migrate flatpages to CMS

pi-sigma:
- [x] add CMS plugin and pages for displaying content of flatpages
- [x] add data migration to move content of flatpages to CMS pages 
- [x] update admin index fixture (remove flatpages)
- [x] update sitemap (replace refs to flatpages with refs to cms pages)

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3288


src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
swrichards, line 1: @pi-sigma a note that adding the CMS pages will trigger the django elasticsearch handler to immediately index these pages, which means the migration will fail if elasticsearch is unavailable. We should look into disable the relevant signal in the `create_cms_pages` handlers.

src/open_inwoner/conf/fixtures/django-admin-index.json
```diff
@@ -201,7 +201,9 @@
     "model": "admin_index.appgroup",
     "fields": {
         "order": 8,
-        "translations": {},
+        "translations": {
+            "nl": "Bestand"
+        },
         "name": "Configuratie",
```
swrichards: This looks a bit strange to me, do we need it? (Bestand is now the name of the menu root)

swrichards:
@pi-sigma there was a question during the sprint review regarding the contactformulier page, whether the surrounding text there is also editable. I wasn't sure whether this related to this PR or is a separate issue, did you look into this at all?

src/open_inwoner/conf/fixtures/django-admin-index.json
pi-sigma: Artifact from developing/debugging (I had to put something in there to save the application groups). Removed it.

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
```diff
@@ -0,0 +1,92 @@
+# Generated by Django 4.2.20 on 2025-06-19 08:58
+
+from django.db import migrations
+
+# from django.contrib.flatpages.models import FlatPage
```
swrichards:
```suggestion
```

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
```diff
@@ -0,0 +1,92 @@
...
+from django.utils.text import slugify
+
+from cms.api import add_plugin, create_page
+
+from ..cms_plugins import CMSFlatPagePlugin
+
+
+def create_cms_pages(apps, schema_editor):
```
swrichards, line 11: Were you able to check for disabling the signals for auto ES indexing here, or was that not feasible?

src/open_inwoner/configurations/models.py
```diff
@@ -664,8 +663,11 @@ def get_accent_color(self):
         return hex_to_hsl(self.accent_color)
 
     @property
-    def get_ordered_flatpages(self):
-        return self.flatpages.order_by("ordered_flatpages")
+    def get_ordered_cms_pages(self):
```
swrichards, line 665:
```suggestion
    def ordered_cms_pages(self):
```
The imperative verb suggests it should be a method: I would either rename or remove `@property` (I am aware this pattern was introduced earlier, but let's try and break it :).

src/open_inwoner/cms/footer/cms_plugins.py
```diff
@@ -1,21 +1,69 @@
...
+@plugin_pool.register_plugin
+class CMSFlatPagePlugin(CMSPluginBase):
+    name = _("CMS Flatpage Plugin")
+    model = CMSFlatPageModel
+    form = CMSFlatPageModelForm
+    render_template = "cms/cms_flatpage.html"
+
+    def render(self, context, instance, placeholder):
+        cms_plugin_placeholder = instance.placeholder
+        cms_page = Page.objects.filter(placeholders=cms_plugin_placeholder).first()
```
swrichards: Should we log an error if there's multiple objects?

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
```diff
@@ -0,0 +1,92 @@
...
+            cms_page=cms_page,
+            order=order,
+        )
+
+
+def delete_flatpages(apps, schema_editor):
+    FlatPage = apps.get_model("flatpages", "FlatPage")
+
+    for flatpage in FlatPage.objects.all():
+        flatpage.delete()
```
swrichards:
```suggestion
    FlatPage.objects.all().delete()
```

src/open_inwoner/cms/footer/models.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CMSFlatPageModel(CMSPlugin):
+    title = models.CharField(
+        verbose_name=_("Title"),
+        max_length=255,
+        blank=True,
+        help_text=_("The title of the page"),
+    )
+    content = models.TextField(
```
swrichards, line 14: To be clear, this is rich content?

swrichards (changes requested):

src/open_inwoner/configurations/models.py
pi-sigma: Was tempted to do this but shied away for fear of making things more confusing. If we're on the same page, though, let's go for it.

src/open_inwoner/cms/footer/cms_plugins.py
pi-sigma: I added a warning. Shouldn't happen but...

src/open_inwoner/cms/footer/models.py
pi-sigma: Yes but it's considered safe because only exposed to staff users.

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
pi-sigma: Which signals do you mean? We have a task for rebuilding search the index on a regular basis; I wasn't aware of signals that trigger rebuilding.

swrichards: Since we added CMS pages to the search index, adding nen CMS pages will trigger one of these:

https://django-elasticsearch-dsl.readthedocs.io/en/latest/es_index.html#signals

The idea is to automatically keep the indexes in sync, but mostly this seems to be a nuisance (we use the Celery job for that).

swrichards: Scratch that, this is for post-index, let me double check what happened on my machine and if I can reproduce.

swrichards (approved):

PR #1801, opened 2025-06-24
[#3316] Limited CSRF cookie age to end of session

linssen814:
Added a setting that limits the CSRF token until the end of the session.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3316?milestone=535

One of the recommendations was that CSRF tokens should rotate on every new session, but this is already done by default. See:
https://docs.djangoproject.com/en/5.2/ref/csrf/#why-might-a-user-encounter-a-csrf-validation-failure-after-logging-in

### DIFF
--- a/src/open_inwoner/conf/base.py
+++ b/src/open_inwoner/conf/base.py
@@ -533,6 +533,7 @@ SESSION_COOKIE_SECURE = IS_HTTPS
 SESSION_COOKIE_HTTPONLY = True
 
 CSRF_COOKIE_SECURE = IS_HTTPS
+CSRF_USE_SESSIONS = True  # Because cookie csrf makes pen testers uncomfortable
 CSRF_FAILURE_VIEW = "open_inwoner.accounts.views.csrf_failure"
 
 if IS_HTTPS:


src/open_inwoner/conf/base.py
```diff
@@ -533,6 +533,7 @@
 SESSION_COOKIE_HTTPONLY = True
 
 CSRF_COOKIE_SECURE = IS_HTTPS
+CSRF_COOKIE_AGE = None  # Expires at end of session
```
swrichards: If I [understand the docs correctly](https://docs.djangoproject.com/en/5.2/ref/settings/#csrf-cookie-age), setting this to `None` is equivalent to setting [`CSRF_USE_SESSIONS`](https://docs.djangoproject.com/en/5.2/ref/settings/#csrf-use-sessions) to `True`. Can you confirm and, if so, I think we should make this value explicit.

It's also worth looking into this note:

> Since the [default error views](https://docs.djangoproject.com/en/5.2/ref/views/#error-views) require the CSRF token, [SessionMiddleware](https://docs.djangoproject.com/en/5.2/ref/middleware/#django.contrib.sessions.middleware.SessionMiddleware) must appear in [MIDDLEWARE](https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-MIDDLEWARE) before any middleware that may raise an exception to trigger an error view (such as [PermissionDenied](https://docs.djangoproject.com/en/5.2/ref/exceptions/#django.core.exceptions.PermissionDenied)) if you’re using CSRF_USE_SESSIONS. See [Middleware ordering](https://docs.djangoproject.com/en/5.2/ref/middleware/#middleware-ordering).

> Change this setting to None to use session-based CSRF cookies, which keep the cookies in-memory instead of on persistent storage.

linssen814: That makes sense. It's also somewhat more secure to safe it in the session with regards to XSS attacks. After looking into it a bit, one caveat could be that if we ever implement (or have implemented?) SSO with different applications on the same domain, then having the CSRF token stored in the session means that other applications would also need to implement the same session-based CSRF method instead of the cookie-based method, but according to django most webframeworks do this in the first place anyway.

Regarding the ordering of the `MIDDLEWARE` setting, I gather from the [official documentation](https://docs.djangoproject.com/en/5.2/ref/middleware/#middleware-ordering) as well as the [sessionprofile documentation](https://django-sessionprofile.readthedocs.io/en/latest/quickstart.html#installation) that the correct order is:
```
    "django.middleware.security.SecurityMiddleware",
    "sessionprofile.middleware.SessionProfileMiddleware",
    "django.contrib.sessions.middleware.SessionMiddleware",
```
at the top.

But I'm not sure if that takes into account the `CSRF_USE_SESSIONS = True` setting.  I'd presume it does.

swrichards: Can confirm this works as expected on our test instance (no more `csrftoken` in the cookie). Note there was a typo: `CSRF_USE_SESSIONS` (plural, you had `CSRF_USE_SESSION`).

swrichards (approved):

PR #1802, opened 2025-06-25
fix: properly use templatetag for branch changing ui conditions

swrichards:
Fixes: [#3320](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3320)


jiromaykin (approved):
When testing with KVK nr. 90000749 which has zero branches, we do see new bugs but this one is already a great improvement.

PR #1812, opened 2025-06-30
:sparkles: #3310 Implement react and web-components

stefrado:
### User story:
[#3353](https://taiga.maykinmedia.nl/project/open-inwoner/task/3353)

### Description:
This PR contains the implementation of react and web-components and a update of the current pipeline.
The major changes of this PR contain
- Webpack bundler for legacy code
- Vite bundler for TS react code (should also bundle webpack bundled code).
- Change karma with vitest for the frontend testsuit.
- Add some documentation for frontend devs.
- Implement i18n for the react components (have own build scripts `npm run makemessages`).
- Write some demo's `http:localhost:8000/react-demo/` & `http:localhost:8000/webcomponents-demo/`

> [!NOTE]
> Make sure to run `npm install` before testing -> `--legacy-peer-deps` might not be needed.

- [x] Implement react. 
- [x] Setup the pipeline for react. 
  - [x] Implement vite typescript compiler.
- [x] Create some demo components. 
- [x] Implement web-components and write a demo for it. 
- [x] Write some FRONTEND.md documentation as a guideline for future features. 
- [ ] Remove the legacy (webpack) compiler and use vite for bundling all assets.


vite.config.js
```diff
@@ -0,0 +1,83 @@
...
+      },
+    }),
+  ],
+
+  build: {
+    outDir: path.resolve(__dirname, paths.jsDir),
+    emptyOutDir: false, // Matches Webpack's behavior (does not wipe output)
+    sourcemap: useSourceMap,
+    minify: isProduction,
+    manifest: 'manifest.json',
```
stefrado: manifest.json might not be needed.

vite.config.js
```diff
@@ -0,0 +1,83 @@
...
+    alias: {
+      '@': path.resolve(__dirname, 'src/open_inwoner/react'),
+    },
+    extensions: ['.js', '.jsx', '.ts', '.tsx', '.json'],
+  },
+
+  // Match Webpack's mode behavior
+  mode: isProduction ? 'production' : 'development',
+
+  server: {
```
stefrado: Should be removed.

webpack.config.js
```diff
@@ -84,15 +85,14 @@ module.exports = {
...
-          },
+          // Use .babelrc
         },
       },
 
-      // .scss
+      // .scss (excluding CKEditor files)
       {
         test: /src\/.*.(sa|sc|c)ss$/,
+        exclude: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
```
stefrado: Might remove this as well.

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+        /* Bundler mode */
+        "moduleResolution": "node",
+        "allowImportingTsExtensions": false,
+        "isolatedModules": true,
+        "moduleDetection": "force",
+        "noEmit": false,
+        "jsx": "react-jsx",
+        "baseUrl": ".",
+        "paths": {
+            "@/*": ["./src/open_inwoner/react/*"]
```
stefrado: Might include some more paths to make this work inside `/js/` and `/webcomponents/`

src/open_inwoner/react/main.ts
```diff
@@ -0,0 +1,5 @@
+import Router from './router/router'
+
+console.log('Vite is configured to render this inside open-inwoner-react.js')
```
stefrado: 🔥 Remove comment

stefrado (commented):
Checked the code myself :)

src/open_inwoner/templates/demo/WebComponentsDemo.html
```diff
@@ -0,0 +1,502 @@
...
+&lt;/utrecht-checkbox&gt;</pre
+              >
+            </div>
+          </div>
+        </div>
+
+        <utrecht-heading-3>Typography & Content</utrecht-heading-3>
+        <div class="demo-grid">
+          <div class="demo-item">
+            <utrecht-heading-4>Headings</utrecht-heading-4>
```
jiromaykin, line 366: The headings are one of the few components that we actually have values for, but strangely enough these do not inherit our values (coming from our dist folder `@import '~@open-inwoner/design-tokens/dist/css/index.css';`

stefrado: Somehow the stylesheet Utrecht design tokens are used instead of the oip tokens. I'll look into it.

src/open_inwoner/react/i18n/locales/nl.json
```diff
@@ -0,0 +1,7 @@
+{
+  "Q6fIiw": {
+    "defaultMessage": "This is the current count",
+    "description": "Counter",
+    "originalDefault": "This is the current count"
```
jiromaykin, line 5: That is not Dutch :-)

Also when I run the project and build etc. I get some extra files, and not sure if we should add these to Gitignore?:

Screenshot 2025-07-07 at 09 51 31

src/open_inwoner/react/components/Count/Count.scss
swrichards, line 1: General comment about the paths: can we put `react` under `src/open_inwoner/js`, to avoid confusion with Django packages and to keep the JS assets in a single hierarchy?

package.json
```diff
@@ -7,13 +7,24 @@
     "doc": "doc"
   },
   "scripts": {
-    "build": "npm run collect && npm run bundle",
+    "build": "npm run collect && npm run bundle && npm run bundle:react",
     "bundle": "webpack",
```
swrichards: I would split this out into `bundle:webpack` and have `bundle` do both, so the various deploy and build pipelines can still rely on just running `npm run bundle`.

package.json
```diff
@@ -7,13 +7,24 @@
...
-    "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss -c",
-    "fix-linting": "prettier src/open_inwoner/js src/open_inwoner/scss -w",
-    "watch": "npm run collect && webpack -w"
+    "test": "vitest",
+    "test:watch": "vitest --watch",
+    "test:ui": "vitest --ui",
+    "test:coverage": "vitest --coverage",
+    "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react -c",
+    "fix-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react -w ",
+    "watch": "npm run collect && webpack -w",
```
swrichards: Same comment as above: perhaps `npm run watch` should run _all_ watch processes, webpack and react?

vite.config.js
```diff
@@ -0,0 +1,83 @@
...
+              return `styles-${match[1]}`
+            }
+          }
+        },
+      },
+    },
+  },
+
+  // Configure base to match Webpack's publicPath
+  base: '/static/bundles/',
```
jiromaykin, line 60: Also a possibility: Let Webpack output to /static/bundles/webpack/and let Vite output to /static/bundles/vite/

package.json
```diff
@@ -44,6 +55,8 @@
     "@joeattardi/emoji-button": "^4.6.4",
     "@open-inwoner/design-tokens": "^0.0.11-alpha.2",
     "@tarekraafat/autocomplete.js": "^10.2.6",
+    "@utrecht/design-tokens": "^3.3.0",
```
jiromaykin: We don't want to use the design-tokens values from 1 municipality. We will have to add our own new values to our own package: @open-inwoner/design-tokens.

.babelrc
```diff
@@ -1,9 +1,13 @@
...
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "usage",
+        "corejs": 3,
+        "targets": "> 0.25%, not dead"
+      }
+    ]
+  ],
+  "plugins": [["@babel/plugin-proposal-decorators", { "version": "2023-11" }]]
```
jiromaykin: Won't we need "@babel/preset-react" here?

src/open_inwoner/react/router/router.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+const loadModule = async (name: string) => {
+  switch (name) {
+    case 'demo':
+      return import('../modules/demo')
+    // some pages don't have an entrypoint at all, so don't throw exceptions
+  }
+}
+
+export default class Router {
```
swrichards: I find `Router` a confusing name for what this does. Something like `ModuleLoader` in `loader.ts` (probably in the root of `modules`) would make more sense to me.

src/open_inwoner/templates/demo/ReactDemo.html
```diff
@@ -0,0 +1,13 @@
+{% load static %}
+
+<html data-page="demo" lang="nl">
```
swrichards: There is a fair bit of indirection in using this. If I understand correctly you have to align:

- the `data-page` attribute
- one or more DOM nodes to mount the module at
- an entry in the `switch` statement of `router.ts`

I wonder if this won't get a bit confusing and error prone when the number of components grow.

Would it perhaps be easier to have something like:

`<html data-mount-modules="['demo', 'button']>` and then have `router.ts` simply select for `data-mount-modules=*` and dynamically mount the modules using `modules/{element-from-array}`?

src/open_inwoner/react/router/router.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Router
+ * Auto loads view based on page name, which is read from html tag data attribute
+ */
+
+/**
+ * Load the relevant module dynamically.
+ * @param name Alias of the page to load the module for.
+ */
+const loadModule = async (name: string) => {
+  switch (name) {
```
jiromaykin: is this just an example or do we want to import multiple later on?
if we want to add more we can avoid a big switch block with something like this:

```
const modules: Record<string, () => Promise<any>> = {
  demo: () => import('../modules/demo'),
  contact: () => import('../modules/anotherone'),
}

const loadModule = async (name: string) => {
  if (modules[name]) {
    return modules[name]()
  }
}
```

src/open_inwoner/webcomponents/custom/CustomButton.js
```diff
@@ -0,0 +1,81 @@
+import { LitElement, html, css } from 'lit'
```
jiromaykin: Yes, I will have to use Lit for storybook too.

src/open_inwoner/webcomponents/custom/CustomButton.js
```diff
@@ -0,0 +1,81 @@
...
+    this.size = 'medium'
+    this.disabled = false
+    this.clickCount = 0
+  }
+
+  increment() {
+    this.clickCount = this.clickCount + 1
+  }
+
+  handleClick = (event) => {
```
jiromaykin: This method is never used here? The button uses @click=${this.increment}.

src/open_inwoner/webcomponents/custom/CustomButton.js
```diff
@@ -0,0 +1,81 @@
...
+      event.preventDefault()
+      return
+    }
+
+    this.clickCount++
+  }
+
+  render() {
+    return html`
+      <button part="button" ?disabled=${this.disabled} @click=${this.increment}>
```
jiromaykin: Maybe replace with method @click=${this.handleClick} or remove handleClick.

tsconfig.json
jiromaykin: In my own PR i do not have paths,
I do have these differences with yours:
"target": "ES6",
"lib": ["DOM", "ESNext"],
"noEmit": true,
"include": ["src/**/*", ".storybook/**/*", "*.ts"],
  "exclude": ["node_modules", "dist"]
  
  Some may be better, some not.

FRONTEND.md
```diff
@@ -0,0 +1,903 @@
+# Frontend Code Contribution Guide
```
swrichards: This file is quite large and covers a lot of ground, a lot of it useful (how to add new components, when to use what), some of it a bit boilerplate'ish/AI-heavy (the unique selling points of `lit`) or not very actionable (the performance considerations).

Can I suggest that:

1. We introduce the documentation in a separate PR and
2. Actually use the `docs` folder for this? (A lot of the current docs are end-user facing, but I don't see a problem with having a development chapter. That would allow us to break this down into more readable chunks.

vite.config.js
```diff
@@ -0,0 +1,83 @@
+import { defineConfig } from 'vite'
```
jiromaykin, line 1: I don't have any other space to put it, but besides Vite, Typescript etc, we also need a `vitest.config.ts` (I have one in my Storybook PR) but all will be a challenge when I rebase Storybook with this.

vite.config.js
```diff
@@ -0,0 +1,83 @@
...
+              ast: true,
+            },
+          ],
+        ],
+      },
+    }),
+  ],
+
+  build: {
+    outDir: path.resolve(__dirname, paths.jsDir),
```
jiromaykin, line 32: Wondering if paths.jsDir is `static/bundles/ `or could be `static/bundles/react` if we want to.

vite.config.js
```diff
@@ -0,0 +1,83 @@
...
+    manifest: 'manifest.json',
+    cssCodeSplit: true,
+    rollupOptions: {
+      input: {
+        [`${paths.package.name}-react`]: `${__dirname}/${paths.reactEntry}`,
+      },
+      output: {
+        entryFileNames: '[name].js',
+        chunkFileNames: '[name].bundle.js',
+        manualChunks: (id) => {
```
jiromaykin, line 45: Wondering if this manualChunks syntax what we want; maybe things will have to be split differently with design-tokens. For optimizing Lit maybe something like this is better:
```
manualChunks(id) {
  if (id.includes('node_modules')) {
    return 'vendor'
  }
}
```
So that looks at any file that’s from node_modules — so React, ReactDOM, Lit, etc. and puts all external/vendor/thirdparty libraries into one shared chunk called vendor.js. Then the browser can cache vendor.js (because it doesn’t change often).

jiromaykin (changes requested):

FRONTEND.md
jiromaykin: @swrichards  A separate PR seems a bit much? In my own PR I actually did put the documentation for Storybook in `docs/storybook.md` but I was actually thinking to remove that one since it's a bit long/redundant, and just add Storybook docs later in frontend docs.

swrichards: I think the document needs a bit of paring down and breaking up here and there, because in its current form it's introducing a _lot_ of text with different kinds of content mixed together, not all of it which I think is useful. I think it would benefit from a proper review and a bit of restructuring, but I don't think it should hold up the PR, hence the proposal for a separate PR (with the perhaps added side-benefit that all this is in motion and we can review it in light of the lessons learned in all PRs in this space for the sprint). Correct me if I'm wrong but I suspect a good chunk of this was AI-generated, which to be clear I think is well and good (it's very useful!), but in those cases I'd like us to apply a bit more scrutiny before merging changes in bulk.

FRONTEND.md
```diff
@@ -0,0 +1,903 @@
...
+
+The project uses **Lit** for building custom webcomponents with modern development patterns:
+
+- **Reactive Properties**: Automatic re-rendering when properties change
+- **CSS-in-JS**: Encapsulated styling with `css` template literals
+- **Template Literals**: Declarative HTML templates with `html` tag
+- **Decorators**: Modern syntax with `@customElement` and `@property`
+- **TypeScript Support**: Type-safe component development
+
+#### Utrecht Design System
```
jiromaykin: I don't think we should mention a 1 municipality 'community' design value like this. We use our own, inheriting from the NL design-system "Estafettemodel". Sometimes that's Utrecht but definitely not always. So change Utrecht mostly into "NL Design System" here.

FRONTEND.md
```diff
@@ -0,0 +1,903 @@
...
+
+Utrecht webcomponents provide standardized Dutch government UI components:
+
+- **Design Tokens**: Consistent theming and spacing
+- **Accessibility**: WCAG 2.1 AA compliant components
+- **Responsive**: Mobile-first responsive design
+- **Interoperable**: Works with any frontend framework
+
+**Available Utrecht Components:**
```
jiromaykin: `**Example of NLDS Community Components:**`

jiromaykin (changes requested):

src/open_inwoner/react/components/Count/Count.stories.tsx
```diff
@@ -0,0 +1 @@
+// Write storybook component
```
jiromaykin: Stories could also go into a separate Stories directory. 
That may be the way to go for imported NLDS React components that are not part of OIP codebase?

package.json
```diff
@@ -54,35 +67,45 @@
     "htmx.org": "^2.0.4",
     "inline-css": "^3.0.0",
     "leaflet": "^1.7.1",
+    "lit": "^3.3.0",
     "material-icons": "^1.13.12",
     "microscope-sass": "latest",
     "proj4leaflet": "^1.0.2",
+    "react": "^19.1.0",
```
jiromaykin: Warning, when i run a check for dependencies:

```
npm info @storybook/react@8.6.14 peerDependencies
{
  react: '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0-beta',
  'react-dom': '^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0-beta',
  storybook: '^8.6.14',
  typescript: '>= 4.2.x',
  '@storybook/test': '8.6.14'
}
```

This means that for the Storybook set-up all the packages we need are Storybook version 8.6.14 (the add=ons are not yet available for storybook 0), and latest React versions may possibly not work. only up to 19.0.0-beta.

src/open_inwoner/webcomponents/custom/CustomButton.js
```diff
@@ -0,0 +1,81 @@
+import { LitElement, html, css } from 'lit'
+import { customElement, property } from 'lit/decorators.js'
+
+/**
+ * Custom Button Web Component using Lit
+ * Example of a custom web component with styling and interactions
+ */
+@customElement('custom-button')
```
jiromaykin: These decorators create a conflict with the Storybook .babelrc file, where we will need
```
"plugins": [
    ["@babel/plugin-proposal-decorators", { "legacy": true }],
    ["@babel/plugin-proposal-class-properties", { "loose": true }]
  ]
```

package.json
stefrado: This is only used inside the web-components demo. A lot of tokens were missing and this made it quick fix for that, but I'll remove it

.babelrc
stefrado: React is compiled with vite inside vite.config.js with `react({ ... })`

vite.config.js
stefrado: Both can use the same output dir, no files are overwritter/removed.

src/open_inwoner/templates/demo/ReactDemo.html
stefrado: I like you suggestion, and will change the logic

FRONTEND.md
stefrado: I agree with @swrichards it is better to split it off, I'll exclude this file for now.

vite.config.js
stefrado: I think for now this is ok, but we need to add this for sure in the future

src/open_inwoner/react/i18n/locales/nl.json
stefrado: Nope we do not add those files to the .gitignore, i added these files so you can remove those files and pull this branch with the latest changes.

src/open_inwoner/react/components/Count/Count.scss
stefrado: To be honest with this expanding codebase i think it is even better to rename `src/open_inwoner/js` to `src/open_inwoner/frontend` and keep all frontend files (js/react/web-components/scss) together in the same directory. What do you think about that idea?

package.json
stefrado: We are now using `react@18.3.1` and `react-dom@18.3.1`, so this should be fixed.

src/open_inwoner/webcomponents/custom/CustomButton.js
stefrado: I moved web-components to the vite bundler so I'll romove this from `.babelrc`.

vite.config.js
stefrado: I'll come back on this tomorrow.

src/open_inwoner/react/components/Count/Count.scss
jiromaykin: I like Stefan's idea if we all agree (this will also have consequences for my PR)

vite.config.js
stefrado: Discussed: we keep the `vitest` config inside `vite.config.js`.

src/open_inwoner/react/components/Count/Count.scss
stefrado: If we are going to implement the frontend folder, we must wait until the storybook PR (#1815) and this PR are merged. Due to a lot of required file changes.

jiromaykin: Agreed.

src/open_inwoner/react/components/Counter/Counter.scss
```diff
@@ -0,0 +1,3 @@
+.counter {
```
jiromaykin, line 1: There is no button or class in the `tsx` component for this style, but we will probably remove this component later anyway.

jiromaykin (approved):

.prettierignore
```diff
@@ -5,3 +5,4 @@ media/
 docs/
 bin/
 deployment/
+src/open_inwoner/react/i18n/compiled/
```
swrichards, line 8: Missing newline!

swrichards (approved):

PR #1813, opened 2025-06-30
[#3321] Only show CMS pages in sitemap if enabled via CMS AppHookConfig

pi-sigma:
`Mijn profiel` can be configured via a CMS apphook so that only certain elements are enabled/disabled. This is picked up in the view + template for the profile, but not in the sitemap, which now incorrectly includes links to all elements of `Mijn profiel`, irrespective of whether they're enabled or not.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3321

### DIFF
--- a/src/open_inwoner/core/views.py
+++ b/src/open_inwoner/core/views.py
@@ -1,3 +1,4 @@
+import logging
 from urllib.parse import urlencode
 
 from django.shortcuts import render
@@ -6,6 +7,7 @@ from django.urls import reverse
 from cms.models import Title
 
 from open_inwoner.accounts.models import User
+from open_inwoner.cms.profile.cms_appconfig import ProfileConfig
 from open_inwoner.cms.utils.page_display import (
     benefits_page_is_published,
     case_page_is_published,
@@ -15,7 +17,11 @@ from open_inwoner.cms.utils.page_display import (
     profile_page_is_published,
 )
 from open_inwoner.configurations.models import SiteConfiguration
+from open_inwoner.openklant.models import KlantenSysteemConfig
 from open_inwoner.pdc.models.category import Category
+from open_inwoner.questionnaire.models import QuestionnaireStep
+
+logger = logging.getLogger(__name__)
 
 
 def _get_category_data_for_user(cat: Category, user: User) -> dict:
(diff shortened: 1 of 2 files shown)


pi-sigma:
@jiromaykin Let me know if you find other pages that should only show up in the sitemap for authenticated users, digid users etc. but the relevant restrictions are missing.

src/open_inwoner/core/views.py
```diff
@@ -2,10 +2,12 @@
 
 from django.shortcuts import render
 from django.urls import reverse
+from django.utils.translation import gettext as _
 
 from cms.models import Title
 
 from open_inwoner.accounts.models import User
+from open_inwoner.cms.profile.cms_appconfig import ProfileConfig
```
jiromaykin, line 11: How is this handled in the sitemap template?
When running the PR i get these errors:
"MultipleObjectsReturned at /sitemap/
get() returned more than one ProfileConfig -- it returned 2!"

src/open_inwoner/core/views.py
```diff
@@ -49,56 +53,122 @@ def sitemap(request):
...
+
     # add "platform pages" (conditional login/register page, cookie info etc.)
     context["platform_pages"] = [
-        {"url_name": "pages-root", "link_text": "Home page"},
-        {"url_name": "contactform", "link_text": "Contactformulier"},
+        {"url_name": "pages-root", "link_text": _("Home page")},
     ]
 
+    klant_config = KlantenSysteemConfig.get_solo()
+    if klant_config.has_contactform_configuration:
```
jiromaykin, line 167: Something weird is going on with the general Contactform - it has disappeared from the space where the flat pages used to be so it does show up correctly in the sitemap, however the page itself is now empty,
also you will need to make the "Onderwerpen" heading conditional in case a municipality makes the Category pages invisible for non-authenticated users.

You can see a living example of both issues here:
https://mijn-acc.groningen.nl/sitemap/

jiromaykin (changes requested):

src/open_inwoner/core/views.py
pi-sigma: I don't think we have a use case for multiple instances of `ProfileConfig`, so I suspect there's some issue with your data. However, I amended the code to handle this situation (can't know for sure if someone did not create multiple profile configurations in prod somehow).

src/open_inwoner/core/views.py
pi-sigma: - The contactform issue should be fixed already (you can try by setting the default Klant backend to null in the KlantSystem configuration); the contactform should disappear from the links.
- Good catch concerning the "Onderwerpen" heading; made it conditional

src/open_inwoner/core/views.py
```diff
@@ -49,56 +56,126 @@ def sitemap(request):
...
+        if profile_config.selected_categories:
+            base_url = reverse("profile:contact_list")
+            begeleiders_url = f"{base_url}?{urlencode({'type': 'begeleider'})}"
+            cms_pages.append(
+                {
+                    "url": begeleiders_url,
+                    "link_text": _("Mijn begeleiders"),
+                },
+            )
+        if profile_config.mentors:
```
jiromaykin, line 99: Isn't this the same s Begeleiders above?
i now see two links called Mijn contacten.

jiromaykin (changes requested):

jiromaykin (approved):
Approved.

My Pycharm IDE does say that: 

```
context = {
        "category_nodes": [
            _get_category_data_for_user(category, request.user)
            for category in root_categories
        ],
    }
```

can/may be rewritten as:

```

    context = {"category_nodes": [
        _get_category_data_for_user(category, request.user)
        for category in root_categories
    ], "cms_pages": []}
```

but that's probably not super necessary.

PR #1814, opened 2025-06-30
feat: upgrade Elasticsearch version and libraries to 9.x/8.x

swrichards:
This also ensures we're using the same elasticsearch image that is used in production, via openinwoner-deployment and our helm charts. This ensures more consistency between environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
           --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
           --name postgres
       elasticsearch:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.9.2
+        image: bitnami/elasticsearch:9.0.3-debian-12-r1
         env:
           discovery.type: single-node
           ES_JAVA_OPTS: '-Xms512m -Xmx512m'
@@ -150,7 +150,7 @@ jobs:
           --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
           --name postgres
       elasticsearch:
-        image: docker.elastic.co/elasticsearch/elasticsearch:7.9.2
+        image: bitnami/elasticsearch:9.0.3-debian-12-r1
         env:
           discovery.type: single-node
           ES_JAVA_OPTS: '-Xms512m -Xmx512m'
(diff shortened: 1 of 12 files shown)


linssen814:
@swrichards It took the whole day, but I've pretty much resolved everything, though the current solution might still potentially need  some improvement.

As a whole, there were actually 3 separate issues:

1. The `http/https` prefix is now required in the `ELASTICSEARCH_DSL` setting, which we didn't do before. Also, since I set it as http, the SSL check needs to be disabled. This is what fixed the E2E tests. I've disabled it using some environment variables in the pipeline. I'm not sure if we ever used it but I can understand if you think flat-out disabling HTTPS/SSL isn't a good solution even if it's just for testing.

2. The `ESMixin` utility class in `search/tests/utils.py`, which is used by ES tests, calls the `search_index` command of elastic-transport, which deletes, creates and repopulates the indices. The issue was that during the 'repopulate' phase, it failed to JSON-serialize the HTML markup of CMS pages and plugins because the HTML was delivered in 'bytes' form., even though it actually has a mechanism built in place to prevent it from trying to json-serialize 'bytes' types.

To illustrate the issue:

elastic-transport's `JsonSerializer` class contains this method, which is eventually called while populating the indices:

```python
   def dumps(self, data: Any) -> bytes:
        # The body is already encoded to bytes
        # so we forward the request body along.
        if isinstance(data, str):
            return data.encode("utf-8", "surrogatepass")
        elif isinstance(data, bytes):
            return data

        try:
            return self.json_dumps(data)
        # This should be captured by the .default()
        # call but just in case we also wrap these.
        except (ValueError, UnicodeError, TypeError) as e:  # pragma: nocover
            raise SerializationError(
                message=f"Unable to serialize to JSON: {data!r} (type: {type(data).__name__})",
                errors=(e,),
            )
```

And this is (sort of) what our data looks like when its delivered to it:
```python
data = {
    'url': '/mijn-profiel/',
    'rendered_html': b'<html>...',  # note that its a 'bytes'
    'description': '',
}
```

So 'isinstance(data, bytes)' check fails here because it only checks whether the data block itself is a 'bytes', but not the 'rendered_html' inside of it.

I'm not sure what changed between versions of elasticsearch/django-elasticsearch-dsl that would suddenly cause this issue to appear (maybe because elastic-transport is at version 8.17.1 while the other packages are at 8.18.1?), but by putting an extra check in our `render_full_page` method in `cms/tests/cms_tools.py` it gets resolved.

There was also another, separate underlying issue with the populating of indices whereby some of the html of plugins could not be rendered because it expected a `user` in the template context that wasn't there, since there isn't an actual request with a user. This didn't affect the tests though so I think that issue might have been there before.

The third/last issue was just that one of the error/exception classes of the client now requires a `message` parameter.

swrichards:
@linssen814 thanks for the comprehensive write up and sleuthing, good work. I want to verify our production settings and ensure we'll still be able to run with SSL there. Will report back here when that's done, we'll leave this in draft until then.

linssen814 (approved):

PR #1816, opened 2025-06-30
Fix prev-login check failing on create_or_update

linssen814:
Added a check that verifies that `update_fields` argument is a list (and not a set).

The `update_or_create` method converts some of its arguments to a `set` before passing it to the `update_fields` parameter. Because of this, the snippet of code I put in the override of the save() method in `accounts/models.py` fails because it tries to call `append` on a set. Currently this doesn't seem to have affected anything yet (except stuff I'm currently working on that uses update_or_create).

### DIFF
--- a/src/open_inwoner/accounts/models.py
+++ b/src/open_inwoner/accounts/models.py
@@ -617,7 +617,10 @@ class User(AbstractBaseUser, PermissionsMixin):
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
         if update_fields and "last_login" in update_fields:
-            update_fields.append("previous_login")
+            if not isinstance(update_fields, set):
+                update_fields = set(update_fields)
+
+            update_fields.add("previous_login")
 
         return super().save(
             force_insert, force_update, using, update_fields=update_fields


src/open_inwoner/accounts/models.py
```diff
@@ -616,7 +616,11 @@ def get_group_managed_categories(self):
     def save(
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
-        if update_fields and "last_login" in update_fields:
+        if (
+            isinstance(update_fields, list)
+            and update_fields
+            and "last_login" in update_fields
+        ):
             update_fields.append("previous_login")
```
swrichards: The [docs say that](https://docs.djangoproject.com/en/4.2/ref/models/instances/#ref-models-update-fields):

> The update_fields argument can be any iterable containing strings. An empty update_fields iterable will skip the save. A value of None will perform an update on all fields.

Technically this means `update_fields` can be a whole bunch of things: `list`, `list`, generators, iterators, etc.

Given that Django doesn't care, I think we can safely convert to a list _first_, which will cover all types of iterable, so:

```suggestion
        if update_fields and "last_login" in update_fields:
            if not isinstance(update_fields, list):
                update_fields = list(update_fields)
                
            update_fields.append("previous_login")
```

swrichards (changes requested):
Good find, and clear description in the PR :+1: .

I'd implement the fix slightly different though (see comment), but this is a good catch and I think it'll warrant a patch release, given this code path is potentially hit a lot with auth flows we may not have fully tested (e.g. OIDC).

swrichards (approved):

PR #1818, opened 2025-07-01
[#3342] Add headers to BRP API to support Centric

pi-sigma:
Taiga:
- https://taiga.maykinmedia.nl/project/open-inwoner/task/3342
- https://taiga.maykinmedia.nl/project/open-inwoner/us/3313?milestone=535


src/open_inwoner/haalcentraal/admin.py
```diff
@@ -7,4 +7,33 @@
...
-    pass
+    fieldsets = (
+        (
+            "Service",
+            {
+                "fields": ("service",),
+            },
+        ),
+        (
+            "Headers for I Connect",
```
swrichards: Let's localize's these.

src/open_inwoner/haalcentraal/api.py
```diff
@@ -122,6 +122,8 @@ def make_request(self, user_bsn: str) -> requests.Response:
             "Accept": "application/json",
             "x-gebruiker": "BurgerZelf",
         }
+
+        # I connect headers
```
swrichards, line 126: The one thing that gives me pause is if, for unclear reasons, values are set for both flavours of API, and are sent along, will the API complain about unknown headers? We could consider adding a choice field indicating which flavour is expected, and only sending the header for the selected type.

swrichards (changes requested):

src/open_inwoner/haalcentraal/api.py
pi-sigma: Good point. What do you think of a constraint (enforced through form validation and model save) to the effect that staff can only define any of the FOO headers if none of the BAR headers are defined and vice versa? Keeps the UI simpler (adding a checkbox is one step in the direction of my original idea).

swrichards: That seems like a better solution indeed.

pi-sigma: I thought it would be a good idea to enforce the constraint not only through form validation but also through the model directly, i.e. calling `self.full_clean()` in the `save` method. Since we didn't have that before, we now have failing tests where the `service` is missing on the haalcentraal config. Do you think it's worth updating the tests (by either providing the service or disconnecting the signals, since Haalcentraal is invoked through signals) or stick with form validation for a relatively minor thing?

swrichards: Happy to stick with form validation for this one.

src/open_inwoner/haalcentraal/models.py
```diff
@@ -26,6 +27,7 @@ class HaalCentraalConfig(SingletonModel):
         null=True,
     )
```
swrichards, line 31: Just a nitpick, but: could we add a `help_text` stating thatr the corresponding headers for the chosen API (and only the chosen API) should be filled in below? I've found that after the fact our config can be a bit cryptic for the admins, the more context the better.

swrichards (approved):

PR #1821, opened 2025-07-07
[#3352] Fix error handling in Mijn Uitkeringen

pi-sigma:
Fixes two small bugs in the uitkeringen app:

- [x] Set `pdf_content` to `null` in case of SSD client error
- [x] Properly handle response contents with `fwi:NietsGevonden` (they are not errors)

https://taiga.maykinmedia.nl/project/open-inwoner/issue/3352

### DIFF
--- a/src/open_inwoner/ssd/views.py
+++ b/src/open_inwoner/ssd/views.py
@@ -58,6 +58,7 @@ class BenefitsFormView(
                         "Please try again later"
                     ),
                 )
+                pdf_content = None
 
             if not pdf_content:
                 return_path = request.get_full_path()
--- a/src/open_inwoner/ssd/xml.py
+++ b/src/open_inwoner/ssd/xml.py
@@ -70,9 +70,6 @@ def _get_report_info(
     if info_response.fwi:
         raise SSDClientException.from_xml_response(xml_response=info_response)
 
-    if info_response.niets_gevonden:
-        raise SSDClientException("unexpected error occured")
-
     return info_response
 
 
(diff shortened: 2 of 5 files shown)


src/open_inwoner/ssd/views.py
```diff
@@ -58,6 +58,7 @@ def post(self, request, *args, **kwargs):
                         "Please try again later"
                     ),
                 )
+                pdf_content = None
```
swrichards, line 61: Wouldn't this be better on `l.42`? Currently it looks like we could still have a unknown symbool error if `form.is_valid() is False`.

src/open_inwoner/ssd/xml.py
```diff
@@ -71,7 +71,7 @@ def _get_report_info(
         raise SSDClientException.from_xml_response(xml_response=info_response)
 
     if info_response.niets_gevonden:
-        raise SSDClientException("unexpected error occured")
+        return None
```
swrichards: Could we get some basic test coverage for these lines? I feel we've burnt our fingers a few times here especially on the error handling, would be good to have an extra layer of security.

swrichards (changes requested):

src/open_inwoner/ssd/views.py
pi-sigma: Where would be accessed to trigger the error? It's only used if `form.is_valid() is True`

src/open_inwoner/ssd/xml.py
pi-sigma: I can add a test with a new response file. However, the mistake was in mis-interpreting the significance of `nietsGevonden`. Since we don't have an API standard to develop against it's sometimes difficult to tell (though in hindsight it makes sense that `niesGevonden` is part of the regular flow.

src/open_inwoner/ssd/views.py
swrichards: Yes my bad, I didn't see that `l.63` had the same indent.

swrichards (approved):

PR #1824, opened 2025-07-10
Apply ruff bandit

swrichards:
We can deprecate the separate `bandit` pipeline by relying on `ruff` which incorporates all the relevant rules, making things faster and more maintainable.

----
📚 Documentation preview 📚: https://open-inwoner--1824.org.readthedocs.build/en/1824/

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -97,24 +97,3 @@ jobs:
           grep ^ruff requirements/ci.txt | uv pip install --system -r /dev/stdin
       - name: Run ruff format on all files
         run: ruff format --diff --target-version py311 src
-
-  bandit:
-    name: Python security check using Bandit
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Install libxml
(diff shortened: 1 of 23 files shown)


swrichards:
~Needs #1787~ Actually, we'll do this PR before the 3.12 upgrade. It makes the latter simpler, because no bandit changes needed.

pi-sigma (approved):
It's odd that `assert` -> conditional `ValueError` is part of this. I thought bandit checks are concerned with security, but according to the the docs it's intended this way for some reason.

swrichards:
> It's odd that assert -> conditional `ValueError` is part of this. I thought bandit checks are concerned with security, but according to the the docs it's intended this way for some reason.

[I suppose the reasoning ](https://docs.astral.sh/ruff/rules/assert/#assert-s101)is that you might be using `assert` to validate user input or some other constraint that is security-sensitive, e.g. if somebody writes `assert user.is_authenticated` as a way to gate access to a view, which might be optimized away depending on the runtime.

PR #1829, opened 2025-07-14
[#3351] Only show questions/contactmomenten initiated by the current user

pi-sigma:
- [x] Only show contactmomenten initiated by the current user (Esuite)
- [x] Only show questions initiated by the current user (OpenKlant)

Taiga: 
- https://taiga.maykinmedia.nl/project/open-inwoner/us/3351
- https://taiga.maykinmedia.nl/project/open-inwoner/task/3364

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1485,7 +1485,17 @@ class OpenKlant2Service(
             klantcontacten,
         )
 
-        return klantcontacten_for_partij
+        # only show questions initiated by the current user
+        klantcontacten_for_initiator = filter(
+            lambda row: True
+            in glom.glom(
+                row,
+                ("_expand.hadBetrokkenen", ["initiator"]),
+            ),
+            klantcontacten_for_partij,
+        )
+
+        return klantcontacten_for_initiator
 
     def questions_for_partij(self, partij_uuid: str) -> list[OpenKlant2Question]:
         answers_for_klantcontact_uuid = {}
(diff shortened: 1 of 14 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -786,6 +786,16 @@ def list_questions_for_zaak(self, zaak: Zaak, user: User) -> list[Question]:
             question: ContactMoment | None = getattr(ocm, "contactmoment", None)
             if question:
                 contactmomenten.append(question)
+
+        # filter contactmomenten via klantcontactmomenten
+        fetch_params = self.get_fetch_parameters(user)
+        klantcontactmomenten = self.fetch_klantcontactmomenten(**fetch_params)
+        cms_with_klantcontact = [kcm.contactmoment for kcm in klantcontactmomenten]
```
swrichards: Could this be a `set` or are the objects not hashable? Would make the lookup below (technically) faster.

swrichards: I think verbose names help here:

```suggestion
        contactmomenten_within_klantcontacten = [kcm.contactmoment for kcm in klantcontactmomenten]
```

src/open_inwoner/openklant/services.py
```diff
@@ -786,6 +786,16 @@ def list_questions_for_zaak(self, zaak: Zaak, user: User) -> list[Question]:
...
+
+        contactmomenten[:] = [
```
swrichards: Maybe we could combine this with the loop on `l.785-788` for efficiency's sake, by fetching the contactmomenten attached to the klantcontacten first?

swrichards (changes requested):
LGTM, just some small suggestions

src/open_inwoner/openklant/services.py
pi-sigma: `ContactMoment` instances are indeed not hashable. This can be avoided though by storing URLs instead of contactmomenten.

swrichards (approved):

PR #1835, opened 2025-07-16
[#1169] Remove xml-related packages from Docker build

pi-sigma:
When `libxml` is installed (which it is through installing libxml2-dev), then pip compiles `lxml` from source against the system `libxml`. This has caused a mismatch between the system `libxml` available at runtime and the `libxml` that `lxml` was compiled against during the build stage. Without `libxml` on the system, pip falls back on pre-built binary wheels, which have their own version of `libxml` built in.

Taiga: https://taiga.maykinmedia.nl/project/maykin-intranet/issue/1169


swrichards:
Taken up in #1834

PR #1837, opened 2025-07-17
[#3370] Add Side Navigation to Home page

jiromaykin:
Add Den Haag React component, JIRA story:
https://dimpact.atlassian.net/jira/software/c/projects/PP/boards/256/backlog?selectedIssue=PP-282

Taiga task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3372

Also uses our updated Design Tokens from this package (so make sure to run NPM install/build):
https://github.com/maykinmedia/open-inwoner-design-tokens

- [x] 1.  Create React Module Structure, probably in src/open_inwoner/react/modules/sidenav/
- [x] 2. Create the Main Sidenav Component src/open_inwoner/react/components/Sidenav/Sidenav.tsx
- [x] 3. Create the Portal Index File src/open_inwoner/react/components/Sidenav/index.tsx
- [x] 4. Add Django Template for deeper Django component src/open_inwoner/components/templates/components/Sidenav/Sidenav.html
- [x] 5. Include the React Module in the Home page Fullwidth CMS template src/open_inwoner/templates/cms/fullwidth.html
- [x] 6. Register the Module in vite config
- [x] Extra: reduce the number of Cards on the Home page from 4 to 3 (still needs a design answer, see also other task https://taiga.maykinmedia.nl/project/open-inwoner/task/3379: 
- [x] to make menu dynamic: 1.  Create a Django Context Processor or Template Data; or use JSON
- [x] Update Sidenav React Component to use CMS menu Data
- [x] add material icons as a mapping to Sidenav tsx component (based on their names)
- [x] correct the wrong names for icons in back-end + remove the "HELP" icons for when municipalities do not want to use icons (make empty)
- [ ] correct/create tablet styles (we do not have a design) and decide if Cards should scale, or drop to next line; below 768px all should be 100%

**Not in this task:** 

- make a story for this component to see it in Storybook with our own styling, [see 3380](https://taiga.maykinmedia.nl/project/open-inwoner/task/3380).
- show 4 Category cards when user is not authenticated, [see 3379](https://taiga.maykinmedia.nl/project/open-inwoner/task/3379).
- dynamic number badge, [see 3385](https://taiga.maykinmedia.nl/project/open-inwoner/task/3385).
- future task, probably different sprint: when all Side Nav pages are removed from the Desktop dropdown, the "Mijn profiel" page needs to only remain in the dropdown and be removed fro mthis sidebar Side Navigation menu, [see 3370](https://taiga.maykinmedia.nl/project/open-inwoner/us/3370).


src/open_inwoner/components/templates/components/Sidenav/menu_json_item.html
stefrado, line 1: This is implemented inside `side_navigation.py` and then added as json_script, right? (might remove the file)

src/open_inwoner/components/templates/components/Sidenav/Sidenav.html
```diff
@@ -0,0 +1,27 @@
+{% load i18n form_tags icon_tags link_tags string_tags solo_tags custom_session_tags menu_tags button_tags side_navigation %}
+
+{% if request.user.is_authenticated %}
+
+    <!-- Set nonce for DenHaag components, or else styling from number-badge will cause CSP errors -->
+    <script nonce="{{ request.csp_nonce }}">
+        window.NONCE = "{{ request.csp_nonce }}"
```
stefrado: By adding this to the head the csp-nonce is set correct on all styles/scripts.

`<meta name="csp-nonce" content="{{ request.csp_nonce }}" />`

src/open_inwoner/components/templates/components/Sidenav/Sidenav.html
```diff
@@ -0,0 +1,27 @@
...
+    </script>
+
+    <div class="side-navigation--oip">
+        <div class="" id="react-openinwoner-sidenav"></div>
+
+        <nav class="primary-navigation" aria-label="Navigatie na inloggen">
+            <ul class="primary-navigation__list subpage-list">
+                {% if has_general_faq_questions %}
```
stefrado: Why isn't this item rendered by react?

src/open_inwoner/react/modules/demo/index.tsx
```diff
@@ -2,6 +2,8 @@ import { Root } from 'react-dom/client'
 import Demo from './Demo'
 import { AbstractPage } from '@react/lib/abstractPage'
 
+console.log('Demo counter module loaded')
```
stefrado: 🔥

src/open_inwoner/react/modules/Sidenav/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { Root } from 'react-dom/client'
+import Sidenav from './Sidenav'
+import { AbstractPage } from '@react/lib/abstractPage'
+
+console.log('Sidenav module loaded')
```
stefrado: 🔥

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
```
stefrado: Just import { FC } to reduce import size (vite handles the base react import)

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { SideNavigation } from '@gemeente-denhaag/side-navigation'
+
+// Material Icons component
+const MaterialIcon: React.FC<{ name: string }> = ({ name }) => (
```
stefrado: Might split this up to different component and use a interface instead of direct adding type direct to the FC (future proof and exportable)

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+  <span className="material-icons-outlined" aria-hidden="true">
+    {name}
+  </span>
+)
+
+const Sidenav: React.FC = () => {
+  // Get menu data from Django
+  const getMenuData = () => {
```
stefrado: This can be placed inside a lib function. It will be used in almost every component

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+        current: item.current,
+        counter: item.counter || undefined,
+      })),
+    ]
+
+    return <SideNavigation items={navigationItems} />
+  }
+
+  // Fallback - should not be appearing
+  const fallbackItems = [
```
stefrado: Implement typing (don't forget the import :) )
 
```suggestion
  const fallbackItems: SideNavigationProps['items'] = [
```

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
stefrado: Generic name and type to allow type errors.
```suggestion
  const getJsonScriptData = <T extends unknown>(id: string): T[] => {
```

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+        console.log('Menu data loaded:', data) // Debug log
+        return data
+      } catch (e) {
+        console.error('Failed to parse menu data:', e)
+      }
+    }
+    return []
+  }
+
+  const menuData = getMenuData()
```
stefrado: If you are going to implement the generic json change the fn to this to make sure menuData has the correct type.
```suggestion
  interface MenuItem {
    href: string
    label: string
    icon?: string
    counter?: number
    current?: boolean
  }

  const menuData = getJsonScriptData<MenuItem>()
```

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  // Transform Django menu data to DenHaag format
+  if (menuData.length > 0) {
+    const navigationItems = [
+      menuData.map((item: any) => ({
```
stefrado: Can be removed if the comments above are implemented. Please avoid the usage of any.

```suggestion
      menuData.map((item) => ({
```

src/open_inwoner/scss/views/_index.scss
```diff
@@ -4,3 +4,4 @@
 @import './Categories.scss';
 @import './product_detail';
 @import './view';
+@import '../components/SideNavigation/SideNavigation';
```
stefrado: Can't we import this in react? (to reduce bundle size)

src/open_inwoner/scss/views/_view.scss
```diff
@@ -7,12 +7,13 @@
...
         }
       }
-      .grid__welcome.plugin {
+      .grid__welcome {
         margin-top: -40px;
         padding: 0;
 
         @media (min-width: 768px) {
           margin-top: -30px;
+          background-color: yellow;
```
stefrado: 🟡

src/open_inwoner/templates/master.html
```diff
@@ -1,5 +1,5 @@
 {% load static i18n card_tags button_tags link_tags notification_tags anchor_menu_tags view_breadcrumbs utils solo_tags session_tags django_htmx cms_tags menu_tags sekizai_tags %}<!DOCTYPE html>
-<html lang="nl" class="view utrecht-document openinwoner-theme {% block view_class %}view--{{ request.resolver_match.namespaces|join:'-' }}-{{ request.resolver_match.url_name }}{% endblock %}">
+<html lang="nl" class="view utrecht-document openinwoner-theme {% block view_class %}view--{{ request.resolver_match.namespaces|join:'-' }}-{{ request.resolver_match.url_name }}{% endblock %}" {% block mount_modules %}{% endblock %}>
```
stefrado: Just a little formatting would be nice i guess.
```suggestion
<html 
    lang="nl" 
    class="view utrecht-document openinwoner-theme {% block view_class %}view--{{ request.resolver_match.namespaces|join:'-' }}-{{ request.resolver_match.url_name }}{% endblock %}" 
    {% block mount_modules %}{% endblock %}
>
```

src/open_inwoner/templates/master.html
```diff
@@ -28,7 +28,7 @@
 
         {% if favicon %}<link rel="icon" href="{{ favicon }}" />
         <link rel="apple-touch-icon" href="{{ favicon }}">{% endif %}
-    
+
```
stefrado, line 35: Extra space?

src/open_inwoner/templates/master.html
```diff
@@ -174,13 +174,13 @@
 
         {% render_block "js" %}
         {% block extra_js %}{% endblock %}
-
```
stefrado, line 177: Less space?

src/open_inwoner/templates/cms/fullwidth.html
```diff
@@ -1,11 +1,21 @@
...
 {% load i18n cms_tags sekizai_tags static %}
 
+{% block mount_modules %} data-mount-modules='["sidenav"]'{% endblock %}
+
+{% block react_modules %}
+{# Including this script will allow the page to use the react router #}
+<script
+  type="module"
+  src="{% static 'bundles/open_inwoner-react.js' %}"
+  defer
```
stefrado: defer is not really needed since the `{% block react_modules %}` is at the end of the body.

src/open_inwoner/templates/cms/fullwidth.html
```diff
@@ -1,11 +1,21 @@
 {% extends 'master.html' %}
 {% load i18n cms_tags sekizai_tags static %}
 
+{% block mount_modules %} data-mount-modules='["sidenav"]'{% endblock %}
+
+{% block react_modules %}
+{# Including this script will allow the page to use the react router #}
```
stefrado: React router? i think you meant to use react components.

```suggestion
{# Including this script will allow the page to use react components #}
```

src/open_inwoner/react/modules/Sidenav/index.tsx
stefrado, line 1: We might want to get and manipulate the data inside this component and pass it as props to the Sidenav. Then we can keep the Sidenav generic (no complicated tasks, just rended this or that).

src/open_inwoner/react/components/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,61 @@
+// PERHAPS THIS SHOULD BE HERE in components? ELSE IN 'modules'
```
stefrado: You might be correct than we can keep split it up like this

React
| — components
| ——— SideNav
| —————— index.ts (component)
| — modules
| ——— SideNav
| —————— index.ts (module)

src/open_inwoner/components/templatetags/side_navigation.py
stefrado, line 1: There are a couple `print()`'s

stefrado (changes requested):

src/open_inwoner/components/templates/components/Sidenav/menu_json_item.html
jiromaykin: Oops, yes, I should have completely removed this file - i first tried to see if I could get the React component to work at all. But this would have been an ugly solution I think.

src/open_inwoner/templates/cms/fullwidth.html
swrichards: Agreed, I'd really like to avoid using `router` for our `loader` because it's ambiguous with the widely used [react-router libary](https://reactrouter.com/).

src/open_inwoner/components/templatetags/side_navigation.py
swrichards: `logger.debug`!

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,143 @@
...
+
+from django import template
+
+from menus.menu_pool import menu_pool
+
+register = template.Library()
+
+
+@register.simple_tag(takes_context=True)
+def react_sidenav_data(context):
```
swrichards, line 44: I don't see the value of making this a templatetag, I think it makes it harder to debug. I think it would make more sense to:

1. Make this a regular utility function
2. Just passing the resulting dict into the templates (possibly via middleware if we expect to need it on all requests)

It should also make the tests a bit simpler (🤞 ).

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,143 @@
...
+        all_nodes = renderer.get_nodes()
+        print(f"Total nodes found: {len(all_nodes)}")
+
+        # Find the "home" node - users are required to set this ID!
+        # else we need to exclude all the other pages from this menu node
+        home_node = None
+        for node in all_nodes:
+            # Check if this is the home page
+            if (
+                (hasattr(node, "id") and node.id == "home")
```
swrichards:
```suggestion
                (getattr(node, "id", None) == "home")
```

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,143 @@
...
+                or (hasattr(node, "reverse_id") and node.reverse_id == "home")
```
swrichards:
```suggestion
                or (getattr(node, "reverse_id", None) == "home")
```

src/open_inwoner/components/templates/components/Sidenav/Sidenav.html
jiromaykin: Because that page/link is not fed by the CMS menu.
It's a remnant: it is currently possible to add an FAQ accordeon (_without linking it to a category_) that then becomes a separate page, like so: ` path("faq/", FAQView.as_view(), name="general_faq"),`

Sicne that page is not part of the CMS pages but should still be present, if the goal is to make both menus (authenticated dropdown & side navigation) look the same, i had to add it here.

If you look at our HTML django-template for the Authenticated Nav menu, you can see it there too.

Maybe this is good to know for @swrichards and @pi-sigma.
Maybe we should remove this functionality in FAQ's since we now have CMS pages that can be linked in the footer. But this particular one may or may not need to be visible in the side navigation tree as well and not in the footer.

But since we do not need to decide this ourselves I mentioned it to Bobby in this issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3374

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,143 @@
...
+    """Template tag to provide menu data for React Sidenav component"""
+    request = context["request"]
+
+    try:
+        # Get the menu renderer
+        renderer = menu_pool.get_renderer(request)
+
+        # Get all menu nodes
+        all_nodes = renderer.get_nodes()
+        print(f"Total nodes found: {len(all_nodes)}")
```
swrichards: Remove `print` statements or make them `logger.debug`.

src/open_inwoner/components/templates/components/Sidenav/Sidenav.html
jiromaykin: I could use `{% block extra_head %}` on this component level.

src/open_inwoner/react/modules/Sidenav/index.tsx
swrichards: I agree that would be nicer, and I think it's an accepted pattern (HOC / higher-order components). I think the menu item fetching logic would fit well in `Page`.

src/open_inwoner/react/modules/Sidenav/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export default class Page extends AbstractPage {
```
swrichards: I don't think this has to be named `Page` right, as we just grab the default export? If so, I would go for something more descriptive: `SideNavPage`

Though generally: I think `Page` is a misnomer, and `Module` and `AbstractModule` would be better.

src/open_inwoner/react/components/Sidenav/Sidenav.tsx
jiromaykin: But then I wonder where we should put the Stories (later).

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const Sidenav: React.FC = () => {
+  // Get menu data from Django
+  const getMenuData = () => {
+    const scriptElement = document.getElementById('sidenav-menu-data')
+    if (scriptElement?.textContent) {
+      try {
+        const data = JSON.parse(scriptElement.textContent)
+        console.log('Menu data loaded:', data) // Debug log
```
swrichards:
```suggestion
        console.debug('Menu data loaded:', data)
```

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
swrichards: const naming, and move this to the global scope given there's no dynamic elements?

```suggestion
  const FALLBACK_ITEMS = [
```

src/open_inwoner/react/modules/Sidenav/index.tsx
jiromaykin: 🙈 Yeah, i'll double check and remove all logs.

src/open_inwoner/scss/views/_view.scss
jiromaykin: 🙈

src/open_inwoner/templates/master.html
jiromaykin: I'll correct the spacing. At first I tried adding stuff to Master instead of Fullwidth but reverted most of them.

src/open_inwoner/react/components/Sidenav/Sidenav.tsx
stefrado: In that case i think this is the structure.

React
| — components
| ——— SideNav
| —————— SideNav.ts (component)
| —————— SideNav.stories.ts (story)
| —————— SideNav.spec.ts (vitest)
| — modules
| ——— SideNavModule
| —————— SideNavModule.ts (module)
| —————— SideNavModule.stories.ts (story)
| —————— SideNavModule.spec.ts (vitest)

src/open_inwoner/components/templatetags/side_navigation.py
jiromaykin: Looks like I can also replace `import traceback ..... traceback.print_exc()`.
I don't know much about this, so hopefully the new commit will work better with Django LOGGING settings.

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
jiromaykin: Ah, but we still have "`import React, { FC } from 'react`'" in `src/open_inwoner/react/components/Counter/Counter.tsx` so i guess I should also change that.

src/open_inwoner/templates/cms/fullwidth.html
jiromaykin: I put that 'router' text there because I copied it from Stefan's src/open_inwoner/templates/demo/ReactDemo.html template, for our portal.

Let me know if I should change both texts 😄

src/open_inwoner/react/components/Sidenav/Sidenav.tsx
jiromaykin: I did not add vitest tests yet, so I guess that also belong in this PR?

src/open_inwoner/react/modules/Sidenav/index.tsx
swrichards: More generally, I think it's a cleaner convention to have an `index.ts` which serves only as an entrypoint for the module, and explicitly exporting the symbols we care about (like how `__init__.py` works in Python).

In this case, the current file would be something like `SidenavPage.tsx` and `index.ts` would just be:

```typescript
import {SidenavPage } from './SidenavPage';

export default SidenavPage;
```

It makes git logs and editor tabs more readable, and helps `index.ts` have a single task, deciding which symbols to export.

But maybe this is better discussed in a separate context than this PR @stefrado @jiromaykin

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Find the "home" node - users are required to set this ID!
+        # else we need to exclude all the other pages from this menu node
+        home_node = None
+        for node in all_nodes:
+            # Check if this is the home page
+            if (
```
swrichards: I find the logic a bit hard to follow due to my ongoing struggles with Django CMS, but FWIW: this does not work on my machine (no Home is found), even though I have a seemingly normal setup.

src/open_inwoner/cms/extensions/constants.py
```diff
@@ -8,10 +8,10 @@ class IndicatorChoices(models.TextChoices):
...
-    person = "person", _("Home")
+    person = "person", _("User")
     description = "description", _("Products")
     inbox = "inbox", _("Inbox")
     inventory_2 = "inventory_2", _("Cases")
     group = "group", _("Collaborate")
-    help_outline = "help_outline", _("Help")
-    euro_outline = "euro_outline", _("Benefits")
+    help = "help", _("Help")
+    euro = "euro", _("Benefits")
```
jiromaykin, line 19: For your information: changes to these faulty icons were not picked up properly when doing migrations,
even when going back a migration and removing the tables.
I had to pick the icons again to make them properly styled in the CMS menus (both our usual authenticated dropdowns, as well as the new sidenav).
@swrichards @stefrado

src/open_inwoner/components/templatetags/side_navigation.py
jiromaykin: Since we have a CMS menu that is dependent on the id=home I thought I could make it work like this.

There's also a way to show ALL cms pages, but then we would need a way to exclude "Flat Pages" etc., and I do not know their dynamic URLS - they could be anything.

Or else we could do something like this:

```
if not home_node:
            print("Home node not found, using all visible nodes")
            # Fallback: use nodes that seem to be main navigation items
            # Filter out footer pages and sub-pages
            navigation_nodes = []
            exclude_urls = [
                "/cookieverklaring/",
                "/persoonsgegevens/",
                "/privacyverklaring/",
                "/toegankelijkheidsverklaring/",
            ]
            for node in all_nodes:
                url = (
                    node.get_absolute_url() if hasattr(node, "get_absolute_url") else ""
                )
                # Skip footer pages and sub-pages
                if url not in exclude_urls and "/onderwerpen/" not in url:
                    navigation_nodes.append(node)
            target_nodes = navigation_nodes
        else:
            # Use children of home node
            target_nodes = getattr(home_node, "children", [])
            print(f"Home node has {len(target_nodes)} children")

        menu_items = []
```

This is not quite front-end I feel. @swrichards @stefrado .

src/open_inwoner/scss/components/SideNavigation/SideNavigation.scss
bddjong, line 1: @jiromaykin 
Het zou handig zijn als we samen een keer kunnen kijken naar de aanpassingen die je hier hebt moeten doen. Dan kunnen we eventueel dingen over brengen als aanpasbaarheid naar het component zelf

src/open_inwoner/cms/extensions/constants.py
swrichards: This needs a data migration to remap the existing values, e.g.:

```
# Generated by Django 4.2.21 on 2025-07-28 14:43

from django.db import migrations, models


def map_icon_values(apps, schema_editor):
    CommonExtension = apps.get_model("extensions", "CommonExtension")

    # Map help_outline => help & euro_outline => euro
    CommonExtension.objects.filter(menu_icon="help_outline").update(menu_icon="help")
    CommonExtension.objects.filter(menu_icon="euro_outline").update(menu_icon="euro")


def reverse_map_icon_values(apps, schema_editor):
    CommonExtension = apps.get_model("extensions", "CommonExtension")

    # Revert to the initial mapping
    CommonExtension.objects.filter(menu_icon="help").update(menu_icon="help_outline")
    CommonExtension.objects.filter(menu_icon="euro").update(menu_icon="euro_outline")


class Migration(migrations.Migration):
    dependencies = [
        ("extensions", "0007_alter_commonextension_requires_auth_bsn_or_kvk"),
    ]

    operations = [
        # Note that the ORM itself does not validate the TextChoices, we can (re)map
        # prior to the field being altered
        migrations.RunPython(
            map_icon_values,
            reverse_map_icon_values,
        ),
        migrations.AlterField(
            model_name="commonextension",
            name="menu_icon",
            field=models.CharField(
                blank=True,
                choices=[
                    ("person", "User"),
                    ("description", "Products"),
                    ("inbox", "Inbox"),
                    ("inventory_2", "Cases"),
                    ("group", "Collaborate"),
                    ("help", "Help"),
                    ("euro", "Benefits"),
                ],
                help_text="Icon in het menu",
                max_length=32,
                verbose_name="Icon",
            ),
        ),
    ]

```

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,144 @@
...
+                (hasattr(node, "id") and node.id == "home")
+                or (hasattr(node, "reverse_id") and node.reverse_id == "home")
+                or (
+                    getattr(node, "title", "").lower() == "overzicht"
+                    and node.get_absolute_url() == "/"
+                )
+            ):
+                home_node = node
+                logger.debug(f"Found home node: {node.title}")
+                break
```
swrichards: I think we should stick to one explicit condition for locating the `home` node, to make the behavior more predictable (and issues less ambiguous to fix):

```suggestion
            node_attr = getattr(node, "attr", {})
            if node_attr.get("reverse_id", None) == "home"):
                home_node = node
                logger.debug(f"Found home node: {node.title}")
                break
```

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,144 @@
...
+            elif hasattr(node, "attr") and hasattr(node.attr, "menu_icon"):
+                icon = node.attr.menu_icon
+                logger.debug(f"Found icon via node.attr.menu_icon: {icon}")
+            else:
+                # Try to get it from the page's CommonExtension
+                try:
+                    if hasattr(node, "id") and node.id:
+                        from cms.models import Page
+
+                        from open_inwoner.cms.extensions.models import CommonExtension
```
swrichards, line 140: Move to the top, I don't think these are circular

src/open_inwoner/scss/components/SideNavigation/SideNavigation.scss
jiromaykin: @bddjong De enige aanpassing op het component van Den Haag betreft hier hun hardcoded breakpoint van 1023px dat voor ons bijna net zo breed is al de gehele container op Desktop maar waar dit component dan tot daar Display none zou krijgen, dus dat is voor ons geen 'mobiel' breakpoint.
De andere styling hier heeft te maken met een niet-dynamische CMS pagina.

src/open_inwoner/cms/extensions/constants.py
jiromaykin: The `help` icon in Material Icons turns out to have a different suffix than all the others, but I'll add a note about that in the code, and change the migration.

src/open_inwoner/components/templatetags/side_navigation.py
jiromaykin: That sounds like back-end :-)

src/open_inwoner/components/templatetags/side_navigation.py
jiromaykin: you mean to move CommonExtension but also Page imports to all other imports? Will do.
EDIT: unless I can use them in fallback only.

src/open_inwoner/scss/views/_index.scss
jiromaykin: needs to be removed completely - i was just testing things in the beginning.

src/open_inwoner/react/modules/Sidenav/Sidenav.tsx
jiromaykin: Added a component + new dir.

src/open_inwoner/react/modules/Sidenav/index.tsx
jiromaykin: The index file is not quite needed if we do this structure, but I guess we should talk about that separately anyway:

React
| — components
| ——— SideNav
| —————— SideNav.tsx (component)
| —————— SideNav.stories.tsx (story)
| —————— SideNav.spec.tsx (vitest)
| — modules
| ——— SideNavModule
| —————— SideNavModule.tsx (module)
| —————— SideNavModule.stories.tsx (story)
| —————— SideNavModule.spec.tsx (vitest)

jiromaykin: Outdated. I've used `export default class SideNavModule extends AbstractPage {`

src/open_inwoner/react/components/SideNav/SideNav.tsx
```diff
@@ -0,0 +1,36 @@
+import { FC } from 'react'
+import { SideNavigation } from '@gemeente-denhaag/side-navigation'
+import { MaterialIcon } from '../MaterialIcons/MaterialIcons'
```
stefrado: Small thing... might need to check this
```suggestion
import { MaterialIcon } from '@react/components/MaterialIcons/MaterialIcons'
```

src/open_inwoner/react/modules/SideNavModule/SideNavModule.tsx
```diff
@@ -0,0 +1,41 @@
+import { Root } from 'react-dom/client'
+import SideNav from '../../components/SideNav/SideNav'
```
stefrado: Please use the aliases we configured 😄 
```suggestion
import SideNav from '@react/components/SideNav/SideNav'
```

stefrado (changes requested):

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,197 @@
...
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
+
+from menus.menu_pool import menu_pool
+
+register = template.Library()
+logger = logging.getLogger(__name__)
+
+
+def get_extra_menu_items(context):
```
stefrado, line 14: Function to add FAQ menu item to react sidenav

src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -0,0 +1,197 @@
...
+            menu_items.append(menu_item)
+            logger.debug(
+                "Added menu item: %s -> %s (icon: %s)",
+                menu_item["label"],
+                menu_item["href"],
+                menu_item["icon"],
+            )
+
+        # Add extra items based on context/conditions
+        extra_items = get_extra_menu_items(context)
```
stefrado, line 182: Add FAQ item :)

stefrado (commented):
Some explanation...

stefrado:
These changes will close #1841

stefrado:
Might need to implement -> https://github.com/maykinmedia/open-inwoner/pull/1840/files#diff-c4d9886ee6d05123c909ad2cc0cd33496d44a1b85b90fdba04b312a536ca6c7b

To make sure the tests pass.

src/open_inwoner/cms/products/tests/test_plugin_categories.py
```diff
@@ -293,8 +293,6 @@ def test_only_published_categories_exist_in_home_page_when_anonymous(self):
                 self.published1,
                 self.published2,
                 self.published3,
```
swrichards, line 295: @stefrado this test is still failing, it's expecting `self.published4`, is that the expected behavior?

stefrado: Yess this is the expected behavior in this case, these are the categories shown for unauthenticated users. Since we do not show these users the sidenav, we show them 4 categories (just like before).

I'll revert these changes inside the test file to make sure it passes with the correct data.

swrichards (approved):
@stefrado LGTM, just needs a rebase

stefrado (approved):

PR #1839, opened 2025-07-23
[#3368] Support questions for non-authenticated users with OpenKlant

pi-sigma:
Support for questions from non-authenticated users submitted via the contactform is added by distinguishing between questions for a 'partij' and questions for a 'betrokkene' (a 'betrokkene' is any party involved in a case, a 'partij' is a formally/officially recognized party).

Taiga https://taiga.maykinmedia.nl/project/open-inwoner/us/3368


src/open_inwoner/openklant/services.py
```diff
@@ -1372,6 +1459,55 @@ def create_question(
...
+        klantcontact = self._create_klantcontact(question=question, subject=subject)
+        self._create_betrokkene_in_klantcontact(
+            klantcontact_uuid=klantcontact["uuid"],
+            betrokkene_data={"wasPartij": {"uuid": partij_uuid}},
+        )
+        self._create_interne_taak(klantcontact_uuid=klantcontact["uuid"])
+
+        return OpenKlant2Question.from_klantcontact_and_answer(klantcontact)
+
+    def create_question_for_betrokkene(
```
swrichards: Nitpick, but: you are  creating a question _and_ a betrokkene here. Perhaps we should reflect that in the name, e.g.:

```suggestion
    def create_question_with_betrokkene(
```

src/open_inwoner/openklant/views/contactform.py
```diff
@@ -197,25 +191,38 @@ def register_by_api(self, form, config: KlantenSysteemConfig):
...
+                phonenumber=phonenumber,
+            )
 
         self.log_system_action(
             "registered question via OpenKlant", user=self.request.user
         )
 
-        # TODO: get email from partij
-        return True, getattr(user, "email", None)
+        return True, email
```
swrichards, line 229: Given this signature, should we catch any exceptions here and return `False, None`?

src/openklant2/_resources/betrokkene.py
```diff
@@ -24,7 +28,12 @@ def create(
         response = self._post(self.base_path, data=data)
         return cast(Betrokkene, self.process_response(response))
 
-    def retrieve(self, /, uuid: str | uuid.UUID) -> Betrokkene:
+    def retrieve(
+        self,
+        /,
+        uuid: str | uuid.UUID,
+        params: Optional[BetrokkeneRetrieveParams] = None,
```
swrichards, line 35: Let's use new-style annotations:

```suggestion
        params: BetrokkeneRetrieveParams | None = None,
```

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -553,3 +556,139 @@ def __init__(self, identificatie):
...
+            email_addresses[0]["verstrektDoorBetrokkene"]["uuid"], betrokkene["uuid"]
+        )
+        self.assertIsNone(email_addresses[0]["verstrektDoorPartij"])
+
+        self.assertEqual(len(phone_addresses), 1)
+        self.assertEqual(phone_addresses[0]["adres"], "0612345678")
+        self.assertEqual(
+            phone_addresses[0]["verstrektDoorBetrokkene"]["uuid"], betrokkene["uuid"]
+        )
+        self.assertIsNone(phone_addresses[0]["verstrektDoorPartij"])
```
swrichards, line 659: It's a matter of preference, but perhaps we could just assert against the full list of expected objects rather than the individual asserts?

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -553,3 +556,139 @@ def __init__(self, identificatie):
...
+                question_kcm_uuid=klantcontact["uuid"],
+                question="A question from an anonymous user",
+                onderwerp="Anonymous inquiry",
+                kanaal=self.openklant2_config.mijn_vragen_kanaal,
+                taal="nld",
+                plaatsgevonden_op=QUESTION_DATE,
+            ),
+        )
+
+    def test_create_question_for_betrokkene_without_contact_info(self):
```
swrichards: Good to have this, but just to note that the form currently disallows this behavior (you have to fill in either an email or a phone number), and that might suggest doing the same for the service method. In our case, we are modelling a domain object (a question) and we know one of the contact details is required, our service should match that, even though the OpenKlant2 _API_ does allow this.

swrichards (changes requested):

src/open_inwoner/openklant/views/contactform.py
pi-sigma: Sounds right, we do the same for the esuite

src/open_inwoner/openklant/tests/test_openklant2_service.py
pi-sigma: Sounds good, but the objects (or dictionaries) contain uuid's which change with every VCR recording.

src/openklant2/_resources/betrokkene.py
swrichards: This one was missed?

src/openklant2/types/resources/partij.py
```diff
@@ -54,7 +55,7 @@ class CreatePartijIdentificatieOrganisatie(TypedDict):
 
 
 class CreatePartijPersoonData(CreatePartijDataBase):
-    soortPartij: Literal["persoon"]
+    soortPartij: Required[Literal["persoon"]]
```
swrichards: Is `Required` not standard?

swrichards: (I guess we did it above as well, I wonder if that's redundant)_

src/openklant2/types/resources/partij.py
```diff
@@ -26,14 +26,15 @@ class CreateContactnaam(TypedDict):
 
 # Note this is polymorphic, concrete types below
 class CreatePartijDataBase(TypedDict):
+    soortPartij: Required[str]
```
swrichards: I think this guy can have a literal (we have it defiend somewhere).

src/open_inwoner/openklant/services.py
```diff
@@ -1200,47 +1208,94 @@ def get_or_create_partij_for_user(self, user: User) -> tuple[Partij | None, bool
...
             for digitaal_adres in adressen
             if digitaal_adres["soortDigitaalAdres"] == soort_digital_adres
         ]
 
-    def get_or_create_digitaal_adres(
+    def filter_digitale_addressen_for_partij(
         self,
         partij_uuid: str,
+        *,
+        soort_digital_adres: str,
```
```suggestion
        soort_digitaal_adres: str,
```

Would a literal make sense here?

src/open_inwoner/openklant/services.py
```diff
@@ -1200,47 +1208,94 @@ def get_or_create_partij_for_user(self, user: User) -> tuple[Partij | None, bool
...
         *,
         soort_digital_adres: str,
         adressen: Iterable[DigitaalAdres] | None = None,
     ) -> list[DigitaalAdres]:
         if adressen is None:
-            adressen = self.retrieve_digitale_addressen_for_partij(partij_uuid)
+            if subject_type == "partij":
+                adressen = self.retrieve_digitale_addressen_for_partij(uuid)
+            else:  # betrokkene
+                adressen = self.retrieve_digitale_addressen_for_betrokkene(uuid)
```
swrichards: I think here we should go for the pattern of `match` with an [`assert_never` as fallthrough](https://typing.python.org/en/latest/guides/unreachable.html#assert-never-and-exhaustiveness-checking)

src/open_inwoner/openklant/services.py
swrichards: And if so, perhaps we can also do the `match`/`assert_never` pattern below.

swrichards (approved):
Just some small nitpicks left

src/openklant2/types/resources/partij.py
pi-sigma: Right, it's only funcionally useful in the context of inheritance with `total=False`, otherwise `Required` is the default.

src/open_inwoner/openklant/services.py
pi-sigma: I've added `Literal`, but I'm not sure about the `match` pattern  here. There's nothing to do in any of the cases, we would just `pass` for the successful ones and `assert_never` for the fall-through...

swrichards (approved):

PR #1845, opened 2025-07-30
[#3389] Sync user profiles with OpenKlant

pi-sigma:
- [x] Sync user profile data (email, standard phone number, altternative phone number) with OpenKlant after profile edits
- [x] Sync user data with OpenKlant on registration

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3389

Note that in the current implementation, the handling of contact information is inconsistent between the registration form and the code for writing to the Klant API's. See https://taiga.maykinmedia.nl/project/open-inwoner/issue/3392

### DIFF
--- a/src/open_inwoner/openklant/models.py
+++ b/src/open_inwoner/openklant/models.py
@@ -351,3 +351,16 @@ class KlantenSysteemConfig(SingletonModel):
             != KlantenServiceType.ESUITE.value
         )
         return self.contact_registration_enabled and contactform_subjects.exists()
+
+    def has_api_service_configured(
+        self, klanten_service_type: KlantenServiceType
+    ) -> bool:
+        match klanten_service_type:
+            case KlantenServiceType.ESUITE:
+                config = ESuiteKlantConfig.get_solo()
+                return getattr(config, "klanten_service", None) is not None
+            case KlantenServiceType.OPENKLANT2:
+                config = OpenKlant2Config.get_solo()
+                return getattr(config, "service", None) is not None
+            case _:
+                return False
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -49,6 +49,7 @@ from open_inwoner.openklant.models import (
     KlantContactMomentAnswer,
     OpenKlant2Config,
 )
+from open_inwoner.openklant.types import PartijUpdateData
 from open_inwoner.openklant.wrap import (
     contactmoment_has_new_answer,
     fetch_klantcontactmoment,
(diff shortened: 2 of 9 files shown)


src/open_inwoner/openklant/models.py
```diff
@@ -351,3 +351,14 @@ def has_contactform_configuration(self) -> bool:
             != KlantenServiceType.ESUITE.value
         )
         return self.contact_registration_enabled and contactform_subjects.exists()
+
+    def has_api_service_configured(self, klanten_service_type: str) -> bool:
```
```suggestion
    def has_api_service_configured(self, klanten_service_type: KlantenServiceType) -> bool:
```

src/open_inwoner/openklant/models.py
```diff
@@ -351,3 +351,14 @@ def has_contactform_configuration(self) -> bool:
...
+        match klanten_service_type:
```
swrichards, line 358: Note this check is also part of the service constructors: you could simply try to instantiate each service and handle the `ImproperlyConfigured` error.

src/open_inwoner/openklant/models.py
pi-sigma: The check uses the `value` from the constant though (a `str`)

swrichards: It could just past the enum though when it's called? We the enum precisely to avoid str comparisons I'd say.

src/open_inwoner/openklant/models.py
pi-sigma: I thought about that, but I find it strange to `try`/`except` for control flow at least in this case. The Klant system is not `ImproperlyConfigured` if the user doesn't want' to use one or more of the supported API's to begin with.

swrichards: Good point.

src/open_inwoner/openklant/tests/data.py
```diff
@@ -23,15 +26,22 @@
...
-    def setUpServices(cls):
-        config = ESuiteKlantConfig.get_solo()
-        config.klanten_service = ServiceFactory(
-            api_root=KLANTEN_ROOT, api_type=APITypes.kc
-        )
-        config.contactmomenten_service = ServiceFactory(
-            api_root=CONTACTMOMENTEN_ROOT, api_type=APITypes.cmc
-        )
-        config.save()
+    def setUpServices(cls, klanten_service_type: str = ""):
```
swrichards: Nitpick, but we can get some extra typing help with

```suggestion
    def setUpServices(cls, klanten_service_type: KlantenServiceType = KlantenServiceType.ESUITE):
```

src/open_inwoner/openklant/services.py
```diff
@@ -1295,37 +1295,64 @@ def update_user_from_partij(self, partij_uuid: str, user: User):
                 content_object=user,
             )
 
-    def update_partij_from_user(self, partij_uuid: str, user: User):
+    def _update_partij(
+        self,
+        partij_uuid: str,
+        user: User,
+        update_data: dict,
```
swrichards: Any chance we could make this a `TypedDict` to make clear our expectations here?

src/open_inwoner/openklant/services.py
```diff
@@ -1295,37 +1295,64 @@ def update_user_from_partij(self, partij_uuid: str, user: User):
...
                 content_object=user,
             )
 
+        return any(updated_fields)
+
+    def update_partij_from_user(
+        self,
+        partij_uuid: str,
+        user: User,
+        user_form_data: dict | None = None,
```
swrichards: I feel `user_form_data` is a bit tightly coupled to the specific call-site, and the method I guess is also slightly misnamed, since we say it's `..from_user` but we need additional data in the form of `user_form_data` and don't actually use `user` except for logging.

I think we have two paths:

- `update_user_from_partij`, which _only_ accepts a `user`, not other fields and;
- `update_partij_details`, which accepts a uuid and a dict.

I think the intent is better captured of we split those use-cases.

swrichards (changes requested):

swrichards (approved):

PR #1852, opened 2025-08-07
fix: properly handle question zaak as instance rather than list

swrichards:
[Taiga 3396](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3396)

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -748,15 +748,7 @@ class eSuiteVragenService(KlantenService):
         if not (kcm := self.fetch_klantcontactmoment(question_uuid, **fetch_params)):
             return None, None
 
-        local_kcm, is_created = KlantContactMomentAnswer.objects.get_or_create(  # noqa
-            user=user, contactmoment_url=kcm.contactmoment.url
-        )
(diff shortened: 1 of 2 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -771,17 +763,22 @@ def retrieve_question(
                     "Unable to find matched contactmomenten zaak in any zgw backend"
                 )
             else:
```
pi-sigma, line 765:
```suggestion
            elif case_count > 1
```
The two conditions belong to the same logical level

pi-sigma (approved):

src/open_inwoner/openklant/services.py
swrichards: But if `case_count != 0` I always want to run the iterator to fetch the first (including `case_count == 1`), but only log the error if `case_count > 1`, so the `else` seemed to make sense to me. It would then have to be `elif case_count >= 1` and I'd _still_ need the extra conditional. The `else` seemed cleaner. Or am I missing something?

pi-sigma: You're right, this makes sense.

PR #1853, opened 2025-08-07
fix: limit openklant2 zaak-questions to those initiated by user

swrichards:
[Taiga 3395](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3395)

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1041,6 +1041,18 @@ class OpenKlant2Service(
 
         return self.client.partij.retrieve(vestiging_pi["identificeerdePartij"]["uuid"])
 
+    def get_partij_for_user(self, user: User) -> Partij | None:
+        if user.kvk:
+            return self.find_organisatie_for_kvk_and_vestiging(
+                kvk=user.kvk,
+                vestigingsnummer=user.vestiging,
+            )
+
+        if user.bsn:
+            return self.find_persoon_for_bsn(user.bsn)
+
+        return None
+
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1041,6 +1041,20 @@ def find_organisatie_for_kvk_and_vestiging(
 
         return self.client.partij.retrieve(vestiging_pi["identificeerdePartij"]["uuid"])
 
+    def get_partij_for_user(self, user: User) -> Partij | None:
+        if user.kvk:
```
pi-sigma, line 1045:
```
if user.kvk:
    return self.find_organisatie_for_kvk_and_vestiging(
    kvk=user.kvk,
    vestigingsnummer=user.vestiging,
    )
```

Shouldn't be necessary to check `if organisatie` since both this and the called method return `Partij | None`.

src/open_inwoner/openklant/services.py
```diff
@@ -1041,6 +1041,20 @@ def find_organisatie_for_kvk_and_vestiging(
...
+            if organisatie := self.find_organisatie_for_kvk_and_vestiging(
+                kvk=user.kvk,
+                vestigingsnummer=user.vestiging,
+            ):
+                return organisatie
+
+        if user.bsn:
+            if persoon := self.find_persoon_for_bsn(user.bsn):
```
pi-sigma:
```suggestion
            return self.find_persoon_for_bsn(user.bsn)
```
See previous comment

pi-sigma (approved):

PR #1856, opened 2025-08-12
[#3367]  Fetch and display tasks with link to external form

pi-sigma:
API models for 'externe taken' are modified to reflect changes in the API schema for tasks with external form; the plugin code for rendering 'taken' is updated accordingly.

The configuration model for the 'TasksPlugin' now has two fields for objecttype version: one for the new 'externformulier taak', one for the old 'url taak'

Bugfix: the objecttype `UUID` was incorrectly retrieved from the `TasksConfig` class instead of an instance, which resulted in a `DeferredAttribute` where we expect a string.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/us/3367


src/open_inwoner/cms/plugins/cms_plugins/tasks.py
```diff
@@ -34,65 +35,98 @@ def render(self, context, instance, placeholder) -> dict[str, str]:
...
+            if task.soort == "externformulier":
+                task = cast(ExternFormulierTaak, task)
+                base_url = task.portaalformulier.formulier.value
+                param = {"initial_data_reference": task.uuid}
+                task_url = f"{base_url}?{urlencode(param)}"
+            elif task.soort == "url":
+                task = cast(UrlTaak, task)
+                task_url = str(task.task_url.uri)
+            else:
+                task_url = None
```
swrichards: I think it'd be cleaner if we just use `match` here, given we can check the instance. The `soort` still feels a bit implicit to me, let's look directly at the types:


```suggestion
           match task:
               case ExternFormulierTaak():
                base_url = task.portaalformulier.formulier.value
                param = {"initial_data_reference": task.uuid}
                task_url = f"{base_url}?{urlencode(param)}"
              case UrlTaak():
                task_url = str(task.task_url.uri)
            else:
                assert_never(task)
```

src/open_inwoner/cms/plugins/cms_plugins/tasks.py
```diff
@@ -34,65 +35,98 @@ def render(self, context, instance, placeholder) -> dict[str, str]:
...
+                    task.verloopdatum if task.verloopdatum is not None else "",
                     format="DATETIME_FORMAT",
                     use_l10n=True,
                 ),
-                "identificatie": task.identificatie.value,
                 "koppeling": task.koppeling,
                 "verwerker_taak_id": str(task.verwerker_taak_id),
                 "eigenaar": task.eigenaar,
+                "task_url": task_url,
             }
```
swrichards: I think there'd be value in having a `TypedDict` for this, just to guard against accidental signature changes.

src/open_inwoner/cms/plugins/api_models/taak.py
```diff
@@ -60,27 +90,59 @@ class Config:
...
-    identificatie: Union[IdentificatieBSN, IdentificatieKVK]
-    koppeling: Optional[Union[KoppelingZaak, KoppelingProduct]]
     verwerker_taak_id: UUID
     eigenaar: str
-    url: Optional[TaakUrl] = None
+    toelichting: str | None = None
+    doorlooptijd: str | None = None
+    verloopdatum: datetime | None = None
+    koppeling: Union[KoppelingZaak, KoppelingProduct] | None = None
+    identificatie: Union[IdentificatieBSN, IdentificatieKVK]
```
swrichards: nitpick, but we can use the ` | ` operator here for more modern Python idioms.

swrichards: (I suspect there's a ruff rule for this).

src/open_inwoner/cms/plugins/cms_plugins/tasks.py
```diff
@@ -34,65 +35,98 @@ def render(self, context, instance, placeholder) -> dict[str, str]:
...
+                    yield UrlTaak(**taak_data)
+                case _:
+                    assert_never(taak_data["soort"])
+
+    def get_tasks_by_bsn(
+        self, instance, user_bsn
+    ) -> Generator[ExternFormulierTaak | UrlTaak, None, None]:
+        tasks = self.get_tasks(instance)
+
+        for task in tasks:
```
swrichards, line 118: As above, use `match`

swrichards (changes requested):

src/open_inwoner/cms/plugins/cms_plugins/tasks.py
```diff
@@ -34,65 +46,101 @@ def render(self, context, instance, placeholder) -> dict[str, str]:
...
+            taak_data["url"] = obj.url
+            taak_data["uuid"] = obj.uuid
+
+            match taak_data["soort"]:
+                case "externformulier":
+                    yield ExternFormulierTaak(**taak_data)
+                case "url":
+                    yield UrlTaak(**taak_data)
+                case _:
+                    assert_never(taak_data["soort"])
```
swrichards: Sorry to be a stickler about this, but I find myself tripping over the pattern of using `soort` to determine the type, it feels fragile and not well capturing the intent. What about:

```suggestion
        factory_map: dict[str, ExternFormulierTaak | UrlTaak] = {}
        if instance.object_type_dimpact:
            factory_map[instance.object_type_dimpact] = ExternFormulierTaak
        if instance.object_type_generieke_dienstverlening:
            factory_map[instance.object_type_generieke_dienstverlening] = UrlTaak

        for object_type_uuid, obj_factory in factory_map.items():
            for obj in objects_api_client.get_objects(
                    object_type_uuid=object_type_uuid
            ):
                taak_data = obj.record["data"] | {"url": obj.url, "uuid": obj.uuid}
                try:
                    yield obj_factory.validate_python(taak_data)
                except ValidationError:
                    logger.exception("Invalid externe taak", extra={"object_type_uuid": object_type_uuid})

```

This also explicitly uses Pydantic's `validate_python` for strong typing (but it would be good to have a simple test that shows a bad object type won't break the whole site).

swrichards (changes requested):

src/open_inwoner/cms/plugins/cms_plugins/tasks.py
pi-sigma: Nice pattern! I worry that if a staff member chooses the wrong object type (e.g. 'afdeling'), this will produce a server error because of failed type validation. That's why I checked with `taak_data.get("soort")` though I'm not sure this would be sufficient either. Anything could come from the Objects API...

swrichards: Oh no, I edited my comment and it might not have come through! I also feared that, but that's why we're using Pydantic -- so I added a `try/except` bloc with a logger call to be sure, might be worth adding in:

```
                try:
                    yield obj_factory.validate_python(taak_data)
                except ValidationError:
                    logger.exception("Invalid externe taak", extra={"object_type_uuid": object_type_uuid})

```

swrichards (approved):

PR #1859, opened 2025-08-18
feat(notifications): add an admin action to deregister webhooks

swrichards:
[Taiga #3410](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3410).

### DIFF
--- a/src/notifications/admin.py
+++ b/src/notifications/admin.py
@@ -14,6 +14,13 @@ class NotificationsConfigAdmin(admin.ModelAdmin):
 def register_webhook(modeladmin, request, queryset):
     for sub in queryset:
         if sub._subscription:
+            messages.warning(
+                request,
+                _(
+                    "Skipping {subscription} because it has already already been"
+                    " registered."
+                ).format(subscription=sub),
+            )
             continue
 
         try:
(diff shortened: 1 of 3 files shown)


src/notifications/admin.py
```diff
@@ -31,7 +31,27 @@ def register_webhook(modeladmin, request, queryset):
 register_webhook.short_description = _("Register the webhooks")  # noqa
 
 
+def deregister_webhook(modeladmin, request, queryset):
+    for sub in queryset:
+        if not sub._subscription:
+            continue
```
pi-sigma: The `deregister` function checks this as well, so it's not necessary here (though doesn't hurt either)

src/notifications/tests/test_webhook.py
```diff
@@ -107,3 +107,106 @@ def test_register_webhook_request_exception(self):
...
+            notifications_api_config=self.config,
+            callback_url="https://example.com/callback",
+            client_id="client_id",
+            secret="secret",
+            channels=["zaken"],
+            _subscription="",
+        )
+
+        # Should not raise an exception
+        subscription.deregister()
```
pi-sigma, line 165: Perhaps an assertion that the `delete` method was not called, since we explicitly try to avoid that:

```
@requests_mock.Mocker()
def test_deregister_webhook_no_subscription(self, m):
    m.delete("")
    ...
    self.assertFalse(m.called)
```

pi-sigma (approved):

src/notifications/admin.py
swrichards: Right, which actually suggests it's better to move this to a queryset filter.

swrichards: Solved this differently, used the extra check to add a warning feedback to the user that we're skipping (and why).

src/notifications/tests/test_webhook.py
swrichards: Good point, updated the other tests too to test the mocks.

PR #1866, opened 2025-08-25
[dimpact #297] Prevent deletion of primary website

pi-sigma:
Instead of preventing deletion of the last website, we now prevent adding sites if one already exists (there is no use case for multiple sites) and prevent deletion altogether.

Refs: https://taiga.maykinmedia.nl/project/dimpact-enschede-ssc-ict-1/issue/297

### DIFF
--- a/src/open_inwoner/configurations/admin.py
+++ b/src/open_inwoner/configurations/admin.py
@@ -1,15 +1,12 @@
 from django import forms
 from django.contrib import admin, messages
-from django.contrib.admin.actions import delete_selected as default_delete_selected
 from django.contrib.flatpages.admin import FlatPageAdmin
 from django.contrib.sites.admin import SiteAdmin
 from django.contrib.sites.models import Site
 from django.core import exceptions
+from django.core.exceptions import ValidationError
 from django.core.validators import URLValidator
-from django.forms import ValidationError
-from django.http import HttpResponseRedirect
-from django.shortcuts import redirect
-from django.urls import resolve, reverse
+from django.urls import resolve
 from django.urls.exceptions import Resolver404
 from django.utils.html import format_html, format_html_join
 from django.utils.translation import gettext_lazy as _
@@ -26,40 +23,14 @@ from ..utils.iteration import split
 from .models import CustomFontSet, SiteConfiguration, SiteConfigurationPage
 
 
-@admin.action(description=_("Delete selected websites"))
-def delete_selected(modeladmin, request, queryset):
-    """
-    Override `delete_selected` action to prevent accidental deletion of all websites
-    """
-    if queryset.count() == Site.objects.all().count():
-        messages.add_message(
-            request,
-            messages.WARNING,
-            _("You cannot delete all websites; at least one site must remain."),
-        )
-        return HttpResponseRedirect(reverse("admin:sites_site_changelist"))
-    return default_delete_selected(modeladmin, request, queryset)
-
-
 class CustomSiteAdmin(SiteAdmin):
-    """
-    Overrides `SiteAdmin` to prevent deletion of last website
-    """
-
     model = Site
-    actions = [delete_selected]
 
-    def delete_model(self, request, obj):
-        if self.model.objects.count() < 2:
-            messages.add_message(
-                request, messages.WARNING, _("You cannot delete the last site.")
-            )
-            return redirect("admin:sites_site_changelist")
-        else:
-            super().delete_model(request, obj)
+    def has_add_permission(self, request):
+        return not Site.objects.exists()
 
-    class Media:
-        css = {"all": ("css/admin/admin_overrides.css",)}
+    def has_delete_permission(self, request, obj=None):
+        return False
 
 
 # re-register `Site` with our CustomSiteAdmin
--- a/src/open_inwoner/configurations/tests/test_admin.py
+++ b/src/open_inwoner/configurations/tests/test_admin.py
@@ -4,7 +4,6 @@ from django.utils.translation import gettext_lazy as _
 
 from django_webtest import WebTest
 from maykin_2fa.test import disable_admin_mfa
-from pyquery import PyQuery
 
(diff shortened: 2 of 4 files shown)


src/open_inwoner/configurations/admin.py
```diff
@@ -29,13 +29,14 @@
 @admin.action(description=_("Delete selected websites"))
 def delete_selected(modeladmin, request, queryset):
     """
-    Override `delete_selected` action to prevent accidental deletion of all websites
+    Override `delete_selected` action to prevent accidental deletion of the primary website
     """
-    if queryset.count() == Site.objects.all().count():
+    primary_site = Site.objects.get(id=1)
```
alextreme: I'd recommend to use settings.SITE_ID instead of hardcoding this

pi-sigma: Good call, though now I've revamped the code completely. If we have no use for more than one site, we can simply block adding more sites and then block deleting altogether (at least via the admin), so there's no need to fish around for the site id.

alextreme (approved):
Seems okay, but let @swrichards do a double-check

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -19,89 +18,43 @@ class TestAdminSite(WebTest):
     def setUpTestData(cls):
         cls.user = UserFactory(is_superuser=True, is_staff=True)
 
-    def test_delete_site_success(self):
-        site2 = Site.objects.create(id=2, domain="example2.com", name="example2")
+    def test_site_add_prevented_if_one_exists(self):
+        """Test that adding a site is prevented if one already exists."""
+        # Make sure we have exactly one site
+        initial_site_count = Site.objects.count()
+        if initial_site_count == 0:
```
swrichards: Shouldn't we assert that it's `0`?

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -19,89 +18,43 @@ class TestAdminSite(WebTest):
...
-        doc = PyQuery(response.content)
+        self.assertEqual(Site.objects.count(), initial_site_count or 1)
 
-        success_msg = doc.find(".success").text()
-        self.assertIn("met succes verwijderd.", success_msg)
+    def test_site_deletion_prevented(self):
+        """Test that site deletion is prevented."""
+        # Make sure we have exactly one site
+        site = Site.objects.first()
+        if not site:
```
swrichards: As above, the conditional above is a bit odd: we should know the initial state of our tests, no?

src/open_inwoner/configurations/admin.py
```diff
@@ -26,40 +23,14 @@
...
-    def delete_model(self, request, obj):
-        if self.model.objects.count() < 2:
-            messages.add_message(
-                request, messages.WARNING, _("You cannot delete the last site.")
-            )
-            return redirect("admin:sites_site_changelist")
-        else:
-            super().delete_model(request, obj)
+    def has_add_permission(self, request):
+        return Site.objects.count() == 0
```
swrichards: I think this better captures the intent (and is theoretically faster because it has to iterate over fewer rows, though that won't really be an issue here given there's only ever going to be a small number of rows):

```suggestion
        return Site.objects.all().exists() is False
```

swrichards (changes requested):

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -19,89 +18,39 @@ class TestAdminSite(WebTest):
     def setUpTestData(cls):
         cls.user = UserFactory(is_superuser=True, is_staff=True)
 
-    def test_delete_site_success(self):
-        site2 = Site.objects.create(id=2, domain="example2.com", name="example2")
+    def test_site_add_prevented_if_one_exists(self):
+        """Test that adding a site is prevented if one already exists."""
+        # Make sure we have exactly one site
+        initial_site_count = Site.objects.count()
```
swrichards, line 24: This needs an assert?

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -19,89 +18,39 @@ class TestAdminSite(WebTest):
...
-        # check message
-        doc = PyQuery(response.content)
+        self.assertEqual(Site.objects.count(), initial_site_count or 1)
 
-        success_msg = doc.find(".success").text()
-        self.assertIn("met succes verwijderd.", success_msg)
+    def test_site_deletion_prevented(self):
+        """Test that site deletion is prevented."""
+        # Make sure we have exactly one site
+        site = Site.objects.first()
```
swrichards:
```suggestion
        site = Site.objects.get()
```

So it'll raise if it's not 1.

src/open_inwoner/configurations/admin.py
```diff
@@ -26,40 +23,14 @@
...
-    def delete_model(self, request, obj):
-        if self.model.objects.count() < 2:
-            messages.add_message(
-                request, messages.WARNING, _("You cannot delete the last site.")
-            )
-            return redirect("admin:sites_site_changelist")
-        else:
-            super().delete_model(request, obj)
+    def has_add_permission(self, request):
+        return not Site.objects.exists()
```
swrichards, line 30: Double check, this doesn't need an extra `.all()`?


```suggestion
        return not Site.objects.all().exists()
```

swrichards (changes requested):
LGTM, sorry to be tedious, but in light of the New Regime (tm): could you add a note to the changelog, and a single line to the docs (noting that at least one Site is needed).

src/open_inwoner/configurations/admin.py
pi-sigma: They're equivalent. Double checked: `.all()` returns a copy of the queryset, and `.all().exists()` is equivalent to calling `.exists()` on the manager itself. 

#one-obvious-way-to-do-it

swrichards (approved):

PR #1867, opened 2025-08-25
[#3419] Implement Storybook and tests for MaterialIcons react component.

stefrado:
## User Story 
https://taiga.maykinmedia.nl/project/open-inwoner/task/3419

## Description
Implement material icon sb and test file

### DIFF
--- a/src/open_inwoner/react/components/MaterialIcons/MaterialIcons.stories.tsx
+++ b/src/open_inwoner/react/components/MaterialIcons/MaterialIcons.stories.tsx
@@ -1,0 +1,127 @@
+import { Meta, StoryObj } from '@storybook/react'
+import { MaterialIcon } from './MaterialIcons'
+import 'material-icons/iconfont/material-icons.css'
+
+type ExtendedMateriaIconProps = typeof MaterialIcon & { nameTool: string }
+
+const meta: Meta<ExtendedMateriaIconProps> = {
+  title: 'React/MaterialIcons',
+  component: MaterialIcon,
+  parameters: {
+    layout: 'padded',
+    docs: {
+      description: {
+        component: `
+The MaterialIcon component renders Google Material Icons using the outlined icon font.
+
+This component provides a consistent way to display Material Design icons throughout the application. All icons are rendered with proper accessibility attributes and use the material-icons-outlined font family.
+
+**Props:**
+- \`name\`: The name of a Material Icon (e.g., 'home', 'search', 'settings')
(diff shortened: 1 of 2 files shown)


src/open_inwoner/react/components/MaterialIcons/MaterialIcons.stories.tsx
```diff
@@ -0,0 +1,236 @@
...
+    name: 'settings',
+  },
+}
+
+export const NotificationsIconExample: Story = {
+  name: 'Notifications Icon Example',
+  parameters: {
+    docs: {
+      description: {
+        story: 'Notifications or alerts functionality icon.',
```
jiromaykin: We don't actually use this icon anywhere in the project :-) but that's OK.

src/open_inwoner/react/components/MaterialIcons/MaterialIcons.stories.tsx
```diff
@@ -0,0 +1,236 @@
+import { Meta, StoryObj } from '@storybook/react'
+import { MaterialIcon } from './MaterialIcons'
+import 'material-icons/iconfont/material-icons.css'
+
+const meta: Meta<typeof MaterialIcon> = {
+  title: 'React/MaterialIcons',
+  component: MaterialIcon,
+  parameters: {
+    layout: 'padded',
+    docs: {
```
jiromaykin, line 12: Instead of having all those separate stories-components for icons that do not really add much of value in a separate window, another option could be to create a bit of a drop down like so:

```
argTypes: {
  name: {
    control: { type: 'select' },
    options: ['home', 'search', 'settings', 'person', 'mail', 'description', 'inbox', 'inventory_2', 'group', 'help_outline', 'euro']
  }
}
```

Then you will need only 1 icon story and you can select different names. Maybe call it "Icon Showcase" .

Also maybe later we can add font-sizes and color variations.

jiromaykin (changes requested):
Not really a 'requested' change but more of a 'suggestion', so it's actually all quite good already.

src/open_inwoner/react/components/MaterialIcons/MaterialIcons.stories.tsx
stefrado: Nice suggestion, I'll implement the dropdown.

The color and font-size etc can later be added once these become requirements for this component.

jiromaykin (approved):

PR #1868, opened 2025-08-25
[#3375] Make long email adresses break properly in cards

jiromaykin:
Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3375 

CSS 'hyphens' is perfect for adding a hyphen and breaking very long titles in the OIP location cards,
but: it is forbidden to be used in emailaddresses because adding a hyphen there. is bad for UX (if people see the address on desktop but want to type it on their phone e.g.):
https://ux.stackexchange.com/questions/31764/hyphenating-urls-and-email-addresses-that-would-overflow-their-container.

Also: hyphens do not work BEFORE an @ sign (IF we would not use break-all in mailaddresses):

Screenshot 2025-08-28 at 11 39 06

### DIFF
--- a/src/open_inwoner/components/templates/components/Card/LocationCard.html
+++ b/src/open_inwoner/components/templates/components/Card/LocationCard.html
@@ -11,7 +11,7 @@
                 {% link href='tel:'|addstr:phonenumber secondary=True text=phonenumber %}
             {% endif %}
             {% if email %}
-                {% link href='mailto:'|addstr:email secondary=True text=email %}
+                {% link href='mailto:'|addstr:email secondary=True text=email extra_classes="link--mailto" %}
             {% endif %}
         </div>
     </div>
--- a/src/open_inwoner/scss/components/Typography/Link.scss
+++ b/src/open_inwoner/scss/components/Typography/Link.scss
@@ -25,6 +25,10 @@
     text-align: right;
   }
 
+  &--mailto {
+    word-break: break-all;
+  }
+
   *[class*='icons'],
   *[class*='Icons'] {
     font-size: 24px;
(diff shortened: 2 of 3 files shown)


src/open_inwoner/components/templates/components/Card/LocationCard.html
```diff
@@ -11,7 +11,7 @@ <h3 class="utrecht-heading-3 card__heading-3">{% link href=location.get_absolute
                 {% link href='tel:'|addstr:phonenumber secondary=True text=phonenumber %}
             {% endif %}
             {% if email %}
-                {% link href='mailto:'|addstr:email secondary=True text=email %}
+                {% link href='mailto:'|addstr:email secondary=True text=email extra_classes="link--mailaddress" %}
```
stefrado: `link--mailto` might be a better classname since this is in line with what type of link href this is.

```suggestion
                {% link href='mailto:'|addstr:email secondary=True text=email extra_classes="link--mailto" %}
```

src/open_inwoner/scss/components/Typography/Link.scss
```diff
@@ -25,6 +25,10 @@
     text-align: right;
   }
 
+  &--mailaddress {
```
stefrado: Ik other feedback is accepted this needs to be changed to `&--mailto`

stefrado (commented):
You could change the classname besides that LGTM

stefrado (approved):

src/open_inwoner/components/templates/components/Card/LocationCard.html
jiromaykin: OK, will do!

PR #1872, opened 2025-08-26
[#3406] Turn contact form into CMS page

pi-sigma:
The contact form is turned into a CMS page. The logic for handling form submissions is kept inside the openklant "ContactFormView" to minimize coupling with the CMS.

Refs: https://taiga.maykinmedia.nl/project/open-inwoner/us/3406


src/open_inwoner/templates/cms/contactform/form_inner.html
```diff
@@ -0,0 +1,12 @@
+{% load cms_tags i18n render_tags %}
+{% block content %}
+    {% if has_form_configuration %}
+        <h1 class="utrecht-heading-1">{% trans "Contactformulier" %}</h1>
+        <p class="utrecht-paragraph">{{ form_description|ckeditor_content|safe }}</p>
+        {% include "components/Contact/ContactForm.html" with form_object=form has_form_configuration=has_form_configuration csrf_token=csrf_token only %}
+    {%else %}
+    <p class="utrecht-paragraph">
+      {% trans "Contact formulier niet geconfigureerd." %}
```
swrichards: I feel this is a bit of an anti-pattern: is there a way we can make the URL registration conditional on the form being configured?

src/open_inwoner/openklant/cms_plugins.py
```diff
@@ -8,16 +8,19 @@
...
 
-#
-# contact form plugin
-#
-
 
 class ContactFormConfig(CMSPlugin):
-    description = models.TextField(
-        _("Description"),
+    description_auth = models.TextField(
```
swrichards: I am happy for more verbose names:

```suggestion
    description_authenticated_user = models.TextField(
```

src/open_inwoner/openklant/cms_plugins.py
```diff
@@ -8,16 +8,19 @@
...
+        _("Description authenticated users"),
+        blank=True,
+        help_text=_("Description of the contact form for authenticated users"),
+    )
+    description_anon = models.TextField(
```
swrichards:
```suggestion
    description_anonymous_user = models.TextField(
```

src/open_inwoner/openklant/cms_apps.py
```diff
@@ -0,0 +1,13 @@
+from django.utils.translation import gettext_lazy as _
+
+from cms.app_base import CMSApp
+from cms.apphook_pool import apphook_pool
+
+
+@apphook_pool.register
+class OpenklantApphook(CMSApp):
```
swrichards, line 8: I guess this matches the module name, but perhaps something referring to `ContactFormulier` would be a clearer name?


```suggestion
class ContactFormulierApphook(CMSApp):
```

src/open_inwoner/cms/footer/cms_plugins.py
```diff
@@ -37,6 +37,12 @@ def render(self, context, instance, placeholder):
                 ):
                     cms_pages.remove(page)
 
+        contactform_page_qs = Page.objects.filter(
+            application_namespace="openklant", publisher_is_draft=False
+        )
+        if contactform_page_qs and klant_config.has_contactform_configuration:
```
swrichards:
```suggestion
        if contactform_page_qs.exists() and klant_config.has_contactform_configuration:
```

swrichards (changes requested):

src/open_inwoner/openklant/cms_apps.py
pi-sigma: I get your point, since we're only creating this for the contact form. However, the name doesn't just reflect the module name, but, by convention, the name of the registered Django app for which the hook is created ("openklant" in this case).

swrichards: In that case we go for consistency. Resolving.

src/open_inwoner/templates/cms/contactform/form_inner.html
pi-sigma: Not directly. The URLs are loaded at startup, if we include them conditionally, changes to the config won't have any effect.

Alternative: always include the URL's, but the view restricts access based on whether the contact form is configured or not (returns `404` in case the form is not configured). Then we don't need this check in the template.

swrichards: I would prefer that approach to be honest, I think clients that do not wish to use this feature should have no external exposure of it (we could add a Timeline log entry just to warn admins that the configuration is disabled but somebody still tried to access it).

src/open_inwoner/cms/footer/cms_plugins.py
```diff
@@ -37,6 +37,12 @@ def render(self, context, instance, placeholder):
                 ):
                     cms_pages.remove(page)
 
+        contactform_page_qs = Page.objects.filter(
+            application_namespace="openklant", publisher_is_draft=False
+        )
+        if contactform_page_qs.exists() and klant_config.has_contactform_configuration:
+            cms_pages.insert(0, contactform_page_qs.get())
```
swrichards: Due to my lack of ready CMS knowledge: what guarantees there will be only one element in the queryset?

src/open_inwoner/openklant/migrations/0028_remove_contactformconfig_description_and_more.py
```diff
@@ -0,0 +1,34 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("openklant", "0027_alter_klantensysteemconfig_primary_backend"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
swrichards: Do we need a data migration here, to move the old description to one of the two other descriptions?

src/open_inwoner/openklant/views/contactform.py
```diff
@@ -120,6 +128,17 @@ def set_result_message(self, success: bool):
...
+        # Double check if the form is configured
+        if not self.klanten_config.has_contactform_configuration:
+            messages.add_message(
+                self.request,
+                messages.ERROR,
+                _(
+                    "Contact formulier is niet geconfigureerd. Neem contact op met de beheerder."
+                ),
+            )
+            return self.form_invalid(form)
```
swrichards: Presumably this will be taken care of by the check in `dispatch` ?

swrichards (changes requested):

src/open_inwoner/cms/footer/cms_plugins.py
pi-sigma: You're right. And it's better to query the CMS pages via the `ContactFormPlugin` plugin and the placeholder(s) containing the plugin.

src/open_inwoner/openklant/migrations/0028_remove_contactformconfig_description_and_more.py
pi-sigma: Sounds right. Which one though? Or we could pre-fill *both* of the new descriptions with the old text, since we don't know which one to choose, and explain this in the changelog.

swrichards: Both sounds sensible.

src/open_inwoner/openklant/views/contactform.py
pi-sigma: This was one of the confusing bits. I removed the `has_contactform_configuration` method altogether, along with this check and the tests for it. I think the more permissive `contact_registration_enabled` is all that's needed: missing esuite subject codes are handled by the `_get_kcm_subject` helper in the esuite service, and the openklant service doesn't need the subject at all.

src/open_inwoner/openklant/migrations/0029_migrate_contactformconfig_descriptions.py
```diff
@@ -0,0 +1,27 @@
...
+
+def copy_contactform_description(apps, _):
+    ContactFormConfig = apps.get_model("openklant", "ContactFormConfig")
+
+    for config in ContactFormConfig.objects.all():
+        old_description = config.description
+
+        config.description_authenticated_user = old_description
+        config.description_anonymous_user = old_description
+        config.save()
```
swrichards: I think it'd be good to merge these into a single migration so it's all applied in a transaction, would make things a bit easier to follow if things go wrong on production.

swrichards (approved):
LGTM, just one comment about squashing the migrations.

src/open_inwoner/openklant/migrations/0029_migrate_contactformconfig_descriptions.py
pi-sigma: I thought it's good to separate schema and data migrations, but the point about transactions makes sense, didn't consider that. I'll merge these into one.

swrichards (approved):

CHANGELOG.rst
```diff
@@ -31,6 +31,10 @@ Nieuwe features
 * [:taiga-us:`3370`, :pr:`1837`, :pr:`1869`]: Het navigatiemenu is nu conform het NL Design
   System op de meeste pagina's aan de linkerkant van het scherm geplaatst. Sommige
   detailpagina's behouden voorlopig het dropdownmenu in afwachting van nieuwe designs.
+* [:taiga-us: `3406`, :pr:`1872`]: Het contactformulier wordt omgezet in een CMS-pagina.
+  In plaats van één beschrijving heeft het contactformulier nu twee beschrijvingen, één voor
+  geauthenticeerde en één voor anonieme gebruikers. De oude beschrijving wordt naar beide
+  nieuwe velden gekopieerd. Pas één of beide beschrijvingen naar behoefte aan.
```
swrichards: This needs to move up in the running changelog section: `1.34` is already released.

swrichards (changes requested):

CHANGELOG.rst
```diff
@@ -6,9 +6,13 @@ Voor een volledig overzicht van alle commits, zie ...
 Nieuwe features
 ---------------
 
-* [:taiga-us:`3370`, :pr:`1888`]: Navigatiemenu-items correct weergegeven op verschillende schermbreedtes; 
-  ``@gemeente-denhaag/side-navigation`` bijgewerkt naar ``4.0.2``, waardoor notificatiebadges in 
+* [:taiga-us:`3370`, :pr:`1888`]: Navigatiemenu-items correct weergegeven op verschillende schermbreedtes;
+  ``@gemeente-denhaag/side-navigation`` bijgewerkt naar ``4.0.2``, waardoor notificatiebadges in
```
swrichards: Just curious: where do the formatting gremlins come from?

(Annoying about RST, no auto-formatters to prevent this, see also below).

pi-sigma: Hm, that's annoying indeed. Must be my editor doing something funky. I'll double-check.

CHANGELOG.rst
```diff
@@ -6,9 +6,13 @@ Voor een volledig overzicht van alle commits, zie ...
...
   navigatiemenu-items de rode indicator niet meer tonen.
+* [:taiga-us: `3406`, :pr:`1872`]: Het contactformulier wordt omgezet in een CMS-pagina.
+  In plaats van één beschrijving heeft het contactformulier nu twee beschrijvingen, één voor
+  geauthenticeerde en één voor anonieme gebruikers. De oude beschrijving wordt naar beide
+  nieuwe velden gekopieerd. Pas één of beide beschrijvingen naar behoefte aan.
```
swrichards, line 15: Can you add in a new section `Deployment notes`  (See https://github.com/maykinmedia/open-inwoner/pull/1897/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR6-R10) with content:

```
* Het contactformulier is omgezet naar een CMS pagina. Verifieer na deployment de beschrijvingen van het contactformulier.
```

CHANGELOG.rst
swrichards, line 1: This will need a documentation update: do you mind creating a task for this in your story so we don't forget?

swrichards (changes requested):

CHANGELOG.rst
pi-sigma: My editor automatically removes white space at the end of a line because of our `.editorconfig` which has `trim_trailing_whitespace = true`. I'll revert and add this change to our `.git-blame-ignore-revs`.

swrichards (approved):

PR #1873, opened 2025-08-26
✨ [#3418] Add custom scripts to Head

jiromaykin:
Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3407

TODO

- [x] add documentation
- [x] add changelog
- [x] we'll need to create an issue in http://github.com/maykinmedia/charts/ to add this to the Helm charts for OIP.

I have added the 'add dangerously' javascript options inside "**Geavanceerde opties**" in Admin.


src/open_inwoner/utils/context_processors.py
```diff
@@ -82,6 +82,9 @@ def settings(request):
         "cookie_link_text": config.cookie_link_text,
         "cookie_link_url": config.cookie_link_url,
         "extra_css": config.extra_css,
+        "custom_javascript_url": config.custom_javascript.url
+        if config.custom_javascript
```
swrichards, line 86: I think the confirmation should also be required to actually show the file:
```suggestion
        if config.custom_javascript and config.custom_javascript_confirmed
```

src/open_inwoner/configurations/admin.py
stefrado, line 1: r. 98 + 1

add -> "custom_javascript": forms.FileInput(attrs={"accept": ".js"}),

Make sure we can only select .js files in the file picker.

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +358,46 @@ def _get_column(props):
...
+    def javascript_preview(self, obj):
+        """Show preview of the uploaded script"""
+        if obj.custom_javascript:
+            try:
+                obj.custom_javascript.seek(0)
+                content = obj.custom_javascript.read().decode("utf-8")
+                obj.custom_javascript.seek(0)
+
+                # Show first 300 characters
+                preview = content[:300] + ("..." if len(content) > 300 else "")
```
stefrado: I would prefer the preview cap at least to 5000 (preview has max-height) - this way a better portion of the file is shown. Keep in mind that a 0.5 MB could contain `~500_000` chats in UTF-8 and `~250_000` in UTF-16

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +358,46 @@ def _get_column(props):
...
+                    '<span style="color:red;">{}: {}</span>',
+                    _("Error reading file"),
+                    str(e),
+                )
+        return _("No JavaScript file uploaded")
+
+    javascript_preview.short_description = _("JavaScript Preview")
+
+    # customize the checkbox widget to make it more prominent
+    def formfield_for_dbfield(self, db_field, request, **kwargs):
```
stefrado: Should this part be included inside admin.py -> class SiteConfigurationAdminForm  ->  class Meta

```python
# r.98 configuration/admin.py
widgets = {
    "extra_css": CSSEditorWidget,
    "custom_javascript": forms.FileInput(attrs={"accept": ".js"}),
    "custom_javascript_confirmed": forms.CheckboxInput(attrs={"style": "margin-bottom: 8px; margin-right: 8px;"}),
}
```

Then this whole part can be removed.

stefrado (changes requested):
Create this file to collapse the fields based on the checkbox:

NOTE: Don't forget to update the checkbox label if this is implemented -> I confirm the JavaScript added ~below~ is safe -> Below the checkbox is nothing if this is implemented.

```javascript
// js/admin/admin-overrides.js

// make sure this file is included in js/admin/django-admin.js -> r 4. import './admin/admin-overrides'

const JS_CONFIRM_FIELD = document.querySelector(
  '.field-custom_javascript_confirmed'
)
const JS_FILE_FIELD = document.querySelector('.field-custom_javascript')
const JS_PREVIEW_FIELD = document.querySelector('.field-javascript_preview')

class CollapseJSField {
  constructor(js_confirm, js_file, js_preview) {
    this.checkbox = js_confirm
    this.file = js_file
    this.preview = js_preview

    this.bindEvents()
    this.render()
  }

  get checkboxInput() {
    return this.checkbox.querySelector('input')
  }

  get checkboxValue() {
    return this.checkbox.querySelector('input').checked
  }

  bindEvents() {
    this.checkboxInput.addEventListener('change', this.render.bind(this))
  }

  render() {
    this.file.classList.toggle('hidden', !this.checkboxValue)
    this.preview.classList.toggle('hidden', !this.checkboxValue)
  }
}

// Start!
if (JS_CONFIRM_FIELD && JS_FILE_FIELD && JS_PREVIEW_FIELD)
  new CollapseJSField(JS_CONFIRM_FIELD, JS_FILE_FIELD, JS_PREVIEW_FIELD)

```

stefrado:
We might not want to configure some dangerous JS and leave this to the admin's, but if we are going to check this is a simple default for dangerous patterns: `dangerous_patterns = ["eval(", "Function(", "document.cookie", "window.location", "fetch(", "XMLHttpRequest"]`.

jiromaykin:
> We might not want to configure some dangerous JS and leave this to the admin's, but if we are going to check this is a simple default for dangerous patterns: ...."

That is why I left my first commit visible for now; it contains some of those validations (in `validators.py`): https://github.com/maykinmedia/open-inwoner/pull/1873/commits/b5aa9471e8f3a0e271c02fd1c5f57bf3b71b66a8 

But depending on what we decide, I will either retrieve and/or squash it all later.

stefrado:
> But depending on what we decide...

I think that even though we provide a dangerous option with a disclaimer, we also should prevent the usage of `eval` and `Function(`. We know that these are dangerous JS patterns and could think of some safer solutions, but some municipalities might want these dangerous patterns as a shortcut instead of the safer solution, because it is easier to implement.

Besides this we might also want a global feature flag (discussed during the stand-up), defined as config setting -> `ALLOW_CUSTOM_JS = config('ALLOW_CUSTOM_JS', default=False)`, to prevent all instances of being 'dangerous'. Only if the admin want custom JS we can toggle the feature for them (during deployment).

src/open_inwoner/configurations/admin.py
jiromaykin: Good point.

src/open_inwoner/configurations/admin.py
jiromaykin: I really like this Preview thing so I'd like to improve it, but it was just some extra functionality I thought was neat, that i added and put up for discussion. So @swrichards  if we agree we want this, then I can improve it.

src/open_inwoner/configurations/models.py
```diff
@@ -638,6 +638,28 @@ class SiteConfiguration(SingletonModel):
...
+        help_text=_(
+            "Upload a custom JavaScript file (.js) that is loaded in the head of every page."
+        ),
+    )
+    # Checkbox to "add script dangerously"
+    custom_javascript_confirmed = models.BooleanField(
+        verbose_name=_("I confirm the JavaScript added below is safe"),
+        default=False,
+        help_text=_(
+            "Let op: Custom JavaScript kan de beveiliging en prestaties van de applicatie beïnvloeden. Ik bevestig dat ik de "
```
jiromaykin: @swrichards Deze tekst is beter Nederlands maar kan inhoudelijk misschien nog anders.

src/open_inwoner/conf/base.py
```diff
@@ -81,6 +81,9 @@
 GEOS_LIBRARY_PATH = config("GEOS_LIBRARY_PATH", None)
 GDAL_LIBRARY_PATH = config("GDAL_LIBRARY_PATH", None)
 
+# Custom JavaScript feature flag
+ALLOW_CUSTOM_JS = config("ALLOW_CUSTOM_JS", default=False)
```
swrichards, line 85: Just a note that we'll need to create an issue in http://github.com/maykinmedia/charts/ to add this to the Helm charts for OIP.

src/open_inwoner/utils/context_processors.py
```diff
@@ -83,7 +83,9 @@ def settings(request):
         "cookie_link_url": config.cookie_link_url,
         "extra_css": config.extra_css,
         "custom_javascript_url": config.custom_javascript.url
-        if config.custom_javascript and config.custom_javascript_confirmed
+        if config.custom_javascript
+        and config.custom_javascript_confirmed
+        and django_settings.ALLOW_CUSTOM_JS
```
swrichards, line 88: I think it would be more readable/testable, given the extra flag (which I think is a good idea!) to expose this check on a method on `SiteConfiguration`, so we get e.g. `config.is_custom_javascript_allowed`. That way the source of truth for that logic is clearly on the model and not tucked away in a template.

src/open_inwoner/configurations/admin.py
swrichards: I am not a huge fan of the preview. It's hard to tell how this will be used (small snippets, gigantic bundles 😱 ), and I don't think clicking the link to see the file is a huge burden. I would instead opt for showing the file size (in human-readable, e.g. Kb) as an admin-only field so people have a quick sense glancing at this whether the file has content, which for me would be the main purpose served by a preview.

src/open_inwoner/configurations/admin.py
```diff
@@ -96,25 +94,38 @@ class Meta:
...
-            else:
-                validate_url = URLValidator()
-                try:
-                    validate_url(redirect_to)
-                except exceptions.ValidationError:
-                    raise ValidationError(_("The entered url is invalid.")) from None
+        if custom_javascript and not settings.ALLOW_CUSTOM_JS:
+            raise ValidationError(
+                _(
+                    "Custom JavaScript upload is disabled. Contact your system administrator to enable this feature."
```
swrichards:
```suggestion
                    "Custom JavaScript upload is disabled. Contact your system administrator to enable this feature by setting the ALLOW_CUSTOM_JS flag to true."
```

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
...
+                    '<pre style="margin:0;font-size:12px;">{}</pre>'
+                    "</div>",
+                    _("File"),
+                    obj.custom_javascript.name,
+                    _("Size"),
+                    obj.custom_javascript.size,
+                    preview,
+                )
+            except Exception as e:
+                return format_html(
```
swrichards:
```suggestion
            except Exception as e:
                logger.exception("Unable to load custom javascript in admin preview")
                return format_html(
```

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
...
+                    '<span style="color:red;">{}: {}</span>',
+                    _("Error reading file"),
+                    str(e),
+                )
+        return _("No JavaScript file uploaded")
+
+    javascript_preview.short_description = _("JavaScript Preview")
+
+    # Feature flag for ALLOW_CUSTOM_JS
+    def get_fieldsets(self, request, obj=None):
```
swrichards, line 356: I think the issue here is that, if `ALLOW_CUSTOM_JS` is _not_ set, the user has no indication that this feature even exists. Ideally, we would add an `else` below to add a read-only field informing the user that Custom javascript is available, but it requires contacting a system administrator to set `ALLOW_CUSTOM_JS` to true.

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
...
+        fieldsets = list(super().get_fieldsets(request, obj))
+
+        # If custom JS is disabled, remove JS fields from advanced options
+        if not settings.ALLOW_CUSTOM_JS:
+            for i, (name, options) in enumerate(fieldsets):
+                if name == _("Advanced display options"):
```
swrichards: Another option is to rename "Advanced display options" to "Custom styles" and "Custom javascript" respectively, which would make it easier also to detect which fields to delete/add.

src/open_inwoner/configurations/models.py
```diff
@@ -638,6 +638,28 @@ class SiteConfiguration(SingletonModel):
...
+        verbose_name=_("Custom JavaScript"),
+        upload_to="custom_scripts/",
+        blank=True,
+        null=True,
+        validators=[
+            FileExtensionValidator(allowed_extensions=["js"]),
+            validate_javascript_file,
+        ],
+        help_text=_(
+            "Upload a custom JavaScript file (.js) that is loaded in the head of every page."
```
swrichards:
```suggestion
            "Upload a custom JavaScript file (.js) that will be loaded on every page."
```

(I am not sure the `<head>` part is relevant).

src/open_inwoner/configurations/models.py
```diff
@@ -638,6 +638,28 @@ class SiteConfiguration(SingletonModel):
...
+        ),
+    )
+    # Checkbox to "add script dangerously"
+    custom_javascript_confirmed = models.BooleanField(
+        verbose_name=_("I confirm the JavaScript added is safe"),
```
swrichards:
```suggestion
        verbose_name=_("I confirm I understand the risks of adding custom Javascript"),
```

src/open_inwoner/configurations/models.py
```diff
@@ -638,6 +638,28 @@ class SiteConfiguration(SingletonModel):
...
+        default=False,
+        help_text=_(
+            "Let op: Custom JavaScript kan de beveiliging en prestaties van de applicatie beïnvloeden. Ik bevestig dat ik de "
+            "mogelijke risico’s begrijp en – indien nodig – dit heb afgestemd met de security-afdeling voordat ik deze inlaad."
```
swrichards:
```suggestion
            "Let op: Custom JavaScript kan de beveiliging en prestaties van de applicatie negatief beïnvloeden. Door het aanvinken van deze optie bevestig ik dat ik de "
            "mogelijke risico’s begrijp en – indien nodig – dit heb afgestemd met de security-afdeling voordat ik deze inlaad. Het aanvinken van deze optie is noodzakelijk om de Javascript daadwerkelijk te tonen."
```

src/open_inwoner/configurations/validators.py
```diff
@@ -18,3 +18,25 @@ def validate_oidc_config(value):
...
+            _("File is too large. Maximum size is %(max_size)s MB")
+            % {"max_size": max_size / (1024 * 1024)}
+        )
+
+    # Read contents for validation
+    file.seek(0)
+    try:
+        content = file.read().decode("utf-8")
+    except UnicodeDecodeError:
+        raise ValidationError(_("File is not a valid JavaScript file")) from None
```
swrichards:
```suggestion
        raise ValidationError(_("File is not a valid JavaScript file (invalid encoding)")) from None
```

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
...
+    def javascript_preview(self, obj):
+        """Show preview of the uploaded script"""
+        if obj.custom_javascript:
+            try:
+                obj.custom_javascript.seek(0)
+                content = obj.custom_javascript.read().decode("utf-8")
+                obj.custom_javascript.seek(0)
+
+                # Show first 5000 characters
+                preview = content[:5000] + ("..." if len(content) > 5000 else "")
```
swrichards: You can use `truncatechars` with `from django.template.defaultfilters import truncatechars` and then do 
```suggestion
                preview = truncatechars(content, 5000)
```

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
             ),
         )
 
+    def javascript_preview(self, obj):
+        """Show preview of the uploaded script"""
+        if obj.custom_javascript:
+            try:
+                obj.custom_javascript.seek(0)
+                content = obj.custom_javascript.read().decode("utf-8")
+                obj.custom_javascript.seek(0)
```
swrichards: Here I think we can use the contextmanager to avoid manual seeks:


```suggestion
                with obj.custom_javascript.open("r") as f:
                    content = f.read().decode("utf-8")
```

src/open_inwoner/configurations/admin.py
```diff
@@ -339,6 +354,76 @@ def _get_column(props):
             ),
         )
 
+    def javascript_preview(self, obj):
+        """Show preview of the uploaded script"""
+        if obj.custom_javascript:
```
swrichards:
```suggestion
        if not obj.custom_javascript:
            return _("No JavaScript file uploaded")
```
Then you don't have to indend the whole file.

src/open_inwoner/configurations/admin.py
swrichards: Adding to this: a simpler option might be to

- Add a new read-only field indicating whether `ALLOW_CUSTOM_JS` is set, _or_ explaining that it is not and how to fix that.
- Keeping all fields present, but making them read-only if `ALLOW_CUSTOM_JS` is not set, as below.

That might give users a better sense of the feature and what is required (and might make the docs a bit easier to write with screenshots).

swrichards (changes requested):

src/open_inwoner/configurations/admin.py
jiromaykin: "indend"?

src/open_inwoner/configurations/admin.py
jiromaykin: But do we want to keep the preview?

src/open_inwoner/configurations/admin.py
jiromaykin: > Another option is to rename "Advanced display options" to "Custom styles" and "Custom javascript" respectively....

I was also wondering (in the description of this PR) what else to do. I do like to hide the advanced CSS and Javascript in the same 'collapsed' section in the Admin. If i'd split them up and give them new names, the question would be which ones we're going to collapse (both?).

For documentation-screenshots I wonder if i should get rid of the `admin-overrides.js` in this PR, but i do think that's better UX; it will just mean an extra screenshot in docs for when the checkbox is/is-not clicked.

src/open_inwoner/configurations/admin.py
swrichards: Indent!

swrichards: I meant: if you get the guard out of the way quickly, the rest of the function can be at regular-indent level.

jiromaykin: Since we are not offering any previews anywhere else in the Admin, I am removing this whole functionality.

src/open_inwoner/configurations/admin.py
jiromaykin: Since we are not offering any previews anywhere else in the Admin, I am removing this whole functionality.

src/open_inwoner/configurations/admin.py
jiromaykin: Since we are not offering any previews anywhere else in the Admin, I am removing this functionality and replacing it with file-size.

src/open_inwoner/configurations/validators.py
```diff
@@ -18,3 +18,25 @@ def validate_oidc_config(value):
                 "in the OpenID Connect configuration."
             )
         )
+
+
+def validate_javascript_file(file):
+    """Validate JavaScript file for size"""
+
+    max_size = 1024 * 512  # 0.5MB
```
jiromaykin, line 25: @swrichards This 0.5MB is just something I picked randomly, so let me know if we want a different maximum limit.

src/open_inwoner/configurations/admin.py
jiromaykin: Yes, agreed. I'll remove the preview logic and add file size.

src/open_inwoner/configurations/admin.py
jiromaykin: Also for the docs I'd guess.

src/open_inwoner/configurations/admin.py
```diff
@@ -310,6 +326,149 @@ def _get_column(props):
...
+
+    def save_model(self, request, obj, form, change):
+        """Override save to validate redirect_to and report contrast ratio"""
+        # Validate redirect_to field
+        redirect_to = self.clean_redirect_to(obj, form, change)
+        if redirect_to is not None:
+            obj.redirect_to = redirect_to
+        else:
+            obj.redirect_to = None
+
```
swrichards:
```suggestion
```

src/open_inwoner/configurations/admin.py
```diff
@@ -310,6 +326,149 @@ def _get_column(props):
...
+                    form.add_error("redirect_to", _("Invalid internal path"))
+                    return None
+            else:
+                # Check if it's a valid URL
+                parsed = urlparse(redirect_to)
+                if not (parsed.scheme and parsed.netloc):
+                    form.add_error("redirect_to", _("Invalid URL format"))
+                    return None
+
+        return redirect_to
```
swrichards: Terug naar form waar die was (d.w.z., dit moet helemaal niet in de diff, geen add/delete)

swrichards:
> We might not want to configure some dangerous JS and leave this to the admin's, but if we are going to check this is a simple default for dangerous patterns: `dangerous_patterns = ["eval(", "Function(", "document.cookie", "window.location", "fetch(", "XMLHttpRequest"]`.

@stefrado @jiromaykin I don't think these checks make a lot of sense, they're not especially effective against most attack vectors, but more importantly, the whole point of the feature is that the user is free to run _whatever Javascript they like_, and the burden is on them to properly vet what they're uploading, after saying very explicitly that they understand the risks. This check just feels like a distraction at best, and a problem at worst (e.g. I can imagine an analytics script using a few of these symbols for legitimate purposes, inasmuch as analytics is _ever_ legitimate).

jiromaykin:
> > We might not want to configure some dangerous JS and leave this to the admin's, but if we are going to check this is a simple default for dangerous patterns: `dangerous_patterns = ["eval(", "Function(", "document.cookie", "window.location", "fetch(", "XMLHttpRequest"]`.
> 
> @stefrado @jiromaykin I don't think these checks make a lot of sense, they're not especially effective against most attack vectors, but more importantly, the whole point of the feature is that the user is free to run _whatever Javascript they like_, and the burden is on them to properly vet what they're uploading, after saying very explicitly that they understand the risks. This check just feels like a distraction at best, and a problem at worst (e.g. I can imagine an analytics script using a few of these symbols for legitimate purposes, inasmuch as analytics is _ever_ legitimate).

Yeah - also: I don't think any of the municipalities are going to write their own scripts. They will be importing third party scripts which they cannot rewrite to check wether the code has "..._dangerous patterns as a shortcut instead of the safer solution, because it is easier to implement._..".  We just provde this as-is. So it is just a yes or a no. 

So i will be removing the validations, also I will make the options for the script upload disappear when the flag is False/empty.

src/open_inwoner/conf/base.py
swrichards: Issue still needed!

swrichards: https://github.com/maykinmedia/charts/issues/287

CHANGELOG.rst
```diff
@@ -31,6 +31,9 @@ Nieuwe features
 * [:taiga-us:`3370`, :pr:`1837`, :pr:`1869`]: Het navigatiemenu is nu conform het NL Design
   System op de meeste pagina's aan de linkerkant van het scherm geplaatst. Sommige
   detailpagina's behouden voorlopig het dropdownmenu in afwachting van nieuwe designs.
+* [:taiga-us:`3407`, :pr:`1873`]: De mogelijkheid voor admin gebruikers om een custom
+  Javascript bestand te uploaden dat automatisch op alle pagina's van de website wordt
+  geladen.
```
stefrado: Feature flag benoemen?

src/open_inwoner/configurations/admin.py
```diff
@@ -67,26 +67,58 @@ class Meta:
         fields = "__all__"
         widgets = {
             "extra_css": CSSEditorWidget,
+            "custom_javascript": forms.ClearableFileInput(attrs={"accept": ".js"}),
```
stefrado: Add init to this model to add correct attrs and keep the original styling of the models.FileField
```py
        self.fields['custom_javascript'].widget.attrs.update({"accept": ".js"})
```

Remove this line
```suggestion
```

src/open_inwoner/scss/admin/_admin_theme.scss
```diff
@@ -343,12 +360,69 @@ form div.help {
...
+      margin-top: 16px;
+      margin-bottom: 16px;
+    }
+
+    [type='checkbox'] {
+      align-self: center;
+      margin-left: 170px;
+      margin-right: 6px;
+    }
+  }
```
stefrado: Can be removed if previous fb is implemented.
```suggestion
```

src/open_inwoner/js/admin/admin-overrides/admin-overrides.js
```diff
@@ -0,0 +1,34 @@
+const JS_CONFIRM_FIELD = document.querySelector(
+  '.field-custom_javascript_confirmed'
+)
+const JS_FILE_FIELD = document.querySelector('.field-custom_javascript')
```
stefrado: Please also hide/show file info/preview if the checkbox is checked/unchecked.
```suggestion
const JS_FILE_FIELD = document.querySelector('.field-custom_javascript')
const JS_FILE_INFO = document.querySelector('.field-custom_javascript_file_info')
```

src/open_inwoner/configurations/admin.py
```diff
@@ -310,6 +345,103 @@ def _get_column(props):
...
+                        # Remove the random suffix added by Django (e.g., "_Yq9uTUy")
+                        original_filename = filename.split("_")[0] + ".js"
+
+                        return format_html(
+                            '{}: "<a href="{}" target="_blank">{}</a>". {}: {} KB',
+                            _("Name"),
+                            obj.custom_javascript.url,
+                            original_filename,  # Use cleaned filename instead of full path
+                            _("Size"),
+                            round(size_kb, 1),
```
stefrado: This is a double link (first in file input and now again)
```suggestion
                            '{}: {} KB',
                            _("Size"),
                            round(size_kb, 1),
```

stefrado (changes requested):

docs/beheerhandleiding/12_configuratie.rst
```diff
@@ -378,8 +378,14 @@ Hier kunt u selecteren of u het onderwerpen-menu en/of de zoekbalk voor anonieme
 12.1.15. Geavanceerde opties
 ----------------------------
 
+Klik op 'Tonen' om deze opties uit te klappen.
+
+**Design/thema stylesheet**
 Hier kunt u extra CSS-stijlen invoegen die op de site gebruikt worden. Let op, er kan enkel gebruik worden gemaakt van een beperkte, veilige subset van CSS-eigenschappen. Niet alle CSS-stijlen worden door het systeem ondersteund. In de aangegeven lijst staan de toegestane CSS-attributen die men eventueel nog aan de site kan toevoegen.
 
+**Custom JavaScript**
+Hier kunt u een eigen Javascript bestand uploaden dat automatisch op alle pagina's van de website wordt geladen. Let op: JavaScript kan de beveiliging en prestaties van de applicatie negatief beïnvloeden. Door het aanvinken van de checkbox bevestigt u dat u de mogelijke risico’s begrijpt.
```
swrichards: I think this would be a good place to also explain the `ALLOW_CUSTOM_JS` variable, and that this needs to be set by whomever manages their deployments.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+    def setUp(self):
+        self.site_config = SiteConfiguration.get_solo()
+        # Ensure required fields are set
+        if not self.site_config.name:
+            self.site_config.name = "Test Municipality"
+            self.site_config.save()
+
+    def test_javascript_field_exists(self):
+        """Test that custom_javascript field exists on model"""
+        self.assertTrue(hasattr(self.site_config, "custom_javascript"))
```
swrichards: This seems overkill to me.
```suggestion
```

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+class SiteConfigurationJavaScriptTests(TestCase):
+    """Test JavaScript functionality in SiteConfiguration model"""
+
+    def setUp(self):
+        self.site_config = SiteConfiguration.get_solo()
+        # Ensure required fields are set
+        if not self.site_config.name:
+            self.site_config.name = "Test Municipality"
+            self.site_config.save()
```
swrichards: This feels out of place. We should not need conditionals in `setUp`, we control the initial state of our tests.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def test_javascript_field_exists(self):
+        """Test that custom_javascript field exists on model"""
+        self.assertTrue(hasattr(self.site_config, "custom_javascript"))
+
+    def test_confirmation_field_exists(self):
+        """Test that custom_javascript_confirmed field exists on model"""
+        self.assertTrue(hasattr(self.site_config, "custom_javascript_confirmed"))
```
swrichards:
```suggestion
```

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+    @override_settings(ALLOW_CUSTOM_JS=True)
+    def test_upload_javascript_without_confirmation_fails(self):
+        """Test that uploading JavaScript without confirmation fails form validation"""
+        from django.utils.datastructures import MultiValueDict
```
swrichards: Move imports to top of file

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+        from open_inwoner.configurations.admin import SiteConfigurationAdminForm
+
+        valid_js = 'console.log("test");'
+        js_file = SimpleUploadedFile(
+            "test.js", valid_js.encode("utf-8"), content_type="application/javascript"
+        )
+
+        form_data = {
+            "name": "Test Municipality",
```
swrichards: Not sure this adds anything.

```suggestion
```

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+        self.assertFalse(form.is_valid())
+        self.assertIn("custom_javascript_confirmed", form.errors)
+
+    @override_settings(ALLOW_CUSTOM_JS=True)
+    def test_upload_javascript_with_confirmation_succeeds(self):
+        """Test that uploading JavaScript with confirmation succeeds"""
+        from django.utils.datastructures import MultiValueDict
+
+        from open_inwoner.configurations.admin import SiteConfigurationAdminForm
```
swrichards: Move to top.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+    def test_upload_javascript_without_confirmation_fails(self):
+        """Test that uploading JavaScript without confirmation fails form validation"""
+        from django.utils.datastructures import MultiValueDict
+
+        from open_inwoner.configurations.admin import SiteConfigurationAdminForm
+
+        valid_js = 'console.log("test");'
+        js_file = SimpleUploadedFile(
+            "test.js", valid_js.encode("utf-8"), content_type="application/javascript"
+        )
```
swrichards: We use this a lot, we could put it on `self.valid_js_file` in `setup`.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+        for field in SiteConfiguration._meta.fields:
+            # Skip file fields and auto fields
+            if field.get_internal_type() not in [
+                "FileField",
+                "AutoField",
+            ] and not field.name.endswith("_file"):
+                value = getattr(config, field.name)
+                if value is not None:
+                    form_data[field.name] = value
```
swrichards: Not sure what this adds -- if we want to test you can upload _other_ fields in addition to the test js fields, we should make that its own test with a descriptive title.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+                "custom_javascript_confirmed": True,
+            }
+        )
+
+        files = MultiValueDict({"custom_javascript": [js_file]})
+        form = SiteConfigurationAdminForm(data=form_data, files=files)
+
+        self.assertTrue(form.is_valid(), f"Form validation failed: {form.errors}")
+
+    def test_javascript_field_validators(self):
```
swrichards: Don't think we need this: we just want to test the behavior of the validators.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+        # Check for file extension validator
+        validator_names = [v.__class__.__name__ for v in field.validators]
+        self.assertIn("FileExtensionValidator", validator_names)
+
+    @override_settings(ALLOW_CUSTOM_JS=False)
+    def test_custom_javascript_disabled_by_feature_flag(self):
+        """Test that custom JavaScript upload fails when feature flag is disabled"""
+        from django.utils.datastructures import MultiValueDict
+
+        from open_inwoner.configurations.admin import SiteConfigurationAdminForm
```
swrichards: Move to top

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+        js_file = SimpleUploadedFile(
+            "test.js", valid_js.encode("utf-8"), content_type="application/javascript"
+        )
+
+        form_data = {"name": "Test Municipality"}
+        files = MultiValueDict({"custom_javascript": [js_file]})
+        form = SiteConfigurationAdminForm(data=form_data, files=files)
+
+        self.assertFalse(form.is_valid())
+        self.assertIn("custom_javascript", form.errors)
```
swrichards: I would prefer matching the entire error message if possible.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+        js_content = 'console.log("test");'  # This is about 20 bytes
+        js_file = SimpleUploadedFile(
+            "test.js", js_content.encode("utf-8"), content_type="application/javascript"
+        )
+        self.site_config.custom_javascript = js_file
+        self.site_config.save()
+
+        info = admin_instance.custom_javascript_file_info(self.site_config)
+        self.assertIn("test.js", info)
+        self.assertIn("KB", info)
```
swrichards: Can we directly assert against the message?

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+        form_data = {}
+
+        for field in SiteConfiguration._meta.fields:
+            if field.get_internal_type() not in [
+                "FileField",
+                "AutoField",
+            ] and not field.name.endswith("_file"):
+                value = getattr(config, field.name)
+                if value is not None:
+                    form_data[field.name] = value
```
swrichards: As above, not sure this adds much to the test.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,318 @@
...
+
+        if form.is_valid():
+            form.save()
+            self.site_config.refresh_from_db()
+
+            # File should be cleared
+            self.assertFalse(self.site_config.custom_javascript)
+
+    @override_settings(ALLOW_CUSTOM_JS=True)
+    def test_teardown_cleanup(self):
```
swrichards: Not sure I follow this test, how does it test the teardown behavior?

CHANGELOG.rst
swrichards: Ja dat lijkt me goed, dit lijkt me een ook een kandidaat voor de "deployment notes" (want environment variabelen moeten op deployment niveau gemanaged worden, functioneel beheeders kunnen dat niet).

src/open_inwoner/configurations/tests/test_custom_javascript.py
swrichards, line 1: All these tests focus on the admin/validation. I think it might make sense to move these to `test_admin.py`, and to this test file add some missing tests:

- That the upoaded javascript is actually included in the head of a page when enabled
- That it is omitted when the setting is disabled
- That it is omitted when the setting is enabled, but the file is missing

swrichards (changes requested):

CHANGELOG.rst
jiromaykin: Goed!

src/open_inwoner/configurations/admin.py
jiromaykin: Yes but it will crash if flag is set to False so needs a check :-)

```
    # Only update the widget if the field exists in the form
    if 'custom_javascript' in self.fields:
        self.fields['custom_javascript'].widget.attrs.update({"accept": ".js"})
```

src/open_inwoner/configurations/tests/test_custom_javascript.py
jiromaykin: This must have been a hallucination.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,147 @@
...
+        <!DOCTYPE html>
+        <html>
+        <head>
+            {self._get_custom_js_script_tag()}
+        </head>
+        <body>
+            <h1>Test Page</h1>
+        </body>
+        </html>
+        """
```
swrichards: I don't think this helps: we should render the _actual_ `master.html` template, because that's where the logic is.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,147 @@
...
+from open_inwoner.configurations.models import SiteConfiguration
+
+
+class CustomJavaScriptInclusionTests(TestCase):
+    """Test that uploaded JavaScript is actually included in rendered pages"""
+
+    def setUp(self):
+        self.site_config = SiteConfiguration.get_solo()
+        self.site_config.name = "Test Municipality"
+        self.site_config.save()
```
swrichards: Does this add anything?

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def _get_custom_js_script_tag(self):
+        """Simulate how the template tag would render the custom JavaScript"""
+        if not settings.ALLOW_CUSTOM_JS:
+            return ""
+
+        if not self.site_config.custom_javascript:
+            return ""
+
+        return f'<script src="{self.site_config.custom_javascript.url}"></script>'
```
swrichards: See above, this feels like a hallucination to me. It's reproducing the logic that we want to actually test in `master.html`.

src/open_inwoner/configurations/tests/test_custom_javascript.py
```diff
@@ -0,0 +1,147 @@
...
+
+        # Test that we can access the file URL
+        file_url = self.site_config.custom_javascript.url
+        self.assertTrue(file_url.endswith(".js"))
+
+        # Verify the file object has correct content type
+        self.assertEqual(js_file.content_type, "application/javascript")
+
+    @override_settings(ALLOW_CUSTOM_JS=True)
+    def test_multiple_page_renders_include_javascript(self):
```
swrichards: I don't think we need this test.

src/open_inwoner/configurations/tests/test_custom_javascript.py
swrichards, line 1: All tests in this module need to be rewritten to:

- Actually render our `master.html` template rather than the hallucinated placeholder
- Test for the _exact_ script tag that we expect, including the URL and nonce (rather than `assertIn` with a snippet).

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -121,3 +129,309 @@ def test_invalid_url_is_not_saved(self):
...
+        # Test with file
+        js_content = 'console.log("test");'  # +/- 20 bytes
+        js_file = SimpleUploadedFile(
+            "test.js", js_content.encode("utf-8"), content_type="application/javascript"
+        )
+        self.site_config.custom_javascript = js_file
+        self.site_config.save()
+
+        info = admin_instance.custom_javascript_file_info(self.site_config)
+        self.assertIn("test.js", info)  # Should find cleaned filename
```
swrichards: Can we test the exact expected message here, including the expected size?

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -121,3 +129,309 @@ def test_invalid_url_is_not_saved(self):
...
+        self.site_config.custom_javascript = self.valid_js_file
+        self.site_config.custom_javascript_confirmed = True
+        self.site_config.save()
+
+        # Verify file exists
+        self.assertTrue(self.site_config.custom_javascript)
+
+        # Now clear the file using the form
+        form_data = {
+            "name": "Test Municipality",
```
swrichards: Unless we're testing that the name is changed/unchanged after submission, remove this and elsewhere in the file, it's noise that makes the tests harder to follow.

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -121,3 +129,309 @@ def test_invalid_url_is_not_saved(self):
...
+        form_data.update(
+            {
+                "custom_javascript-clear": True,  # This clears the file
+                "custom_javascript_confirmed": True,
+            }
+        )
+
+        form = SiteConfigurationAdminForm(data=form_data, instance=self.site_config)
+
+        if form.is_valid():
```
swrichards: We know whether we expect a test to be valid or invalid, and should assert it. This test would pass if the form turns out to be invalid.

src/open_inwoner/configurations/tests/test_admin.py
swrichards, line 1: There's still a few paths missing according to the coverage: I am ok skipping the edge cases of e.g. `IOError` when opening the file. But we're currently missing a test that actually renders the whole admin in both the flag enabled and disabled scenarios, as per the coverage report.

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -121,3 +129,309 @@ def test_invalid_url_is_not_saved(self):
...
+        }
+
+        form = SiteConfigurationAdminForm(data=form_data, instance=self.site_config)
+
+        if form.is_valid():
+            form.save()
+            self.site_config.refresh_from_db()
+
+            # File should be cleared
+            self.assertFalse(self.site_config.custom_javascript)
```
swrichards: As noted in another comment, this is an anti-pattern in a test: we either expect it to be valid or invalid, but the assertions should not be conditional (and indeed these lines are not run according to the coverage report).

CHANGELOG.rst
```diff
@@ -58,3 +58,7 @@ Onderhoud
   waarmee een `beveiligingsissue met severity "high"
   <https://www.djangoproject.com/weblog/2025/sep/03/security-releases/>`_ wordt
   opgelost.
+* [:taiga-us:`3407`, :pr:`1873`]: De mogelijkheid voor admin gebruikers om een custom
+  Javascript bestand te uploaden dat automatisch op alle pagina's van de website wordt
+  geladen. Deze functionaliteit wordt gecontroleerd door de ``ALLOW_CUSTOM_JS``
+  environment variabele die op deployment niveau moet worden ingesteld.
```
swrichards: These are in the wrong place: `1.34` has already been released. Move them to the running changelog section.

CHANGELOG.rst
```diff
@@ -39,7 +39,7 @@ Bugfixes
   bij het ophalen van contactmomenten uit OpenKlant2 zijn opgelost.
 * [:taiga-is:`3389`, :pr:`1845`]: De synchronisatie van gebruikersprofielen met
   OpenKlant gebeurt nu ook direct na registratie, niet alleen na inloggen.
-* [:taiga-is:`3375`, :pr:`1852`, :pr:`1868` ]: Correct afbreken van lange woorden en
+* [:taiga-is:`3375`, :pr:`1868`]: Correct afbreken van lange woorden en
```
swrichards, line 66: Move to running changelog section.

swrichards (changes requested):
The tests still need a bit of work. There were two rounds of reviews where I felt like I had to wade through a lot of hallucinations in the tests especially. This is apparently an area where Claude seems to struggle there, please keep that in mind in future.

CHANGELOG.rst
jiromaykin: ~~? you mean Nieuwe features? But this is a bug fix.~~

Edit: this bug fix was  actually in there, right, so it is part of 1.34 still, i just corrected the link to Taiga for this old issue.

src/open_inwoner/js/admin/admin-overrides/admin-overrides.js
stefrado, line 1: It might be better to rename this file to `override-custom-js-script.js` or something like that. And import it in a `index.js` file (needs to be created). This allows us to add more script to admin-overrides and not having to do this change at that time.

src/open_inwoner/js/django-admin.js
```diff
@@ -1,3 +1,4 @@
 import './admin/ckeditor'
+import './admin/admin-overrides/admin-overrides'
```
stefrado: > And import it in a `index.js` file (needs to be created)

Then this could be simplified

```suggestion
import './admin/admin-overrides'
```

src/open_inwoner/js/admin/admin-overrides/admin-overrides.js
```diff
@@ -0,0 +1,41 @@
...
+
+  bindEvents() {
+    this.checkboxInput.addEventListener('change', this.render.bind(this))
+  }
+
+  render() {
+    this.file.classList.toggle('hidden', !this.checkboxValue)
+    if (this.info) {
+      this.info.classList.toggle('hidden', !this.checkboxValue)
+    }
```
stefrado: This might be a little cleaner

```suggestion
  bindEvents() {
    this.checkboxInput.addEventListener('change', this.render.bind(this))
  }

  render() {
    this.file.classList.toggle('hidden', !this.checkboxInput.checked)
    if (this.info) {
      this.info.classList.toggle('hidden', !this.checkboxInput.checked)
    }
```

stefrado (changes requested):

src/open_inwoner/configurations/tests/test_admin.py
jiromaykin: 'name' seems to be a required field though (which is why tests were not passing initially.

src/open_inwoner/configurations/tests/test_custom_javascript.py
jiromaykin: You mean just something like:

```
        template_content = """
        {% block custom_scripts %}
            {% if custom_javascript_url %}
                <script{% if request.csp_nonce %} nonce="{{ request.csp_nonce }}"{% endif %} src="{{ custom_javascript_url }}" defer></script>
            {% endif %}
        {% endblock custom_scripts %}
        """
```

src/open_inwoner/configurations/tests/test_custom_javascript.py
jiromaykin: The master.html template includes Django CMS template tags that require a complete CMS runtime environment to function. When I tried to render it in a test I got failures on required toolbar/user/page/etc. So i added a test for only the custom JavaScript inclusion logic.

stefrado (approved):

src/open_inwoner/configurations/tests/test_custom_javascript.py
swrichards: No I mean that we have to render our _actual_ `master.html` content, not this or another snippet, because that's not what runs in our actual code in https://github.com/maykinmedia/open-inwoner/pull/1873/files#diff-b5f48e6039303a60de4a5f8f0c7a6f4ac8ff161b1979b539a5cb6340be1f52dcR67-R74.

We could either render a simple page (e.g. "/", which renders `master.html`), or use something like `render_to_string` with `master.html` as the template.

src/open_inwoner/configurations/tests/test_custom_javascript.py
swrichards: Correct me if I am wrong, but it seems to me that we are not actually testing our code, but code that exists _only in our tests_. The question is, if we were to remove the custom JS logic from our `master.html` template, would these tests fail?

I can the CMS issue is annoying, but at the same time, that is actually where a lot of our pages run so it might not hurt to go through the setup (or alternatively, maybe we could use something like https://docs.djangoproject.com/en/5.2/topics/templates/#django.template.loader.render_to_string ? e.g. `render_to_string("master.html", request_from_factory, ...` ).

If there's no easy solution we can take a look on Monday.

swrichards: Try something like this:

```python
from cms import api

from open_inwoner.cms.tests.cms_tools import render_full_page

    def setUp(self):
        self.site_config = SiteConfiguration.get_solo()
        self.page = api.create_page(
            title="Test Page",
            template="cms/fullwidth.html", 
            language="nl",
            in_navigation=True
        )
        self.page.publish("nl")

    def test_foo(self):
        # Set up the custom js flags for the test
        rendered_html = render_full_page(self.page)
```

The `cms/fullwidth.html` template inherits `master.html` so we'll be testing the actual production code.

src/open_inwoner/configurations/admin.py
```diff
@@ -310,6 +350,104 @@ def _get_column(props):
...
+                        return format_html(
+                            "{}: {} KB",
+                            original_filename,  # Show filename
+                            round(size_kb, 1),  # Show size
+                        )
+                    else:
+                        return _("File uploaded but feature is disabled")
+                else:
+                    return _("File missing from storage")
+            except (OSError, IOError):
```
swrichards: Let's catch all exceptions, as discussed today, we don't want a failure here to tear down the admin:


```suggestion
            except Exception:
```

src/open_inwoner/configurations/tests/test_admin.py
```diff
@@ -116,8 +115,394 @@ def test_invalid_url_is_not_saved(self):
...
+        js_file = SimpleUploadedFile(
+            "missing.js",
+            js_content.encode("utf-8"),
+            content_type="application/javascript",
+        )
+        self.site_config.custom_javascript = js_file
+        self.site_config.save()
+
+        # Now simulate the file being missing from storage by mocking storage.exists
+        from unittest.mock import patch
```
swrichards: Move to the top

swrichards (changes requested):
Just the point about the tests, I left a suggestion as to how we might render the actual template in the tests but let me know if it doesn't work.

swrichards (approved):

PR #1875, opened 2025-08-27
feat: validate ES connection settings at startup

swrichards:
Due to bumping the underlying elasticsearch libraries, the ES host string now requires a scheme and a port, which many of our production deploys do not include. Adding this runtime check means this issue will be detected early and given some clearer context so it can be dealt with at deploy time.

### DIFF
--- a/src/open_inwoner/search/apps.py
+++ b/src/open_inwoner/search/apps.py
@@ -1,0 +1,35 @@
+import logging
+
+from django.apps import AppConfig
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+
+from elasticsearch.dsl.connections import connections
+
+logger = logging.getLogger(__name__)
+
+
+class SearchAppConfig(AppConfig):
+    name = "open_inwoner.search"
+
+    def ready(self):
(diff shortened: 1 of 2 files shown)


pi-sigma (approved):
Should we move the valiation to a separate file perhaps, so that we could do

```
from .validators import validate_host

ELASTICSEARCH_DSL = {"default": {"hosts": validate_host(config("ES_HOST"))}}
```

Otherwise the config file might be poised to become cluttered with code in the future.

swrichards:
> Should we move the valiation to a separate file perhaps, so that we could do
> 
> ```
> from .validators import validate_host
> 
> ELASTICSEARCH_DSL = {"default": {"hosts": validate_host(config("ES_HOST"))}}
> ```
> 
> Otherwise the config file might be poised to become cluttered with code in the future.

Fair point, ended up solving this differently by adding the runtime check in the app ready hook. The ES client itself raises an exception for this validation, but 1) only does so when you actually try to perform a search / build the index and 2) is a bit opaque, might not be clear to an end-user what caused the error. So this PR does the check at runtime and adds some context to the exception.

PR #1881, opened 2025-09-01
:sparkles: [#3424] Search and select for large number of business-branches

jiromaykin:
## Summary

A React Search/Select component with better UX for the log-in screen in case of very large number of business branches, with a rudimentary NLDS implementation in order to make it possible to give the component back to the community.

The term 'combobox' comes from the [ARIA specification](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/) (Web Accessibility Initiative - Accessible Rich Internet Applications). The official ARIA pattern defines a combobox as having:

- An input element with role="combobox"
- A popup listbox with role="listbox"
- Options with role="option"

Figma design: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6908-13562&t=Or9BM4hccARh03hS-1

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3408
- Taiga Issue: task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3424

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/react/components/MaterialIcons/MaterialIcons.tsx
```diff
@@ -4,7 +4,7 @@ export interface MaterialIconProps {
   name: string
 }
 
-// Material Icons component
+// Material Icons component - icons should never be clickable, always surround with buttons
```
stefrado:
```suggestion
/** 
 * Material Icons component
 * - Icons should never be clickable, always surround with buttons
 */
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,345 @@
...
+  vestigingInfo?: string
+  rechtspersoonInfo?: string
+  addressInfo?: string
+  cityInfo?: string
+  vestigingsnummer?: string
+  type?: string
+  [key: string]: any
+}
+
+interface Props {
```
stefrado: The name `Props` is too generic. I would reneema it to `SelectComboBoxProps`. When we later add a export to the type (too use it in the test file for example) Props is too generic and can mean anything.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,345 @@
...
+        setIsOpen(false)
+        break
+
+      case 'Tab':
+        setIsOpen(false)
+        break
+    }
+  }
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
stefrado: We might want to add a debounce to the input changes - when there are many results each input change could be a heavy rendering. With a debounce we could ease this process by waiting until the user is done typing. 

> React debounces could be a little tricky...

src/open_inwoner/templates/pages/kvk/branches.html
```diff
@@ -1,75 +1,50 @@
...
-
-                </ul>
+                {# React combobox with data only - translations handled by react-intl #}
+                <div
+                    data-react-module="selectcombobox"
+                    data-id="branch-selector"
+                    data-label="{% trans 'Selecteer de vestiging waarmee u wilt inloggen' %}"
+                    data-name="branch_number"
+                >
+                    <script type="application/json" class="branch-data">{% react_combobox_data company_branches user.vestiging %}</script>
```
stefrado: I must say - i do like the effect that data script disappears if the react root mounts...

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,345 @@
...
+                  <div className="utrecht-listbox__content">
+                    <div className="utrecht-listbox__heading">
+                      {item.label}
+                      {selectedId === item.id && (
+                        <span
+                          aria-hidden="true"
+                          className="material-icons-outlined"
+                        >
+                          check
+                        </span>
```
stefrado:
```suggestion
                        <MaterialIcon name="check" />
```

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -0,0 +1,345 @@
...
+          placeholder={intl.formatMessage({
+            id: 'combobox.placeholder',
+            description: 'Placeholder text for branch search input',
+            defaultMessage: 'Vul naam, adres of vestigingsnummer in...',
+          })}
+          autoComplete="off"
+        />
+
+        {hasText ? (
+          <button
```
stefrado, line 288: Might want to add a `TODO` to this file that mentions these buttons and that they will be replaced when the `Button` react component is created.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,345 @@
...
+            id={`${id}-listbox`}
+            aria-label={label}
+          >
+            <ul className="utrecht-listbox__list" ref={listRef}>
+              {filtered.map((item, i) => (
+                <li
+                  key={item.id}
+                  role="option"
+                  aria-selected={selectedId === item.id}
+                  className={`utrecht-listbox__option ${
```
stefrado: we might want to add a util lib like `clsx` or `classNames` to handle classNames.

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+    # This corresponds to the current user's vestiging selection
+    data = {"items": items, "selected_id": escape(str(selected_id or ""))}
+
+    # We need mark_safe here because the JSON needs to be output as-is in the template
+    # The data is already escaped above, so this is safe
+    return mark_safe(json.dumps(data, ensure_ascii=True))  # noqa: S308
+
+
+@register.simple_tag
+def react_combobox_data(branches, selected_id=None):
```
pi-sigma: Are these two functions duplicates?

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+@register.simple_tag
+def react_branch_data(branches, selected_id=None):
+    """
+    Convert Django branch data to React-compatible JSON format
+    This matches exactly what SelectComboBoxModule expects
+
+    This function creates JSON data that gets embedded in HTML and read by React.
+    The 'selected_id' corresponds to the branch_number field that Django expects
+    when the form is submitted via POST request.
+    """
```
pi-sigma:
```
if not branches or not isinstance(branches, (list, tuple)):
    return mark_safe(json.dumps({"items": [], "selected_id": ""}))
```

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+    items = []
+    for branch in branches:
+        # Handle the "entire company" case (empty vestigingsnummer)
+        # This becomes the 'id' that React stores and sends back as 'branch_number'
+        branch_id = branch.get("vestigingsnummer", "") or "entire-company"
```
pi-sigma:
```suggestion
        branch_id = branch.get("vestigingsnummer", "rechtspersoon")
```

You can add the default as second argument to `get`. I would use the Dutch "rechtspersoon" because it is clear ("entire company" occurs elsewhere but really shouldn't have been used to begin with; it's not very clear)

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+        rechtspersoon_info = ""
+
+        if branch.get("vestigingsnummer"):
+            vestiging_info = f"Vestiging: {branch['vestigingsnummer']}"
+            if branch.get("type") == "hoofdvestiging":
+                vestiging_info += " (Hoofdvestiging)"
+        else:
+            # Show "(Rechtspersoon)" as separate line for entire company option
+            # This handles both explicit rechtspersoon entries AND generated entire-company options
+            rechtspersoon_info = "(Rechtspersoon)"
```
pi-sigma:
```suggestion
        if branch_id == "rechtspersoon":
            rechtspersoon_info = "(Rechtspersoon)"
        else:
            vestiging_info = f"Vestiging: {branch['vestigingsnummer']}"
            if branch.get("type") == "hoofdvestiging":
                vestiging_info += " (Hoofdvestiging)"
```

you can get the `branch_id` above, no need to `get` again here

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+            # Add house number if available
+            if adres.get("huisnummer"):
+                address_parts.append(str(adres["huisnummer"]))
+            # Add house number addition if available (e.g., "A", "bis")
+            if adres.get("huisnummerToevoeging"):
+                address_parts.append(adres["huisnummerToevoeging"])
+            address_info = " ".join(address_parts)
+
+        if adres.get("plaats"):
+            city_info = adres["plaats"]
```
pi-sigma: Could be exported into a helper:

```

def format_address(address_data: Dict[str, Any]) -> str:
    """Format address data into a human-readable string."""
    if not address_data:
        return ""

    if address_data.get("straatHuisnummer"):
        return address_data["straatHuisnummer"]

    address_parts = []
    if address_data.get("straatnaam"):
        address_parts.append(address_data["straatnaam"])

    if address_data.get("huisnummer"):
        address_parts.append(str(address_data["huisnummer"]))

    if address_data.get("huisnummerToevoeging"):
        address_parts.append(address_data["huisnummerToevoeging"])

    return " ".join(address_parts)
```

Then in the template tag:
```
address_info = format_address(adres)
```

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -166,10 +166,20 @@ def test_post_branches_page_with_incorrect_vestigingsnummer(
...
 
-        # check that pseudo-branch representing company as a whole has been added
-        self.assertEqual(len(branch_inputs), 3)
+        # Verify JSON data structure embedded in script tag for React component
+        # React reads branch data from this JSON rather than parsing DOM elements
+        branch_data_script = doc.find("script.branch-data")
+        self.assertEqual(len(branch_data_script), 1)
+
+        # Parse the JSON data to verify it contains the expected branches
+        import json
```
pi-sigma: `import` should be at the top

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -300,18 +323,35 @@ def test_get_branches_page(self, mock_solo, mock_kvk):
 
         doc = PyQuery(response.content)
 
-        branch_inputs = doc.find("[name='branch_number']")
+        # Check for React component data instead of DOM inputs
+        branch_data_script = doc.find("script.branch-data")
+        self.assertEqual(len(branch_data_script), 1)
+
+        # Parse the JSON data to verify it contains the expected branches
+        import json
```
pi-sigma: `import` should be at the top (no need to repeat)

src/open_inwoner/kvk/tests/test_template_tags.py
```diff
@@ -0,0 +1,85 @@
...
+            {"kvkNummer": "12345678", "naam": "Test Company", "type": "rechtspersoon"},
+            {
+                "kvkNummer": "12345678",
+                "vestigingsnummer": "1234",
+                "naam": "Test Company",
+                "type": "hoofdvestiging",
+                "adres": {
+                    "binnenlandsAdres": {
+                        "straatnaam": "Teststraat",
+                        "huisnummer": 42,
```
pi-sigma, line 22: Could use:

```
"huisnummerToevoeging": "A",
```

and then update the assert below

src/open_inwoner/kvk/tests/test_template_tags.py
```diff
@@ -0,0 +1,85 @@
+import json
+
+from django.test import TestCase
+
+from open_inwoner.components.templatetags.kvk_tags import react_combobox_data
+
+
+class KvKTemplateTagsTestCase(TestCase):
```
pi-sigma: Could use a test for nevenvestiging:

```
def test_react_combobox_data_with_non_main_branch(self):
    """Test formatting for a non-main branch (nevenvestiging)"""
    branches = [
        {
            "naam": "Test Company Branch",
            "vestigingsnummer": "9876",
            "type": "nevenvestiging",  # Not hoofdvestiging
            "adres": {
                "binnenlandsAdres": {
                    "straatnaam": "Branchstraat",
                    "huisnummer": 10,
                    "plaats": "Rotterdam",
                }
            },
        }
    ]

    data = json.loads(react_combobox_data(branches))

    # Should not contain "Hoofdvestiging" text
    self.assertNotIn("Hoofdvestiging", data["items"][0]["vestigingInfo"])
    self.assertEqual(data["items"][0]["vestigingInfo"], "Vestiging: 9876")
```

Alternatively: add the data to `test_react_combobox_data_with_branches` below and update the asserts

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,180 @@
...
+        # Properly escape all string values to prevent XSS
+        # These fields will be displayed in the React component dropdown
+        items.append(
+            {
+                "id": escape(
+                    str(branch_id)
+                ),  # This becomes the form value sent back to Django
+                "label": escape(
+                    str(branch.get("naam", ""))
+                ),  # Main display text in dropdown
```
pi-sigma: I think the comments are unnecessary (except for the first maybe "form value sent back...")

src/open_inwoner/react/components/SelectComboBox/types.ts
```diff
@@ -0,0 +1,13 @@
+// types for KVK branches
+export interface Branch {
```
stefrado: Is this interface used somewhere?

pi-sigma (changes requested):
I also noted:

Typing "Donald" (the company name is "BV Donald") followed by a space i immediately get the nevenvestiging suggested as the only option. The hoofdvestiging somehow disappears. Even "Donald hoofd" or similar doesn't bring the main branch back.

(using KVK `68750110` for testing)

fix-json-newlines.js
```diff
@@ -0,0 +1,18 @@
...
+    const filePath = path.join(compiledDir, file)
+    const content = fs.readFileSync(filePath, 'utf-8')
+
+    // Add trailing newline if missing
+    if (!content.endsWith('\n')) {
+      fs.writeFileSync(filePath, content + '\n')
+      console.log(`Fixed ${file}: added trailing newline`)
+    }
+  }
+})
```
stefrado: And this file has no new line... 🙈

package.json
```diff
@@ -14,7 +14,7 @@
     "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
     "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -c",
     "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
-    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled",
+    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled && node fix-json-newlines.js",
```
stefrado: Did you run this script? The json files do not have a new line ending

stefrado (changes requested):

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: That'll be fun.

fix-json-newlines.js
jiromaykin: ughhh, well spotted, apparently nothing is checking it...

package.json
jiromaykin: My PR was checked by .pre-commit-config.yaml so that's when i found i had to format the React compiled messages etc.

src/open_inwoner/components/templatetags/kvk_tags.py
jiromaykin: i was mostly trying to clarify it for myself :-) here is where front-end and back-end meet, and can be a bit difficult for both sides, but i can remove most of them.

src/open_inwoner/components/templatetags/kvk_tags.py
jiromaykin: oops yes, i first tried to think of the component as a general component for all sorts of data to be re-used later, so i wanted to create 2, so i've been moving theses tags around in different places, but now i'll just make it KVK specific.

jiromaykin: So this second one is actually the one that needs review - because that's the one that gets injected into branches.html - i'll update it first.

package.json
jiromaykin: Correction: actually those files DO have the newline, you just don't see the newline in github - but if it would NOT be there, you'd see a red circle with a minus, 

like in the old version: https://github.com/maykinmedia/open-inwoner/commit/50e637f07d232890bb5b4039822ed63eb630693a#diff-5e4d410309bd105ae381c12d2e5bfc6d21e4bd0645ddd842d77e8abd5a05d17eL8

scroll down to the src/open_inwoner/react/i18n/compiled/en.json file etc.

fix-json-newlines.js
jiromaykin: after checking again: it does have a newline, hence why github doesn't show red warning sigs :-)

src/open_inwoner/kvk/tests/test_template_tags.py
jiromaykin: I guess option 2 is better, so there may be less code duplication.

src/open_inwoner/components/templatetags/kvk_tags.py
jiromaykin: But if I do this than I will lose the Rechtspersoon in case of empty string, so if vestigingsnummer is missing, use 'rechtspersoon'". BUT if vestigingsnummer exists and is an empty string (""), it would use that empty string instead of "rechtspersoon"

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: Seems like i will need useEffect for it, and some kind of hook.

fix-json-newlines.js
swrichards, line 1: The `pre-commit` end of file fixer should auto-fix this, let's not include this in the repo. You should already have it with your dependencies, so you can just run `pre-commit install` to have this run locally whenever you commit.

src/open_inwoner/templates/pages/kvk/branches.html
jiromaykin: Well, here i could have this side-effect because I didn't make a deeper separate html component for the Combobox only, outside of the page, but still wondering if that is actually a good idea.

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,125 @@
...
+    if address_data.get("huisnummer"):
+        address_parts.append(str(address_data["huisnummer"]))
+    if address_data.get("huisnummerToevoeging"):
+        address_parts.append(address_data["huisnummerToevoeging"])
+
+    return " ".join(address_parts)
+
+
+@register.simple_tag
+def react_combobox_data(branches, selected_id=None):
```
swrichards: I think `react_combobox_data` is a bit misleading, it makes it sounds generic, but this specifically and only handles branches. `vestigingen_combox_data` would be more explicit.

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,125 @@
...
+        "items": items,
+        "selected_id": escape(
+            str(selected_id or "")
+        ),  # Pre-selected branch for React component
+        # No translations here - handled by react-intl in the React component
+    }
+
+    # We need mark_safe here because JSON needs to be output as-is in the template
+    # All individual data values are already escaped above for security
+    return mark_safe(json.dumps(data, ensure_ascii=True))  # noqa: S308
```
swrichards: Why is `ensure_ascii` needed?

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,369 @@
...
+  }, [value, delay])
+
+  return debouncedValue
+}
+
+/**
+ * Accessible combobox component for branch selection with search filtering.
+ * Supports keyboard navigation, mouse/touchscreen, and form submission.
+ */
+export const SelectComboBox: FC<SelectComboBoxProps> = ({
```
swrichards: As in the template: this component makes assumptions about the content, it;s not a generic `SelectComboBox` but a `VestigingSelector` or some such.

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -250,20 +260,33 @@ def test_get_branches_page_one_branch_found(self, mock_solo, mock_kvk):
 
         doc = PyQuery(response.content)
 
-        branch_inputs = doc.find("[name='branch_number']")
+        # Verify branch data is properly serialized as JSON for React consumption
+        branch_data_script = doc.find("script.branch-data")
+        self.assertEqual(len(branch_data_script), 1)
+
+        # Parse the JSON data to verify it contains the expected branches
+        import json
```
swrichards:
```suggestion
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,369 @@
...
+export interface ComboBoxItem {
+  id: string
+  label: string
+  vestigingInfo?: string
+  rechtspersoonInfo?: string
+  addressInfo?: string
+  cityInfo?: string
+  vestigingsnummer?: string
+  type?: string
+  [key: string]: any
```
swrichards: Doesn't this more or less negate the rest of the type?

jiromaykin:
> Typing "Donald" (the company name is "BV Donald") followed by a space i immediately get the nevenvestiging suggested as the only option. The hoofdvestiging somehow disappears. 

@pi-sigma  Yeah the name of the branch will not have a space after it. This would not be as big of a search-functionality as Elasticsearch, but I'll need to check to see how much of a 'combined' search functionality we could to put in.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,369 @@
...
+  const filtered = branches.filter(
+    (b) =>
+      b.label.toLowerCase().includes(debouncedQuery.toLowerCase()) ||
+      b.vestigingInfo?.toLowerCase().includes(debouncedQuery.toLowerCase()) ||
+      b.rechtspersoonInfo
+        ?.toLowerCase()
+        .includes(debouncedQuery.toLowerCase()) ||
+      b.addressInfo?.toLowerCase().includes(debouncedQuery.toLowerCase()) ||
+      b.cityInfo?.toLowerCase().includes(debouncedQuery.toLowerCase())
+  )
```
swrichards: Vestiging have a natural key on `vestigingsnummers`, couldn't we just use that?

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -0,0 +1,369 @@
...
+          <div
+            className="utrecht-listbox oip-combobox__popover oip-combobox__popover--block-end"
+            role="listbox"
+            id={`${id}-listbox`}
+            aria-label={label}
+          >
+            <ul className="utrecht-listbox__list" ref={listRef}>
+              {filtered.map((item, i) => (
+                <li
+                  key={item.id}
```
swrichards, line 327: If we use the vestigingsnummer here, it'll be order-independent, avoiding potentially nasty edge-cases of stale renders due to sorting issues induced elsewhere.

src/open_inwoner/templates/pages/kvk/branches.html
swrichards: Would `id` make more sense? Is there a use-case for multiple instances?
```suggestion
                    <script type="application/json" id="branch-data">{% react_combobox_data company_branches user.vestiging %}</script>
```

src/open_inwoner/react/modules/SelectComboBoxModule/SelectComboBoxModule.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { SelectComboBox } from '@react/components/SelectComboBox/SelectComboBox'
+import { getIntlProviderProps } from '@react/i18n/i18n'
+
+export interface ComboBoxItem {
+  id: string
+  label: string
+  subLabel?: string
+  vestigingsnummer?: string
+  type?: string
+  [key: string]: any
```
swrichards: Defeats the purpose:

```suggestion
```

src/open_inwoner/react/modules/KVKBranchSelectorModule/KVKBranchSelectorModule.tsx
swrichards, line 1: As above, I think we should name this for what it is: `VestigingSelecorModule` or some such.

src/open_inwoner/components/templatetags/kvk_tags.py
```diff
@@ -0,0 +1,125 @@
...
+        address_parts.append(address_data["straatnaam"])
+    if address_data.get("huisnummer"):
+        address_parts.append(str(address_data["huisnummer"]))
+    if address_data.get("huisnummerToevoeging"):
+        address_parts.append(address_data["huisnummerToevoeging"])
+
+    return " ".join(address_parts)
+
+
+@register.simple_tag
```
swrichards: Using a tag here I think introduces unnecessary complexity. The tag converts `company_branches` into the format required by the react component, but `company_branches` _only exists_ for this page, so the desired format can be achieved by directly changing the data pushed into the template context at: https://github.com/maykinmedia/open-inwoner/blob/674eaf097ec8d39ae7a33cafa0eb8548312c2680/src/open_inwoner/kvk/views.py#L22 (this can be a new method on that view class, to convert the raw API data to what you need).

src/open_inwoner/react/components/MaterialIcon/MaterialIcon.tsx
swrichards, line 1: Should we rename the file/folder to `MaterialIcon` as that's the component name?

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,369 @@
...
+  FC,
+  useState,
+  useRef,
+  useEffect,
+  ChangeEvent,
+  KeyboardEvent,
+} from 'react'
+import { useIntl } from 'react-intl'
+import clsx from 'clsx'
+import { MaterialIcon } from '../MaterialIcons/MaterialIcons'
```
swrichards: I can't remember for sure, but didn't we alias `@` so we can do something like:


```suggestion
import { MaterialIcon } from '@/components/MaterialIcons/MaterialIcons'
```

src/open_inwoner/react/components/MaterialIcon/MaterialIcon.tsx
```diff
@@ -4,7 +4,10 @@ export interface MaterialIconProps {
   name: string
 }
 
-// Material Icons component
+/**
+ * Material Icons component
+ * Icons should never be clickable (aria-hidden="true"), always surround with buttons
+ */
 export const MaterialIcon: React.FC<MaterialIconProps> = ({ name }) => (
```
swrichards, line 11: If we add an `index.ts` file at this level, and in there do `export { MaterialIcon } from './MaterialIcons.tsx'`, we don't need the double `MaterialIcons/MaterialIcon` import.

src/open_inwoner/react/modules/SelectComboBoxModule/SelectComboBoxModule.tsx
```diff
@@ -0,0 +1,85 @@
...
+        let root: Root | undefined = (el as any)[rootKey]
+
+        // Extract branch data from embedded JSON script tag
+        const script = el.querySelector<HTMLScriptElement>('script.branch-data')
+        let items: ComboBoxItem[] = []
+        let selectedId: string | undefined
+
+        if (script) {
+          try {
+            const data = JSON.parse(script.textContent || '{}')
```
swrichards: Missing `textContent` should be an error case right? Do we want to handle it with `|| {}`?

src/open_inwoner/react/modules/SelectComboBoxModule/SelectComboBoxModule.tsx
```diff
@@ -0,0 +1,85 @@
...
+            items = data.items || []
+            selectedId = data.selected_id
+          } catch (err) {
+            console.error(
+              '[SelectComboBoxModule] Failed to parse branch data:',
+              err
+            )
```
swrichards: I wonder if it makes sense to proceed beyond this point: if no branches, the whole exercise doesn't make sense and we should show a user-friendly error message.

src/open_inwoner/react/modules/SelectComboBoxModule/SelectComboBoxModule.tsx
```diff
@@ -0,0 +1,85 @@
...
+              branches={items}
+              selectedBranchId={selectedId}
+            />
+          </IntlProvider>
+        )
+
+        root.render(element)
+      } catch (err) {
+        console.error('[SelectComboBoxModule] Failed to mount component:', err)
+        el.innerHTML = `<div style="color:red;">⚠️ Failed to load SelectComboBox</div>`
```
swrichards: I think this should be something user-friendly, along the lines of "We cannot show your vestiging due to a technical error"

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -0,0 +1,369 @@
...
+    <div
+      className="utrecht-form-field utrecht-form-field--text"
+      ref={containerRef}
+    >
+      <label htmlFor={id} className="utrecht-form-label">
+        {label}
+      </label>
+
+      <div className="oip-combobox">
+        <input
```
swrichards, line 263: Should we do some kind of ellipsis so the close icon isn't rendered over the text within the text input?

Image

src/open_inwoner/react/lib/loader.ts
```diff
@@ -1,22 +1,20 @@
+// src/open_inwoner/react/lib/loader.ts
```
swrichards:
```suggestion
```

src/open_inwoner/templates/pages/kvk/branches.html
```diff
@@ -1,75 +1,50 @@
 {% extends "master.html" %}
-{% load i18n static custom_session_tags form_tags grid_tags button_tags card_tags button_tags %}
+{% load i18n static form_tags grid_tags button_tags card_tags kvk_tags %}
+
+{% block mount_modules %}
+    data-mount-modules='["selectcombobox"]'
+{% endblock %}
+
+{% block react_modules %}
+{# Including this script will allow the page to use the react component #}
```
swrichards: I think this is clear from the block name.
```suggestion
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,369 @@
...
+  }, [value, delay])
+
+  return debouncedValue
+}
+
+/**
+ * Accessible combobox component for branch selection with search filtering.
+ * Supports keyboard navigation, mouse/touchscreen, and form submission.
+ */
+export const SelectComboBox: FC<SelectComboBoxProps> = ({
```
swrichards: Two comments (but they might go against the design):

- I find `(Rechtspersoon`) a bit sparse. I would expect something like "Wissel naar de rechtspersoon zonder een specifieke vestiging te selecteren"
- I find it slightly strange that the KvK number isn't visible anywhere, it seems like valuable context (And would fill up the Rechtspersoon selector a bit

Image

swrichards (changes requested):

src/open_inwoner/kvk/tests/test_views.py
```diff
@@ -166,10 +166,20 @@ def test_post_branches_page_with_incorrect_vestigingsnummer(
...
+
+        branch_data = json.loads(branch_data_script.text())
+        branch_items = branch_data.get("items", [])
+
+        # Should have 3 items: entire company + 2 branches
+        self.assertEqual(len(branch_items), 3)
```
swrichards: This is a test so we're in control of the conditions: we should know whether we expect there to be a valid object or not, e.g.:

```suggestion
        # Should have 3 items: entire company + 2 branches
        self.assertEqual(len(branch_data["items"]), 3)
```

src/open_inwoner/react/components/MaterialIcon/MaterialIcon.tsx
jiromaykin: Yeah, it's also a singular export.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: That's indeed a bunch of design decisions. 
Looks like the design actually forgot we have a Rechtspersoon at all - maybe we need that to be much more clear anyway? Because it has different consequences - should the rechtspersoon be part of the dropdown? Currently it is separate from the actual vestigingen. I think I'd like to put that in front of the UX'er without blocking this PR.

But then the question remains on what to do now.❓

KVK nr is not visible currently in old design, nor is it in this new design.
Not until user navigates to Profile.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: Yes but a bit different,  "@react/" and   "@webcomponents/".

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: Note: This `item.id` actually comes from kvk_tags here:

```
branch_id = branch.get("vestigingsnummer") or "rechtspersoon"

[.....]
        # React will display these values in the interactive dropdown
        items.append(
            {
                "id": escape(
                    str(branch_id)
                ),  # Used as form value for branch_number field
```

And no sorting happens on the front-end. React won't have outdated stale stuff here because the key is unique, and is already the vestigingsnummer in most cases.
And AFAIK the Django view always puts Rechtspersoon at position 0. If there are no Vestigingen at all, we have a redirect, right?

src/open_inwoner/templates/pages/kvk/branches.html
jiromaykin: In that case we would have to remove it from all the other 9 templates that also have this line @stefrado

src/open_inwoner/components/templatetags/kvk_tags.py
jiromaykin: it could be set to False, I guess, so we can still have a readable version of "Crémière & Fils du Spécialités Françaises"

src/open_inwoner/react/modules/KVKBranchSelectorModule/KVKBranchSelectorModule.tsx
jiromaykin: Not sure if that conflicts with the NLDS community.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: "Ackchyually" that text is also the user-typing-in text, so for accessibility reasons  you're not allowed an ellipsis there. Not sure what happens when a screenreader is reading out the typed text and suddenly a bunch of dots gets injected. Here you would want the entire name to always be able to be read out loud. 
Also: browsers don't apply`text-overflow: ellipsis` to `<input>` elements because the text inside is meant to be editable and scrollable.

I can however make the input less wide so it doesn't look sad and painful.

src/open_inwoner/components/templatetags/kvk_tags.py
jiromaykin: for back-end improvement, i've created this sub-task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3490

src/open_inwoner/react/components/MaterialIcon/MaterialIcon.tsx
jiromaykin: That's a nice-to-have that's not fully related to the combobox -> we have multiple React dirs currently that do not yet have this. Not sure if all these refactors should just be a new issue.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: in case we need more fields later, but can be removed.

src/open_inwoner/kvk/tests/test_views.py
jiromaykin: Can be handled in sub-task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3490

src/open_inwoner/templates/pages/kvk/branches.html
jiromaykin: Not if we would rename it to something like SelectComboBoxVestigingen, but we may in future have multiple comboboxes in zaak filtering etc. So for this 1 component ID would be fine, but not for a future generic component, because "you never know...".

src/open_inwoner/react/modules/SelectComboBoxModule/SelectComboBoxModule.tsx
jiromaykin: Yeah! I could add a console Error if the script tag exists but is empty.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: What if I rename it to KVKBranchSelector?

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
stefrado: The best way to ressolve this is to add a padding-right as large as the icon + the gap we want between text and icon.

jiromaykin: I already did that with a calc().

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+  branches: ComboBoxItem[]
+  selectedBranchId?: string
+}
+
+/**
+ * Custom hook for debouncing a value
+ * Delays updating the returned value until after the specified delay
+ * has elapsed since the last time the input value changed
+ */
+function useDebounce<T>(value: T, delay: number): T {
```
stefrado: This should not be written inside this component, but inside a file named `utils/hooks.ts` or `lib/hooks.ts`

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+  const inputRef = useRef<HTMLInputElement>(null)
+  const containerRef = useRef<HTMLDivElement>(null)
+
+  // Find the selected branch to display its name in the input
+  const selectedBranch = branches.find((b) => b.id === selectedId)
+
+  // Filter branches based on debounced search query across all text fields
+  // This expensive operation only runs after user stops typing for 300ms
+  // Split query and make substrings-tolerant search
+  const filtered = branches.filter((b) => {
```
stefrado: This looks like a heavy calculation which is not placed inside a useMemo

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+    const rechtspersoonBranch = branches.find((b) => b.id === 'rechtspersoon')
+    return rechtspersoonBranch ? 'rechtspersoon' : undefined
+  })
+  const [focusedIndex, setFocusedIndex] = useState<number>(0)
+  const listRef = useRef<HTMLUListElement>(null)
+  const inputRef = useRef<HTMLInputElement>(null)
+  const containerRef = useRef<HTMLDivElement>(null)
+
+  // Find the selected branch to display its name in the input
+  const selectedBranch = branches.find((b) => b.id === selectedId)
```
stefrado: Might place this inside a useMemo, we do not want to recalculate the `selectedBranch` every re-render

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+                <li
+                  key={item.id}
+                  role="option"
+                  aria-selected={selectedId === item.id}
+                  className={clsx(
+                    'utrecht-listbox__option',
+                    selectedId === item.id &&
+                      'utrecht-listbox__option--selected',
+                    focusedIndex === i && 'utrecht-listbox__option--active'
+                  )}
```
stefrado: With clsx you could do something like this
```suggestion
                  className={clsx(
                    'utrecht-listbox__option', {
                      ['utrecht-listbox__option--selected']: selectedId === item.id,
                      ['utrecht-listbox__option--active']: focusedIndex === i,
                    }
                  )}
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+
+  // Sync selection with hidden form input for Django form submission
+  useEffect(() => {
+    const form = inputRef.current?.closest('form')
+    if (!form) return
+
+    // Create or update hidden input with selected branch value
+    let hiddenInput = form.querySelector(
+      `input[name="${name}"]`
+    ) as HTMLInputElement
```
stefrado:
```suggestion
    let hiddenInput = form.querySelector<HTMLInputElement>(`input[name="${name}"]`)
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+
+    // Map 'rechtspersoon' ID to empty string for Django backend
+    if (selectedId !== undefined) {
+      hiddenInput.value = selectedId === 'rechtspersoon' ? '' : selectedId
+    }
+
+    // Enable/disable submit button based on selection state
+    const submitButton = form.querySelector(
+      'button[name="submit"]'
+    ) as HTMLButtonElement
```
stefrado:
```suggestion
    const submitButton = form.querySelector<HTMLButtonElement>(
      'button[name="submit"]'
    )
```

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -0,0 +1,388 @@
...
+      setQuery(selectedBranch.label)
+    }
+  }, [selectedBranch, query])
+
+  // Scroll focused option into view for keyboard navigation
+  useEffect(() => {
+    if (isOpen && listRef.current && listRef.current.children[focusedIndex]) {
+      const el = listRef.current.children[focusedIndex] as HTMLElement
+      el.scrollIntoView({ block: 'nearest', behavior: 'smooth' })
+    }
```
stefrado, line 212: Might want to create a helper function for the smooth scroll -> `smoothScrollIntoView(element)`.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    if (isOpen) {
+      document.addEventListener('mousedown', handleClickOutside)
+    }
+
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside)
+    }
+  }, [isOpen])
```
stefrado: containerRef is also a changing dependency. Might want to include it inside the dependency array.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+    const newIsOpen = !isOpen
+    setIsOpen(newIsOpen)
+    if (newIsOpen) {
+      setFocusedIndex(0)
+      // Only focus input when opening to prevent conflict with close action
+      if (inputRef.current) inputRef.current.focus()
+    }
+  }
+
+  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
```
stefrado:
```suggestion
  const handleKeyDown: KeyboardEventHandler<HTMLInputElement> = (e) => {
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+
+  const handleClearQuery = () => {
+    setQuery('')
+    setSelectedId(undefined)
+    setFocusedIndex(0)
+    setIsOpen(true)
+    if (inputRef.current) inputRef.current.focus()
+  }
+
+  const handleToggleDropdown = (e: React.MouseEvent<HTMLButtonElement>) => {
```
stefrado:
```suggestion
  const handleToggleDropdown: MouseEventHandler<HTMLButtonElement> = (e) => {
```

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
```diff
@@ -0,0 +1,388 @@
...
+        setIsOpen(false)
+        break
+
+      case 'Tab':
+        setIsOpen(false)
+        break
+    }
+  }
+
+  const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
```
stefrado:
```suggestion
  const handleInputChange: ChangeEventHandler<HTMLInputElement> = (e) => {
```

src/open_inwoner/react/components/SelectComboBox/index.ts
```diff
@@ -0,0 +1 @@
+export { SelectComboBox } from './SelectComboBox'
```
stefrado: No new line

stefrado (changes requested):

src/open_inwoner/react/components/SelectComboBox/index.ts
jiromaykin: Yes there is.

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: If we will have multiple hooks in the future let's create a Hooks dir - probably in 'lib' because we are using that already.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: You mean in lib/smoothScrollIntoView.ts kind of file?
A helper function for just 1 line seems a bit like overkill right now?

src/open_inwoner/react/components/SelectComboBox/SelectComboBox.tsx
jiromaykin: Yes, this will be heavy, will wrap it.

CHANGELOG.rst
```diff
@@ -26,12 +26,12 @@ Nieuwe features
 * [:taiga-us:`3461`, :taiga-ta:`3473`, :pr:`1932`]: Basisinfrastructuur voor
   OpenTelemetry toegevoegd, inclusief logging en metrics ondersteuning voor
   observability.
+* [:taiga-us:`3408`, :pr:`1881`]: Nieuw selectiescherm voor eHerkenning-inlog met zoek-
+  en dropdownfunctie voor betere UX bij grote aantallen vestigingen.
```
stefrado: Might want to include the switching of branches as well

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -0,0 +1,369 @@
...
+  useEffect,
+  MouseEventHandler,
+  KeyboardEventHandler,
+  ChangeEventHandler,
+  useMemo,
+} from 'react'
+import { useIntl } from 'react-intl'
+import { useDebounce } from '@react/lib/hooks/useDebounce'
+import clsx from 'clsx'
+import { MaterialIcon } from '@react/components/MaterialIcon'
```
stefrado, line 14: I think we should make it a requirement to sort the TS imports:

Please run `TS sort imports`. Now sure how that works in PyCharm

src/open_inwoner/react/components/SideNav/SideNav.spec.tsx
```diff
@@ -6,7 +6,7 @@ import SideNav from './SideNav'
 import '@testing-library/jest-dom'
 
 // Mock MaterialIcon to simplify rendering and testing
-vi.mock('../MaterialIcons/MaterialIcons', () => ({
+vi.mock('@react/components//MaterialIcon/MaterialIcon', () => ({
```
stefrado: Not sure if this is the correct source - it might be:

```suggestion
vi.mock('@react/components/MaterialIcon', () => ({
```

src/open_inwoner/react/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedValue(value)
+    }, delay)
+
+    // Cleanup timer if value or delay changes before timeout
+    return () => {
+      clearTimeout(handler)
+    }
```
stefrado: Same functionality shorter code :)

```suggestion
  useEffect(() => {
    const handler = setTimeout(() => setDebouncedValue(value), delay)

    // Cleanup timer if value or delay changes before timeout
    return () => clearTimeout(handler)
```

src/open_inwoner/react/lib/hooks/useDebounce.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Delays updating the returned value until after the specified delay
+ * has elapsed since the last time the input value changed
+ *
+ * @template T - The type of the value being debounced.
+ * @param value - The current value to debounce.
+ * @param delay - The debounce delay in milliseconds.
+ * @returns The debounced value that only updates after the delay period.
+ */
+
+export function useDebounce<T>(value: T, delay: number): T {
```
stefrado: Might include a default value

```suggestion
export function useDebounce<T>(value: T, delay: number = 300): T {
```

src/open_inwoner/react/modules/KVKBranchSelectorModule/KVKBranchSelectorModule.tsx
```diff
@@ -0,0 +1,82 @@
...
+        let root: Root | undefined = (el as any)[rootKey]
+
+        // Extract branch data from embedded JSON script tag
+        const data = getJsonFromScriptTag<{
+          items: ComboBoxItem[]
+          selected_id?: string
+        }>('branch-data')
+
+        if (!data || !data.items || data.items.length === 0) {
+          console.error('[KVKBranchSelectorModule] No branches available')
```
stefrado: I usually add `new Error`, bc this gives some context about the stack trace.

```suggestion
          console.error(new Error('[KVKBranchSelectorModule] No branches available'))
```

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+    justify-content: space-between;
+    width: 100%;
+  }
+
+  &__list {
+    margin-block-start: 0;
+    margin-block-end: 0;
+    padding-block-start: 0;
+    padding-block-end: 0;
+    padding-inline-start: 0;
```
stefrado: Is there a reason why this is split up.
```suggestion
    margin-block: 0;
    padding-block: 0;
    padding-inline-start: 0;
```

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+  &__popover {
+    position: var(--oip-combobox-popover-position);
+    inset: var(--oip-combobox-popover-inset);
+    z-index: var(--oip-combobox-popover-z-index, 1);
+    overflow: var(--oip-combobox-popover-overflow);
+    block-size: var(--oip-combobox-popover-block-size);
+    inline-size: var(--oip-combobox-popover-inline-size);
+    border: none;
+    border-top: 1px solid var(--oip-color-fg);
+    border-top-color: var(--oip-color-fg);
```
stefrado: This is already included in the `border-top` right?

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+      border-radius: 6px 6px 0 0;
+    }
+  }
+
+  &__popover {
+    position: var(--oip-combobox-popover-position);
+    inset: var(--oip-combobox-popover-inset);
+    z-index: var(--oip-combobox-popover-z-index, 1);
+    overflow: var(--oip-combobox-popover-overflow);
+    block-size: var(--oip-combobox-popover-block-size);
```
stefrado, line 54: Nice one - didn't knew it exists - now i do.

Same for `inline-size`

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+    padding-block-end: var(--utrecht-textbox-padding-block-end);
+    padding-block-start: var(--utrecht-textbox-padding-block-start);
+    width: calc(100% - (2 * var(--spacing-extra-large)));
+  }
+
+  .button {
+    color: var(--oip-color-fg);
+    position: absolute;
+    right: 2px;
+    top: 2px;
```
stefrado: Is there a reason for these `2px`

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+    &--active {
+      background-color: var(--utrecht-listbox-option-hover-background-color);
+    }
+  }
+
+  &__heading {
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    color: var();
```
stefrado: Empty var.

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
```diff
@@ -0,0 +1,139 @@
...
+
+  .button {
+    color: var(--oip-color-fg);
+    position: absolute;
+    right: 2px;
+    top: 2px;
+
+    // Only rotate the dropdown button, not the clear button
+    &[aria-expanded='true'] {
+      transition: transform 0.3s;
```
stefrado: I prefer `ms` over seconds `s` in CSS - not sure if this is ever discussed

stefrado (changes requested):
Besides these code changes there are also a couple of UX issues.

### Issue met het leegmaken van de input 

https://github.com/user-attachments/assets/8201ec77-d90b-4d36-b186-f1572e131b09

### Opties worden gefilterd na het selecteren

Ik verwacht hier dat na het selecteren van een optie de combobox sluit.

https://github.com/user-attachments/assets/b0ca8977-7234-4c5e-a59b-9a2de8e96661

### Dubbele border-bottom

Image

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: Yes, there's a PyCharm key combo (Ctrl+Alt+O).

src/open_inwoner/react/i18n/locales/en.json
```diff
@@ -3,5 +3,20 @@
...
+  "combobox.clear": {
+    "defaultMessage": "Clear",
+    "description": "Clear search button label",
+    "originalDefault": "Wissen"
+  },
+  "combobox.placeholder": {
+    "defaultMessage": "Enter name, address or branch number",
+    "description": "Placeholder text for branch search input",
+    "originalDefault": "Vul naam, adres of vestigingsnummer in..."
+  },
```
swrichards: Nitpick: the names are generic, it might make more sense for them to match the new (and improved!) component name: `kvkbranchselector.placeholder` etc.

src/open_inwoner/react/components/SideNav/SideNav.spec.tsx
jiromaykin: Find-&-replace gone wrong! Funny how it still passed the tests?

swrichards (approved):
Just a nitpick, but no more comments from me. Once this is in I'll do the backend improvements.

src/open_inwoner/react/modules/KVKBranchSelectorModule/KVKBranchSelectorModule.tsx
jiromaykin: Nice.

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
jiromaykin: If you search for "0.3s" you can see that multiple times in this project. I'm fine with moving to milliseconds, but not for all of them in this PR - i'd like to make a refactor issue for multiples of these things that are not fully related to the search-select component.

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
jiromaykin: The goodness of NLDS!

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
jiromaykin: They are usually all split up and defined separately in separate designtokens for NLDS components. So they never use the shorthand versions. So if we would want to "give this component back to the community" we would have to supply all the separate values.

src/open_inwoner/scss/components/SelectComboBox/SelectComboBox.scss
jiromaykin: I should actually make new design tokens for these. With "`border-style.solid`" etc.

src/open_inwoner/react/i18n/locales/en.json
jiromaykin: New and improved!

pi-sigma (approved):

src/open_inwoner/react/i18n/compiled/en.json
stefrado, line 1: These are not the english translations

stefrado (approved):
@swrichards Approved - but the english translation should be fixed (could be done in backend PR) :)

swrichards:
@stefrado @jiromaykin two UX points I noticed (don't have to hold up the PR, can be a later improvement if there's time):

1. If the branch has the same name as the rechtspersoon, there's no distinguishing characteristic in the input after selection (see screenshot, both rechtspersoon and vestiging selection show `Test BV Donald`)
2. I would expect focusing the input to open the dropdown again, but it doesn't -- you have to delete the existing value first. Might be wide off the mark here, but my brain seems conditioned to expect focusing the input to also open the dropdown.


https://github.com/user-attachments/assets/62bfc210-f35a-4595-a9b2-a728bc0e75cc

swrichards:
> @stefrado @jiromaykin two UX points I noticed (don't have to hold up the PR, can be a later improvement if there's time):
> 
>     1. If the branch has the same name as the rechtspersoon, there's no distinguishing characteristic in the input after selection (see screenshot, both rechtspersoon and vestiging selection show `Test BV Donald`)
> 
>     2. I would expect focusing the input to open the dropdown again, but it doesn't -- you have to delete the existing value first. Might be wide off the mark here, but my brain seems conditioned to expect focusing the input to also open the dropdown.
> 
> 
> Screencast.from.2025-10-08.16-34-09.mp4

And more UX nitpick (but this might be a feature rather than a bug):

- Focusing the input and hitting backspace until the field is empty, resets the state to the full name of the selected item


https://github.com/user-attachments/assets/f9eaa0d4-ce9f-44e6-8b2d-73157976938c

src/open_inwoner/kvk/tests/test_views.py
swrichards: Not resolved

src/open_inwoner/react/i18n/compiled/en.json
jiromaykin: I'll fix them.

PR #1890, opened 2025-09-08
Support OpenZaak 1.20 style rol parameters for company case list fetching

swrichards:
[Taiga 3317](https://taiga.maykinmedia.nl/project/open-inwoner/us/3317).


docs/beheerhandleiding/09_datakoppelingen.rst
```diff
@@ -184,7 +184,29 @@ Deze API haalt de documenten op die bij een bepaalde zaak horen.
...
+wordt:
+
+**Maak gebruik van het RSIN voor ophalen eHerkenning zaken**: indien aangevinkt, zullen
+eHerkenning gebruikers geidentificeerd worden via de RSIN in plaats van het KVK nummer.
+
+**Klanten API back-end**: De backend waarin de klantgegevens voor deze API set staan. In
+de regel zal dit eSuite zijn voor eSuite zaaksystemeen, en Openklant 2 voor Open Zaak.
+Laat dit veld leeg indien u geen gebruik maakt van een klant systeem.
+
+**Fetch Zaken for users authenticated with eHerkenning using OpenZaak 1.20 query
```
pi-sigma, line 200: Are the newly supported query params the standard for OpenZaak >= 1.20? Then we should modify the text in the docs and the help text accordingly (something like "...using new query parameteres for OpenZaak 1.20+"). Current text makes it seem as if these are specific to v1.20.

pi-sigma (approved):
The text for the docs and model could be made clearer (though perhaps this is just my impression). Otherwise lgtm

docs/beheerhandleiding/09_datakoppelingen.rst
swrichards: Made a few tweaks, thanks.

PR #1891, opened 2025-09-09
build(deps): run npm audit fix

swrichards:


swrichards:
Bit of an experiment, but the drawback is it fixes everything at once, making for an unreadable diff and no breadcrumbs in the changelog or commit message. Best to do this either via dependabot or manual updates for specific packages.

PR #1902, opened 2025-09-15
[#3443] Update documentation for CMS contactform

jiromaykin:
Update documentation for #1872 

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3443

----
📚 Documentation preview 📚: https://open-inwoner--1902.org.readthedocs.build/en/1902/


jiromaykin:
@pi-sigma I've also added you as a reviewer so you can check if my story here is factually correct. 
I am not sure about the naming of the to-be-connected "openklant application" in advanced settings of the CMS page + configuring the klantensysteem back-end with some "contactformulier onderwerpen." 
Additionally it looks like the front-end will show an ugly dropdown icon if zero "contactformulier onderwerpen" are in the dropdown...

docs/beheerhandleiding/11_cms.rst
```diff
@@ -420,10 +420,26 @@ Wanneer u de paginalijst in de footer wilt weergeven, selecteert u deze plugin.
 11.2.10. Contactformulier aanpassen via CMS
 -------------------------------------------
 
-Het contactformulier van Open Inwoner kan worden aangepast via het Django CMS. Dit doet u door naar het CMS menu te gaan. U krijgt dan frontend van het Open Inwoner de opties om pagina’s te wijzigen, maar als u op de pagina van het contactformulier bent vindt u i het plugin menu ook de contactformulier plugin.
+Het contactformulier van Open Inwoner kan worden aangepast via het Django CMS. Hiervoor dient eerst een primaire back-end te zijn geconfigureerd, met een koppeling naar een klanten systeem. Ook dient deze back-end een aangemaakt "contact formulier onderwerp" te hebben, anders worden deze formulieronderwerpen niet zichtbaar in de keuzelijst van het gepubliceerde contactformulier.
```
swrichards: > Hiervoor dient eerst een primaire back-end te zijn geconfigureerd, met een koppeling naar een klanten systeem.

@pi-sigma correct me if I'm wrong, but it's also possible to use this without a klanten systeem configured but with the "send requests as email" option enabled, right?

swrichards (approved):
Very nice! Just one comment about whether or not the klanten systeem is mandatory, I'll let @pi-sigma weigh in.

docs/beheerhandleiding/11_cms.rst
pi-sigma: That's right. We register contactmomenten via email or API or both. So API service/klanten systeem configuratie is not necessary if the email registration is set.

pi-sigma: It's also not strictly necessary to create a "contact formulier onderwerp". Although in that case, as you point out in a comment, the dropdown will be empty.

jiromaykin: Ooh, so I should probably mention that here? I haven't looked if there is another section in the Documentation that also talks about setting a preferred back-end for this.

swrichards: I would just go for a generic "... dient u eerst de configuratie voor openklant te voltooien" (maybe with a ref to `9.4`).

pi-sigma (changes requested):
The section about the API service/email and the "contact formulier onderwerp" should be adjusted, otherwise looking good!

swrichards (approved):

pi-sigma (approved):

PR #1903, opened 2025-09-15
[#3455] feat: add timeline logging around zgw notification handling

swrichards:
[Taiga 3455](https://taiga.maykinmedia.nl/project/open-inwoner/task/3455).


swrichards:
Discussed with @alextreme -- the timeline logger is important for auditing purposes, and an auto-prune is therefore a bit of an anti-pattern. The underlying desire -- to be able to use timeline logger more liberally as an operational logger -- remains, but we'll push that towards the upcoming OTEL/Grafana tasks.

PR #1908, opened 2025-09-16
[#3459] Fetch CMS page for contactformulier based on template

pi-sigma:
The CMS page for the contact form is fetched by first getting the CMS plugin and the associated placeholders. However, if no plugin has been created yet, the link to the contact form page won't show up in the footer; it is only accessible by manually entering the URL. This is fixed by fetching the CMS page based on the template, which works because we have a dedicated template for the contact form.

Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3459


swrichards (approved):

swrichards (commented):
Sorry to be a stickler: could you add this pr to the existing changelog entry?

PR #1914, opened 2025-09-16
Add pre commit qa checks to ci

swrichards:
## Summary

Prompted by the recent end-of-line-whitespace/diffing issues with regards to the changelog, I was finally motivated to add some proper tooling to ensure basic formatting hygiene. I used some standard `pre-commit` hooks for this, to checK

- end-of-file-fixer (adds a newline at the end of each file, if present)
- trailing-whitespace (removes trailing whitespaces)
- check-toml (ensure that all TOML is parseable)
- check-yaml (ensure that YAML is parseable)
- mixed-line-ending (detect mixed posix/win32 line endings)
- check-added-large-files (don't accidentally commit a disk image)
- check-merge-conflict (warn if you forgot to resolve a merge conflict)

As part of this, I've also migrated the CI checks for ruff to `pre-commit`, and added a check to ensure our `pre-commit` `ruff` version is in sync with the version declared in the dependencies.


pi-sigma (approved):
Where do the changes related to the `svg` files come from, are they necessary?

swrichards:
> Where do the changes related to the `svg` files come from, are they necessary?

@pi-sigma SVG files are also text, so can have extraneous whitespace, missing EOF, etc. This formats them consistently -- it's not strictly speaking necessary, but it can't hurt Id say.

.pre-commit-config.yaml
```diff
@@ -1,9 +1,16 @@
...
-    - id: ruff
-    # Run the formatter.
-    - id: ruff-format
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.11.7
+    hooks:
+      - id: ruff
+      - id: ruff-format
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
```
jiromaykin: Looks like pre-commit-hooks currently is around version 6, so why use v2?

.pre-commit-config.yaml
```diff
@@ -1,9 +1,16 @@
...
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: check-toml
+      - id: check-yaml
+      - id: mixed-line-ending
+      - id: check-added-large-files
+      - id: check-merge-conflict
```
jiromaykin, line 16: Maybe also add `check-case-conflict` (case-sensitive issues?) and/or `check-json`?

.github/workflows/code-quality.yml
```diff
@@ -66,36 +66,36 @@ jobs:
...
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.12'
-      - name: Install dependencies
-        run: |
-          pip install uv
-          grep ^ruff requirements/ci.txt | uv pip install --system -r /dev/stdin
-      - name: Run ruff check on all files
-        run: ruff check --output-format=github src
+      - uses: actions/checkout@v3
```
jiromaykin, line 73: A bit up higher on line 61 (non diff) we use: actions/checkout**@v2** for Prettier, so I'm wondering if V2 versus V3 could cause problems.

.github/workflows/code-quality.yml
```diff
@@ -66,36 +66,36 @@ jobs:
...
-      - name: Install dependencies
+      - uses: actions/checkout@v3
+      - name: Check ruff versions match
         run: |
-          pip install uv
-          grep ^ruff requirements/ci.txt | uv pip install --system -r /dev/stdin
-      - name: Run ruff format on all files
-        run: ruff format --diff --target-version py311 src
+          # Extract ruff version from ci.txt
+          CI_VERSION=$(grep "^ruff==" requirements/ci.txt | cut -d'=' -f3)
```
jiromaykin: This works perfectly for things like if the line is exactly like "ruff==0.11.7" but what if there are extra spaces in there or something like "ruff[blabla]==0.11.7" with brackets? Although this may never occur in real life.

jiromaykin (changes requested):
Not actually a hard request for changes, instead mostly just asking questions and a suggestion.

.pre-commit-config.yaml
swrichards: `check-json` currently fails on our `.jshintrc` because it has comments, which this hook doesn't like. But more generally, I was hoping to fix this with #1929 -- if we run prettier on the whole repo, it will also validate JSON validity (as well as YAML actually, so we can remove that hook here once we've expanded the reach of prettier).

`check-case-conflict` is a great idea but we'd currently fail it due to the openklant2 fixture issue, which is still on my backlog to fix :/ Once that's done though we can add that hook.

.github/workflows/code-quality.yml
swrichards: They'll run in isolated jobs, so they shouldn't affect each other (but I might do a separate PR to bump the prettier one to `v3` for consistency).

.github/workflows/code-quality.yml
swrichards: This is a good point. Won't matter much for `ruff` I suspect, but I am potentially going to add more of these so it's good to have a comprehensive pattern in place for future re-use. Added.

.pre-commit-config.yaml
swrichards: _Very_ old copy pasta from a private repo :scream:. Thanks, good catch.

.github/workflows/code-quality.yml
swrichards: Actually I'll bump them all to _v4_ separately.

PR #1928, opened 2025-09-18
:sparkles: [#2538] Add NLDS styles per component

jiromaykin:
## Summary

Use the preferred way of working with NLDS components: import them one by one so styles and HTML/React will become easier to maintain in future updates. 

Install and run `npm run build` to properly see if _none_ of the actual styles have changed. Everything needs to look the same after this update.
Note this is using the **keys** from utrecht NLDS packages, but the **values** come from our own NLDS package.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/2538
- GitHub: Many these 'atoms' are needed for [the KVK branch selection screen](https://github.com/maykinmedia/open-inwoner/pull/1881).

## Checklist

- [x] No breaking changes for deployment or otherwise


swrichards (approved):
No visual gremlins as far as I can see 👍

pi-sigma (approved):

PR #1947, opened 2025-09-29
Do not show dropdown menu items on case list page

swrichards:
## Summary

As part of the missing case list exception, I also refactored this code out of the template into a tag, for clarity and to avoid having too much business logic in templates.

## Issue References

- Taiga Issue: [#3486](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3486)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/components/templatetags/link_tags.py
+++ b/src/open_inwoner/components/templatetags/link_tags.py
@@ -1,3 +1,5 @@
+import logging
+
 from django import template
 from django.urls import NoReverseMatch, reverse
 from django.utils.html import format_html
@@ -5,6 +7,7 @@ from django.utils.html import format_html
 from furl import furl
 
 register = template.Library()
+logger = logging.getLogger(__name__)
 
 
 @register.inclusion_tag("components/Typography/Link.html")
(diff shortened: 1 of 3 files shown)


swrichards:
@stefrado @jiromaykin Could mainly use a double check on the list of excluded URLs to make sure we're now in line with the expectation. @pi-sigma can do the actual code review.

src/open_inwoner/components/templatetags/link_tags.py
```diff
@@ -125,3 +128,59 @@ def addnexturl(href, next_url):
...
+
+    # Exclude specific namespace:url_name combinations where menu items
+    # are already shown in sidebar or other navigation
+    urls_with_minimal_dropdown_menu = {
+        "pages-root",
+        "general_faq",
+        "collaborate:plan_list",
+        "ssd:uitkeringen",
+        "products:category_list",
+        "inbox:index",
```
jiromaykin: Is this 'inbox' for our Mijn Berichten chat thing? We do not have a sidenavigation there yet because it already has a sidebar, so it should have the full dropdown menu (which it has, so doesn't seem to do anything here?).

I'm missing Mijn Balie-afspraken UserAppointments(CMSPlugin) here , but not sure yet if that also has a sidenavigation, since i can't get it to work properly locally.

And perhaps we should not wake sleeping dogs but we also have the http://localhost:8000/mijn-profiel/actions/ page that may need a side navigation.

In future we will need sidenaviagtion and limited dropdownmenu for MijnDocumenten and MijnProducten/MijnServices.

jiromaykin (changes requested):

src/open_inwoner/components/templatetags/link_tags.py
stefrado: > I'm missing Mijn Balie-afspraken UserAppointments(CMSPlugin) here...

Appointment has a side-navigation and should be added to the list - now it has the same duplicate sidenavigation and top-left menu

http://localhost:8000/mijn-profiel/appointments

stefrado (changes requested):

src/open_inwoner/components/templatetags/link_tags.py
swrichards: Removed the inbox, with the caveat discussed (the page seems to ignore the templatetag anyway, but the default behavior is what we want).

pi-sigma (approved):

stefrado (approved):

PR #1949, opened 2025-09-30
[#3454] Replace CKEditor with Prosemirror

pi-sigma:
## Summary

+ Replaced CKEditor rich text fields and textfields with CKEditorWidget with
  ProsemirrorModelField.

+ Updated templatetags + tests for new ProsemirrorModelField

+ Added serializers for ProsemirrorFieldDocument in Elasticsearch indexing

  The Product.content and Category.description fields were converted to
  ProsemirrorModelField, but the Elasticsearch document indexing was trying
  to serialize the ProsemirrorFieldDocument objects directly, causing
  SerializationError.

  Added prepare_content() and prepare_categories() methods to convert
  ProsemirrorFieldDocument objects to HTML strings before indexing.

+ Modules affected:
  - configurations
  - cms
  - ssd
  - questionnaire
  - openklant
  - openzaak
  - pdc

## Issue References

Taiga user story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3450?milestone=548
Taiga task: https://taiga.maykinmedia.nl/project/open-inwoner/task/3454


src/open_inwoner/ssd/migrations/0007_ssdconfig_textfields.py
```diff
@@ -0,0 +1,232 @@
...
+
+    # format italics
+    res = re.sub(r"_(.*?)_", r"<em>\1</em>", res)
+
+    # format links
+    res = re.sub(
+        r"\[([^\]]+)\]\(([^)]+)\)",  # Match [link_text](link)
+        r'<a href="\2" title="\1">\1</a>',  # Replace with <a href="link" title="link_text">link_text</a>
+        res,
+    )
```
swrichards: Note we have `markdown==3.3.6` in our requirements, you could use that to do the conversion.

swrichards (commented):
Looks good! I wonder if it would be useful to add a `.safe_html` shorthand in django prosemirror to make the templates a bit easier to use.

I think you still need to add the dependency.

src/open_inwoner/configurations/migrations/0086_alter_siteconfiguration_warning_banner_text.py
```diff
@@ -0,0 +1,97 @@
...
+        html_content = f"<p>{html_content}</p>"
+
+    # Convert the HTML to a proper ProseMirror document
+    try:
+        doc = html_to_doc(html_content, schema=prosemirror_config.schema)
+        setattr(config, field_name, doc)
+        config.save()
+    except (IndexError, ValueError) as e:
+        # Skip records with problematic HTML content
+        print(f"Warning: Could not convert {field_name} for {model_name}: {e}")
```
swrichards: Maybe a `logger.warning` would be more appropriate here.

src/open_inwoner/pdc/migrations/0069_product_content.py
```diff
@@ -0,0 +1,168 @@
...
+    Image = apps.get_model("filer", "Image")
+
+    def process_content(content):
+        for node in content:
+            if node.get("type") == "filer_image":
+                attrs = node.get("attrs", {})
+                src = attrs.get("src", "")
+
+                # Look up image by URL to get the correct ID
+                if "media" in src.lower() and "imageId" not in attrs:
```
swrichards: I wonder if we need the `media` check, feels possible fragile -- can we use `MEDIA_PATH` and only use `Image.objects.get`? We handle the non-existing case anyway, don't think we have to limit this conditional further beyond checking for `imageId`.

src/open_inwoner/utils/html.py
```diff
@@ -39,14 +39,26 @@ def convert_first_row_to_th(html_tables):
...
-    md = markdown.Markdown(extensions=["tables"])
-    # remove weird undocumented \\< escape/prefix generated by CKeditor
-    content = content.replace("\\<", "<")
-    html = md.convert(content)
+    # Handle empty content
+    if not content:
+        return ""
+
+    # Check if this is a ProseMirror field
+    if hasattr(content, "doc"):
```
swrichards: I think you can be more explicit here: I believe the fields will return a `ProsemirrorFieldDescriptor` instance, which you can import from `django_prosemirror.fields` for `isinstance`

pi-sigma: The `isinstance` test fails for both `ProsemirrorFieldDescriptor` and `ProsemirrorModelField`. Not sure if this is due to a mismatch between test environment and the real world.

src/open_inwoner/components/templates/components/Faq/Faq.html
```diff
@@ -14,7 +14,7 @@ <h3 class="utrecht-heading-3">
                     {% firstof 'question-'|add:id|add:'-answer' as answer_id %}
                     {% link bold=True href='#'|add:answer_id icon='keyboard_arrow_down' secondary=True text=question.question toggle="open"  %}
                 </h3>
-                <div id="{{ answer_id }}" class="faq__answer">{{ question.answer|ckeditor_content|safe }}</div>
+                <div id="{{ answer_id }}" class="faq__answer">{{ question.answer|safe }}</div>
```
swrichards: No `.html`?

src/open_inwoner/pdc/models/faq.py
```diff
@@ -24,7 +26,18 @@ class Question(OrderedModel):
         null=True,
     )
     question = models.CharField(verbose_name=_("Vraag"), max_length=250)
-    answer = models.TextField(verbose_name=_("Antwoord"))
+    answer = ProsemirrorModelField(
+        _("Antwoord"),
+        allowed_node_types=[NodeType.PARAGRAPH],
```
swrichards: It seems this field sometimes uses blockquote:

Image

```suggestion
        _("Antwoord"),
        allowed_node_types=[NodeType.PARAGRAPH, NodeType.BLOCKQUOTE],
```

src/open_inwoner/cms/plugins/models/text.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Text(CMSPlugin):
+    body = ProsemirrorModelField(
+        _("Body"),
+        help_text=_("The text content"),
+    )
+
+    def __str__(self):
+        return str(self.body)[:50] if self.body else ""
```
swrichards: Guess we might as well do `self.body.html` here, otherwise it's just a bunch of JSON.

src/open_inwoner/pdc/utils.py
```diff
@@ -6,19 +6,27 @@
...
-    md = markdown.Markdown()
-    html_string = md.convert(content)
+    # Check if content is a string
+    if isinstance(content, str):
+        # Legacy markdown conversion for old TextField content
+        md = markdown.Markdown()
+        html_string = md.convert(content)
+    else:
+        # ProsemirrorModelField - use .html property
+        html_string = content.html
```
swrichards, line 23: Just want to double-check we have no better way of doing check than looking for the `.html` property (I know you tried, no worries if not, just jumped out at me again).

src/open_inwoner/cms/footer/migrations/0003_rename_content_cmsflatpagemodel_content_tmp.py
```diff
@@ -0,0 +1,101 @@
...
+        # Skip empty or whitespace-only content
+        if not content or not content.strip():
+            flatpage.content = None
+            flatpage.save()
+            continue
+
+        # If content doesn't look like HTML, wrap it in paragraphs
+        content = content.strip()
+        if not content.startswith("<"):
+            content = f"<p>{content}</p>"
```
swrichards: What happens if we don't do this?

src/open_inwoner/configurations/migrations/0086_alter_siteconfiguration_warning_banner_text.py
swrichards, line 1: Two general comments:

1. Could we extract the conversion to a `migration_utils` helper, and re-use it? (Injecting the dependencies for the specific field)
2. I would still be inclined to merge these migrations if we can: I'd prefer they either all succeed or all fail, given how little we know about the state of the targets where this will be running.

swrichards (approved):
Great job on this, couldn't really find any major issues except for the missing blockquote, and one or two questions/nitpicks, but otherwise it looks good to me.

src/open_inwoner/cms/footer/migrations/0003_rename_content_cmsflatpagemodel_content_tmp.py
pi-sigma: Nothing bad should happen, `html_to_doc` handles plain text correctly. This was defensive programming and should help to make the intent clear: make sure we're feeding HTML to django-prosemirror's parser. It's also perhaps a bit faster if the parser doesn't have to infer the structure. I could remove it though.

src/open_inwoner/pdc/utils.py
pi-sigma: Did you mean some other part of the code? Here I'm using the `.html` property to assign the content, not to check for something.

swrichards: Sorry, I mean checking if `content` is an instance of a prosemirror field. I was wondering whether we could check that directly, rather than inferring it from the presence of the `.html` attribute.

src/open_inwoner/configurations/migrations/0086_alter_siteconfiguration_warning_banner_text.py
pi-sigma: ~~Re 2: I found the issue. The migrations created automatically by Django have `history=True` for the addition of `ProsemirrorModelField`, which creates database triggers to track changes to the field. Postgres thenblocks the removal of the field in the subsequent migration because the triggers from the first field haven't been committed. The options are to do `atomic=False` for the addition of `ProsemirrorModelField`. which defeats the purpose of grouping them together, or `history=False`. I think we don't need `history=True`, which saves historical copies of the field; it's a singleton model used for configuration and we audit the change ourselves in the review. What do you think?~~

Disregard, the explanation is wrong and the change only helped accidentally

src/open_inwoner/pdc/migrations/0069_category_description_data.py
swrichards, line 1: Why can't this file use the helper?

src/open_inwoner/pdc/tests/test_logging.py
```diff
@@ -27,54 +27,78 @@ def setUp(self):
...
     def test_change(self):
-        self.product.save()
-        self.product.categories.add(self.category.id)
-        response = self.app.get(
-            reverse("admin:pdc_product_change", kwargs={"object_id": self.product.id}),
-            user=self.user,
+        # Create product and change it to generate a log entry
+        from django.contrib.admin.models import CHANGE, LogEntry
+        from django.contrib.contenttypes.models import ContentType
+
```
swrichards: Why the inline imports?

src/open_inwoner/pdc/migrations/0069_category_description_data.py
pi-sigma: Oversight or lost in rebase; fixed

src/open_inwoner/pdc/tests/test_logging.py
pi-sigma: Forgot to move them after experimenting

src/open_inwoner/utils/migration_operations.py
```diff
@@ -1,59 +1,337 @@
...
+        str: HTML with empty tags removed
+    """
+    # Replace self-closing tags with proper closing tags
+    # <tag /> -> <tag></tag>
+    html = re.sub(r"<(\w+)\s*/>", r"<\1></\1>", html)
+
+    # Remove empty tags iteratively (nested empty tags need multiple passes)
+    # Matches: <tag></tag> or <tag> </tag> or <tag attrs></tag>
+    prev = None
+    max_iterations = 10  # Safety limit
```
swrichards: What kind of limits do we run up against? If we do reach our limit, if I understand this correctly, we'll still have to deal with the `IndexError`s later. We're not recursing, it might just take a long time. If the fear is infinite loop, I would set the limit much higher so we do eventually time out if we reach an infinite edge case, but will hopefully cover all practical scenarios.

swrichards (approved):
Good work here, turns out (as ever) there were more edge cases than we realized. I think we've done all we reasonable can to smooth the migration.

Just a few questions/nitpicks, can do a quick follow up review and merge before tomorrow morning's meeting.

src/open_inwoner/cms/footer/migrations/0003_rename_content_cmsflatpagemodel_content_tmp.py
swrichards: It's fine, defensive is good here.

src/open_inwoner/utils/migration_operations.py
pi-sigma: Meant as a defense against regex edge cases or malformed HTML that would get the function stuck in a loop. Shouldn't happen but well... I've set it to 50. Seems enough? Real-world HTML shouldn't contain that many levels of tag nesting (though who knows).

src/open_inwoner/utils/migration_operations.py
```diff
@@ -1,59 +1,339 @@
...
+        # After cleaning, double-check again if anything is left
+        if not content or not content.strip():
+            setattr(instance, field_name, None)
+            instance.save()
+            continue
+
+        try:
+            # Convert markdown to HTML if needed
+            if convert_markdown:
+                content = markdown.markdown(content, extensions=["extra"])
```
swrichards: Final thought on this: markdown and HTML are mutually exclusive right? And if so, is there a chance the earlier cleaning operations will generate errors (because they're working on markdown, not HTML), or does thew built-in HTML parser silently ignore everything in that case?

swrichards (approved):
Epic PR, I think we're there. Just one final question, probably not an issue but just wanted to make sure.

src/open_inwoner/utils/migration_operations.py
pi-sigma: Good catch! They're not exclusive (markdown can contain inline HTML) and the cleaning operations work on HTML not markdown, but there could be bugs: e.g. if the cleaning operations modify/strip inline HTML and the markdown conversion is feeded corrupted markdown. I'll fix it, thanks.

swrichards (approved):

PR #1953, opened 2025-10-06
Add more opentelemetry metrics

swrichards:
## Summary

In addition to adding the metrics, this PR involves moving all logging calls into semantically named logging mixins, so we have a clear central hook for logs and metrics (they're a similar concern), and so that the business methods don't have to worry about implementation details.

## Issue References

- Taiga User Story: [#3468](https://taiga.maykinmedia.nl/project/open-inwoner/us/3468)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/accounts/views/mixins.py
+++ b/src/open_inwoner/accounts/views/mixins.py
@@ -1,5 +1,3 @@
-from functools import partial
-
 from django.http import HttpRequest
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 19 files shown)


src/open_inwoner/accounts/views/mixins.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class ContactmomentLogMixin(LogMixin):
+    request: HttpRequest
+
+    def log_contactmoment_registered_by_email(self, success: bool):
+        """
+        Log registering a contactmoment by email
+        """
+        log = partial(self.log_system_action, user=self.request.user)
```
pi-sigma: Use of `partial` is a bit overkill for this case, no? Same below.

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -14,24 +22,98 @@
...
+        )
+        case_document_downloads.add(1)
+
+    def log_case_document_uploaded(self, case: UniformCase, filename: str):
+        """
+        Log uploading a document to a case
+        """
+        self.log_user_action(
+            self.request.user,
+            _("Document was uploaded for {case}: {filename}").format(
```
pi-sigma, line 55: English here, Dutch above.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -830,7 +824,7 @@ def handle_no_permission(self):
         raise PermissionDenied()
 
 
-class CaseDocumentUploadFormView(CaseAccessMixin, LogMixin, FormView):
+class CaseDocumentUploadFormView(CaseAccessMixin, CaseLogMixin, FormView):
```
pi-sigma, line 827:
```suggestion
class CaseDocumentUploadFormView(CaseLogMixin, CaseAccessMixin, FormView):
```
or change above

pi-sigma (changes requested):
Couple minor points.

Re our discussion earlier: I wonder if the signal approach is not better. The pattern would be almost identical with only minor complexity added (need need extra modules for signals/receivers). E.g.

```python
# Signal
contactmoment_registered = Signal(providing_args=["user", "contactmoment_success", "objectcontactmoment_success"])

# in the view
...business logic
...
contactmoment_registered.send(
            sender=self.__class__,
            user=user,
            contactmoment_success=contactmoment_success,
            objectcontactmoment_success=objectcontactmoment_success,
        )

# receiver
@receiver(contactmoment_registered)
def log_contactmoment_registration(sender, user, contactmoment_success, objectcontactmoment_success, **kwargs):
    success = True

    def log_action(message):
        # Assuming a system action log method or just logger.info here:
        logger.info(f"User {user} - {message}")

    if contactmoment_success:
        log_action("registered contactmoment by API")
    else:
        success = False
        log_action("error while registering contactmoment by API")

    if objectcontactmoment_success is not None:
        if objectcontactmoment_success:
            log_action("registered objectcontactmoment by API")
        else:
            success = False
            log_action("error linking objectcontactmoment to contactmoment by API")

    # Update whatever metrics tracking system is used (example)
    case_contact_form_registrations.add(1, {"channel": "api", "success": success})
```

Decoupling would simplify our logging tests; they're currently intertwined with tests for the business logic which is often annoying.

However, we could also run with the mixin approach

src/open_inwoner/cms/cases/views/mixins.py
swrichards: True, but for now my goal was not change to the logging, just to move it, so I'll leave what we had for now.

swrichards:
@pi-sigma sticking with the mixin for now, given the lack of time. I also thought an alternative might be to just make the loggers separate classes (`AuditLogger`) and instantiating it, as a middle-ground. But that'll have to wait until we have some space to fix this.

pi-sigma (approved):

PR #1954, opened 2025-10-06
fix: use paginator object in contactmomenten list

swrichards:
## Summary

The contactmomenten list template did not properly use the paginator object to control view, pagination and conditional logic. This commit fixes all references to the raw questions list to use the paginator object.  

## Issue References

- Taiga Issue: [#3493](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3493)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/templates/pages/contactmoment/list.html
+++ b/src/open_inwoner/templates/pages/contactmoment/list.html
@@ -16,7 +16,7 @@
 {% endblock %}
 
 {% block content %}
-    <h1 class="utrecht-heading-1" id="contactmomenten">{{ page_title }} ({{ questions|length }})</h1>
+    <h1 class="utrecht-heading-1" id="contactmomenten">{{ page_title }} ({{ paginator.count }})</h1>
 
     {% if contactmoment_contact_form_enabled %}
     {# title + scroll down button #}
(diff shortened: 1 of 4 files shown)


src/open_inwoner/templates/pages/contactmoment/list.html
```diff
@@ -81,9 +81,7 @@ <h2 class="card__heading-2">{{ question.subject }}</h2>
         {% endrender_grid %}
     </div>
 
-    {% if questions %}
-        {% pagination page_obj=page_obj paginator=paginator request=request %}
-    {% endif %}
+    {% pagination page_obj=page_obj paginator=paginator request=request %}
```
pi-sigma, line 84: The pagination will always be shown even with 0 items. Could be wrapped:
```
{% if paginator.count %}
    {% pagination page_obj=page_obj paginator=paginator request=request %}
{% endif %}
```

pi-sigma (approved):

src/open_inwoner/templates/pages/contactmoment/list.html
swrichards: I was half-minded to make this a feature and let the pagination tag handle this, which I think would be the proper place rather than having each list view itself worry about this edge case. But perhaps it leads to ugly results right now?

swrichards: Going with this for now, as I think this edge case should indeed be handled in the template tag if we care about hiding it with zero items. The tag has all the state it needs to make this determination.

PR #1956, opened 2025-10-07
fix: ensure openklant2 question subject always has a value for zaak q…

swrichards:
We defaulted to using zaak.omschrijving for the question subject in our openklant2 question-for-zaken method, but that can be an empty string, which triggers a Bad Request error.

## Issue References

- Taiga Issue: [#3495](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3495)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1557,11 +1557,12 @@ class OpenKlant2Service(
         self,
         partij_uuid: str,
         question: str,
-        subject: str,
         zaak: Zaak,
     ) -> OpenKlant2Question:
         ok2_question = self.create_question_for_partij(
-            partij_uuid=partij_uuid, question=question, subject=subject
+            partij_uuid=partij_uuid,
+            question=question,
+            subject=f"Vraag voor zaak: {zaak.identificatie}",
         )
 
         self.log_system_action(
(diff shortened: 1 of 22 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1557,11 +1557,12 @@ def create_question_for_zaak(
...
         partij_uuid: str,
         question: str,
-        subject: str,
         zaak: Zaak,
     ) -> OpenKlant2Question:
         ok2_question = self.create_question_for_partij(
-            partij_uuid=partij_uuid, question=question, subject=subject
+            partij_uuid=partij_uuid,
+            question=question,
+            subject=f"Vraag voor zaak: {zaak.identificatie}",
```
pi-sigma, line 1565:
```suggestion
            subject=_("Question for case: %(identificatie)s") % {"identificatie": zaak.identificatie},
```
or:
```suggestion
            subject=_("Vraag voor zaak: %(identificatie)s") % {"identificatie": zaak.identificatie},
```

pi-sigma (approved):

src/open_inwoner/openklant/services.py
swrichards: I don't think it makes sense to translate this. It's not a UI text, but something that will end up in another repository. Using the target working language directly makes more sense to me.

PR #1957, opened 2025-10-07
feat: have klanten configuration models mirror admin structure

swrichards:
## Summary

The `django-setup-configuration` models for the klantensysteem had gotten a bit stale, with unclear nesting. I've refactored this to match the current admin structure, with a single configuration model at the root, and optional sub-models for esuite and openklant.

## Issue References

- Taiga Issue: [#3496](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3496)

## Checklist

- [x] CHANGELOG.rst updated
  - [x] Deployment notes added 
- [x] Updated the `docker-compose.yml` environment variables
    variables


src/open_inwoner/configurations/bootstrap/openklant.py
```diff
@@ -153,28 +172,3 @@ def execute(self, model: OpenKlant2Configuration):
 
         config.full_clean()
```
pi-sigma: The `_configure_esuite` method wraps this in a `try/except` and raises `ConfigurationRunFailed` if a `ValidationError` is excepted. Why not here?

src/open_inwoner/configurations/bootstrap/openklant.py
```diff
@@ -65,18 +67,46 @@ class Meta:
...
+    config_model = KlantenSysteemConfigurationModel
+
+    def execute(self, model: KlantenSysteemConfigurationModel):
+        # Configure eSuite if provided
+        if model.esuite_config:
+            self._configure_esuite(model.esuite_config)
+
+        # Configure OpenKlant2 if provided
+        if model.openklant2_config:
+            self._configure_openklant2(model.openklant2_config)
```
pi-sigma, line 90: I note that if `_configure_esuite` succeeds but `_configure_openklant2` fails, the esuite config is saved but the klantensysteem config is not. Is that the behavior we want? Otherwise consider wrapping in a transaction.

pi-sigma (approved):

src/open_inwoner/configurations/bootstrap/openklant.py
swrichards: Blind copy pasta, will make this consistent, thanks.

src/open_inwoner/configurations/bootstrap/openklant.py
swrichards: The library itself wraps the whole step in a transaction, so we don't have to do that explicitly.

PR #1960, opened 2025-10-10
Implement design header + mobile overlay

stefrado:
## Summary

Implement design header + mobile overlay

## Issue References
https://taiga.maykinmedia.nl/project/open-inwoner/task/3502

- Taiga User Story:
https://taiga.maykinmedia.nl/project/open-inwoner/us/3492

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/scss/components/Header/Header.scss
+++ b/src/open_inwoner/scss/components/Header/Header.scss
@@ -5,17 +5,24 @@ $hm: max(calc((100vw - var(--container-width)) / 2), var(--spacing-large));
 $vm: var(--spacing-large);
 
 .header {
+  --header-height: 53px;
+
   position: relative;
-  min-height: 81px;
+  min-height: var(--header-height);
   background-color: var(--color-white);
   width: 100%;
   z-index: 1003;
+  margin-bottom: var(--spacing-large);
 
   @media (min-width: 768px) {
+    --header-height: 63px;
     display: flex;
     align-items: center;
-    min-height: 90px;
     margin-bottom: var(--spacing-tiny);
+  }
+
+  @media (min-width: 911px) {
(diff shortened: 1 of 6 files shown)


swrichards (approved):

src/open_inwoner/scss/components/Header/Header.scss
```diff
@@ -5,19 +5,26 @@ $hm: max(calc((100vw - var(--container-width)) / 2), var(--spacing-large));
 $vm: var(--spacing-large);
 
 .header {
+  --header-height: 53px;
```
jiromaykin, line 8: We don't usually work with local variables like this in the project, but I understand there's an unhappy value needed here.

But: it's not working properly for mobile (should be way less high).

Design is this:

Image

But looks like this now:

Image

jiromaykin (changes requested):
We do not yet have a design for what the banner images should look like on all the other pages, so we only have a design now for the 'overlay' indent for the welcome page.

src/open_inwoner/scss/components/Header/Header.scss
jiromaykin: issue: if you are logged in - on mobile - and you navigate to http://localhost:8000/accounts/register/ then you will see the notifications pushing things down (only on the Home page) so that's a problem.

You could look at this old PR to remove/adjust that code: https://github.com/maykinmedia/open-inwoner/pull/1491/files

which also added a bunch of exceptions in the _views.SCSS file because this was the only page with a different notifications style.

jiromaykin (approved):
Good to see we have task https://taiga.maykinmedia.nl/project/open-inwoner/issue/3521 now so this can go through.

PR #1961, opened 2025-10-10
feat: add a basic healthcheck endpoint

swrichards:
## Summary

Triggered by https://github.com/maykinmedia/charts/pull/303 -- we might as well have an explicit health check endpoint which is simpler, quicker and less ambiguous than the admin page (which involves template interpolation, redirects, etc.).

## Checklist

- [x] CHANGELOG.rst updated


swrichards:
Made redundant by health checks in maykin common

PR #1963, opened 2025-10-14
:art: [#3507] CHANGE active/hover color for the side nav

stefrado:
## Summary

Als gebruiker wil ik dat de hover en active state kleuren van het sidenav in de huisstijl zijn die gekozen in door de gemeente.

## Notes

I changed it inside this repo, because this does not work: `Reference doesn't exist`

```json
"hover": {
  "color": {"value": "{color.primary}"}
},
"active": {
  "color": {"value": "{color.primary}"},
  "font-weight": {"value": "bold"}
}
```

## Issue References

[Taiga User Story](https://taiga.maykinmedia.nl/project/open-inwoner/us/3507?milestone=550)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/scss/components/SideNavigation/SideNavigation.scss
+++ b/src/open_inwoner/scss/components/SideNavigation/SideNavigation.scss
@@ -2,6 +2,8 @@
 
 :root {
   --denhaag-side-navigation-mobile-display: flex !important;
+  --denhaag-side-navigation-link-active-color: var(--color-primary);
+  --denhaag-side-navigation-link-hover-color: var(--color-primary);
 }
 
 .side-navigation--oip {
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -84,6 +84,8 @@ Bugfixes
   Prosemirror-velden niet waren gekopieerd.
 * [:taiga-is:`3519`: :pr:`1966`]: Probleem opgelost in de side menu waarbij de
   'mijn vragen' item niet geselecteerd wordt als huidige pagina.
+* [:taiga-is:`3507`: :pr:`1963`]: Verander kleur van menu items bij actieve en hover status
+  naar de correcte primaire kleur.
 
 Onderhoud
 ---------


stefrado:
@swrichards not sure where to put this inside the `CHANGELOG`

swrichards:
> @swrichards not sure where to put this inside the `CHANGELOG`

@stefrado Als het een fix is op een reeds bestaande toevoeging in de running changelog kan je het wat mij betreft zo laten.

Anders: ik begreep van het gesprek vanochtend dat dit eigenlijk een bugfix is? Dan zou ik het daarbij zetten, met een nieuwe issue waar expliciet in staat dat dit een fix is (de PR verwijst naar een user story, in dat geval zou het een feature zijn, maar de teneur klonk toch meer naar "bugfix" vanochtend.)

src/open_inwoner/scss/components/SideNavigation/SideNavigation.scss
```diff
@@ -2,6 +2,8 @@
 
 :root {
   --denhaag-side-navigation-mobile-display: flex !important;
+  --denhaag-side-navigation-link-active-color: var(--color-primary);
```
jiromaykin, line 5: But you actually do not need to do anything at all. 
Did we check if Dimpact has uploaded CSS in the admin without knowing?

It's already working. I know it looks weird that values are hard-coded in designtokens but that will always happen: https://github.com/maykinmedia/open-inwoner-design-tokens/blob/main/src/community/denhaag/side-navigation.tokens.json

CSS values will be hardcoded in all design-tokens bij any municipality, and are indeed unrelated to our App.SCSS styles, but in OIP we are eventually overwriting them, in the Master templates and it turns into 

```
--color-primary: hsl(
    var(--color-primary-h),
    var(--color-primary-s),
    var(--color-primary-l)
  );
```

So because of the master template, it's already functional.

BUT municipalities CAN overwrite these styles IF they use the CSS upload function in our Admin!

So if you upload a CSS file that contains this, then you will have the last final resulting style - maybe that is some that Dimpact did without realising it?

```
.openinwoner-theme {
  --denhaag-side-navigation-link-active-color: orange;
  --denhaag-side-navigation-link-hover-color: aqua;
}

```

stefrado: This is needed to make it work with the color settings under 'Algemene configuratie'.

Scherm­afbeelding 2025-10-14 om 14 17 12

Currently the only way to use the primary/secondary color is to write custom css and add a piece of code similar as what you wrote, but this is not what is requested.

> ALs gebruiker wil ik dat de hover en current state van het zijmenu in de huisstijl zijn die gekozen in sdoor de gemeente.
 
Looking at the the value has to be the same as  `--color-primary` not `--oip-color-primary` (which could only be overwritten using the custom CSS).

Besides this the `--oip-color-primary` is not used at all in our project - only in some NLDS components (which are nowhere used in our project).

jiromaykin: Okay, yes. Making the color-picker work "with" the tokens is a different issue perhaps than this story. 

The actual solution would be to make the primary color in OIP App.scss into something like this (with existing tokens that should be overwritten by Master template):

```
  --color-primary-h: var(--oip-color-primary-h);
  --color-primary-s: var(--oip-color-primary-s);
  --color-primary-l: var(--oip-color-primary-l);
```

I can make a new issue. And we probably need a better kind of developer documentation so next time it's easier to know what to do.

stefrado: As i said the `--oip-color-primary` is currently not in use. If we are going to use this color this is not an issue, but a complete user story and i doubt that this is what is required. So if we are going to change this, we'll need to remove the `Kleur` section from the admin and make a lot of changes.

jiromaykin (approved):
Approved but i will need to connect separated HSL colors to the Design-token package and then connect them in OIP App.scss; because we DO want to use those colors (in future municipalities will use JSON tokens to update colors).

PR #1965, opened 2025-10-14
[#345] Add error message for phone number DB constraint

pi-sigma:
## Summary

Also removed the form-layer validation and error messages. We have different login/registration flows and different flows; it is cumbersome and error-prone to add the validation to some but not others. The new approach is also in line with other constraint checks for users, which are implemented at DB layer.

## Issue References

Taiga Dimpact Issue: https://taiga.maykinmedia.nl/project/dimpact-enschede-ssc-ict-1/issue/345

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/accounts/models.py
```diff
@@ -359,6 +359,9 @@ class Meta:
             models.CheckConstraint(
                 check=~Q(phonenumber__exact="") | Q(phonenumber_alternative__exact=""),
                 name="phonenumber_alt_requires_phonenumber_primary",
+                violation_error_message=_(
+                    "Primary phone number is required for setting alternative phone number"
```
swrichards: If we keep the verbatim error message we shouldn't need a new translation:


```suggestion
                    "A primary phone number is required for setting an alternative phone number"
```

swrichards (approved):
One nitpick but otherwise LGTM, good cleanup.

src/open_inwoner/accounts/migrations/0091_remove_user_phonenumber_alt_requires_phonenumber_primary_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+        migrations.AddConstraint(
+            model_name="user",
+            constraint=models.CheckConstraint(
+                check=models.Q(
+                    models.Q(("phonenumber__exact", ""), _negated=True),
+                    ("phonenumber_alternative__exact", ""),
+                    _connector="OR",
+                ),
+                name="phonenumber_alt_requires_phonenumber_primary",
+                violation_error_message="Primary phone number is required for setting alternative phone number",
```
swrichards: Think you'll have to regenerate the migration

PR #1968, opened 2025-10-15
:bug: [#3519] fix active side nav item determination

swrichards:
## Summary

The active menu item check failed for Mijn Vragen, because we hadn't properly implemented the redirect-url use case. That prompted a general refinement of this logic to make the check more thorough and less depended on hard-coding edge cases (such as Mijn Vragen).

## Issue References

- Taiga Issue: [#3519](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3519)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/components/templatetags/side_navigation.py
+++ b/src/open_inwoner/components/templatetags/side_navigation.py
@@ -1,9 +1,10 @@
 import logging
-from typing import List, TypedDict
+from typing import List, TypedDict, cast
 
 from django import template
 from django.core.exceptions import ImproperlyConfigured
-from django.urls import NoReverseMatch, resolve, reverse
+from django.http import HttpRequest
+from django.urls import NoReverseMatch, Resolver404, ResolverMatch, resolve, reverse
 from django.utils.translation import get_language, gettext_lazy as _
 
 from cms.models import Page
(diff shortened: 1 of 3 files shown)


src/open_inwoner/components/templatetags/side_navigation.py
```diff
@@ -142,23 +145,106 @@ def _extract_counter(self, node: NavigationNode) -> int | None:
...
+            node_qualified_name = self._get_qualified_url_name(resolved_node_path)
+            logger.debug(
+                "Node '%s' qualified URL name: '%s'", node, node_qualified_name
+            )
+
+            is_match = current_qualified_name == node_qualified_name
+
+            # For CMS pages (and other catch-all patterns), URL names match but paths differ.
+            # When URL names match, also verify the actual paths match.
+            if is_match and current_qualified_name == node_qualified_name:
```
pi-sigma:
```suggestion
            if is_match:
```

second conjunct is redundant (l. 209: `is_match = current_qualified_name == node_qualified_name`)

pi-sigma (changes requested):

src/open_inwoner/components/templatetags/side_navigation.py
swrichards: Bad rebase, thanks!

pi-sigma (approved):

PR #1969, opened 2025-10-15
fix: cull cases from list that cannot be fully resolved

swrichards:
## Summary

We're not sufficiently defensive in our case resolutions. For a misconfigured zaken systeem (as often happens with new vendors), a failure to properly resolve part of the case leads to downstream errors. This is a defensive fix to ensure we just skip cases we can't properly resolve. Eventually, though, we'll want to make this entire flow more resilient. 

## Issue References

- Taiga Issue: [#3522](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3522)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/cms/cases/views/services.py
+++ b/src/open_inwoner/cms/cases/views/services.py
@@ -2,6 +2,7 @@ import concurrent.futures
 import enum
 import logging
 import threading
+from collections import defaultdict
 from dataclasses import dataclass
 from typing import Iterable, TypedDict, cast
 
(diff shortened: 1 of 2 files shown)


src/open_inwoner/cms/cases/views/services.py
```diff
@@ -27,6 +28,12 @@
 logger = logging.getLogger(__name__)
 
 
+class ResolveCaseException(Exception):
+    """Exception indicatin we were unable to resolve a cases type/status/result."""
```
pi-sigma:
```suggestion
    """Exception indicating we were unable to resolve a cases type/status/result."""
```

pi-sigma (approved):

PR #1978, opened 2025-10-20
fix: truncated dropdown should not determine items based on CMS position

swrichards:
## Summary

Previously, when the dropdown menu was truncated, we would assume "My Profile" was the first item in the CMS ordering, and we used a CSS selector to only show the first item. This turned out to be fragile: the order of CMS pages, as so many things in Django CMs, is non-deterministic and we have no guarantee that the profile page is the first. I've moved this logic into a separate template tag to handle this logic explicitly in Python, and consolidated the tags a little to be more coherent after the recent sidenav menu changes.

## Issue References

- Taiga Issue: [#3525](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3525)

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/scss/components/Header/PrimaryNavigation.scss
```diff
@@ -153,14 +153,13 @@
...
       justify-content: flex-end;
-      /* Wide screens: show only first item */
 
       body:has(main.container.container--fullwidth):has(
           .denhaag-side-navigation
         )
         & {
         @media (min-width: 1024px) {
-          &:not(:first-child, &--always-visible),
+          &:not(&--always-visible),
```
swrichards: @stefrado @jiromaykin heads up, I've moved this logic (mostly) out of CSS.

pi-sigma: This makes the dropdown disappear completely

src/open_inwoner/components/templatetags/menu.py
```diff
@@ -420,3 +420,90 @@ def has_sidenav_items(context):
...
+        True if the item should be shown, False otherwise
+    """
+    # If the full dropdown menu should be shown, all items are visible
+    if show_full_dropdown_menu(context):
+        return True
+
+    # When side nav is active, only show "My Profile"
+    try:
+        profile_detail_url = reverse("profile:detail")
+        return profile_detail_url in menu_item_url
```
pi-sigma: Possible false negative due to trailing slashes:

`profile_detail_url == "/mijn-profiel/`, `menu_item_url == /mijn-profiel`

use `rstrip('/')` on both items? Don't know how plausible this is, but I've stumbled over it a few times.

pi-sigma (changes requested):

src/open_inwoner/components/templatetags/menu.py
swrichards: Good point, went for the more comprehensive solution, which involves resolving the URLs and comparing the url names directly.

src/open_inwoner/components/templatetags/menu.py
```diff
@@ -420,3 +420,104 @@ def has_sidenav_items(context):
...
+        else current_url_name
+    )
+
+    # The following URLs are expected to have the sidenav, so they only need a minimal
+    # dropdown menu.
+    urls_with_minimal_dropdown_menu = {
+        "pages-root",
+        "general_faq",
+        "collaborate:plan_list",
+        "ssd:uitkeringen",
```
stefrado, line 466: Found two missing minimal dropdown pages.

```python
"ssd:yearly_benefits_index",
"ssd:monthly_benefits_index",
```

If one of these pages is the current page, i think "ssd:uitkeringen" (mijn uitkeringen) should stay active in the sidemenu. I do not think that we currently haven't build this logic.

We could add a custom check inside `_is_current_page` that looks for the current route and the namespace. If the current route is part of this namespace, the item. 

Example: 
```python
active_item_if_qualified_name = {
  "ssd:uitkeringen": "ssd:yearly_benefits_index",
  "ssd:uitkeringen": "ssd:monthly_benefits_index",
} 


special_route = active_item_if_qualified_name[node_qualified_name]

if special_route and current_qualified_name == special_route:
  current = True
```

> The example is not final and should be updated to handle this special page case.

stefrado (changes requested):

src/open_inwoner/components/templatetags/menu.py
swrichards: Good point, _but_: the lines you're commenting on are not new, they were being moved from another module. Could you dump this in a separate Taiga issue and tag me? I'll do it in a followup.

src/open_inwoner/components/templatetags/menu.py
```diff
@@ -420,3 +420,104 @@ def has_sidenav_items(context):
...
+    if show_full_dropdown_menu(context):
+        return True
+
+    # When side nav is active, only show these items
+    allowed_url_names_in_minimal_dropdown = {
+        "profile:detail",
+    }
+
+    try:
+        resolved_menu_item = resolve(menu_item_url)
```
pi-sigma: Maybe far-fetched, but: `resolve` works on paths, null full URL's. The `redirect_url` (where `menu_item_url` comes from) is configurable by admins. If they enter a fill URL for *any* CMS page, this page will raise a `Resolver404` and will be displayed as result. Fix: parse the URL and use only the path:

```
from urllib.parse import urlparse

parsed = urlparse(menu_item_url)
# Use only the path component for resolve()
resolved_menu_item = resolve(parsed.path)
```

pi-sigma (approved):
One possible issue. Approving anyway because it might be far-fetched.

src/open_inwoner/components/templatetags/menu.py
swrichards: Good idea!

src/open_inwoner/scss/components/Header/PrimaryNavigation.scss
swrichards: Fixed with help from @stefrado

src/open_inwoner/components/templatetags/menu.py
swrichards: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3530

PR #1979, opened 2025-10-20
[#3508] Use structlog for logging

pi-sigma:
## Summary

Refactored logging to use structlog instead of buil-in logging

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3508

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added 
- [ ] Documentation updated 
    variables


src/open_inwoner/cms/cases/views/cases.py
```diff
@@ -91,9 +91,9 @@ def get_context_data(self, **kwargs):
                     statuses.append(CaseFilterFormOption(status))
                 except ValueError:
                     logger.error(
-                        "Invalid data (%s) for case filtering by %s",
-                        self.request.GET,
-                        self.request.user,
+                        "Invalid data for case filtering by",
+                        data=str(self.request.GET),
```
swrichards: As a general pattern, I don't think we should be doing this: we want the params to be actual Python objects, and the formatters will take care of invoking `__str__`

pi-sigma: structlog passes arguments to logger calls through `JSONRenderer`, which expects JSON-serializable objects, so this doesn't work. Alternatively, we could hook into the `json.dumps` flow from `JSONRenderer` and define a `default_serializer`  which simply returns `repr(obj)`. That way we can continue to pass arbitrary objects.

src/open_inwoner/accounts/tests/test_auth.py
```diff
@@ -2031,11 +2031,13 @@ def test_update_klant_via_openklant_partij_creation_fails(
 
         # Assertions
         self.assertEqual(response.status_code, 302)
+
+        user.refresh_from_db()
```
swrichards: Curious, why was this needed?

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -152,13 +152,14 @@ def get_context_data(self, **kwargs):
                     # TODO: This can happen. Ideally, we would present the user with
                     # warning noting that not all questions might be visible.
                     logger.warning(
-                        "Connection error for service %s",
-                        service_type,
+                        "Connection error for service",
+                        service_type=str(service_type),
```
swrichards: I think we could use `service_type.value` here

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -152,13 +152,14 @@ def get_context_data(self, **kwargs):
...
                         exc_info=True,
                     )
                 except BaseException:
                     logger.exception(
-                        "Unable to fetch questions for service %s", service_type
+                        "Unable to fetch questions for service",
+                        service_type=str(service_type),
```
swrichards: As above, possibly use `.value`

src/open_inwoner/accounts/signals.py
```diff
@@ -144,7 +144,8 @@ def _update_eherkenning_user_from_kvk_api(user: User):
         else:
             logger.error(
```
```suggestion
            logger.exception(
```

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -233,13 +233,14 @@ def get_context_data(self, **kwargs):
...
-                            service_type,
+                            "Connection error for service",
+                            service_type=str(service_type),
                             exc_info=True,
                         )
                     except BaseException:
                         logger.exception(
-                            "Unable to fetch questions for service %s", service_type
+                            "Unable to fetch questions for service",
+                            service_type=str(service_type),
```
swrichards: We can use `.value` for the enums.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -515,9 +515,9 @@ def is_file_upload_enabled_for_case_type(self) -> bool:
             ).exists()
         )
         logger.info(
-            "Case {url} has case type file upload: {case_upload_enabled}".format(
-                url=self.case.url, case_upload_enabled=case_upload_enabled
-            )
+            "File upload status for case (enabled/disabled)",
+            case=str(self.case),
```
swrichards: Could we replace these (and below) with a reference to `self.case.identificatie`, or does it have a good `__str__` implemented that only shows the salient details?

src/open_inwoner/cms/tests/cms_tools.py
```diff
@@ -115,36 +115,38 @@ def render_all_placeholders(
...
-                "rendered content: %s",
-                rendered_content
-                if len(rendered_content) < 128
-                else rendered_content[:128] + "...",
+                "rendered content",
+                content=str(
+                    rendered_content
+                    if len(rendered_content) < 128
+                    else rendered_content[:128] + "..."
+                ),
```
swrichards: You could use `from django.template.defaultfilters import truncatechars`

src/open_inwoner/components/templatetags/render_tags.py
```diff
@@ -26,5 +26,5 @@ def prosemirror_content(content):
     try:
         return get_rendered_content(content)
     except (AttributeError, TypeError) as e:
-        logger.warning("Could not render content: %s", e)
+        logger.warning("Could not render content", error=str(e))
```
swrichards: I assume we can still use `exc_info=True`:
```suggestion
        logger.warning("Could not render content", exc_info=True)
```

src/open_inwoner/conf/base.py
```diff
@@ -493,6 +523,29 @@
...
+    ],
+    logger_factory=structlog.stdlib.LoggerFactory(),
+    cache_logger_on_first_use=True,
+)
+
+#
+# DJANGO-STRUCTLOG
+#
+DJANGO_STRUCTLOG_IP_LOGGING_ENABLED = False
+DJANGO_STRUCTLOG_CELERY_ENABLED = True
```
swrichards, line 546: I think some extra love will be needed for Celery: https://github.com/open-formulieren/open-forms/blob/644413b91d391691cb0c5368aa9d4fe062664312/src/openforms/celery.py#L60-L130

src/open_inwoner/haalcentraal/api.py
```diff
@@ -85,8 +85,8 @@ def fetch_data(self, user_bsn: str) -> dict | None:
                 verify=False,
             )
             return get_json_response(response)
-        except (RequestException, ClientError) as e:
-            logger.exception("exception while making request", exc_info=e)
+        except (RequestException, ClientError):
+            logger.exception("exception while making request")
```
swrichards: Let's add a note to the string noting which service is involved, e.g. `unexpected error connecting to Haal Centraal`.

src/open_inwoner/haalcentraal/api.py
```diff
@@ -173,8 +173,8 @@ def fetch_data(self, user_bsn) -> dict | None:
         try:
             response = self.make_request(user_bsn)
             return get_json_response(response)
-        except (RequestException, ClientError) as e:
-            logger.exception("exception while making request", exc_info=e)
+        except (RequestException, ClientError):
+            logger.exception("exception while making request")
```
swrichards: Same as above, let's add a bit more verbosity here.

src/open_inwoner/laposta/client.py
```diff
@@ -30,8 +30,8 @@ def get_lists(self) -> list[LapostaList]:
         try:
             response = self.get("list")
             data = get_json_response(response)
-        except (RequestException, ClientError) as e:
-            logger.exception("exception while making request", exc_info=e)
+        except (RequestException, ClientError):
+            logger.exception("exception while making request")
```
swrichards: Add something about laposta in the message

src/open_inwoner/openklant/clients.py
swrichards, line 1: A lot of generic exception message here, would be good to at least include the type of API here as well.

src/open_inwoner/openklant/services.py
swrichards, line 1: Repeated: generic error messages.

src/open_inwoner/openzaak/clients.py
swrichards, line 1: Generic API errors, include service name.

src/open_inwoner/openzaak/import_export.py
```diff
@@ -431,7 +433,9 @@ def from_jsonl_stream_or_string(cls, stream_or_string: IO | str) -> Self:
                                 source_config, exclude_fields, line
                             )
                 except ZGWImportError as exc:
-                    logger.error(exc)
+                    logger.error(
```
swrichards: use `exc_info=True`

src/open_inwoner/openzaak/utils.py
```diff
@@ -45,9 +44,8 @@ def is_info_object_visible(
     """
     if info_object.status not in ["definitief", "gearchiveerd"]:
         logger.info(
-            "Ignoring informatieobject %s as not visible for user: status is neither "
-            "'definitief' nor 'gearchiveerd'",
-            info_object.url,
+            "Ignoring informatieobject as not visible for user: status is neither definitief nor gearchiveerd",
+            info_object_url=info_object.url,
```
swrichards, line 48: Let's include `info_object_status=info_object.status`

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -58,9 +58,9 @@ def import_catalog_configs() -> list[CatalogusConfig]:
     if result.has_errors:
         for response in result.failing_responses:
             logger.exception(
-                "Client %s encountered an exception. Import will continue, for any other configured clients",
-                response.client,
-                exc_info=response.exception,
+                "ZGW client encountered an exception. Import will continue, for any other configured clients",
+                client=str(response.client),
+                exc_info=str(response.exception),
```
swrichards: Does the API accept passing the exception to `exc_info`?

src/open_inwoner/search/utils.py
```diff
@@ -13,7 +13,7 @@ def load_synonyms() -> list:
     try:
         synonyms = [s.synonym_line() for s in Synonym.objects.all()]
     except DatabaseError as exc:
-        logger.warning(f"Synonyms for elasticsearch were not loaded: {exc}")
+        logger.warning("Synonyms for elasticsearch were not loaded", error=str(exc))
```
swrichards:
```suggestion
        logger.warning("Synonyms for elasticsearch were not loaded", exc_info=True)
```

src/open_inwoner/userfeed/management/commands/process_feed_updates.py
```diff
@@ -24,6 +24,6 @@ def auto_expire_items():
         completed_at__isnull=True,
     )
     for data in qs:
-        logger.info(f"automatically expired feed item: {data.id} {data}")
+        logger.info("automatically expired feed item:", id=data.id, data=data)
```
swrichards, line 27: Might as well be defensive here too:

```suggestion
        logger.info("automatically expired feed item:", id=data.id, data=str(data))
```

src/open_inwoner/utils/decorators.py
```diff
@@ -128,17 +129,19 @@ def wrapped(*args, **kwargs) -> RT:
                         raise
 
             cache_key = cache_key_with_attr_placeholders.format(**key_kwargs)
-            logger.debug("Resolved cache_key `%s` to `%s`", key, cache_key)
+            logger.debug(
+                "Resolved cache_key", original_key=str(key), cache_key=str(cache_key)
```
swrichards: I think both are guaranteed to be strings:

```suggestion
                "Resolved cache_key", original_key=key, cache_key=cache_key
```

src/open_inwoner/utils/logentry.py
```diff
@@ -70,9 +71,10 @@ def addition(request, object, message=""):
     Log that an object has been successfully added.
     """
     logger.info(
-        ("Added: {object}, {message}. \n{request}").format(
-            object=object, message=message, request=request
-        )
+        "Added",
+        object=str(object),
+        message=message,
+        request=str(request),
```
swrichards:
```suggestion
         "Timeline logger entry created",
        object=str(object),
        message=message,
        request=str(request),
        action_flag=LOG_ACTIONS[models.ADDITION  # Maybe store this separately
```

src/open_inwoner/utils/logentry.py
```diff
@@ -128,9 +132,10 @@ def user_action(request, object, message):
     aren't appropriate.
     """
     logger.info(
-        ("User action: {object}, {message}. \n{request}").format(
-            object=object, message=message, request=request
-        )
+        "User action",
+        object=str(object),
+        message=message,
+        request=str(request),
```
swrichards: Follow pattern above, using a generic error message but with more attrs.

src/open_inwoner/celery.py
```diff
@@ -25,7 +25,7 @@
 # Use django's logging settings as these are reset by Celery by default
 @setup_logging.connect()
 def config_loggers(*args, **kwargs):
-    from logging.config import dictConfig
+    from logging.config import dictConfig  # noqa: TID251
 
     dictConfig(settings.LOGGING)
```
swrichards: See my other comment re Celery: I am not sure this will suffice, we might have slightly different configs. I would follow what OF does here.

src/openklant2/_resources/base.py
```diff
@@ -130,9 +130,8 @@ def row_iterator(
             if next_url := _data.get("next"):
                 if max_requests and num_requests >= max_requests:
                     logger.info(
-                        "Number of requests while retrieving paginated results reached "
-                        "maximum of %s requests, returning results",
-                        max_requests,
+                        "Number of requests while retrieving paginated results reached maximum, returning results",
+                        max_requests=max_requests,
```
swrichards, line 134:
```suggestion
                        max_requests=max_requests, next_url=next_url,
```

CHANGELOG.rst
```diff
@@ -46,6 +46,7 @@ Nieuwe features
 * [:taiga-us:`3492`, :pr:`1960`]: Header en logo zijn kleiner gemaakt, onderwerpen en
   zoekveld zijn gewisseld van plaats, mobiele navigatie is nu een overlay en de mobiele
   welkom tekst is hoger geplaatst.
+* [:taiga-us:`3508`]: structlog toegevoegd voor betere logging
```
```suggestion
* [:taiga-us:`3508`, :pr:`1979`]: Alle logging aangepast om gebruikt te maken van ``structlog``, waardoor de logging data beter kan worden verwerkt door andere observability tools zoals Grafana.
```

swrichards (changes requested):
Good job, a lot of work here. I have a few comments here and there where I think it might make sense to improve the logging now, otherwise I didn't look too much at the calls and messages themselves, that's for later.

src/open_inwoner/accounts/tests/test_auth.py
pi-sigma: The `User` instance in the test becomes stale after form submission which updates the database. Previously, we passed the `User` instance itself to the logger and asserted that the logger was called with the instance - which works because equality in Django is based on primary key, not equivalence of field values. Now we're passing `str(user)` and assert against the string - but `str(user)` in the test is different from `str(user)` in the view. The latter contains updated field values, the former does not.

src/open_inwoner/cms/cases/views/status.py
pi-sigma: I've added
```
case_identificatie=self.case.identification,
case_uuid=str(self.case.uuid),
```

src/open_inwoner/cms/tests/cms_tools.py
pi-sigma: Nice, didn't know about this!

src/open_inwoner/openzaak/zgw_imports.py
pi-sigma: Yep

src/open_inwoner/utils/logentry.py
pi-sigma: ~~I don't think "Timeline logger entry created" captures the intent. The log is not about timeline logger entries, it's about the object (whatever it is, `User` instance or something else) that was created.~~

I misunderstood the suggestion, will apply the change. Though `models.ADDITON` won't be very useful (it's just a number). I'll add `action="Object created", similarly for the other cases.

swrichards (approved):

PR #1980, opened 2025-10-20
Cache decorator support for kwarg defaults and None/"None" conflation

swrichards:
## Summary

Our cache decorator did not handle function kwargs with defaults, which is needed for another feature. I also took the opportunity to handle a small edge case: `None` vs `"None"` confusion, which seems hypothetical in our code base, but is nevertheless still a potential nasty bug.

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/utils/tests/test_utils.py
```diff
@@ -244,3 +246,61 @@ def returns_none():
...
+        result1 = func(None)  # cache key: "dynamic:None"
+        result2 = func("")  # cache key: "dynamic:''"
+        result3 = func("None")  # cache key: "dynamic:'None'"
+
+        self.assertEqual(m.call_count, 3)
+        self.assertEqual(result1, "result:None")
+        self.assertEqual(result2, "result:")
+        self.assertEqual(result3, "result:None")
+
+    def test_literal_none_string_vs_none_value(self):
```
pi-sigma: The previous test already tests for `None` and `"None"`. Is this one necessary?

pi-sigma (changes requested):
Nice! I think the second test is redundant and can be removed though.

src/open_inwoner/utils/tests/test_utils.py
swrichards: Good catch, I think this was another rebase snafu. Will delete the redundant one, thanks.

pi-sigma (approved):

PR #1981, opened 2025-10-20
Feature/3528 redesign homepage actions

stefrado:
## Summary

Implement new design for homepage actions. Figma: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6658-20035&t=fNnLAPzDCf4LgzUl-1

UPDATE: the design has changed, while Groningen is still testing. Status labels do not exist in the NLDs component so we are adding a split Title/Message view:
https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=8337-9173&t=oNTEm6DNPwaK9pjB-1

> Depending on https://github.com/maykinmedia/open-inwoner-design-tokens/pull/22

## Issue References

https://taiga.maykinmedia.nl/project/open-inwoner/task/3528

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3513

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/react/lib/loader.ts
```diff
@@ -39,4 +46,16 @@ export default class ModuleLoader {
...
   }
+
+  // auto-detect the usage of a webcomponent on the page.
+  // if there is one import the needed script and define the component.
+  static async detectWebComponents() {
+    for (const wc of Object.entries(webcomponents)) {
+      const [name, im] = wc
+      const found = document.querySelector(name)
+      if (!found) continue
+      const { ActionListWebComponent } = await im()
```
stefrado: This should be generic... instead of the actual component.

src/open_inwoner/react/components/ActionList/ActionList.scss
```diff
@@ -0,0 +1,17 @@
+action-list {
+  margin-top: 24px;
```
jiromaykin: Wondering if we could start using equivalents like `margin-block-start `so we can start working the same way as NLDS.

src/open_inwoner/react/components/ActionList/ActionList.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+**Props:**
+- \`actions\`: Array of feed items with type, title, message, status info, and action details
+
+**Item Structure:**
+- \`title\`: Item title
+- \`message\`: Item description/message
+- \`action_text\`: Text for the action button
+- \`action_url\`: URL for the action link
+
```
jiromaykin: If you look at the documentation text inspiration from https://nl-design-system.github.io/denhaag/?path=/docs/react-action--docs
you can see they also added a hover state version in it, which would be nice.
We do not yet have a strict focus style in the project, so we usually just use browser default.

src/open_inwoner/react/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface IActionListProps {
+  actions: IActionProps[]
+}
+
+const ActionList: FC<IActionListProps> = ({ actions = [] }) => {
+  return actions?.map(
+    ({ title, message, action_url: url, action_text: text }, index) => {
+      return (
+        <Action
+          key={index}
```
jiromaykin: Right now the actual link area is much too small (just the button) while in  Den Haag Storybook the entire action is the link - which also has its own distinctive hover style and focus style:
https://nl-design-system.github.io/denhaag/?path=/story/react-action--hover

Did you see their stuff in a different Storybook maybe?

package.json
```diff
@@ -46,6 +46,7 @@
     "@emotion/react": "^11.5.0",
     "@emotion/styled": "^11.3.0",
     "@fortawesome/fontawesome-free": "^6.4.2",
+    "@gemeente-denhaag/action": "^4.0.1",
```
jiromaykin, line 49: And you can now also update our own design-tokens to 0.0.16 :-)

jiromaykin (changes requested):

src/open_inwoner/react/components/ActionList/ActionList.tsx
stefrado: I did, but then it is impossible to add `Ga naar actie` next to the arrow...

CHANGELOG.rst
```diff
@@ -51,6 +51,8 @@ Nieuwe features
 * [:taiga-us:`3508`, :pr:`1979`]: Alle logging aangepast om gebruikt te maken van ``structlog``,
   waardoor de logging data beter kan worden verwerkt door andere observability tools zoals
   Grafana.
+* [:taiga-us:`3519`, :pr:`1981`]: Nieuw design vsn openstaande acties plugin op de
```
jiromaykin:
```suggestion
* [:taiga-us:`3519`, :pr:`1981`]: Nieuw design van openstaande acties plugin op de
```

src/open_inwoner/react/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export interface IActionListProps {
+  actions: IActionProps[]
+}
+
+const ActionList: FC<IActionListProps> = ({ actions = [] }) => {
+  return actions?.map(({ title, message, action_url: url }, index) => {
+    return (
+      <ActionSingle key={index} link={url}>
```
jiromaykin, line 18: Not sure where the icon is actually coming from, because it differs from our own OIP icon, which is a little bit of a longer arrow.

So: instead of material icon [arrow forward](https://mui.com/material-ui/material-icons/?query=arrow&theme=Outlined&selected=ArrowForwardOutlined) we actually use '[east arrow](https://mui.com/material-ui/material-icons/?query=east&theme=Outlined&selected=EastOutlined)' in all of the designs, or maybe the design should be changed instead.

Also we are missing the "Ga naar actie" text now, but... this actually doesn't even exist in the DenHaag component:
https://nl-design-system.github.io/denhaag/?path=/story/react-action--list

so i guess that text should be removed from our OIP design?

src/open_inwoner/react/components/ActionList/ActionList.scss
```diff
@@ -0,0 +1,33 @@
...
+
+.denhaag-action {
+  border-left: none !important;
+  border-right: none !important;
+  font-size: 14px;
+  font-family: var(--oip-typography-sans-serif-font-family);
+
+  // Use browser default focus-state until there is a design for it.
+  &:focus {
+    outline: revert !important;
```
jiromaykin, line 13: nice,
i still would like to see a hover-color style as in https://nl-design-system.github.io/denhaag/?path=/story/react-action--list too though.
We are using that very light-grey also in our  KVK search-select hover/active colors (`var(--oip-color-gray-lightest)`)

jiromaykin (changes requested):

src/open_inwoner/react/components/ActionList/ActionList.scss
stefrado: The hover style is implemented in https://github.com/maykinmedia/open-inwoner-design-tokens/pull/23 and can be included when that PR is merge

src/open_inwoner/react/components/ActionList/ActionList.tsx
stefrado: Yes, the icon of this action component is not configurable - when you want the whole `Action` to be clickable or without creating a whole custom component (using the `Link` prop). We do use the icon `arrow_forward` more on the home page than we use the icon `east`, which makes the usage of `arrow_forward` more consistent.

Scherm­afbeelding 2025-10-27 om 11 15 29

Same for the `Ga naar actie` text. This is also not possible with the default `Action` component.

So i am not sure what you want me to do, besides that if we accept these changes that the design needs to be changed

jiromaykin: Yeah, let's go for the NLDS design then.

src/open_inwoner/react/lib/getJsonScriptData.ts
```diff
@@ -26,7 +26,8 @@ export function parseJsonSafely<T = unknown>(
   }
 
   try {
-    return JSON.parse(text) as T
+    const json = text.replace(/'/g, '"')
```
jiromaykin: For me this is breaking text that have inner apostrophes, like `"Aanvraagstatus is veranderd naar 'Aanvraag is in behandeling'"`

But also any of the userfeed messages get single quotes (like as if they are directly coming from Python) instead of double quotes in my browser, so this will never parse correctly:

`getJsonScriptData.bundle.js:1 Failed to parse JSON: [{'title': 'Bijzondere bijstand aanvragen', 'message': 'Wij hebben documenten van u nodig', 'action_text': '', 'action_url': 'http://localhost:8000/mijn-aanvragen/'}, {'title': 'Bijzondere status aanvragen', 'message': 'Zaakstatus is veranderd naar &#x27;<span class="status">Geen data beschikbaar</span>&#x27;', 'action_text': '', 'action_url': '/mijn-aanvragen/1/None/status/'}, {'title': 'Status gewijzigd', 'message': 'Antwoord nodig', 'action_text': '', 'action_url': 'http://localhost:8000/'}, {'title': 'Gemeentelijk e foo bar', 'message': 'Foo bar status message', 'action_text': '', 'action_url': 'http://localhost:8000/mijn-aanvragen/'}, {'title': 'Aanvraag witgoedregeling', 'message': 'Aanvraagstatus is veranderd naar Aanvraag is in behandeling', 'action_text': '', 'action_url': 'http://localhost:8000/mijn-aanvragen/'}] 

SyntaxError: Expected ',' or '}' after property value in JSON at position 268 (line 1 column 269)`

jiromaykin: Seems the back-ticks in the console error should be single quotes instead.

src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -11,34 +11,8 @@ <h2 class="utrecht-heading-2 {% if userfeed.action_required %}indicator{% endif
...
-                        <span class="status">{{ item.message }}</span>
-                    </p>
-                    <span class="button button--icon-before button--transparent">
-                    {% icon icon="east" icon_position="after" primary=True outlined=True %}
-                    </span>
-                </div>
-            </a>
-        {% endfor %}
+    <div class="card-container card-container--columns-1 plugin-card">
+        <action-list actions='{{ userfeed_item_list_json|safe }}' />
```
swrichards: The JSON should be escaped, it'll be un-escaped in the React component:

```suggestion
        <action-list actions='{{ userfeed_item_list_json }}' />
```

src/open_inwoner/react/lib/getJsonScriptData.ts
```diff
@@ -26,9 +26,21 @@ export function parseJsonSafely<T = unknown>(
...
-    return undefined
+    // If that fails, try to fix Python-style single quotes
+    try {
+      // More careful replacement that handles HTML entities
+      const fixed = text
+        .replace(/&#x27;/g, "'") // Decode HTML entity for apostrophe first
+        .replace(/&quot;/g, '\\"') // Decode HTML entity for quote
+        .replace(/&amp;/g, '&') // Decode HTML entity for ampersand
+        .replace(/'/g, '"') // Then replace single quotes with double
+      return JSON.parse(fixed) as T
```
swrichards: I don't think this is needed anymore if we remove ` | safe` from the `userfeed.html` as above.

src/open_inwoner/react/lib/getJsonScriptData.ts
swrichards: I've noted this in another comment, but this is a hacky consequence of the ` | safe` filter. We should just have Django escape it, then I don't think we need any of these fallbacks.

src/open_inwoner/react/lib/web-component/abstract.tsx
swrichards, line 1: As a general comment, `wc` feels needlessly cryptic. On the basis of the maxim that "code should be easy to read, not write", can't we use `lib/web-component`?

src/open_inwoner/cms/plugins/cms_plugins/userfeed.py
```diff
@@ -21,6 +23,8 @@ def render(self, context, instance, placeholder):
             {
                 "instance": instance,
                 "userfeed": feed,
+                "userfeed_item_list": feed.items_as_list(),
```
swrichards: This is not used
```suggestion
```

src/open_inwoner/react/lib/wc/wc.spec.tsx
```diff
@@ -0,0 +1,490 @@
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest'
+import { webcomponents, registerWebComponents } from './register'
+import GenericReactWebComponent, { AbstractWebComponent } from './abstract'
+import { normalizeAttribute } from './utils'
+
+describe('Test webComponents folder', () => {
+  describe('Test utils.tsx file', () => {
+    describe('Test normalizeAttribute', () => {
+      it('should convert kebab-case to camelCase', () => {
```
swrichards: Given that we're testing the same function with a different input/output spec: could we use `it.each` or some such?

src/open_inwoner/react/lib/wc/wc.spec.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+      it('should have adoptedCallback method', () => {
+        const instance = Object.create(AbstractWebComponent.prototype)
+        expect(typeof instance.adoptedCallback).toBe('function')
+      })
+
+      it('should have attributeChangedCallback method', () => {
+        const instance = Object.create(AbstractWebComponent.prototype)
+        expect(typeof instance.attributeChangedCallback).toBe('function')
+      })
```
swrichards: Other than the first: I am not sure what value these tests add given that we also have Typescript?

src/open_inwoner/react/lib/wc/wc.spec.tsx
```diff
@@ -0,0 +1,490 @@
...
+        class TestWebComponent extends GenericReactWebComponent<any> {
+          constructor() {
+            super(TestComponent)
+          }
+        }
+
+        const elementName = 'test-extend-element'
+        if (!customElements.get(elementName)) {
+          customElements.define(elementName, TestWebComponent)
+        }
```
swrichards: This is a test, we should know our initial conditions: why is the conditional needed?

src/open_inwoner/react/lib/wc/wc.spec.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+        const elementName = 'test-json-element'
+        if (!customElements.get(elementName)) {
+          customElements.define(elementName, TestWebComponent)
+        }
+
+        const jsonData = { key: 'value', nested: { prop: 42 } }
+        const div = document.createElement('div')
+        div.innerHTML = `<${elementName} data='${JSON.stringify(jsonData)}'></${elementName}>`
```
swrichards: The backend will HTML escape the JSON string. We should expand the test `jsonData` to include a bunch of strings that include HTML characters, and use something like https://lodash.info/doc/escape here to simulate what actually happens when Django passes JSON into the template:


```suggestion
        div.innerHTML = `<${elementName} data='${_.escape(JSON.stringify(jsonData))}'></${elementName}>`
```

src/open_inwoner/react/lib/wc/wc.spec.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+        expect(instance.props).toBeDefined()
+        expect(typeof instance.props).toBe('object')
+      })
+    })
+  })
+
+  describe('Test register.tsx file', () => {
+    describe('Test webcomponents constant', () => {
+      it('should have action-list component', () => {
```
swrichards: I'd expect a different test up for this: tests will fail if we ever decide to rename or remove the `action-list` wc, if I understand this correctly. Is there a way to test the register functionality with a test registry of webcomponents?

src/open_inwoner/userfeed/adapter.py
```diff
@@ -65,3 +65,14 @@ def action_url(self) -> str:
...
+    def to_dict(self) -> dict:
+        """Serialize the feed item to a dictionary for JSON serialization."""
+        # List of property names to serialize
+        fields = [
+            "title",
+            "message",
+            "action_text",
+            "action_url",
+        ]
+        return {field: getattr(self, field) for field in fields}
```
swrichards: I would move this to `cms_plugins.userfeed` as the needs are specific to that call site.

src/open_inwoner/userfeed/feed.py
```diff
@@ -34,6 +34,9 @@ def total_items(self) -> int:
     def __bool__(self):
         return self.has_display()
 
+    def items_as_list(self):
+        return [item.to_dict() for item in self.items]
+
```
swrichards: This can be inlined in  `cms_plugins.userfeed`

swrichards (changes requested):

src/open_inwoner/react/lib/web-component/abstract.tsx
jiromaykin: Agreed

src/open_inwoner/react/lib/wc/wc.spec.tsx
jiromaykin: We don't use lodash yet in the project, but i could add it.

src/open_inwoner/react/lib/wc/wc.spec.tsx
jiromaykin: Yeah, TypeScript already checks the method's existence.
(I did not write these 😄)

src/open_inwoner/react/lib/wc/wc.spec.tsx
jiromaykin: Like a 'mock' version of webcomponents used only in tests? I've attempted something in the new commit.

jiromaykin (approved):

src/open_inwoner/cms/plugins/cms_plugins/userfeed.py
```diff
@@ -17,10 +19,24 @@ class UserFeedPlugin(CMSPluginBase):
     def render(self, context, instance, placeholder):
         request = context["request"]
         feed = get_feed(request.user)
+
+        # Inline the serialization logic here (instead of calling items_as_list)
```
swrichards: GPT speak?

```suggestion
```

src/open_inwoner/cms/plugins/cms_plugins/userfeed.py
```diff
@@ -17,10 +19,24 @@ class UserFeedPlugin(CMSPluginBase):
...
+                "action_url": item.action_url,
+            }
+            for item in feed.items
+        ]
+
         context.update(
             {
                 "instance": instance,
                 "userfeed": feed,
+                "userfeed_item_list": feed_items_list,
```
swrichards: Is this still used?

src/open_inwoner/react/lib/getJsonScriptData.ts
```diff
@@ -28,7 +28,7 @@ export function parseJsonSafely<T = unknown>(
   try {
     return JSON.parse(text) as T
   } catch (error) {
-    console.error(`Failed to parse JSON: ${text}`, error)
+    console.error('Failed to parse JSON:', text, error)
     return undefined
   }
 }
```
swrichards: Does this function really add much over inlining `JSON.parse`? The checks for falsiness and type seem redundant, and if we're worried about no results or an exception, I think it'd be clearer to handle this in the function. Feels a bit like legacy from the earlier approach.

src/open_inwoner/react/lib/web-component/utils.ts
swrichards, line 1: Nitpick, but this does not contain JSX: `s/utils.tsx/utils.ts`

package.json
```diff
@@ -77,6 +78,7 @@
     "htmx.org": "^2.0.4",
     "leaflet": "^1.7.1",
     "lit": "^3.3.0",
+    "lodash.escape": "^4.0.1",
```
swrichards: This is a `devDependency` I think, as we only use it in the tests (so far).

src/open_inwoner/react/lib/web-component/web-component.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+          }),
+      })
+
+      it('registerWebComponents() should register all matching DOM elements as custom elements', async () => {
+        createTestRegistry()
+        const element = document.createElement('test-component-a')
+        document.body.appendChild(element)
+
+        // Test the registration logic with the test registry
+        expect(typeof registerWebComponents).toBe('function')
```
swrichards: I don't see how this assert matches the intent of the test (i.e. that `test-component-a` has been registered).

Maybe we can use something like https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements to verify the component has been registered? Makes the test less implementation dependent too (we just care about the result).

src/open_inwoner/react/lib/web-component/web-component.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+        document.body.removeChild(element)
+      })
+
+      it('should handle registry with dynamic components', () => {
```
swrichards: I find this test hard to follow, it's basically testing the type signature of webcomponents, which seems to me to be doing Typescript's work for it.

I would focus on the previous test and actually assert against the result we care about: registered components are in fact usable in the dom and present in the custom element registry.

src/open_inwoner/react/components/ActionList/index.ts
swrichards, line 1: Nitpick: no jsx -> `index.ts`.

src/open_inwoner/react/components/ActionList/ActionList.spec.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const specialActions = [
+      {
+        title: 'Test & Review',
+        message: 'Check <data> & verify',
+        action_url: '/test?id=1&type=review',
+      },
+    ]
+
+    render(<ActionList actions={specialActions} />)
+    expect(screen.getByText('Test & Review')).toBeInTheDocument()
```
swrichards, line 138: I think it'd make more sense to actually assert against the `message` here to show what we're expecting the component to do with special/HTML characters.

swrichards (changes requested):

src/open_inwoner/react/lib/web-component/utils.ts
jiromaykin: Yeah, besides that it's kinda smart.

src/open_inwoner/react/lib/web-component/web-component.spec.tsx
jiromaykin: Ah yes, tests should verify behavior, not types.

src/open_inwoner/react/lib/getJsonScriptData.ts
jiromaykin: Yes!

swrichards: Not the function does even less 😇 

Can't we replace all invocations of parseJsonSafely with `JSON.parse` and wrap that in a `try/catch` bloc?

swrichards (approved):
LGTM, just a final question re the JSON parsing

src/open_inwoner/react/lib/getJsonScriptData.ts
jiromaykin: yes, but it was used sooooo often in other code files, so i wanted to make it a bit more DRY :-)

jiromaykin:
@swrichards Is this approved now? (I can't pick you as re-re-re-re-re-viewer currently).

swrichards (approved):

PR #1986, opened 2025-10-23
Remove old modules and use react web components.

stefrado:
## Summary

Remove old module mounting react method and use cleaner web component mounting method.

## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added 
- [ ] Documentation updated 
    variables


src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -11,34 +11,10 @@ <h2 class="utrecht-heading-2 {% if userfeed.action_required %}indicator{% endif
...
-                        <span class="status">{{ item.message }}</span>
-                    </p>
-                    <span class="button button--icon-before button--transparent">
-                    {% icon icon="east" icon_position="after" primary=True outlined=True %}
-                    </span>
-                </div>
-            </a>
-        {% endfor %}
+    <div class="card-container card-container--columns-1 plugin-card">
+        <action-list>
```
swrichards, line 15: @stefrado can we explore whether we can do away with the entire parse-JSON-from-tag approach, and pass it directly as an attribute?


```suggestion
        <action-list data-items="{{ feed_item_list_json }}">
```

Where `feed_item_list_json` is `json.dumps`'ed in the backend.

Then in the web component you could do something like:

```
const items = JSON.parse(this.getAttribute('data-config'));
```

I think this would make the code simpler and easier to follow.

swrichards: (Same for the KvK branch selector)

stefrado:
Outdated

PR #1988, opened 2025-10-27
[#3511] Create CMS plugin for links

pi-sigma:
## Summary

Created a CMS plugin for displaying links, using an extended `LinkModel` to add icons

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3510?milestone=550

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added 
- [ ] Documentation updated 
    variables


src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,76 @@
...
+        context.update(
+            {
+                "instance": instance,
+            }
+        )
+        return context
+
+
+# Replace original CMS LinkPlugin model with extended Model
+plugin_pool.unregister_plugin(OriginalLinkPlugin)
```
swrichards: Shouldn't this be moved to the top of the module? Or does `CMSLinkPlugin` need to be registered before it can be removed?

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,76 @@
...
+                    ("external_link",),
+                    "icon",
+                )
+            },
+        ),
+        (
+            _("Link settings"),
+            {
+                "classes": ("collapse",),
+                "fields": (("target",),),
```
swrichards: Nested tuple?

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,76 @@
...
+    model = ExtendedCMSLink
+    form = CustomLinkForm
+
+    fieldsets = [
+        (
+            None,
+            {
+                "fields": (
+                    "name",
+                    ("external_link",),
```
swrichards: Why a tuple?

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,76 @@
...
+
+from open_inwoner.cms.plugins.models.links import CMSLinkPluginConfig, ExtendedCMSLink
+
+
+@plugin_pool.register_plugin
+class CMSLinkPlugin(CMSPluginBase):
+    model = CMSLinkPluginConfig
+    render_template = "cms/plugins/links/links.html"
+    module = _("General")
+    name = _("CMS link plugin")
```
swrichards: The name is quite generic, what about: "External Link Plugin"? The fact that it's CMS is not that relevant to users I'd say.

swrichards: Or "'Ga Naar' Links", using the domain language

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,76 @@
...
+            {
+                "fields": (
+                    "name",
+                    ("external_link",),
+                    "icon",
+                )
+            },
+        ),
+        (
+            _("Link settings"),
```
swrichards: Maybe `Advanced settings` ?

swrichards (approved):
Just some nitpicks

src/open_inwoner/cms/plugins/cms_plugins/links.py
pi-sigma: I wanted to keep the registering/unregistering together, but it can be moved to the top as well

src/open_inwoner/cms/plugins/cms_plugins/links.py
pi-sigma: typo

src/open_inwoner/cms/plugins/cms_plugins/links.py
pi-sigma: The plugin *is* pretty generic. It can be extended to cover internal links, file, links, and other things. The display name for the plugin is customizable ("Ga naar" by default). However, we can go with "External Link Plugin" since that's the only function we support for now

src/open_inwoner/cms/plugins/cms_plugins/links.py
pi-sigma: `Advanced settings` are the ones I've removed because they're not needed. This is the default name. Though might as well re-use "Advanced settings" for this section

src/open_inwoner/cms/plugins/cms_plugins/links.py
jiromaykin: I do think 'External Link Plugin' is better because we may end up with different looking 'link plug-ins' if in the future Product Pages would become part of the CMS.

CHANGELOG.rst
```diff
@@ -51,6 +51,8 @@ Nieuwe features
 * [:taiga-us:`3508`, :pr:`1979`]: Alle logging aangepast om gebruikt te maken van ``structlog``,
   waardoor de logging data beter kan worden verwerkt door andere observability tools zoals
   Grafana.
+* [:taiga-us:`3511`, :pr:`1988`]: CMS-plugin toegevoegd voor links naar andere portalen van de
+  gemeente
```
```suggestion
  gemeente.
```

swrichards (approved):

src/open_inwoner/templates/cms/plugins/links/external-links.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n icon_tags %}
+
+<section class="plugin">
+    <h2 class="utrecht-heading-2">{{ instance.title }}</h2>
+    <div class="card-container card-container--columns-2">
```
jiromaykin, line 5: I worry a bit about the generic link name of this template. Perhaps also name this "external-link.html"?
In future we might need to add more types of link plug-ins if product pages would become CMS pages.

jiromaykin (changes requested):
Little note: I have heard many users use the word 'CMS' different than how we sue it :-) When they see the default Django Admin, they call that '_the_' CMS which is kind of true, so for them there is no distinction with the Django-CMS package.

src/open_inwoner/templates/cms/plugins/links/external-links.html
pi-sigma: This is pretty standard, but I take your point. I'll update the link name

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -0,0 +1,75 @@
...
+from djangocms_link.cms_plugins import LinkPlugin as OriginalLinkPlugin
+
+from open_inwoner.cms.plugins.models.links import CMSLinkPluginConfig, ExtendedCMSLink
+
+# Replace original CMS LinkPlugin model with extended Model
+plugin_pool.unregister_plugin(OriginalLinkPlugin)
+
+
+@plugin_pool.register_plugin
+class CMSLinkPlugin(CMSPluginBase):
```
jiromaykin, line 15: I've noticed something odd: you can add 'plug ins' to each link with its plus icon, which of course is not functional. I guess that's a fun Django-CMS side effect you'll need to disable somehow. 
The other plug-ins do not have this though:

Image

jiromaykin (changes requested):

src/open_inwoner/cms/plugins/cms_plugins/links.py
pi-sigma: Good catch, it is annoying. I've disabled it.

jiromaykin (approved):

swrichards (approved):

PR #1992, opened 2025-10-28
Emit outgoing request log messages as structured events

swrichards:
## Summary

This PR adds a handler to emit outgoing requests as debug entries with all the attributes passed as parameters that `structlog` can then format. We might want more granularity in due course (especially avoiding the inclusion of bodies, which might include sensitive data).

## Issue References

- Taiga Issue: [#3556](https://taiga.maykinmedia.nl/project/open-inwoner/task/3556)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/utils/logging.py
+++ b/src/open_inwoner/utils/logging.py
@@ -1,0 +1,131 @@
+import time
+from collections import OrderedDict
+from datetime import timedelta
+from logging import Handler  # noqa: TID251 -- needed for base class of handler
+from typing import Mapping, Optional, cast
+from urllib.parse import urlparse
+
+from django.utils import timezone
+
+import structlog
+from log_outgoing_requests.compat import format_exception
+from log_outgoing_requests.typing import (
+    AnyLogRecord,
+    ErrorRequestLogRecord,
+    RequestLogRecord,
+    is_request_log_record,
+)
+from requests import PreparedRequest, Request, RequestException, Response
+
+logger = structlog.stdlib.get_logger(__name__)
+
+
+class StructlogOutgoingRequestsHandler(Handler):
+    """Emit an outgoing request log as structured data."""
+
+    def emit(self, record: AnyLogRecord):
+        # skip requests not coming from the library requests
+        if not record or not is_request_log_record(record):
+            return
+
+        from log_outgoing_requests.models import OutgoingRequestsLogConfig
+        from log_outgoing_requests.utils import process_body
+
+        config = cast(OutgoingRequestsLogConfig, OutgoingRequestsLogConfig.get_solo())
+        record = cast(RequestLogRecord, record)
+
+        # check if we're dealing with success or error state
+        request: PreparedRequest | Request | None
+        response: Response | None
+        exception = record.exc_info[1] if record.exc_info else None
+        if (response := getattr(record, "res", None)) is not None:
+            # we have a response - this is the 'happy' flow (connectivity is okay)
+            request = record.req
+        elif isinstance(exception, RequestException):
+            record = cast(ErrorRequestLogRecord, record)
+            # we have an requests-specific exception
+            request = exception.request or None
+            response: Optional[Response] = exception.response  # likely None
+        else:
+            logger.debug("Received log record that cannot be handled", record=record)
+            return
+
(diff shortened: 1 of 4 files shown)


src/open_inwoner/utils/logging.py
```diff
@@ -0,0 +1,133 @@
...
+                    "res_content_type": processed_response_body.content_type,
+                    "res_body": processed_response_body.content,
+                    "res_body_encoding": processed_response_body.encoding,
+                }
+            )
+
+        # Emit the actual log message
+        logger.debug("outgoing_request", **kwargs)
+
+    def format_headers(self, headers):
```
pi-sigma: Where is this used? Might be used somewhere in the library (code looks familiar) but I don't see it being called here.

src/open_inwoner/utils/logging.py
```diff
@@ -0,0 +1,133 @@
...
+                "url": request.url if request else "(unknown)",
+                "hostname": parsed_url.netloc if parsed_url else "(unknown)",
+                "timestamp": timestamp,
+                "response_ms": (
+                    int(response.elapsed.total_seconds() * 1000)
+                    if response is not None
+                    else None
+                ),
+                # "req_headers": scrubbed_req_headers,
+                # "res_headers": response.headers if response is not None else {},
```
pi-sigma: Can be removed (or do we need them?)

src/open_inwoner/utils/logging.py
```diff
@@ -0,0 +1,133 @@
...
+        ):
+            kwargs.update(
+                {
+                    "req_content_type": processed_request_body.content_type,
+                    "req_body": processed_request_body.content,
+                    "req_body_encoding": processed_request_body.encoding,
+                }
+            )
+
+        # # check response
```
pi-sigma:
```suggestion
        # check response
```

src/open_inwoner/utils/logging.py
```diff
@@ -0,0 +1,133 @@
...
+                "timestamp": timestamp,
+                "response_ms": (
+                    int(response.elapsed.total_seconds() * 1000)
+                    if response is not None
+                    else None
+                ),
+                # "req_headers": scrubbed_req_headers,
+                # "res_headers": response.headers if response is not None else {},
+                "trace": "\n".join(format_exception(exception)) if exception else None,
+                "params": parsed_url.params if parsed_url else "(unknown)",
```
pi-sigma, line 78: Do we want to add `"query": parsed_url.query if parsed_url else "(unknown)"`?

src/open_inwoner/utils/logging.py
```diff
@@ -0,0 +1,133 @@
...
+        # check if we're dealing with success or error state
+        exception = record.exc_info[1] if record.exc_info else None
+        if (response := getattr(record, "res", None)) is not None:
+            # we have a response - this is the 'happy' flow (connectivity is okay)
+            request = record.req
+        elif isinstance(exception, RequestException):
+            record = cast(ErrorRequestLogRecord, record)
+            # we have an requests-specific exception
+            request: Optional[PreparedRequest] = exception.request
+            response: Optional[Response] = exception.response  # likely None
```
pi-sigma: Types for `request` and `response` are declared in this branch but not the `if` branch. After the `if/elif`, which types do they have? It's hard to tell from the annotation. I would declare them before the branching:
```
request: Optional[PreparedRequest] = None
response: Optional[Response] = None

exception = ...
if (response ...
```

pi-sigma (changes requested):

src/open_inwoner/utils/logging.py
swrichards: Good catch, redundant copy-pasta from the original code (which was intended to format the thing as text). Will remove.

src/open_inwoner/utils/logging.py
swrichards: Yup nice

src/open_inwoner/utils/logging.py
swrichards: Improved this.

pi-sigma (approved):
Good to go once CI passes

swrichards:
> Good to go once CI passes

It's the stupid flaky linkchecker. I'll merge this and fix that separately.

PR #1993, opened 2025-10-28
[#3516] add color-picker colors as override to NLDS values

jiromaykin:
## Summary

Connect OIP color-picker to HSL NLDS Design-tokens so that NLDS values are not set in stone and municipalities can use the color picker for their Huisstijl. The color picker always needs to be overwriting the NLDS values. And: “geavanceerde opties” in Admin then needs to again overwrite everything else, in case a municipality uses their own design-token values.

Review this by looking at the 'active' color in the SideNavigation and see if it responds to color picker changes.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3516

## Checklist

- [x] CHANGELOG.rst updated
- [x] Documentation updated docs/beheerhandleiding/12_configuratie.rst

### DIFF
--- a/src/open_inwoner/scss/components/Pagination/Pagination.scss
+++ b/src/open_inwoner/scss/components/Pagination/Pagination.scss
@@ -30,11 +30,12 @@
     text-decoration: none;
 
     &--active {
-      color: var(--color-white);
+      color: var(--color-font-primary);
       background-color: var(--color-primary);
       border-radius: var(--border-radius-rounded);
 
       &:hover {
+        color: var(--color-white);
         background-color: var(--color-body);
       }
     }
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,6 +113,8 @@ Bugfixes
 * [:taiga-ta:`3557`, :pr:`1987`]: Ontbrekende verplichte velden toegevoegd aan de
   Subscription detail admin, waardoor het weer mogelijk is om nieuwe Subscription
   objecten aan te maken.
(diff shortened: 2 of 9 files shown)


src/open_inwoner/scss/components/Pagination/Pagination.scss
```diff
@@ -30,11 +30,12 @@
     text-decoration: none;
 
     &--active {
-      color: var(--color-white);
+      color: var(--color-font-primary);
       background-color: var(--color-primary);
       border-radius: var(--border-radius-rounded);
 
       &:hover {
+        color: var(--color-white);
```
swrichards, line 38: Shouldn't this also be a variable for the contrasting color?

swrichards: (Or also primary as we had before)

swrichards (approved):

pi-sigma (approved):

src/open_inwoner/scss/components/Pagination/Pagination.scss
jiromaykin: No, I thought so too at first, but this hovered page-number has background color "--color-body" which is a static dark gray.
If a municipality would have a very light primary color, they can then set the text colors for Buttons as 'dark' in the OIP color picker.

This 'dark' var(--color-font-primary) text-color would then not have enough contrast with this gray hover style.

We would have a bit of a bigger problem here for combining static colors with ANY kind of color-picker combination . 
We may want to add completely separate color pickers for those primary/secondary/accent **Text** colors, but maybe that's also a bit too much.

PR #1995, opened 2025-10-28
fix: CMS Categories plugin should only query zaken of a backend is co…

swrichards:
## Summary

Not all instances have a ZGW backend configured, and some of those have the CMS plugin configured. However, initializing a ZGW client without groups triggers an exception. This PR ensures we only fetch cases if there is a backend with which to do so.

## Issue References

- Taiga Issue: [#3561](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3561)

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/cms/products/cms_plugins.py
+++ b/src/open_inwoner/cms/products/cms_plugins.py
@@ -7,7 +7,7 @@ from cms.plugin_pool import plugin_pool
 
 from open_inwoner.cms.utils.plugin_mixins import CMSActiveAppMixin
 from open_inwoner.components.templatetags.menu import react_sidenav_data
-from open_inwoner.openzaak.models import OpenZaakConfig
+from open_inwoner.openzaak.models import OpenZaakConfig, ZGWApiGroupConfig
 from open_inwoner.pdc.forms import ProductFinderForm
 from open_inwoner.pdc.models import Category, ProductCondition, ProductLocation
 from open_inwoner.questionnaire.models import QuestionnaireStep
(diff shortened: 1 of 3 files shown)


src/open_inwoner/cms/products/cms_plugins.py
```diff
@@ -77,6 +77,9 @@ def render(self, context, instance, placeholder):
                 config.enable_categories_filtering_with_zaken
                 and request.user.is_authenticated
                 and (request.user.bsn or request.user.kvk)
+                and
+                # There might be no ZGW backend configured
+                ZGWApiGroupConfig.objects.count()
```
pi-sigma:
```suggestion
                ZGWApiGroupConfig.objects.exists()
```
More efficient query

src/open_inwoner/cms/products/cms_plugins.py
```diff
@@ -77,6 +77,9 @@ def render(self, context, instance, placeholder):
                 config.enable_categories_filtering_with_zaken
                 and request.user.is_authenticated
                 and (request.user.bsn or request.user.kvk)
+                and
+                # There might be no ZGW backend configured
```
pi-sigma:
```suggestion
                # There might be no ZGW backend configured
                and ZGWApiGroupConfig...
```

pi-sigma (approved):
Minor suggestions, otherwise good to go

src/open_inwoner/cms/products/cms_plugins.py
swrichards: You're right!

PR #1996, opened 2025-10-28
[#3511] Add front-end NLDS template for 'Ga naar' external-links plug-in

jiromaykin:
## Summary

Adding HTML template + design tokens for the 'Ga naar' external links homepage plugin.
Figma: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=6954-11139&m=dev 

This is an external link plug-in for the Home page, meant for municipalities who are not (yet) able to add theses portals to the Open-Inwoner platform.

externe-links-voltooid

## Issue References

- Taiga story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3511
- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3531
- ~~TODO~~ DONE: add new designtokens (this component does not exist in NLDS so we are delivering our own): https://github.com/maykinmedia/open-inwoner-design-tokens/pull/28

Fixed: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3583

## Checklist

- [x] CHANGELOG.rst updated
- [x] Storybook component created (only for viewing, no 'states' to test)


CHANGELOG.rst
```diff
@@ -53,15 +53,15 @@ Nieuwe features
   Grafana.
 * [:taiga-us:`3519`, :pr:`1981`]: Nieuw design van openstaande acties plugin op de
   homepagina geïmplementeerd.
-* [:taiga-us:`3511`, :pr:`1988`]: CMS-plugin toegevoegd voor links naar andere portalen van de
```
jiromaykin, line 56: @pi-sigma I had to remove your old one here (it had the wrong userstory number) and you actually added the correct changelog already at line 63.

src/open_inwoner/react/components/HTMLstories/ExternalLinks.stories.tsx
swrichards, line 1: I am uneasy about merging this, because the principle of storybook is that the stories draw on the same code that is used in your application, whereas this is just a copy-paste. This will be fixed if and when our experiment with webcomponents works, but on the off-chance we don't get that done soon, I'd be in favour of leaving this out, because we have no checks against this going out of sync.

swrichards (changes requested):

src/open_inwoner/react/components/HTMLstories/ExternalLinks.stories.tsx
jiromaykin: 👍 Okay - i just thought this would be nice if in future we would add a Github link for municipalities to view OIP Storybook with 'working/responsive' components, because we get questions from municipalities not understanding the 'flow' of components in Figma. 
But I agree it would be better if it's a maintainable thing. 
Although Utrecht and DenHaag do actually also add ordinary HTML components in their Storybook (they have sections coming from different instances like Webcomponents, Angular, etcetera.

src/open_inwoner/templates/cms/plugins/links/external-links.html
```diff
@@ -1,17 +1,22 @@
...
-            {% for link in link_plugin.child_plugin_instances %}
-                <a href="{{ link.get_link }}" class="card" {% if link.target %}target="{{ link.target }}" {% endif %}>
-                    <span class="button button--icon-before button--transparent">
-                        {% icon icon=link.icon %}
+    <section class="plugin external-links">
+       <h2 class="utrecht-heading-2">{{ link_plugin.title }}</h2>
+       <ul class="card-container card-container--columns-2 external-links__list">
+        {% for link in link_plugin.child_plugin_instances %}
+            <li class="external-links__list-item">
+                <a href="{{ link.get_link }}" class="external-link" {% if link.target %}target="{{ link.target }}" {% endif %}>
```
pi-sigma:
```suggestion
                <a href="{{ link.get_link }}" class="external-link" {% if link.target %}target="{{ link.target }}" rel="noopener noreferrer" aria-label="{{ link.name.text }} {% trans '(opens in new window)' %}"{% endif %}>
```

pi-sigma (changes requested):
The link could use an aria label in case it opens a new tab (also prevent sending referrer info). Otherwise looking good!

src/open_inwoner/templates/cms/plugins/links/external-links.html
jiromaykin: Ah yes, at first I thought this would not work, because aria-labels overwrite everything that's inside of them, but adding the actual text + hint  (not html hopefully) would work great.

jiromaykin: But: it had to be `name_text`, not `name.text`

pi-sigma (approved):

swrichards (approved):

PR #1997, opened 2025-10-29
Implement OIDC EIDAS support

swrichards:
## Summary

This PR implements eIDAS/OIDC support. The implementation is slightly awkward, because our existing DigiD/eHerkenning OIDC backends rely on the `digid_eherkenning` packages, and we have to implement the same patterns using the underlying `mozilla_django_oidc` library, because the former does not yet support `digid_eherkenning`. We'll have an opportunity to rationalize this a bit during the planned upgrade of the `mozilla_django_oidc_db` library.

## Issue References

- Taiga User Story: [#3509](https://taiga.maykinmedia.nl/project/open-inwoner/us/3509)

## Checklist

- [x] CHANGELOG.rst updated
  - [x] Deployment notes added


src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        ):
+            return None
+
+        updated = User.objects.filter(id=user.id).update(**create_kwargs)
+
+        if not updated:
+            # TODO Log update error
+            return user
+
+        return user
```
swrichards, line 499: @pi-sigma only if you have some spare time, this is the function to look at, together with the user model changes (everything else is messy and incomplete, so don't worry about style, comments and typos, it's mainly about the high-level approach).

The basic idea is that EIDAS claims can lead you to branch on three different types of user:

1. Natural person, with pseudo id but no bsn
2. Natural person, with pseudo id _and_ bsn
3. Legal person, with pseudo id

Pseudo id is the common denominator, and required by the spec, so I've chosen that as the unique ID for EIDAS users.

I've added unique constraints for this as well, as well as new login types.

Cleanup, config, actual UI for the login button etc. to follow.

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        # Do we have a BSN? If so, add it and update login type to reflect this
+        if claim_values["bsn"]:
+            person_kwargs = person_kwargs | {
+                "bsn": claim_values["bsn"],
+                "login_type": LoginTypeChoices.eidas_person_bsn,
+            }
+
+        return person_kwargs
+
+    def filter_users_by_claims(self, claims):
```
pi-sigma, line 459: The func is supposed to `filter_user_by_claims`, but then actually only filters by `pseudo_id`, despite there being three types of users with different kinds of claims. Perhaps I misunderstand what the function is supposed to do?

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        claims = self._extract_eidas_claim_values(claims)
```
pi-sigma:
```suggestion
        claims_values = self._extract_eidas_claim_values(claims)
```
Avoid shadowing

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        if not claim_values["pseudo_id"]:
+            msg = (
+                "Claims did not include a value for the configured pseudo identifier claim. "
+                "Pseudo identifier is required for all eIDAS authentications."
+            )
+            self.log_system_error(msg)
+            raise SuspiciousOperation(msg)
+
+        # Validate that at least one identifying claim is present (excluding pseudo_id)
+        identifying_claims = {k: v for k, v in claim_values.items() if k != "pseudo_id"}
```
pi-sigma, line 403: On the one hand, the docstring says "WE rely on the Pseudo ID, which is guaranteed to be unique per service provider/user". But here you raise an error if the identifying claims consist only of `pseudo_id`'s. Is this a contradiction? I guess I don't fully grok the purpose and concept of `pseudo_id`.

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+                "login_type": LoginTypeChoices.eidas_company,
+                "company_name": claim_values["company_name"],
+                "eidas_company_id": claim_values["legal_entity_id"],
+            }
+
+        # Assume this is a person
+        person_kwargs = base_create_kwargs | {
+            "login_type": LoginTypeChoices.eidas_person_pseudo_id,
+            "first_name": claim_values["first_name"] or "",
+            "last_name": claim_values["family_name"] or "",
```
pi-sigma, line 447: Anything that has only `pseudo_id` but not `first_name` or `last_name` is by default a person. Is that right?

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+
+        base_create_kwargs = {
+            "eidas_pseudo_id": claim_values["pseudo_id"],
+            "email": generate_email_from_string(
+                claim_values["pseudo_id"], domain="localhost"
+            ),
+        }
+
+        # Try company first, as the required fields are unambiguous
+        if claim_values["company_name"] and claim_values["legal_entity_id"]:
```
pi-sigma: If only one is present, the code falls through to the person case. Should check that if one present the other present as well.

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,183 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+
+        return user
+
+    def update_user(self, user: User, claims: JSONObject):
+        if not (
+            create_kwargs := self._construct_user_creation_kwargs_from_claims(claims)
+        ):
+            return None
+
+        updated = User.objects.filter(id=user.id).update(**create_kwargs)
```
pi-sigma: Bypasses model validation (`clean` checks bsn-users for duplicate emails). Use `user.save()` instead.

src/open_inwoner/accounts/backends.py
swrichards: It's a required method inherited from the base-class, which uses the count to determine whether a user exists, needs to be updated, or throws an error if more than 1 user is returned. The purpose is thus basically to give you the "unique" queryset for the given set of claims.

src/open_inwoner/accounts/backends.py
swrichards: It's a case of "There's a standard, but we can't count on anybody following it" so have to be defensive, or more charitably, if an admin configures the incorrect/insufficient OIDC scopes, we might have a case where only a few of the claims are there.

src/open_inwoner/accounts/models.py
```diff
@@ -103,6 +105,138 @@ def get_callback_view(self):
...
+
+    company_name_claim = ClaimField(
+        verbose_name=_("Company name claim"),
+        default=ClaimFieldDefault("company_name"),
+        blank=True,
+        help_text=_(
+            "Claim path for company/organization legal name. Source: eIDAS urn:etoegang:1.12:EntityConcernedID:LegalName"
+        ),
+    )
+    company_representative_pseudo_identifier_claim = ClaimField(
```
pi-sigma: Where is this used?

src/open_inwoner/accounts/models.py
```diff
@@ -349,12 +495,16 @@ class Meta:
         verbose_name = _("User")
         verbose_name_plural = _("Users")
 
+        # TODO: Update unique constraints
         constraints = [
             UniqueConstraint(
```
pi-sigma, line 489: The constraint is getting long and complex. Might be worth extracting the conditions and give them some descriptive name:

```
DUPLICATE_EMAILS_ALLOWED_LOGIN_TYPES = [
    LoginTypeChoices.digid,
    LoginTypeChoices.eherkenning,
    LoginTypeChoices.eidas_person_bsn,
    LoginTypeChoices.eidas_person_pseudo_id,
    LoginTypeChoices.eidas_company,
]

UniqueConstraint(
    fields=["email"],
    condition=~Q(login_type__in=DUPLICATE_EMAILS_ALLOWED_LOGIN_TYPES),
    name="unique_email_for_regular_users",
)
```

src/open_inwoner/accounts/models.py
```diff
@@ -103,6 +105,138 @@ def get_callback_view(self):
...
+        blank=True,
+        help_text=_(
+            "Claim path for PseudoID - unique pseudonymous identifier for natural persons. "
+            "Use this to identify users across authentication sessions when BSN is not available. "
+            "Source: eHerkenning/eIDAS urn:etoegang:1.12:EntityConcernedID:PseudoID"
+        ),
+    )
+
+    # Level of Assurance claim
+    loa_claim = ClaimField(
```
pi-sigma, line 201: This is displayed in the admin, but do we use this anywhere for validation?

src/open_inwoner/accounts/backends.py
swrichards: Yes, but we still need the keys to be present (the idea is that the keys are mutually exclusive for natural and legal persons).

src/open_inwoner/accounts/models.py
swrichards: I solved this by simply inverting the condition, only the regular user needs a unique id.

src/open_inwoner/accounts/models.py
swrichards: We haven't implemented yet, but it's an accepted part of the standard and I suspect we'll have to deal with it soon'ish. It's good to have it here so admins only have to configure this once, as it's a painful configuration.

src/open_inwoner/accounts/models.py
swrichards: You're right, it's not used yet as we haven't thought about delegated access for the login forms. Deleting.

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,164 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+
+    def update_user(self, user: User, claims: JSONObject):
+        if not (
+            create_kwargs := self._construct_user_creation_kwargs_from_claims(claims)
+        ):
+            return None
+
+        for attr, val in create_kwargs.items():
+            setattr(user, attr, val)
+            user.save()
```
pi-sigma:
```suggestion
        for attr, val in create_kwargs.items():
            setattr(user, attr, val)           
        try:
            user.full_clean()
        except ValidationError:
            ....
        else:       
            user.save()
```

src/open_inwoner/accounts/eidas_urls.py
```diff
@@ -0,0 +1,16 @@
...
+from mozilla_django_oidc.urls import urlpatterns
+from mozilla_django_oidc_db.views import OIDCCallbackView
+
+from .views import eidas_init, eidas_logout
+
+app_name = "eidas_oidc"
+
+
+urlpatterns = [
+    # XXX: generic callback view, this can move to a single URL.
```
pi-sigma: XXX ?

pi-sigma (changes requested):

src/open_inwoner/accounts/backends.py
pi-sigma: Makes sense, perhaps adapt the comment here to avoid this confusion

src/open_inwoner/accounts/backends.py
swrichards: :facepalm:

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,186 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+
+        try:
+            user.full_clean()
+        except ValidationError as exc:
+            raise SuspiciousOperation(
+                "Unable to update user due to a validation error"
+            ) from exc
+        else:
+            logger.info(
+                "Updating new eIDAS user", pseudo_id=create_kwargs["eidas_pseudo_id"]
```
pi-sigma:
```suggestion
                "Updating eIDAS user", pseudo_id=create_kwargs["eidas_pseudo_id"]
```
or was the "new" intentional?

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,186 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        self, claims: JSONObject
+    ) -> dict | None:
+        """
+        Extract user filter and create kwargs based on eIDAS claim patterns.
+        """
+        # Extract all claim values
+        claim_values = self._extract_eidas_claim_values(claims)
+        logger.info("eidas_oidc_claims_received", claim_keys=list(claims.keys()))
+
+        # Validate presence of requried pseudo_id
```
pi-sigma:
```suggestion
        # Validate presence of required pseudo_id
```

src/open_inwoner/accounts/models.py
```diff
@@ -103,6 +105,128 @@ def get_callback_view(self):
...
+        verbose_name=_("Date of birth claim"),
+        default=ClaimFieldDefault("birthdate"),
+        blank=True,
+        help_text=_(
+            "Claim path for date of birth in format YYYY-MM-DD. Source: eHerkenning/eIDAS"
+        ),
+    )
+
+    # Legal person (company) claims
+    ## Note this is not pseudonomized, and _separate_ from the psuedo ID
```
```suggestion
    ## Note this is not pseudonymized, and _separate_ from the pseudo ID
```

src/open_inwoner/accounts/backends.py
```diff
@@ -314,3 +314,186 @@ def update_user(self, user: AbstractUser, claims: JSONObject):
...
+        # Do we have a BSN? If so, add it and update login type to reflect this
+        if claim_values["bsn"]:
+            person_kwargs = person_kwargs | {
+                "bsn": claim_values["bsn"],
+                "login_type": LoginTypeChoices.eidas_person_bsn,
+            }
+
+        return person_kwargs
+
+    def filter_users_by_claims(self, claims):
```
pi-sigma, line 459: Am I correct that, if a user first logs into OIP with via DigiD (`bsn` but no `pseudo_id`) and later via eIDAS (`bsn` plus `pseudo_id`),  they will be treated as different users, a new account will be created?

Perhaps that's intentional (they use different auth methods after all), but should be documented somewhere.

pi-sigma (approved):
Couple of typos, otherwise good to go

src/open_inwoner/accounts/backends.py
swrichards: This is indeed intentional. Our system currently does not true bsn as an identifying attribute for _auth_ purposes, though it does for polling separate systems. I think this design makes sense (because auth and users are closely linked). Perhaps it some point we'll get a feature request to consolidate the `bsn` users, but for now this seems cleaner to me.

PR #2001, opened 2025-10-30
[#3563] Fix header overlay desktop, and mobile header

jiromaykin:
## Summary

Remove transparent Welcome Banner overlay on desktop, fix mobile Welcome header height + also fix both for when Notifications are popped up on  http://localhost:8000/accounts/register/  

Also: removing a script that was adding a class for Home URL only.
We are now allowed to use the CSS `has:` attribute now that Firefox also supports it.

UPDATE: updated design: there is now a maximum image height on mobile and when user is not logged is, also an indent on the other 'welcome' text (not part of the plug-in).

## Issue References

- original Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3492
- Taiga fix Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3563

## Checklist

- [x] CHANGELOG.rst updated


pi-sigma (approved):

src/open_inwoner/scss/components/Grid/Grid.scss
```diff
@@ -37,16 +37,14 @@
...
+        $image-height-mobile: 140px;
+        margin-top: -$image-height-mobile;
         padding: var(--spacing-large);
         padding-bottom: 0;
         box-sizing: border-box;
         width: 100%;
+        border-top-left-radius: var(--border-radius);
+        border-top-right-radius: var(--border-radius);
         background-color: rgba(255, 255, 255, 0.6); /* fallback */
         @supports (background-color: color-mix(in srgb, white, transparent)) {
```
jiromaykin: @pi-sigma I have to change this color to white instead of a transparency (for accessibility reasons).

pi-sigma (approved):

PR #2005, opened 2025-11-03
Avoid removal of newlines when running 'npm run build'

pi-sigma:
'npm run build' triggers 'npm run compilemessages', which in turn runs 'formatjs'. The latter removes trailing newlines, ignoring .editorconfig and all other config options. Added a command to add trailing newlines when running 'formatjs'.

## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added 
- [ ] Documentation updated 
    variables

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
     "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -c",
     "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
-    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled",
+    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled && pre-commit run end-of-file-fixer --files src/open_inwoner/react/i18n/compiled/*.json || true",
     "fix-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -w",
     "makemessages": "conc npm:makemessages:*",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
     "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -c",
     "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
-    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled",
+    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled && sed -i -e '$a\\' src/open_inwoner/react/i18n/compiled/*.json",
```
jiromaykin: instead of `sed`,
perhaps use `"compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled && pre-commit run end-of-file-fixer --files src/open_inwoner/react/i18n/compiled/*.json || true",`

this does  these things:
-at compile/build it creates JSON without newlines because that's how it goes,
-then immediately runs pre-commit's end-of-file-fixer on those specific files.
-The `|| true` prevents NPM from treating it as a failure when pre-commit modifies files. You'll see what happens if you leave that out.

jiromaykin: Or instead: always run `pre-commit run -a` before committing :-)

pi-sigma: Nice, thanks!

package.json
```diff
@@ -14,7 +14,7 @@
     "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
     "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -c",
     "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
-    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled",
+    "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled && pre-commit run end-of-file-fixer --files src/open_inwoner/react/i18n/compiled/*.json || true",
```
jiromaykin, line 17: Actually we do not need this PR at all,
you should just run `pre-commit` locally and you' ll see it's solved.

We do have this new large formatting PR here in order to put all of the formatting together: https://github.com/maykinmedia/open-inwoner/pull/2011

PR #2006, opened 2025-11-04
ci: ignore nodejs.org in linkcheck due to false positives

swrichards:
----
📚 Documentation preview 📚: https://open-inwoner--2006.org.readthedocs.build/en/2006/


swrichards:
Made redundant by #2012

PR #2008, opened 2025-11-04
[#3321] Avoid duplicate entries in sitemap

pi-sigma:
## Summary

Some links (privacy, cookies) showed up both under "Platform links" and (incorrectly) under "Mijn Profiel". The PR fixes this and refactors the code to improve readability.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3321

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/core/tests/test_views.py
+++ b/src/open_inwoner/core/tests/test_views.py
@@ -1,13 +1,33 @@
 from django.contrib.auth.models import AnonymousUser
 from django.test import TestCase
+from django.test.utils import override_settings
+from django.urls import reverse
 from django.utils.crypto import get_random_string
+from django.utils.translation import gettext as _
+
+from cms import api
 
 from open_inwoner.accounts.tests.factories import (
     DigidUserFactory,
+    UserFactory,
     eHerkenningUserFactory,
 )
+from open_inwoner.cms.benefits.cms_apps import SSDApphook
+from open_inwoner.cms.cases.cms_apps import CasesApphook
+from open_inwoner.cms.collaborate.cms_apps import CollaborateApphook
+from open_inwoner.cms.extensions.models import CommonExtension
+from open_inwoner.cms.inbox.cms_apps import InboxApphook
+from open_inwoner.cms.products.cms_apps import ProductsApphook
+from open_inwoner.cms.profile.cms_appconfig import ProfileConfig
+from open_inwoner.cms.profile.cms_apps import ProfileApphook
+from open_inwoner.cms.tests import cms_tools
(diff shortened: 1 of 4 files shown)


src/open_inwoner/core/tests/test_views.py
```diff
@@ -271,3 +291,297 @@ def test_display_category_visibility(self):
...
+
+    def test_sitemap_hides_login_for_authenticated_users(self):
+        self.client.force_login(self.user)
+
+        response = self.client.get(reverse("sitemap"))
+        content = response.content.decode()
+
+        self.assertNotIn(_("Login or create an account"), content)
+
+    from unittest.mock import patch
```
swrichards:
```suggestion
```

src/open_inwoner/core/tests/test_views.py
```diff
@@ -271,3 +291,297 @@ def test_display_category_visibility(self):
...
+@override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
+class SitemapViewTest(TestCase):
+    def setUp(self):
+        self.user = UserFactory()
+
+        # Homepage is required for sitemap
+        create_homepage()
+
+        # Clear default SiteConfiguration
+        SiteConfiguration.objects.all().delete()
```
swrichards: Why is this needed?

src/open_inwoner/core/tests/test_views.py
```diff
@@ -271,3 +291,297 @@ def test_display_category_visibility(self):
                 res = _get_category_data_for_user(root_category, eherkenning_user)
 
                 self.assertEqual(len(res["sub_categories"]), res_company)
+
+
+@override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
+class SitemapViewTest(TestCase):
```
swrichards, line 297: This is comprehensive, which is good, but I am a bit worried that the sitemap logic will remain perenially fragile, because there is no procedural way to register what should be in the sitemap. I don't have a better solution right now, beyond a vague sense that some kind of registration mechanism might be needed. For future discussion!

swrichards (approved):
Just nitpicks, LGTM.

src/open_inwoner/core/tests/test_views.py
pi-sigma: Not at all. Leftover from earlier iteration.

src/open_inwoner/core/tests/test_views.py
pi-sigma: Agreed. Perhaps next time this breaks due to changing requirements we bring it up for refinement

PR #1930, opened 2025-09-22
Apply djlint for templates

swrichards:
## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added 
  - [ ] Breaking changes flagged 
- [ ] Documentation updated 
- [ ] Codecov report reviewed for untested lines
- [ ] Storybook component updated/created 
- [ ] Vitest tests added/updated 
- [ ] Updated the `docker-compose.yml` environment variables
  - [ ] Created an issue in https://github.com/maykinmedia/charts/ for new environment
    variables

PR #2011, opened 2025-11-06
Run prettier on everything except ignore and update proseWrap and semicolon settings

swrichards:
## Summary

This PR changes our prettier and CI config to:

- Run prettier on everything, _except_ ignored files, which is recommended by the prettier docs and lowers the chance of us skipping files (it also ensures consistency in files ignored until now, such as config files and fixtures)
- Runs prettier as part of the `pre-commit` pipeline
- Simplifies the `.prettierignore` to remove the redundancy with `.gitignore`
- Changes the config to add `proseWrap: always`, for consistent markdown formatting
- Changes the config to set `semi: true`, because frankly I find semi-colonless JS ghastly (but YMMV).

## Issue References

Triggered by our i18n files not (always) being formatted.


.pre-commit-config.yaml
```diff
@@ -4,13 +4,18 @@ repos:
     hooks:
       - id: ruff
       - id: ruff-format
```
jiromaykin, line 7: unrelated but wondering: should we add "ruff check --fix" too for ordering imports?

swrichards: `ruff == ruff check`, but I added the `--fix` argument to the config (and I agree, let's do all the cleanups here in one go).

swrichards:
@jiromaykin Laten we deze mergen nu meeste frontend PRs door de pijplijn zijn

.prettierignore
```diff
@@ -1,8 +1,4 @@
...
-static/
-src/open_inwoner/static/
-media/
-docs/
-bin/
-deployment/
-src/open_inwoner/react/i18n/compiled/
+# Templates, which cannot be formatted by prettier due to
+# Django's templating language
+**/templates/**/*.html
```
jiromaykin, line 3: Does this mean we have to decide ourselves how to 'format'/type HTML?

package.json
```diff
@@ -12,10 +12,8 @@
     "bundle": "conc npm:bundle:*",
     "bundle:webpack": "webpack",
     "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
-    "check-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -c",
     "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
     "compilemessages": "formatjs compile-folder --ast src/open_inwoner/react/i18n/locales src/open_inwoner/react/i18n/compiled",
-    "fix-linting": "prettier src/open_inwoner/js src/open_inwoner/scss src/open_inwoner/react src/open_inwoner/webcomponents -w",
```
jiromaykin, line 18: You've removed my baby 😁 but that's okay - i guess this was convenient for developers to be able to run only 1 part of all formatters without having to run the others.

jiromaykin (approved):
I approve of this, despite some questions :-)

package.json
swrichards: You can still do that, and because we specify the ignore the direct command is actually fewer characters!

```
npm run check-linting
npx prettier -c .

npm run fix-linting
npx prettier -w .
```

.prettierignore
swrichards: Prettier doesn't speak django template, but that's the next project! #1930

PR #2016, opened 2025-11-11
⚗️ (experiment) Add web-component (without React) to bundle

jiromaykin:
## Summary

In case where we have very simple OIP web-components without any state,
in situations where the 'pattern' doesn't quite exist in NLDS yet,
like these 'Ga naar' links.

## Issue References

loosely related to https://taiga.maykinmedia.nl/project/open-inwoner/task/3531

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/templates/cms/plugins/links/external-links.html
+++ b/src/open_inwoner/templates/cms/plugins/links/external-links.html
@@ -1,22 +1,9 @@
 {% load i18n icon_tags %}
 
 {% if link_plugin and link_plugin.child_plugin_instances %}
-    <section class="plugin external-links">
-       <h2 class="utrecht-heading-2">{{ link_plugin.title }}</h2>
-       <ul class="card-container card-container--columns-2 external-links__list">
-        {% for link in link_plugin.child_plugin_instances %}
-            <li class="external-links__list-item">
-                <a href="{{ link.get_link }}" class="external-link" {% if link.target %}target="{{ link.target }}" rel="noopener noreferrer" aria-label="{{ link.name_text }} {% trans '(opens in new window)' %}"{% endif %}>
-                    <span class="external-link__custom-icon">
-                    {% icon icon=link.icon outlined=True %}
-                    </span>
-                    <span class="external-link__content">{{ link.name.html|safe }}</span>
-                    <span class="external-link__arrow">
-                    {% icon icon="east" icon_position="after" primary=True outlined=True %}
-                    </span>
-                </a>
-            </li>
-        {% endfor %}
-       </ul>
-    </section>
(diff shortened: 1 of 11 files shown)


src/open_inwoner/webcomponents/components/ExternalLinks.ts
```diff
@@ -0,0 +1,88 @@
+import { LitElement, html, css } from 'lit'
+import { customElement, property } from 'lit/decorators.js'
+import 'src/open_inwoner/static/bundles/open_inwoner-css.css'
```
swrichards: @jiromaykin If I [read the docs correctly](https://vite.dev/guide/features#css-pre-processors), something like this should work (will need `npm add -D sass-embedded`):

```suggestion
import styles from 'src/open_inwoner/scss/components/ExternalLink/ExternalLink.scss?inline'
```

Then in the component attributes:

`static styles = css(styles)` (or perhaps `unsafeCSS`)

jiromaykin: i've moved the SCSS to the components own directory + i had to add more than 1 SCSS file due to Material Icons and shadow-DOM workings, so it becomes styles array.

jiromaykin:
So @swrichards @stefrado this is a bit of an experiment - in case we have designs/components that do not quite exist in NLDS and it is just 'dumb' HTML. 
We could still just write Django-html, or use pure web-components like this. But we'd still need to figure out how we can add another Storybook 'instance' for the web-components directory, and if it makes sense to write front-end tests for something as simple as this.

src/open_inwoner/templates/cms/plugins/links/external-links.html
```diff
@@ -1,22 +1,9 @@
...
-                    </span>
-                </a>
-            </li>
-        {% endfor %}
-       </ul>
-    </section>
+  <external-links
+    id="external-links-{{ link_plugin.id }}"
+    title="{{ link_plugin.title }}"
+    links='{{ links_json|safe }}'
```
jiromaykin, line 7: Should not be safe here.

jiromaykin:
TODO:
- skeleton CSS
- Playwright testing (true connection between back- and front-end)
- Storybook frontend testing
- directory layout frontend/components
- registration/entry structure

General idea:
- everything will eventually be a web-component
- Lit may be clean but wondering whether it's not too heavy, or look more at https://www.webcomponents.org/libraries

stefrado:
@jiromaykin I think we can close this PR, right?

jiromaykin:
> @jiromaykin I think we can close this PR, right?

Yass.

PR #2019, opened 2025-11-11
[#3572] Homogenise colors and styles on homepage (and elsewhere)

jiromaykin:
## Summary

Colors in code need to conform to the ones in Figma for 'call-to-action' buttons and "action" links (they should have the primary color and be bold instead).This is a bigger issue (since we do not have 'primary' styled Buttons-that-look-like-links) but at least this solves the color differences in styles on the Home page - in case a municipality uses a specific color for secondary colors, different from the primary color.
Also all Primary "action" links should be bold.
(like: the "Bewerk" links on Profile page)

## Issue References

- Taiga Issue (with screenshot): https://taiga.maykinmedia.nl/project/open-inwoner/issue/3572

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/scss/components/Button/Button.scss
+++ b/src/open_inwoner/scss/components/Button/Button.scss
@@ -103,6 +103,7 @@
     background-color: transparent;
     border-color: transparent;
     color: var(--color-primary);
+    font-weight: bold;
 
     &.button--primary {
       border-color: transparent;
--- a/src/open_inwoner/templates/pages/cases/list_inner.html
+++ b/src/open_inwoner/templates/pages/cases/list_inner.html
@@ -58,7 +58,7 @@
                             </li>
                             {% list_item text=_("Openstaande zaak") caption=_("Status") compact=True strong=False %}
                         {% endrender_list %}
-                        <span class="link link--icon link--secondary">
+                        <span class="link link--icon link--bold link--primary">
                         <span class="link__text">{% trans "Zaak afronden" %}</span>
                         {% icon icon="arrow_forward" icon_position="after" primary=True outlined=True %}
                     </span>
@@ -88,7 +88,7 @@
                             </li>
                         {% endrender_list %}
 
-                        <span class="link link--icon link--secondary">
+                        <span class="link link--icon link--bold link--primary">
                         <span class="link__text">{{ case.statustype_config.case_link_text|default:"Bekijk zaak" }}</span>
                         {% icon icon="arrow_forward" icon_position="after" primary=True outlined=True %}
                     </span>
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -152,6 +152,8 @@ Bugfixes
   vanuit de eSuite.
 * [:taiga-is:`3590`: :pr:`2030`]: Bug opgelost waarbij tablet gebruikers geen volledig uitgevuld
   navigatie menu zagen.
+* [:taiga-is:`3572`: :pr:`2019`]: De kleuren van call-to-action links en knoppen in tegels komen
+  nu overeen met de primaire kleur zoals die in de Figma designs vastgelegd is.
 * [:taiga-is:`3588`, :pr:`2029`]: Ontbrekende logout URL voor reguliere gebruikers is
   toegoevegd, en de logica voor alle login types is opgeschoond.
 * [:taiga-is:`3599`, :pr:`2045`]: Bij het bekijken van de bron code van componenten in Storybook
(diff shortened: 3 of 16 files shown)


jiromaykin:
@pi-sigma @stefrado At some point between 2021 and 2024 the OIP default colors were 'inverted' in the designs,
but in App.SCSS you can still see that the 'old' static HSL Primary color was actually green instead of blue.
(And vice-versa for the secondary color).

```
  --color-primary-h: 138;
  --color-primary-s: 58%;
  --color-primary-l: 33%;
  --color-primary: hsl(
    var(--color-primary-h),
    var(--color-primary-s),
    var(--color-primary-l)
  );
= HSL(138,58%,33%)
```

The new NLDS components all use the **primary** colors for links and links in cards.

But in OIP we used the **secondary** colors in cards, meaning we're getting two different colors on the Home page (see screenshot for this issue in Taiga).

I actually do not know if _any_ municipality uses a different color in the color-picker for primary versus secondary, but this PR is meant to just make all of those cards and "Buttons-that-look-like-links" appear with the primary color, instead of secondary.

Also they are all Bold in the designs. But the icons are not bold.

So I can really use an extra pair of eyes to see if this looks better now in the PR and is more consistent.

src/open_inwoner/templates/pages/cases/list_inner.html
```diff
@@ -88,7 +88,7 @@ <h2 class="card__heading-2">{{ case.description }}</h2>
                             </li>
                         {% endrender_list %}
 
-                        <span class="link link--icon link--secondary">
+                        <span class="link link--icon link--bold link--bold link--primary">
```
pi-sigma:
```suggestion
                        <span class="link link--icon link--bold link--primary">
```

src/open_inwoner/scss/components/Button/Button.scss
```diff
@@ -129,6 +130,10 @@
     }
   }
 
+  &--bold {
+    font-weight: bold;
+  }
```
pi-sigma: Where is this used? All the buttons in this PR are styled bold via `&--transparent`, which already has `font-weight: bold`. Is it used elsewhere for styling other buttons?

pi-sigma (changes requested):
I note that the footer links still use secondary color. Is this intentional?

src/open_inwoner/scss/components/Button/Button.scss
jiromaykin: Thanks for noticing; this was a leftover from when I added another template-tag but we don't want to add more of those.

jiromaykin:
> I note that the footer links still use secondary color. Is this intentional?

@pi-sigma  Yes, that's intentional because ordinary "non call-to-action" links inside body texts should still have the secondary color (just like the links in product pages etc.) and they should not be bold either (for now). This PR should only be about making the code conform to the Figma design colors.

pi-sigma (approved):

src/open_inwoner/components/templatetags/button_tags.py
```diff
@@ -62,6 +62,7 @@ def button(text, **kwargs):
         - aria_label: string | The HTML aria-label attribute if different from the text.
         - extra_classes: string | Extra classes that need to be added to the button
         - extra_attributes: dict | Extra attributes that need to be added to the button
+        - bold
```
stefrado: A comment would be nice
```suggestion
        - bold: bool | if the button text is bold
```

src/open_inwoner/scss/components/Button/Button.scss
```diff
@@ -103,6 +103,7 @@
     background-color: transparent;
     border-color: transparent;
     color: var(--color-primary);
+    font-weight: bold;
```
stefrado, line 106: Transparent always bold?

stefrado (changes requested):
Small extra comment would be nice... 

---

Also did you forgot this one or is thins on purpose?
Image

src/open_inwoner/components/templatetags/button_tags.py
jiromaykin: oh no's - I should have deleted this, we're not using this.

src/open_inwoner/scss/components/Button/Button.scss
jiromaykin: Yes, this transparent button (that just looks like a link) turned bold in the 'new' design in 2023, and that is the moment when all the colors suddenly swapped... 
At that moment the OIP primary color looked green in the designs and secondary looked blue (which at some unknown point got reversed in Figma).

jiromaykin:
> Also did you forgot this one or is thins on purpose? Image

@stefrado that is on purpose in this PR,
in Figma breadcrumbs still have the secondary color.

stefrado (approved):

PR #2020, opened 2025-11-12
[#3575] Create 'Mijn zaken' CMS plugin

pi-sigma:
## Summary

Created CMS plugin for zaken

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3575

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/cms/plugins/cms_plugins/zaken.py
```diff
@@ -0,0 +1,51 @@
...
+            LoginTypeChoices.digid,
+            LoginTypeChoices.eherkenning,
+            LoginTypeChoices.oidc,
+            LoginTypeChoices.eidas_person_bsn,
+        ):
+            return context
+
+        case_service = CaseListService(context["request"])
+
+        open_submissions: Sequence[UniformCase] = case_service.get_submissions()
```
swrichards: @pi-sigma one thing to keep in mind: this is going to make the landing page terribly slow, because this doesn't just fetch the cases, but also resolves them. I feel this may be a bit of a shock. I see two remediation options:

1. Adding a new `post_login` signal handlers, which seeds the ZGW cache by calling these methods in a Celery task, hopefully meaning some requests will result in cache hits by the time the user ends up on the landing page (but that might be too slow to have any effect).
2. Implementing this as with HTMX logic, so the case loading does not block loading the whole profile page.

We could also check with Bobby in the ticket if he wants this to be part of the story, or a follow-up if people actually complain about hit.

src/open_inwoner/cms/plugins/cms_plugins/zaken.py
```diff
@@ -0,0 +1,51 @@
...
+
+@plugin_pool.register_plugin
+class CMSZakenPlugin(CMSPluginBase):
+    model = CMSZakenPluginConfig
+    name = _("Zaken Plugin")
+    render_template = "cms/plugins/zaken/zaken.html"
+    cache = False
+
+    def render(self, context, instance, placeholder) -> dict[str, str]:
+        if not context["request"].user.login_type in (
```
swrichards: Let's be a bit more defensive here in case it's an anonymous user etc., checking separately for request, user and login_type.

src/open_inwoner/cms/plugins/cms_plugins/zaken.py
pi-sigma: Yes, the zaken page is one of the slowest we have, so embedding this functionality in the homepage is problematic. I'll look into HTMX for this one. We already have that elsewhere, it should be possible to use the same pattern here.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+        case_dicts = [
+            case.process_data()
+            for case in itertools.chain(open_submissions, preprocessed_cases)
+        ]
+        context = {
+            "zaken": case_dicts,
+        }
+
+        return render(request, "cms/plugins/zaken/zaken.html", context)
```
swrichards, line 70: We might want a timeline logger here as well, as we do in the main view where we have:

```
self.log_case_list_accessed(case_dicts)
```

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,39 @@
...
+class ZakenPluginContentView(RequiresHtmxMixin, View):
+    """
+    HTMX endpoint that fetches and returns zaken content for the `CMSZakenPlugin`
+    """
+
+    def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
+        case_service = CaseListService(request)
+
+        open_submissions: Sequence[UniformCase] = case_service.get_submissions()
+        preprocessed_cases: Sequence[UniformCase] = case_service.get_cases()
```
swrichards, line 55: I can't remember what our conclusion was re limiting the number of results?

I am a bit concerned that this feature will be very difficult to use if you have more than, say, 5 case: it'll take long to load and overwhelm the landing page.

@jiromaykin @stefrado I think we briefly discussed this in the weekly but I can't remember the conclusion, but would it be possible if we only load N cases (say 5) and a flag "has_more", so the HTML can indicate that there are more cases than the user can see?

src/open_inwoner/cms/plugins/cms_plugins/zaken.py
```diff
@@ -0,0 +1,42 @@
...
+    render_template = "cms/plugins/zaken/zaken.html"
+    cache = False
+
+    def render(self, context, instance, placeholder) -> dict[str, str]:
+        request = context["request"]
+
+        if getattr(request.user, "login_type", None) not in (
+            LoginTypeChoices.digid,
+            LoginTypeChoices.eherkenning,
+            LoginTypeChoices.oidc,
```
swrichards: OIDC does not necessarily have a `bsn`

swrichards: I would be happy just checking for a non-zero BSN choice here. You can use the new `user.is_bsn_user` property for this.

CHANGELOG.rst
```diff
@@ -67,6 +67,8 @@ Nieuwe features
 * [:taiga-us:`3511`, :taiga-is:`3531`, :pr:`1996`]: Front-end voor CMS-plugin voor links naar
   andere portalen gebouwd, inclusief nieuwe design-tokens volgens NLDS conventies.
 * [:taiga-us:`3509`, :pr:`1997`]: Ondersteuning voor eIDAS login via OIDC is toegevoegd.
+* [:taiga-us:`3575`, :pr:`2020`]: CMS plug-in 'Mijn Zaken' angemaakt voor het insluiten van zaken
```
```suggestion
* [:taiga-us:`3575`, :pr:`2020`]: CMS plug-in 'Mijn Zaken' angemaakt voor het ontsluiten van zaken
```

src/open_inwoner/cms/plugins/cms_plugins/zaken.py
```diff
@@ -0,0 +1,42 @@
...
+
+@plugin_pool.register_plugin
+class CMSZakenPlugin(CMSPluginBase):
+    model = CMSZakenPluginConfig
+    name = _("Zaken Plugin")
+    render_template = "cms/plugins/zaken/zaken.html"
+    cache = False
+
+    def render(self, context, instance, placeholder) -> dict[str, str]:
+        request = context["request"]
```
swrichards: We can add an extra bit of defensiveness here: at least one ZGW API group (some plugins have caused problems there before).

swrichards (changes requested):

src/open_inwoner/cms/plugins/views.py
jiromaykin: Right now none of the plug-ns on the home page have a 'more' styling, just a links pointing to "Bekijk alle zaken",
all of the plugins should look similar, I'd think.
also i don't think we need a Spinner anymore for 3-4 cases on Home? So no HTMX necessary?

swrichards: > also i don't think we need a Spinner anymore for 3-4 cases on Home? So no HTMX necessary?

We can't specify a page size when fetching (the API doesn't support it), so we'll have twenty cases at least. And each of those will have to have their zaak/result type resolved, so that's another 3 requests for each of those. Basically what the Case List does now, and that can take quite some time.

The limiting 3-4 would really just be truncating.

I am aware the plugins currently don't show anything, but in this case it is starting to feel very necessary. I don't think showing _all_ cases is a good solution (Some have 40-50!, rare but it happens). So we could truncate to the first N (even though we fetch more), and then just hope that users don't interpret that as "ah these are _all_ my cases".

That might be the sensible approach here for now, but something to discuss perhaps in the design meeting.

pi-sigma: I think we agreed on 4 for the plugin, don't remember if we reached a conclusion about adding a flag for showing more.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+from open_inwoner.cms.cases.views.mixins import CaseLogMixin
+from open_inwoner.cms.cases.views.services import CaseListService
+from open_inwoner.htmx.mixins import RequiresHtmxMixin
+from open_inwoner.openzaak.types import UniformCase
+
+logger = structlog.stdlib.get_logger(__name__)
+
+
+NUM_ZAKEN_DISPLAY = 4
```
swrichards: Given that we actually have a config model easily accessible (or not? We have a `plugin_id`) we could make the number of cases a plugin config variable.

We could also do that and pass the number of cases in the HTMX call via query parameter.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,47 @@
...
+    def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
+        case_service = CaseListService(request)
+
+        open_submissions: Sequence[UniformCase] = case_service.get_submissions()
+        preprocessed_cases: Sequence[UniformCase] = case_service.get_cases()
+
+        zaken_dicts = [
+            zaak.process_data()
+            for zaak in itertools.islice(
+                itertools.chain(open_submissions, preprocessed_cases), NUM_ZAKEN_DISPLAY
```
swrichards: If we are truncated, I feel it would make sense to first sort the last by date, so we get only the latest submissions from both categories.

Though actually come to think of it: aren't the submissions already part of a different plugin? The open tasks?

swrichards (changes requested):

swrichards:
@pi-sigma one more thought here, triggered by the other PR. Perhaps we can work with a lower num_requests for this plugin? We'd have to make that a parameter on the `CaseListService`, but I suspect it'd be worth it here to have `num_requests=1` so we don't needlessly resolve a whole bunch of cases we have no intention of showing.

The only edge case I can think of is if the first page includes only cases that would be filtered out, whereas the second page has zaken we can actually show. Not quite sure how we'd handle that.

src/open_inwoner/cms/plugins/views.py
pi-sigma: No, those are the "openstaande taken", which are different from "openstaande-inzendingen" (the submissions).

src/open_inwoner/cms/plugins/views.py
pi-sigma: Made it configurable

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,45 @@
...
+        case_service = CaseListService(request)
+
+        open_submissions: Sequence[UniformCase] = case_service.get_submissions()
+        preprocessed_cases: Sequence[UniformCase] = case_service.get_cases()
+
+        zaken_dicts = [
+            zaak.process_data()
+            for zaak in itertools.islice(
+                itertools.chain(open_submissions, preprocessed_cases),
+                int(request.GET.get("num_zaken", "4")),
```
swrichards: Might need an extra guard: could raise a `ValueError` for bad values. I would do:


```
try:
    num_zaken = int(request.GET.get("num_zaken", "4"))
except (ValueError, TypeError):
    logger.exception("Bad value for num_zaken")
    num_zaken = 5
```

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -0,0 +1,46 @@
...
+                        </span>
+                    </div>
+                    <p class="userfeed-card__description">
+                        <span class="status">{{ zaak.identification }}</span>
+                    </p>
+                </a>
+            {% endfor %}
+        </div>
+    {% else %}
+        <p class="userfeed-card__description">{% trans "Geen zaken gevonden." %}</p>
```
swrichards:
```suggestion
        <p class="userfeed-card__description">{% trans "U heeft op dit moment geen lopende zaken." %}</p>
```

src/open_inwoner/cms/plugins/views.py
swrichards: Also move this check to the top, before we do the fetching (to avoid DDOS attack surfaces)

swrichards (changes requested):

src/open_inwoner/cms/plugins/models/zaken.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class CMSZakenPluginConfig(CMSPlugin):
+    title = models.CharField(
+        _("Title"),
+        max_length=250,
+        help_text=_("The title of the zaken plugin block"),
+        default=_("Mijn Zaken"),
+    )
+    num_zaken = models.IntegerField(
```
swrichards, line 18: I think it makes sense to enforce a max here (say 10), and to also enforce that max when we get the query parameter. We would store the default as `MAX_CASES_DEFAULT` here so we could import it in the plugin handler.

src/open_inwoner/cms/plugins/views.py
pi-sigma: Good point, I forgot that although the value for `num_zaken` is validated by the model, it becomes subject to manipulation because it's passed as query string.

I think it's also worth adding validation for min and max value to the model. Less than 1 doesn't make sense, and more than 100 should also be out of the question for a homepage plugin.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,70 @@
...
+                num_zaken = MAX_CASES_DEFAULT
+        except (ValueError, TypeError) as exc:
+            client_ip, _ = get_client_ip(request)
+            logger.warning(
+                "Invalid num_zaken parameter, using default",
+                value=request.GET.get("num_zaken"),
+                error=str(exc),
+                client_ip=client_ip,
+            )
+            num_zaken = MAX_CASES_DEFAULT
```
swrichards, line 52: Let's move this to the top of the function, so we don't trigger any requests if the input is invalid.

I think you can skip the IP lookup: with django structlog, there will be a request_id bound to the logger, which also has remote ip and such things.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,70 @@
...
+                    "Invalid num_zaken value, using default",
+                    num_zaken=num_zaken,
+                    client_ip=client_ip,
+                )
+                num_zaken = MAX_CASES_DEFAULT
+        except (ValueError, TypeError) as exc:
+            client_ip, _ = get_client_ip(request)
+            logger.warning(
+                "Invalid num_zaken parameter, using default",
+                value=request.GET.get("num_zaken"),
```
swrichards:
```suggestion
                value=request.GET.get("num_zaken", None),
```

And maybe move this to the top too. We want to separate out the "absent" case from the "bad value" case.

swrichards (approved):
LGTM, just a nitpick to address

src/open_inwoner/cms/plugins/views.py
swrichards: I think we need this change because it could still raise an exception if it's missing, or better yet:


```suggestion
                value=num_zaken,
```

src/open_inwoner/cms/plugins/views.py
```diff
@@ -0,0 +1,66 @@
...
+class ZakenPluginContentView(RequiresHtmxMixin, CaseLogMixin, View):
+    """
+    HTMX endpoint that fetches and returns zaken content for the `CMSZakenPlugin`.
+
+    The view accepts and validates a `num_zaken` query parameter to limit the number
+    of cases displayed.
+    """
+
+    def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
+        case_service = CaseListService(request)
```
swrichards, line 35: We could also use the zero api groups check here, just so no 500s can be triggered.

swrichards (approved):
Sorry, two more nitpicks, I'm a bit slow today 🙈

swrichards (approved):

PR #1797, opened 2025-06-23
Almelo #53 -- DONT MERGE: Workarounds for Almelo PoC

alextreme:

PR #2021, opened 2025-11-12
[#2175] Move to pypi openklant client

### LINKED ISSUE #2175: Migrate from internal openklant api to the packaged version
See #2021 -- this would be a good moment to use the common API.

swrichards:
## Summary

Replaced vendored openklant2 client with pypi version

## Issue References

Closes #2175 

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added


swrichards:
@pi-sigma This might also be needed, apparently we don't have coverage for this: https://github.com/maykinmedia/open-inwoner/pull/1797/commits/e919d8cdefbfaf2834e7dd16de7ceec567752a30

pi-sigma:
@swrichards Can you check the branch protection rules and remove the openklant check?

swrichards:
@pi-sigma done!

requirements/base.txt
```diff
@@ -454,6 +454,8 @@ objects-api-client-django==0.5.0
     # via -r requirements/base.in
 odfpy==1.4.1
     # via tablib
+open-klant-client==0.1.0
```
swrichards: We're up to `0.4.0`

src/open_inwoner/openklant/tests/test_openklant2_service.py
swrichards, line 1: I am sad to see the VCR tests go -- I think they added a lot of resilience, but I suspect that was a bit too difficult to lift out in a clean way @pi-sigma ?

swrichards (commented):
@pi-sigma I think this is fine to get us going but I would like a quick chat regarding replacing the VCR tests with mocks, I am wondering whether we can't retain some of that functionality in-repo.

requirements/base.txt
swrichards: Upgraded.

src/open_inwoner/openklant/tests/test_openklant2_service.py
swrichards: As discussed, this is really the responsibility of the library and we should be able to rely on the types as a library consumer. The VCR tests might give extra security, but that'd be more of an end-to-end test, which we might re-introduce later.

PR #2028, opened 2025-11-17
[#3589] Check for existing permissions when reverting migration

pi-sigma:
## Summary

The function for reverting the migration which introduced 'Permission' instances for editing parts of the 'SiteConfiguration' failed to check if Permission instances exist.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/issue/3589

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/configurations/migrations/0085_create_partial_admin_edit_permissions.py
+++ b/src/open_inwoner/configurations/migrations/0085_create_partial_admin_edit_permissions.py
@@ -2,6 +2,10 @@
 
 from django.db import migrations
 
+import structlog
+
+logger = structlog.stdlib.get_logger(__name__)
+
 
 def create_partial_admin_edit_permissions(apps, _):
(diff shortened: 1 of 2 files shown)


src/open_inwoner/configurations/migrations/0085_create_partial_admin_edit_permissions.py
```diff
@@ -34,7 +38,14 @@ def delete_partial_admin_edit_permissions(apps, _):
         "siteconfig_fieldset_page_texts",
         "siteconfig_fieldset_help_texts",
     ):
-        Permission.objects.get(codename=codename).delete()
+        try:
+            Permission.objects.get(codename=codename).delete()
```
swrichards:
```suggestion
            Permission.objects.filter(codename=codename).delete()
```

Would not require the catch. I think it returns the number of deleted rows, you could print that.

swrichards (approved):

PR #2029, opened 2025-11-17
fix: properly match the logout url to the login types

swrichards:
## Summary

Inadvertently broke the logout url during the EIDAS work. Ensured that we always return the default, and handles the special cases in a cleaner way.

## Issue References

- Taiga User Story:
- Taiga Issue: [#3588](https://taiga.maykinmedia.nl/project/open-inwoner/issue/3588)
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/accounts/models.py
+++ b/src/open_inwoner/accounts/models.py
@@ -1,5 +1,6 @@
 import os
 from datetime import timedelta
+from typing import Any, assert_never, cast
 from uuid import uuid4
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


jiromaykin (approved):
Now it "works on my machine" :-)

src/open_inwoner/accounts/models.py
```diff
@@ -678,23 +678,21 @@ def require_necessary_fields(self) -> bool:
...
+                    else reverse("digid:logout")
+                )
+            case LoginTypeChoices.eherkenning:
                 return reverse("eherkenning_oidc:logout")
-            return reverse("logout")
-        elif self.is_eidas_user:
-            if OpenIDEIDASConfig.get_solo().enabled:
+            case LoginTypeChoices.oidc:
+                return reverse("oidc_logout")
+            case _ if self.is_eidas_user:
```
```suggestion
            case LoginTypeChoices.eidas_person_bsn | LoginTypeChoices.eidas_person_pseudo_id | LoginTypeChoices.eidas_company
```

a bit unusual to have two catch-all cases, one with a guard clause.

pi-sigma (approved):

src/open_inwoner/accounts/models.py
swrichards: Good point, and generally we should aim for exhaustiveness checking when using enums/textchoices. Your change makes that clearer, I added an `assert_never` as the fallthrough so each case has to be explicitly handled.

PR #2032, opened 2025-11-18
Add admin visibility and retryability to ZGW webhook notifications

### LINKED ISSUE #1900: Improve admin observability of webhook handling
Some problems with the current webhook handling:

- You can only really tell if it's working by checking the Celery tasks (indicating a `process_zaken_notification` task result) and indirectly through the Timeline logger (it might list reasons for _not_ successfully processing a notification, e.g. missing role) or the Yubin mail log (mail sent as a consequence of a webhook).
- You cannot easily tell to which subscription a Celery task result for a notification belongs, which is painful if you're trying to figure out if any have gone stale.
- You cannot at all see any sign in the admin for webhooks being received from stale subscriptions, because they'll be refused at the HTTP handler level, but not logged anywhere (the Celery task is only spawned for _accepted_ notifications).

A lot of this could be improved by logging more to the Timeline logger for diagnostics, but that would significantly increase the volume, so #1899 would be needed first.

swrichards:
This PR implements a big refactor of our ZGW webhook notification handling. Currently, the webhook handler immediately triggers a Celery task to process a notification. There are a few issues with that approach:

1. When configuring a system, it's hard to see if notifications are working and if not, why not, without consulting Celery logs which are often sparse
2. You cannot retry notifications notification. This is an issue if you know failures are transient (e.g. due to a temporary backend system being down), in which case you might want to retry notifications in the last N hours. It's also important for migrating a live system.
3. No clear trace of incoming webhooks or their request bodies (again, unless you go digging through logs). Making it hard to triage and debug.
4. No clear success/fail indicators. This is mixed currently: certain failures are pushed to the system log (e.g. bad credentials), but can then absolutely flood the system log with a firehose of not-very-relevant spam (e.g. for a misfiring subscription).

This PR changes the mechanism. Incoming webhooks are first stored in a Django model, which persists the notification metadata and content, and also tracks the processing state. The Celery task works on these entries, rather than being fired directly with the notification body from Celery. This allows us to expose all incoming webhooks to the admin panel, and to provide some tools for manual retrying and such.

The complexity lies in the concurrency management (ensuring each notification is handled serially, we use `select_for_update` for this), and ensuring that data is pruned, as the table can fill up quickly.


src/open_inwoner/openzaak/tasks.py
```diff
@@ -26,7 +37,55 @@ def import_zgw_data():
...
+    except NotificationRecord.DoesNotExist:
+        logger.error(
+            "Attempted to process unknown Notification Record",
+            record_pk=record_pk,
+        )
+        raise
+    except (NotificationRecordLockError, NotificationAlreadyProcessedError) as e:
+        # Already locked or processed - don't count as failure (idempotent/concurrent access)
+        logger.info(str(e), record_pk=record_pk)
+        raise
```
pi-sigma, line 93: Why `raise` instead of `return` when you don't want to count it as failure? Will it not be classified as failed task if we raise?

src/open_inwoner/openzaak/api/views.py
```diff
@@ -26,10 +33,28 @@ class NotificationsWebhookBaseView(WebhookLogMixin, APIView):
...
+    def handle_notification(self, notification: NotificationRecord) -> None:
         raise NotImplementedError("extend and override this method")
 
     def post(self, request):
+        # Check payload size before processing
+        config = NotificationProcessingConfig.get_solo()
+        max_payload_size = config.max_payload_size
+
+        content_length = request.META.get("CONTENT_LENGTH")
+        if content_length and int(content_length) > max_payload_size:
```
pi-sigma: Invalid `content_length` header will raise an exception here. Perhaps `try/except (ValueError, TypeError)`

src/open_inwoner/openzaak/api/views.py
```diff
@@ -40,30 +65,98 @@ def post(self, request):
...
+                is_valid=False,
+            )
+        except ParseError:
+            # We don't store invalid JSON
+            return Response(
+                {"detail": "payload is not valid json"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        except Exception:
+            return Response(
```
pi-sigma, line 87: Add logging

src/open_inwoner/openzaak/api/views.py
```diff
@@ -40,30 +65,98 @@ def post(self, request):
                 status=status.HTTP_401_UNAUTHORIZED,
             )
 
+        # Create initial log, so we have a record if the incoming webhook
```
pi-sigma: Incomplete comment

src/notifications/query.py
```diff
@@ -1,7 +1,174 @@
...
+
+
+class NotificationRecordManager(models.Manager):
+    def prune_old_records(
+        self, retention_days: int, *, dry_run: bool = False
+    ) -> PruneResult:
+        """
+        Prune notification records older than the specified retention period.
+
+        Only deletes records in final states (SUCCESS, FAILED, SKIPPED).
```
pi-sigma: If a worker picks up a task which gets status `PROCESSING`  but the worker is killed before finishing? Are these tasks automatically pruned? If not, we should maybe have an extra command to clean up tasks with this status (there might be a few of them over time).

src/open_inwoner/openzaak/notifications.py
```diff
@@ -314,7 +313,7 @@ def _handle_zaakinformatieobject_update(
         return
 
     # let's not spam the users
-    period = timedelta(seconds=settings.ZGW_LIMIT_NOTIFICATIONS_FREQUENCY)
+    period = timedelta(seconds=OpenZaakConfig.get_solo().notification_frequency_limit)
```
pi-sigma, line 312: Now we're doing a DB call for every user. The `OpenZaakConfig` is already fetched at L.109 and could be passed as argument to `_handle_zaakinformatieobject_update`

pi-sigma (changes requested):

src/open_inwoner/openzaak/tasks.py
swrichards: It will be classed as a _Celery task_ failure, which is what we want (so it's visible in the tasks overview), but _not_ in the actual notification processing logs (because we didn't actually attempt to process it). Raising is a way to keep visibility on the degree of contention.

src/open_inwoner/openzaak/api/views.py
swrichards: Changed this, and also handled the case where content length is missing (e.g. due to chunked requests).

src/open_inwoner/openzaak/notifications.py
swrichards: Given that this happens async with no further Celery task splitting, I'd say the overhead is acceptable (it'll also likely be cached).

src/notifications/query.py
swrichards: This was a very good point. I added an extra setting to track this, and the prune task will now also remove in-flight tasks that are older than the configured setting.

src/notifications/admin.py
```diff
@@ -1,16 +1,131 @@
...
+        "id",
+        "subscription",
+        "payload",
+        "kanaal",
+        "received_at",
+        "updated_at",
+        "last_processed_at",
+        "processing_error",
+    )
+    fields = (
```
pi-sigma, line 65: `process_started_at` would be useful to have, it's the only indicator of slow/potentially stuck tasks.

src/notifications/admin.py
```diff
@@ -1,16 +1,131 @@
...
+    )
+    readonly_fields = (
+        "id",
+        "subscription",
+        "payload",
+        "kanaal",
+        "received_at",
+        "updated_at",
+        "last_processed_at",
+        "processing_error",
```
pi-sigma, line 62: `processing_output` is not in `readonly_fields`, which means an admin can manually temper with it. Should be added here?

src/notifications/admin.py
```diff
@@ -1,16 +1,131 @@
...
+                .filter(status__in=ProcessingStatus.retryable_statuses())
+                .only("pk", "status")
+            )
+
+            # Get PKs before update
+            record_pks = list(retryable.values_list("pk", flat=True))
+
+            # Bulk update status to PENDING
+            count = retryable.update(
+                status=ProcessingStatus.PENDING, processing_error=""
```
pi-sigma: Should we also clear `processing_output` (i.e. `processing_output=""`)? If notifications are disabled first, the output is "Case notifications are disabled in configuration". An admin manually runs the task, it succeeds, but `processing_output` is not updated, hence in the admin it will read "Case notifications are disabled in configuration" even when the task has now run and succeeded?

pi-sigma (approved):
Couple minor suggestions, otherwise LGTM

src/notifications/admin.py
swrichards: Good point

PR #2033, opened 2025-11-19
[#3580] Make search optional

pi-sigma:
## Summary

The search functionality is made optional via `SiteConfiguration.search_enabled`

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3580

## Checklist

- [x] CHANGELOG.rst updated
- [x] Documentation updated


pi-sigma:
@swrichards I've renamed the "Zoekfilter opties" section under "Configuratie" to "Zoeken", which now includes not only the filters but also the enable/disable option, both in the admin and in the documentation. There is one more search-related setting: `hide_search_from_anonymous_users`, which is currently under "Display options for anonymous users". Perhaps this could also be moved to "Zoeken", then all options related to search would be grouped together.

Perhaps this goes beyond the scope of the PR though and needs discussion.

src/open_inwoner/components/templatetags/header_tags.py
```diff
@@ -19,3 +19,30 @@ def accessibility_header(request, **kwargs):
...
+    """
+    request = context.get("request")
+    config = SiteConfiguration.get_solo()
+
+    if not config.search_enabled:
+        return False
+
+    if request.user.is_authenticated:
+        cms_apps = context.get("cms_apps", {})
+        return bool(cms_apps.get("products"))
```
swrichards: This is because we index these presumably: will the indexing break if this is not configured? If so, I'd be more inclined to move this check to the search index logic: if products is not enabled, we shouldn't return the results. This feels like an out-of-place check here.

It's also a bit strange authenticated users could then potentially search for less than anonymous users?

src/open_inwoner/components/templatetags/header_tags.py
```diff
@@ -19,3 +19,30 @@ def accessibility_header(request, **kwargs):
     config = SiteConfiguration.get_solo()
     kwargs["help_text"] = config.get_help_text(request)
     return {**kwargs, "request": request}
+
+
+@register.simple_tag(takes_context=True)
+def display_search(context):
```
swrichards: Let's underline in the naming it can differ per user.

```suggestion
def display_search_for_user(context):
```

swrichards (changes requested):
Mostly LGTM:

1. I agree the anonymous search function makes more sense here, but let's check with Bobby too on Monday
2. A bit unsure about the products thing and whether it belongs here, but I might be missing something

src/open_inwoner/components/templatetags/header_tags.py
pi-sigma: You're right about my thinking here, but that was wrong. I removed it

docs/beheerhandleiding/12_configuratie.rst
swrichards, line 1: Do we need to move any of the descriptions for the search options between sections, given that we've moved them in the admin?

swrichards (approved):

docs/beheerhandleiding/12_configuratie.rst
pi-sigma: Not really, I thought it would make the code more accessible. Is it a bad idea because it distorts `git blame`?

swrichards: I meant that something like `search_help_text` is described in https://docs.openinwoner.nl/en/latest/beheerhandleiding/12_configuratie.html#helpteksten (the "helpteksten" section), but would now be in the "search" section. It 'd be good to make sure the moved settings are listed under the correct section.

pi-sigma: I overlooked that part of the documentation, I'll update it accordingly

swrichards (approved):

PR #2034, opened 2025-11-20
✨ Add front-end for Zaken plug-in

jiromaykin:
## Summary

Adding case-cards + grid for Zaken plug-in with webcomponent on Home page.
Figma variant "A": https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=7243-21103&t=tVS77KLAzH9LoV69-1

Some NLDS WCAG recommendations: do not make the Card a link: https://github.com/orgs/nl-design-system/discussions/305 but use CSS to cover the Card-as-link.

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3576

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/scss/components/Button/Button.scss
+++ b/src/open_inwoner/scss/components/Button/Button.scss
@@ -1,4 +1,4 @@
-@import '~@utrecht/button-css/dist/index.css';
+@import '@utrecht/button-css/dist/index.css';
 
 .button {
   align-items: center;
--- a/src/open_inwoner/scss/views/_view.scss
+++ b/src/open_inwoner/scss/views/_view.scss
@@ -1,4 +1,10 @@
 .view {
+  &--cases-index {
+    // prevent large content shift after spinner disappears
+    .loader-container {
+      height: 65vh;
+    }
+  }
   &---pages-root {
     &:has(.notifications .notification) {
       .notifications {
--- a/src/open_inwoner/templates/cms/plugins/zaken/zaken.html
+++ b/src/open_inwoner/templates/cms/plugins/zaken/zaken.html
@@ -1,4 +1,5 @@
 {% load i18n icon_tags button_tags django_htmx %}
+
 <!-- start zaken plugin -->
 {% if show_zaken_plugin %}
     {% if request.htmx %}
@@ -10,7 +11,6 @@
                     identificatie="{{ zaak.identification }}"
                     detail-url="{{ zaak.url }}"
                     >
-                    <div slot="continue-section"></div>
                 </oip-zaken-plugin-zaak-item>
             {% endfor %}
         </oip-zaken-plugin-container>
@@ -18,11 +18,17 @@
         <oip-homepage-plugin-section
             title="{{ plugin_title }}"
             next-url="{{ mijn_zaken_url }}"
-            next-url-label="Bekijk alle zaken">
-                <div hx-get="{{ hx_get_url }}">
-                    <div>Laden...</div>
-                </div>
+            show-indicator="false"
+            next-url-label="{% trans 'Bekijk alle zaken' %}">
+            <div hx-get="{{ hx_get_url }}"
+                 hx-trigger="load"
+                 hx-swap="outerHTML">
+                <oip-loading-spinner loading-text="{% trans 'Zaken laden...' %}"></oip-loading-spinner>
+            </div>
         </oip-homepage-plugin-section>
     {% endif %}
+    <!-- end zaken plugin -->
+{% else %}
+    <!-- Unable to render Mijn Zaken Plugin due to a configuration issue-->
 {% endif %}
 <!-- end zaken plugin -->
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -84,6 +84,8 @@ Nieuwe features
   ontsluiten van zaken in de startpagina.
 * [:taiga-us:`3574`, :pr:`2026`]: Aantal requests aan zaken API's configureerbaar gemaakt via
   omgevingsvariabele.
+* [:taiga-us:`3576`, :pr:`2034`]: Front-end voor CMS plug-in 'Mijn Zaken' gebouwd met design-tokens
+  volgens het NLDS principe.
 * [:taiga-us:`3596`, :pr:`2036`]: Hints (zoals ``id_token_hint`` en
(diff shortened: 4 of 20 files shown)


src/open_inwoner/templates/cms/plugins/zaken/zaken.html
swrichards, line 33:
```suggestion
{{ zaken|json:'zaken-items' }}
<mijn-zaken-plugin htmx={{ request.htmx }} zaken-json-script-id="zaken-items"  />
```

src/open_inwoner/cms/plugins/views.py
```diff
@@ -61,8 +62,26 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
...
+                "description": zaak.get("naam", ""),  # Using 'naam' based on your log
+                "case_type": zaak.get("case_type", ""),
+                "url": zaak.get("url") or zaak.get("vervolg_link", ""),
+            }
+            cases_for_component.append(case_data)
+
         context = {
-            "zaken": zaken_dicts,
+            "zaken": zaken_dicts,  # For backward compatibility
+            "cases_json": json.dumps(cases_for_component),
```
swrichards:
```suggestion
```

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
swrichards: @jiromaykin besproken dat @swrichards deze apart implementeert zodat daarop gerebased kan worden.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -61,8 +62,26 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
...
+            # Based on data structure
+            case_data = {
+                "uuid": zaak.get("uuid", ""),
+                "api_group_id": str(zaak.get("api_group", {}).id)
+                if zaak.get("api_group")
+                else "",
+                "identification": zaak.get("identification", ""),
+                "description": zaak.get("naam", ""),  # Using 'naam' based on your log
+                "case_type": zaak.get("case_type", ""),
+                "url": zaak.get("url") or zaak.get("vervolg_link", ""),
```
swrichards: Should not be defensive here at the level of _attributes_: if necessary attribures are missing, we need to log to Sentry, but we'll use `try/except` to exclude it from `cases_for_component`.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -61,8 +62,26 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
...
+            }
+            cases_for_component.append(case_data)
+
         context = {
-            "zaken": zaken_dicts,
+            "zaken": zaken_dicts,  # For backward compatibility
+            "cases_json": json.dumps(cases_for_component),
         }
```
swrichards, line 120:
```suggestion
context = {
            "zaken": cases_for_component,
            "errorMessage": error_message or None
}
```

src/open_inwoner/cms/plugins/views.py
swrichards: Defensive try/except: plugins should not have the power to drag down the whole page.

src/open_inwoner/cms/plugins/views.py
swrichards:
```
{
   "data": dict | None,
    "ok": True | False,
    "errorMessage": str | None
}
```

src/open_inwoner/cms/plugins/cms_plugins/zaken.py
```diff
@@ -19,7 +19,51 @@ class CMSZakenPlugin(CMSPluginBase):
...
+                "data": {
+                    "cmsZakenPluginInstance": {
+                        "pk": str,     # Plugin instance ID (for unique container ID)
+                        "title": str,  # Display title for the plugin header
+                    },
+
+                    "ok": bool,  # Success indicator (True if config exists and user is authenticated)
+
+                    "hxget": str,  # URL for HTMX to fetch zaken data from
+                        # Format: "/cms-plugins/zaken/{plugin_id}/content/?num_zaken={num}"
```
jiromaykin: is that the correct HTMX url? (it's not working for me yet)

swrichards: Should be. It gives a 404?

jiromaykin: 500: GET http://localhost:8000/cms-plugins/zaken/429/content/?num_zaken 500 (Internal Server Error)

some evil chatprogram tells me may be a mistake in the CMS plugin Views at the  `"api_group_id"` which may need to be `zaak["api_group"]` but i am not sure.

swrichards (approved):
🎊

src/open_inwoner/scss/components/Button/Button.scss
```diff
@@ -1,4 +1,4 @@
-@import '~@utrecht/button-css/dist/index.css';
+@import '@utrecht/button-css/dist/index.css';
```
stefrado: Does this still work with webpack?

src/open_inwoner/scss/views/_view.scss
```diff
@@ -1,4 +1,10 @@
 .view {
+  &--cases-index {
+    // prevent large content shift after spinner disappears
+    .loader-container {
+      height: 65vh;
```
stefrado, line 5: Maybe i like it more when the height is are large as the spinner, bc we have a spinner the layout shift minimal.

Also i noticed that the amount of cards is manually set. We can do something with the cards and skeletons based on this number. But thats a nice to have.

stefrado (commented):
Just a suggestion - very good. Now let's rebuild this with the new structure😸

src/open_inwoner/scss/views/_view.scss
jiromaykin: I personally also would like it more when the Spinner is not super-large. But I remember we had to make this Spinner on very large on Zaken listview only because 1 gemeente did not like they could see a scrollbar appear when number of zaken is larger than 9 and was finally loaded. (When loading there was of course no scrollbar).

With "manually set" you mean the plug-in on Home itself?

PR #2042, opened 2025-11-24
Refactor CMS zaken plugin for use with web component

pi-sigma:
## Summary

`CMSZakenPlugin` and `ZakenPluginContentView` are refactored to return data suitable for consumption by a web component 

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3575

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -80,9 +80,8 @@ Nieuwe features
   voor links naar andere portalen gebouwd, inclusief nieuwe design-tokens volgens NLDS conventies. Opmaak is
   verbeterd voor inhoud met langere linkteksten die naar een nieuwe regel worden afgebroken.
 * [:taiga-us:`3509`, :pr:`1997`]: Ondersteuning voor eIDAS login via OIDC is toegevoegd.
-* [:taiga-us:`3575`, :pr:`2020`, :pr:`2037`]: CMS plug-in 'Mijn Zaken' angemaakt voor het
-  ontsluiten van zaken in de startpagina
-
+* [:taiga-us:`3575`, :pr:`2020`, :pr:`2037`, :pr:`2042`]: CMS plug-in 'Mijn Zaken' aangemaakt voor het
+  ontsluiten van zaken in de startpagina.
 * [:taiga-us:`3574`, :pr:`2026`]: Aantal requests aan zaken API's configureerbaar gemaakt via
   omgevingsvariabele.
 * [:taiga-us:`3596`, :pr:`2036`]: Hints (zoals ``id_token_hint`` en
--- a/src/open_inwoner/cms/plugins/views.py
+++ b/src/open_inwoner/cms/plugins/views.py
@@ -3,15 +3,17 @@ from typing import Sequence
 
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import render
+from django.urls import reverse
+from django.utils.translation import gettext as _
 from django.views import View
 
 import structlog
+from furl import furl
 
 from open_inwoner.cms.cases.views.mixins import CaseLogMixin
 from open_inwoner.cms.cases.views.services import CaseListService
 from open_inwoner.cms.plugins.models.zaken import MAX_CASES_DEFAULT, MIN_CASES
 from open_inwoner.htmx.mixins import RequiresHtmxMixin
-from open_inwoner.openzaak.models import ZGWApiGroupConfig
 from open_inwoner.openzaak.types import UniformCase
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -19,27 +21,19 @@ logger = structlog.stdlib.get_logger(__name__)
 
 class ZakenPluginContentView(RequiresHtmxMixin, CaseLogMixin, View):
     """
-    HTMX endpoint that fetches and returns zaken content for the `CMSZakenPlugin`.
-
-    The view accepts and validates a `num_zaken` query parameter to limit the number
-    of cases displayed.
+    HTMX endpoint that fetches and returns content for the `CMSZakenPlugin`.
     """
 
     def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
-        if not ZGWApiGroupConfig.objects.exists():
-            logger.warning(
-                "ZGWApiGroupConfig required to fetch zaken for CMS zaken plugin"
-            )
-            return render(request, "cms/plugins/zaken/zaken.html", {})
-
         case_service = CaseListService(request)
 
-        num_zaken = request.GET.get("num_zaken", None)
+        # determine no. of zaken to be displayed
+        num_zaken = request.GET.get("num_zaken", 0)
         try:
             num_zaken = int(num_zaken)
             if num_zaken < MIN_CASES or num_zaken > MAX_CASES_DEFAULT:
                 logger.warning(
-                    "Invalid num_zaken value, using default",
+                    "Missing or invalid num_zaken value, using default",
                     num_zaken=num_zaken,
                 )
                 num_zaken = MAX_CASES_DEFAULT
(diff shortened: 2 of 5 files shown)


src/open_inwoner/cms/plugins/views.py
```diff
@@ -21,19 +24,56 @@ class ZakenPluginContentView(RequiresHtmxMixin, CaseLogMixin, View):
...
+                # None: No issues
+                # String: User-facing error/warning message
+                # Can be present when ok=True (partial success) or ok=False (failure)
+        }
+
+    Response States:
+        Success with results:     ok=True,  data=[...],  message=None
+        Success, no results:      ok=True,  data=[],     message=None
+        Partial success:          ok=True,  data=[...],  message="Some results missing..."
+        Complete failure:         ok=False, data=None,   message="Technical difficulties..."
```
swrichards: Generally, I think `data` will always be a `dict`, so in this case it would be something like: `data={zaken: [...]`

src/open_inwoner/cms/plugins/views.py
```diff
@@ -51,8 +91,20 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
             )
             num_zaken = MAX_CASES_DEFAULT
 
-        open_submissions: Sequence[UniformCase] = case_service.get_submissions()
-        preprocessed_cases: Sequence[UniformCase] = case_service.get_cases()
+        # fetch zaken + openstaande inzendingen
+        try:
+            open_submissions: Sequence[UniformCase] = case_service.get_submissions()
+            preprocessed_cases: Sequence[UniformCase] = case_service.get_cases()
+        except Exception:
```
swrichards, line 66: We should wrap both separately, so that if one fails, the other might succeed (if we're serious about partial success)

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,30 +1,11 @@
...
-                    </p>
-                </a>
-            {% endfor %}
-        </div>
-    {% else %}
-        <p class="userfeed-card__description">{% trans "U heeft op dit moment geen lopende zaken." %}</p>
-    {% endif %}
+    {# web component #}
+    <zaken-list
+        data="{{ data|json }}"
```
swrichards: This guy goes into a separate JSON script tag:

```
data|json:'element-id-of-the-script-tag-that-this-creates`
```

and the attribute then becomes:

```
<zaken-list component-data-id="element-id-of-the-script-tag-that-this-creates" ...
```

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,30 +1,11 @@
...
-                        <span class="status">{{ zaak.identification }}</span>
-                    </p>
-                </a>
-            {% endfor %}
-        </div>
-    {% else %}
-        <p class="userfeed-card__description">{% trans "U heeft op dit moment geen lopende zaken." %}</p>
-    {% endif %}
+    {# web component #}
+    <zaken-list
```
swrichards: We'll distinguish inner and outer
```suggestion
    <zaken-list-inner
```

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,30 +1,11 @@
...
+        data="{{ data|json }}"
+        ok="{{ ok|lower }}"
+        message="{{ message }}">
+    </zaken-list>
 {% else %}
     {% if instance %}
```
swrichards: Let's also invoke the outer version of the web-component here.

swrichards (commented):
Looks good, this tracks what we've discussed, just a few small comments.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,46 +1,30 @@
...
-                    </div>
-                </div>
-            </div>
-        </section>
+        </oip-homepage-plugin-section>
     {% endif %}
+{% else %}
+    {# Missing configuration or user not authorized #}
+    {# Leave a DOM trace for debugging purposes #}
+    <!-- start zaken plugin --><!-- end zaken plugin -->
```
swrichards: This guy would go at the top and bottom of the template: we'd want to render these always, regardless what happens in between (if there was _nothing_ in between, we'd know we tried to render, but hit this branch. I guess we can then ditch the `{% else %}`

swrichards (commented):
Looks good!

Let's leave it open for now so @jiromaykin can rebase on this and give us feedback.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,46 +1,30 @@
...
     {% else %}
-        <p class="userfeed-card__description">{% trans "U heeft op dit moment geen lopende zaken." %}</p>
-    {% endif %}
-{% else %}
-    {% if instance %}
-        <section class="plugin userfeed">
-            <header class="userfeed__header">
-                <div class="heading-2__indicator">
-                    <h2 class="utrecht-heading-2">{{ instance.title }}</h2>
+        <oip-homepage-plugin-section
```
jiromaykin, line 18: aksuwally the entire Section should be wrapping the plugin (because we have a parent-grid that defines the spacing between these sections / spacing should be defined by parents, not children) and then the re-usable oip-homepage-plugin-section should contain a re-usable oip-homepage-plugin-Header-Group or sometning, but that's my job in my PR.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -1,46 +1,30 @@
...
+            title="{{ plugin_title }}"
+            next-url="{{ mijn_zaken_url }}"
+            next-url-label="Bekijk alle zaken">
+                <div hx-get="{{ hx_get_url }}">
```
jiromaykin, line 22: This won't do anything on pageload without an HTMX trigger, 
like `hx-trigger="load"` and `hx-swap="outerHTML"`
but i've already added that to my PR that i have rebased on this one, so changes will just conflict.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -51,22 +45,81 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
...
+        # format data for web component
+        cases_for_component = []
+        for zaak in zaken_dicts:
+            try:
+                f_url = furl(
+                    reverse(
+                        "cases:case_detail",
+                        kwargs={
+                            "object_id": zaak["uuid"],
+                            "api_group_id": zaak["api_group_id"],
```
jiromaykin: I think this breaks it for me (500 error) so the HTMX url's seam to break also, a "api_group_id": zaak["api_group"], conflict?

jiromaykin: when i log the Zaak keys, i get: ['identification', 'url', 'uuid', 'naam', 'vervolg_link', 'datum_laatste_wijziging', 'eind_datum_geldigheid', 'case_type', 'api_group']

So: no `api_group_id`

jiromaykin (changes requested):

src/open_inwoner/cms/plugins/views.py
swrichards: Try changing this line to: ` "api_group_id": str(zaak["api_group"].id),`

pi-sigma: thanks for catching, the id needs to be retrieved from `api_group`, it's not stored by itself.

jiromaykin (approved):
With these changes I can now make my web-components appear in my own PR, so I think these are done and correct now?

src/open_inwoner/cms/plugins/views.py
```diff
@@ -51,22 +45,81 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
...
+                            "object_id": zaak["uuid"],
+                            "api_group_id": str(zaak["api_group"].id),
+                        },
+                    )
+                )
+                case_data = {
+                    "uuid": zaak["uuid"],
+                    "url": f_url.url,
+                    "identification": zaak.get("identification", ""),
+                    "description": zaak.get(
```
jiromaykin: wait, i feel like something is still missing: in cases that are not submissions, i do not receive the name, but the case-number instead.

Also: for submissions the 'description' seems to both contain a name and a casenumber.

CHANGELOG.rst
```diff
@@ -80,7 +80,7 @@ Nieuwe features
   voor links naar andere portalen gebouwd, inclusief nieuwe design-tokens volgens NLDS conventies. Opmaak is
   verbeterd voor inhoud met langere linkteksten die naar een nieuwe regel worden afgebroken.
 * [:taiga-us:`3509`, :pr:`1997`]: Ondersteuning voor eIDAS login via OIDC is toegevoegd.
-* [:taiga-us:`3575`, :pr:`2020`, :pr:`2037`]: CMS plug-in 'Mijn Zaken' angemaakt voor het
+* [:taiga-us:`3575`, :pr:`2020`, :pr:`2037`, :pr:`2042`]: CMS plug-in 'Mijn Zaken' angemaakt voor het
```
jiromaykin: "angemaakt" should be "aangemaakt" + there is an extra newline below this change that should be removed.

jiromaykin (changes requested):

src/open_inwoner/cms/plugins/views.py
pi-sigma: The first is a bug, we need to handle the description for submissions and cases differently; will be fixed.

I don't think we can do much about the second. If the "name" of an "open submission" contains both, that's what we show.

jiromaykin: Okay, on my PR it looks like this now which is not very informative (and i do see 'names' or perhaps 'types' for non-submission Cases in the case listview page, which looks a bit more informative):

Screenshot 2025-12-02 at 11 36 56

jiromaykin: Ah wait, with the latest change it is much improved, i will approve also.

jiromaykin (approved):

swrichards (approved):

PR #2043, opened 2025-11-24
Implement Preact + Web Components

stefrado:
## Summary

- Use **Preact**, instead of React.
- Create web components and storybook for all unexisting components.
- This PR refactored the following components:
  - `material-icon` (not yet used as web component).
  - `side-navigation`
  - `kvk-branch-selector`
  - `action-list`
  - `oip-homepage-plugin-section`
  - `oip-homepage-plugin-card` (this includes the required changes for balie afspraken)
  - `oip-loading-spinner`
- Pin Node in the CI to 20 - using the actual `.nvmrc` content

## Example rendering

Pass data to the component, and render the list in the _frontend_
```django
{{ data|json_script:"id" }}
<action-list actions-id="id"></action-list>
```

Render list with Django _backend_ (recommended usage)
```django
<oip-homepage-plugin-section>
  {% for card in cards %}
    <oip-homepage-plugin-card some-attribute="{{ card.some_attribute }}"><oip-homepage-plugin-card/>
  {% endfor %}
</oip-homepage-plugin-section>
```

> It is recommended to do the forloop in the django template rather than the frontend.

### Architecture

```mermaid
flowchart TD
    A[Django Backend]
    B[HTML Template]
    B2[Web Component Registry]
    C[Web Component Rendering]
    D[Preact Component]
    E[Rendered HTML]
    E2[HTMX Response]

    %% Rendering
    A --> B --> B2 --> C --> D --> E
    E2 --> B

    %% TESTING 
    F[Storybook Testing]
    F2[Vitest Testing]
    A2[PyTest Testing]

    F -->|default story| D
    F -->|story decorator| B2

    F2 --> D
    A2 --> B
```

## Issue References

This PR includes the following stories

- Balie afspraken homepage plugin design -> https://taiga.maykinmedia.nl/project/open-inwoner/us/3578
- Mijn zaken homepage plugin design -> http://taiga.maykinmedia.nl/project/open-inwoner/us/3576

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables


src/open_inwoner/react/lib/web-component/web-component.spec.tsx
```diff
@@ -1,138 +1,138 @@
 import { describe, expect, it, beforeEach, afterEach } from 'vitest';
-import { registerWebComponents } from './register';
-import GenericReactWebComponent, { AbstractWebComponent } from './abstract';
-import { normalizeAttribute } from './utils';
-import escape from 'lodash.escape';
+import { registerWebComponents } from './loader';
+// import GenericReactWebComponent, { AbstractWebComponent } from './abstract';
```
jiromaykin: This is all commented out - do you wish to remove it or is this a TODO for later?

src/open_inwoner/react/i18n/i18n.ts
```diff
@@ -25,7 +25,7 @@ const getIntlProviderProps = async (): Promise<IntlConfig> => {
   const lang = getLocale();
   const messages = await loadLocaleData(lang);
   return {
-    messages,
+    messages: messages.default,
```
jiromaykin: Can you check (after makemassages/Buil;d/Bundle) if the `pre-commit run -a` process still correctly adds an empty line to end of file?

I don't know if it is just me but i get file diffs in src/open_inwoner/react/i18n/compiled/nl.json... etc, etc.

src/open_inwoner/templates/pages/cases/list_outer.html
```diff
@@ -1,8 +1,6 @@
 {% extends 'master.html' %}
 {% load i18n icon_tags static %}
 
-{% block mount_modules %} data-mount-modules='["sidenav"]'{% endblock %}
-
 {% block react_modules %}
 {# Including this script will allow the page to use the react component #}
```
jiromaykin: Nice, so each individual page no longer needs the data-mount-module, so: it's only the including of `{% static 'bundles/open_inwoner-react.js' %}` that is needed for placement?

In another PR we decided that the repeating comment `"{# Including this script will allow the page to use the react component #}"` can be deleted everywhere, or if you want you could change into something useful like "Including this will place the web-component here" which could be useful for new/future developers.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.stories.tsx
```diff
@@ -1,29 +1,12 @@
-import type { Meta, StoryObj } from '@storybook/react';
-import { IntlProvider } from 'react-intl';
-import { KVKBranchSelector } from '../KVKBranchSelector/KVKBranchSelector';
-import 'material-icons/iconfont/material-icons.css';
 import '@open-inwoner/design-tokens/dist/css/index.css';
+import type { Meta, StoryObj } from '@storybook/preact';
+// import 'material-icons/iconfont/material-icons.css';
```
jiromaykin: commenting out the icon styles here: did you want to use the global styles elsewhere?

src/open_inwoner/templates/cms/plugins/userfeed/userfeed.html
```diff
@@ -12,7 +12,8 @@ <h2 class="utrecht-heading-2 {% if userfeed.action_required %}indicator{% endif
     </header>
 
     <div class="card-container card-container--columns-1 plugin-card">
-        <action-list actions='{{ userfeed_item_list_json }}'></action-list>
+        {{userfeed_item_list_json|json_script:"userfeed_data"}}
```
jiromaykin, line 15: i see here for json_script you're using an `underscore` in userfeed_data (where you're using a `hyphen` in a different tempalte, like in kvk/branches.html and i am wondering if that has something to do with the failign test "...TypeError: the JSON object must be str, bytes or bytearray, not list..."

vite.config.js
```diff
@@ -45,7 +47,6 @@ export default defineConfig({
     rollupOptions: {
       input: {
         [`${paths.package.name}-react`]: `${__dirname}/${paths.reactEntry}`,
```
jiromaykin, line 47: so for Web-components this is now also the Preact entry?

package.json
```diff
@@ -95,25 +97,23 @@
     "@chromatic-com/storybook": "^3.2.7",
     "@formatjs/cli": "^6.7.2",
     "@formatjs/ts-transformer": "^3.14.0",
+    "@preact/preset-vite": "^2.10.2",
```
jiromaykin, line 98: perhaps of later concern but does this PR allow for writing of SCSS as a scss file in each web-component/Preact component?

in my other PR i'm using "sass-embedded" node module, not sure if we can throw that away.

src/open_inwoner/react/lib/web-component/registry.tsx
```diff
@@ -0,0 +1,37 @@
...
+  [SIDE_NAV]: () => import('@react/components/SideNav/SideNav'),
+  [KVK_BRANCH_SELECTOR]: () =>
+    import('@react/components/KVKBranchSelector/KVKBranchSelector'),
+  [MATERIAL_ICON]: () => import('@react/components/MaterialIcon/MaterialIcon'),
+};
+
+/**
+ * Global plugins that run for all web components
+ */
+export const wcPluginRegistry: WebComponentPlugin[] = [
```
jiromaykin: in a previous PR we decided not to use "wc" anymore :-)

src/open_inwoner/react/lib/web-component/utils.ts
```diff
@@ -1,3 +1,82 @@
-export const normalizeAttribute = (attribute: string) => {
-  return attribute.replace(/-([a-z])/g, (_, letter) => letter.toUpperCase());
+import register from 'preact-custom-element';
+import { AnyComponent } from 'preact';
+import { wcRegistry as wcRegistry } from './registry';
+import { WebComponentContext } from './types';
+import { withIntl } from './plugins';
+
+/**
+ * Extended options for web component registration
```
jiromaykin: so this actually converts Preact components into web components? that's a bit unclear to me , i think i'd like that to be added to the doc/comment if that's correct.

src/open_inwoner/react/lib/web-component/utils.ts
```diff
@@ -1,3 +1,82 @@
...
+  | {
+      shadow: false;
+      /**
+       * Whether to automatically wrap the component with IntlProvider for i18n support
+       * @default false
+       */
+      i18n?: boolean;
+    }
+  | {
+      shadow: true;
```
jiromaykin: what is happening here that it needs a shadow DOM again?

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -288,6 +316,24 @@ export const KVKBranchSelector: FC<KVKBranchSelectorProps> = ({
     }
   }, [selectedId, name]);
 
+  if (!data || data.length === 0) {
+    console.error(new Error('[KVKBranchSelector] No branches available'));
```
jiromaykin, line 318: "new Error" is not needed here? could just be console.error.

src/open_inwoner/react/components/SideNav/SideNav.tsx
```diff
@@ -14,19 +17,35 @@ export interface MenuItem {
...
    * with multiple lists that have a gap in between.
    */
-  items: MenuItem[][];
+  items?: MenuItem[] | MenuItem[][];
+  itemsId?: string;
 }
 
-const SideNav: FC<SideNavProps> = ({ items }) => {
+const SideNav: FC<SideNavProps> = ({ items, itemsId }) => {
+  if (!items && !itemsId) return <></>;
```
jiromaykin, line 29: here also wondering if return null; is better

src/open_inwoner/react/components/SideNav/SideNav.tsx
```diff
@@ -14,19 +17,35 @@ export interface MenuItem {
...
+  const rawData = usePropsOrScriptData<MenuItem[] | MenuItem[][]>(
+    items,
+    itemsId
+  );
+  if (!rawData) return <></>;
+
+  // Normalize to MenuItem[][] - check if first element is an array
+  const normalized =
+    rawData.length > 0 && Array.isArray(rawData[0])
+      ? (rawData as MenuItem[][])
```
jiromaykin, line 41: oh dear, is this an array of an array? then i wonder if the OR/ELSE is needed

src/open_inwoner/react/components/SideNav/SideNav.tsx
```diff
@@ -14,19 +17,35 @@ export interface MenuItem {
...
+  // Get data from props or script tag
+  const rawData = usePropsOrScriptData<MenuItem[] | MenuItem[][]>(
+    items,
+    itemsId
+  );
+  if (!rawData) return <></>;
+
+  // Normalize to MenuItem[][] - check if first element is an array
+  const normalized =
+    rawData.length > 0 && Array.isArray(rawData[0])
```
jiromaykin, line 40: so it is always at least 1 item, but what if What if rawData is empty? (not sure if we have that situation)

jiromaykin (changes requested):

src/open_inwoner/react/lib/web-component/web-component.spec.tsx
stefrado: Yes its a TODO - i still need to write some tests for all the functions i wrote

src/open_inwoner/templates/pages/cases/list_outer.html
stefrado: I removed the `{% block react_modules %}` and added this script to `master.html`.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.stories.tsx
stefrado: The global `openinwoner-css.css` styling (which includes material-icons.css) is imported inside `preview.tsx`

package.json
stefrado: Yes it does allow that (see ActionList.tsx).

So we might have to remove `sass-embedded` later if we are going to merge your PR.

vite.config.js
stefrado: Yes :)

stefrado: Eventually we will have only have the `openinwoner-js.js` again :) (once webpack is removed...)

.storybook/preview.tsx
```diff
@@ -1,32 +1,20 @@
...
+  withThemeClass,
+} from '../src/open_inwoner/react/lib/decorators';
 import '../src/open_inwoner/static/bundles/open_inwoner-css.css';
-
-const withThemeClass = (Story: StoryFn) => {
-  document.body.classList.add('openinwoner-theme');
-
-  return <Story />;
-};
+import '../src/open_inwoner/static/bundles/open_inwoner-react.js';
```
swrichards: Out of interest: no `@` prefixing in Storybook? I thought that was in the tsconfig, so it should be picked up.

src/open_inwoner/cms/plugins/cms_plugins/userfeed.py
```diff
@@ -34,7 +32,7 @@ def render(self, context, instance, placeholder):
             {
                 "instance": instance,
                 "userfeed": feed,
-                "userfeed_item_list_json": json.dumps(feed_items_list),
+                "userfeed_item_list_json": feed_items_list,
```
swrichards, line 35: As discussed, I think we should ditch this way of doing things and compose in the backend. But this is more of a note to us for a follow-up refactor, outside of scope here.

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,60 @@
...
+  // Get the import function.
+  const importer = wcRegistry[name];
+  if (!importer) throw new Error(`"${name}" has no web component import`);
+
+  // Get the context of the web components
+  const contexts = createContextsForComponent(name);
+
+  try {
+    // Run plugins `beforeLoad` hooks
+    await Promise.all(contexts.map((context) => runBeforeLoadHooks(context)));
```
swrichards:
```suggestion
    await Promise.all(contexts.map(runBeforeLoadHooks));
```

src/open_inwoner/react/lib/web-component/utils.ts
```diff
@@ -1,3 +1,60 @@
...
+import {
+  WebComponentContext,
+  WebComponentKey,
+  WebComponentRegisterOptions,
+} from './types';
+import { withIntlWc } from '@react/lib/decorators';
+
+/**
+ * Find all unique web component names on the current page
+ * @returns a unique array of strings from the founded component names.
```
swrichards: Most functions on this page seem to me to be only used (and useful) in `wcLoader`. Would it make sense to inline them in that file, and to not `export` them to keep the external facing API simpler?

src/open_inwoner/react/components/ActionList/ActionList.tsx
```diff
@@ -9,20 +12,37 @@ export interface IActionProps {
...
 
+// This wrapper allows lazy loading of the component
+export function loader() {
+  registerWebComponent(
+    ActionList,
+    WEB_COMPONENT_NAME,
+    ['actions', 'actionsId'],
+    {
+      shadow: false,
+    }
```
swrichards: Given that we have a fairly substantive bit of tooling around loading and configuring webcomponents, wouldn't it be cleaner if the web component name, props, and extra options are exported declaratively in something like:

```typescript
interface WebComponentDefinition {
   compentName: string;
   expectedAttributes: Array<string>;
   options?: WebComponentRegisterOptions;
}

```

Otherwise the setup code is quite fragmented. Seems cleaner to me to be declarative here, and do all the imperative stuff in the loader, including registering the webcomponent (maybe we then also don't have to do any magic of finding the components on the page to apply operations to them).

src/open_inwoner/react/components/ActionList/constants.ts
```diff
@@ -0,0 +1 @@
+export const WEB_COMPONENT_NAME = 'action-list' as const;
```
swrichards: See comment above about a declarative spec, but we might then also have the convention that this happens in a `web-component-spec.ts` file within the module.

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,60 @@
...
+  wcRegistry,
+  WebComponentKey,
+} from '.';
+
+/**
+ * Load a single web component with plugin lifecycle
+ */
+export const wcLoader = async (name: WebComponentKey): Promise<void> => {
+  // Skip if already defined
+  if (customElements.get(name)) return;
```
swrichards: Some debug tooling couldn't hurt:

```suggestion
  if (customElements.get(name)) {
      console.debug(`Custom element ${name} is already defined, aborting loading for this element`);
  }
```

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,60 @@
...
+  runBeforeLoadHooks,
+  runErrorHooks,
+  wcRegistry,
+  WebComponentKey,
+} from '.';
+
+/**
+ * Load a single web component with plugin lifecycle
+ */
+export const wcLoader = async (name: WebComponentKey): Promise<void> => {
```
swrichards: I think this can be private, making the API easier to follow. `registerWebComponents` is the public entrypoint.

```suggestion
const wcLoader = async (name: WebComponentKey): Promise<void> => {
```

src/open_inwoner/react/lib/web-component/types.ts
```diff
@@ -0,0 +1,78 @@
...
+      shadow: true;
+      mode?: 'open' | 'closed';
+      adoptedStyleSheets?: CSSStyleSheet[];
+      serializable?: boolean;
+      /**
+       * Whether to automatically wrap the component with IntlProvider for i18n support
+       * @default false
+       */
+      i18n?: boolean;
+    };
```
swrichards: Can we simplify this to?:
```suggestion
export type WebComponentRegisterOptions = {
  /**
   * Whether to automatically wrap the component with IntlProvider for i18n support
   * @default false
   */
  i18n?: boolean;
  shadow: boolean;
} & {
  shadow: true;
  mode?: 'open' | 'closed';
  adoptedStyleSheets?: CSSStyleSheet[];
  serializable?: boolean;
};

```

swrichards (commented):
Some preliminary comments, let me know if you think we should discuss. Otherwise good work here.

src/open_inwoner/react/lib/web-component/types.ts
stefrado: i18n is in both types. I could refactor it to: 
```suggestion
export type WebComponentRegisterOptions =
  | {
      shadow: false;
    }
  | {
      shadow: true;
      mode?: 'open' | 'closed';
      adoptedStyleSheets?: CSSStyleSheet[];
      serializable?: boolean;
    }
    & {
       /**
       * Whether to automatically wrap the component with IntlProvider for i18n support
       * @default false
       */
      i18n?: boolean;
    }
```

I'll take another look

src/open_inwoner/react/lib/web-component/loader.tsx
stefrado: It might be better readable to create registerWebComponents as a class instead of function based.

src/open_inwoner/react/components/ActionList/ActionList.tsx
stefrado: Yes i'll refactor this logic a bit

src/open_inwoner/react/lib/web-component/utils.ts
stefrado: Looking at the other comment - same applies here

> It might be better readable to create registerWebComponents as a class instead of function based.

src/open_inwoner/cms/plugins/cms_plugins/userfeed.py
stefrado: Yes it'll require complete component refactorings. I kept it simple for now by using the old logic.

.storybook/preview.tsx
stefrado: TBH i did not check this, but the `.storybook` folder may not be included in the tsconfig

stefrado: Also we do not have a prefix for our bundles folder (i might add this as well)

src/open_inwoner/react/components/SideNav/SideNav.tsx
stefrado: Sometimes returning null returns in a `0` being rendered on the page...

Haven't tested it in this case

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
stefrado: This adds some error tracing

src/open_inwoner/react/lib/web-component/utils.ts
stefrado: This is to work with conditional types

package.json
```diff
@@ -125,6 +125,7 @@
     "cssnano": "latest",
     "formatjs": "^0.1.1",
     "jest": "^30.0.3",
+    "jsdom": "^27.2.0",
```
jiromaykin, line 127: Apparently doesn't support the Dialog tag yet:
https://github.com/jsdom/jsdom/issues/3294

src/open_inwoner/react/i18n/i18n.ts
stefrado: checked it and it seems ok after pre-commit

package.json
stefrado: Interesting... will be fun when we decide to implement dialogs

src/open_inwoner/react/components/ActionList/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const WEB_COMPONENT_NAME = 'action-list' as const;
+
+export const ACTION_LIST_DEFINITION: WebComponentDefinition<IActionListProps> =
+  {
+    tagName: WEB_COMPONENT_NAME,
+    propNames: ['actions', 'actionsId'],
+    options: { shadow: false },
+    importer: () => import('./ActionList'),
+  };
```
swrichards: Quick drive by comment: Why export the component name separately, if people can simply import the definition and get it from there?

src/open_inwoner/react/components/ActionList/ActionList.tsx
```diff
@@ -9,20 +12,27 @@ export interface IActionProps {
...
           {message} |{' '}
-        </span>{' '}
+        </span>
         <span className="denhaag-action__content--oip-title">{title}</span>
       </ActionSingle>
     );
   });
 };
 
+export const loader = WebComponentLoader.loadWC(WEB_COMPONENT_NAME, ActionList);
```
swrichards: What's the value of defining this in this file, rather than at the point of registering the component?

stefrado: The usage is to make sure the web-component can be loaded in a storybook story. But maybe i can fix that part a little - then we can define the `WEB_COMPONENT_NAME` inside the config as well.

src/open_inwoner/react/components/ActionList/constants.ts
stefrado: Short answer: circular imports. But you other comment gave me inspiration to prevent this.

package.json
```diff
@@ -137,14 +138,15 @@
     "prettier-stylelint": "^0.4.2",
     "raw-loader": "^4.0.2",
     "regenerator-runtime": "^0.14.1",
+    "rollup": "^4.53.3",
```
jiromaykin: Looks like we still need Rollup for Vite, but when i look here: https://stackoverflow.com/questions/79048814/github-action-is-failing-due-to-rollup-rollup-linux-x64-gnu

It sugests you don't need to change CI yml, but juts add 
"... this to the package.json ..."

```
"optionalDependencies": {
  "@rollup/rollup-linux-x64-gnu": "*"
}
```
could be worthwhile to test and see if that works instead.

src/open_inwoner/react/components/HomepagePluginSection/HomepagePluginSection.tsx
```diff
@@ -0,0 +1,58 @@
...
+export type IHomepagePluginSectionProps = RenderableProps<{
+  title?: string;
+  nextUrl?: string;
+  nextUrlLabel?: string;
+  showIndicator?: boolean;
+  columns: number;
+  children?: ComponentChildren;
+}>;
+
+const HomepagePluginSection: FC<IHomepagePluginSectionProps> = ({
```
jiromaykin: A general question, also to @swrichards : this is a very rudimentary component that either exists or doesn't so do we not need front-end spec tests for these?

Are we going to work in such a way most testing is done by the back-end for these type of simple components?

src/open_inwoner/cms/plugins/tests/test_appointments.py
```diff
@@ -35,20 +35,20 @@ def test_plugin(self, m):
         pyquery = PyQuery(html)
 
         # test item
-        items = pyquery.find(".card-container .card")
+        items = pyquery.find("oip-homepage-plugin-card")
```
jiromaykin: Suggestion: not about to this test file, but adding to the `_/cms/plugins/cms_plugins/appointments.py` file: I would love to have a proper mock for the front-end .

If I copy the text from the Figma design and use this code, I think it could be better @swrichards ?:

```
from django.utils.translation import gettext as _

import structlog
from cms.plugin_base import CMSPluginBase
from cms.plugin_pool import plugin_pool

from open_inwoner.cms.plugins.models.appointments import UserAppointments
from open_inwoner.qmatic.client import NoServiceConfigured, qmatic_client_factory

@plugin_pool.register_plugin
class UserAppointmentsPlugin(CMSPluginBase):
    model = UserAppointments
    module = _("General")
    name = _("My appointments")
    render_template = "cms/plugins/appointments/appointments.html"

    def render(self, context, instance, placeholder):
        request = context["request"]
        if not request.user.is_authenticated or not request.user.has_verified_email():
            appointments = []
        else:
            try:
                client = qmatic_client_factory()
            except NoServiceConfigured:
                logger.exception("Error occurred while creating Qmatic client")
                appointments = []
            else:
                appointments = client.list_appointments_for_customer(
                    request.user.verified_email
                )

        # MOCK DATA
        if not appointments:
            from datetime import datetime

            class MockBranch:
                timeZone = "Europe/Amsterdam"

            class MockAppointment:
                def __init__(self, service_name, date_str, address_line, city, public_id):
                    self.title = service_name
                    self.start = datetime.fromisoformat(date_str)
                    self.services = [type('Service', (), {'name': service_name})()]
                    self.branch = type('Branch', (), {
                        'timeZone': "Europe/Amsterdam",
                        'addressLine2': address_line,
                        'addressCity': city,
                    })()
                    self.publicId = public_id

            appointments = [
                MockAppointment("rijbewijs ophalen", "2025-09-21T13:15:00+02:00", "Raadhuisplein 10", "Haren", "rijbewijs-001"),
                MockAppointment("VOG aanvragen", "2025-09-30T15:45:00+02:00", "Kreupelstraat 1", "Groningen", "vog-001"),
            ]

        context.update(
            {
                "instance": instance,
                "appointments": appointments,
            }
        )
        return context


```

jiromaykin (changes requested):
Almost ready to merge :-)

src/open_inwoner/cms/plugins/tests/test_appointments.py
stefrado: I used this to mock the appointments:

```python
import factory

from open_inwoner.qmatic.tests.factories import (
    AppointmentFactory,
    BranchDetailFactory,
    QmaticServiceFactory,
)

appointments_mocks = [
    AppointmentFactory.build(
        title=factory.Faker("catch_phrase").evaluate(None, None, {"locale": "nl_NL"}),
        start=factory.Faker(
            "date_time_between", start_date="now", end_date="+30d"
        ).evaluate(None, None, {"locale": "nl_NL"}),
        notes=factory.Faker("text", max_nb_chars=120).evaluate(
            None, None, {"locale": "nl_NL"}
        ),
        branch=BranchDetailFactory.build(
            name=factory.Faker("company").evaluate(None, None, {"locale": "nl_NL"}),
            timeZone=factory.Faker("timezone").evaluate(
                None, None, {"locale": "nl_NL"}
            ),
            addressCity=factory.Faker("city").evaluate(None, None, {"locale": "nl_NL"}),
            addressLine1=factory.Faker("street_name").evaluate(
                None, None, {"locale": "nl_NL"}
            ),
            addressLine2=factory.Faker("street_address").evaluate(
                None, None, {"locale": "nl_NL"}
            ),
            addressZip=factory.Faker("postcode").evaluate(
                None, None, {"locale": "nl_NL"}
            ),
        ),
        services=[
            QmaticServiceFactory.build(
                name=factory.Faker("sentence", nb_words=2).evaluate(
                    None, None, {"locale": "nl_NL"}
                )
            )
        ],
    )
    for i in range(0, 10)
]
```

src/open_inwoner/react/components/HomepagePluginSection/HomepagePluginSection.tsx
stefrado: I did not write a test for this component yet (time), but in my opinion every should this component have a test.

src/open_inwoner/react/components/SideNav/SideNav.tsx
stefrado: This is a layer to work with the actual accepted props for the MenuItems.

With this we accept their standard `MenuItem[][]`, this renders mutliple lists with spacing.

package.json
stefrado: This also did the trick for us...

src/open_inwoner/cms/plugins/views.py
```diff
@@ -27,6 +28,14 @@ class ZakenPluginContentView(RequiresHtmxMixin, CaseLogMixin, View):
     def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
         case_service = CaseListService(request)
 
+        # Get plugin instance for title
+        try:
+            plugin_instance = CMSZakenPluginConfig.objects.get(pk=plugin_id)
+            plugin_title = plugin_instance.title
+        except CMSZakenPluginConfig.DoesNotExist:
+            logger.warning("Plugin instance not found", plugin_id=plugin_id)
+            plugin_title = _("Mijn Zaken")
```
swrichards, line 37: @pi-sigma not per se for this PR (we'd do it separately), but: I wonder if we should handle this case more gracefully. If no plugin, then we probably shouldn't try to fetch cases, and might want to degrade more gracefully.

swrichards: Alternatively, we can just raise and return a `500` here: we're the only callers of this endpoint, and definitionally, we only do it if the calling page exists, which it only does if there is a plugin.

src/open_inwoner/react/components/HomepagePluginSection/HomepagePluginSection.tsx
swrichards: I agree with @stefrado -- there should at the very least be some kind of smoke test, to ward against e.g. renames, import errors, and the likes, just so there's coverage.

I'd say this might be a good use-case for snapshot testing (possibly via Storybook, e.g. using Storyshots).

swrichards: Given the large scope of this PR, I think a quick smoke test (`it('renders without errors', () =>...`) would suffice.

src/open_inwoner/react/components/HomePluginCardItem/HomePluginCardItem.tsx
```diff
@@ -0,0 +1,54 @@
+import { AnyComponent as FC } from 'preact';
+import type { HomepageCardTypes } from './types';
+import './HomePluginCardItem.scss';
+import { MaterialIcon } from '../MaterialIcon';
+import clsx from 'clsx';
+import { kebabCase } from 'lodash';
```
swrichards: Targeted import for better tree shaking? I forget the form, something like `lodash/kebab-case`

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+const HomePluginCardItem: FC<HomepageCardTypes> = ({
+  title,
+  description,
+  identificatie,
+  detailUrl,
+  date,
+  address,
+  renderAsHeading = false,
+}) => {
+  const Heading = String(renderAsHeading) !== 'false' ? 'h3' : 'p';
```
swrichards, line 17: What is `String`?

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.stories.tsx
swrichards: So is this line in or out? :)

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
```diff
@@ -1,17 +1,15 @@
+import { MaterialIcon } from '@react/components/MaterialIcon';
+import { usePropsOrScriptData } from '@react/lib/json/getJsonScriptData';
+import { useDebounce } from '@react/lib/hooks/useDebounce';
+import clsx from 'clsx';
 import {
-  ChangeEventHandler,
-  FC,
+  AnyComponent as FC,
```
swrichards: What does the rename add?

src/open_inwoner/react/i18n/i18n.tsx
```diff
@@ -0,0 +1,102 @@
+import { AnyComponent as FC } from 'preact';
+import { useEffect, useState } from 'preact/hooks';
+import { IntlConfig, IntlProvider } from 'react-intl';
+
+/** Cache of locale string and IntlConfig */
+const cache: Map<string, IntlConfig> = new Map();
+
+/**
+ * i18n
+ * Auto loads translations based on html document language, which is read from html lang attribute.
```
swrichards, line 10: > ...which is read from html lang attribute

I note we hardcode this attribute, so this makes sense, but it raises the question of whether the `en` langauge is used at all.

src/open_inwoner/react/lib/decorators/test-lib-decorators.spec.tsx
swrichards, line 1: I only noticed now, but my sense is that hyphens are commonly used in frontend land then underscores (different from Python, where you end up importing the module name, not a string literal pointing to a file). I am ok with this, but there's something to be said for keeping the hyphen convention if you agree it is more commonplace (see also e.g. `web-component.tsx`).

src/open_inwoner/react/lib/json/json.ts
swrichards, line 1: I think naming the file kebab-cased after a symbol is not generally a pattern except for components: maybe just abolish this directory and call it `json.ts` ?

src/open_inwoner/react/lib/web-component/plugins/error-boundary.ts
```diff
@@ -0,0 +1,15 @@
+import type { WebComponentPlugin } from '../types';
```
swrichards, line 1: Would love to have sentry integration here! (Not now, in future).

src/open_inwoner/react/lib/web-component/plugins/error-boundary.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Silent error plugin
+ * Hides components that fail to load without showing error messages
+ */
+export const silentErrorPlugin: WebComponentPlugin = {
+  name: 'silent-error',
+
+  onError: ({ element }, error) => {
+    if (!error) return;
+    element.style.display = 'none';
```
swrichards, line 12: Would a `console.error` here help things, or is the call site expected to already do this?

src/open_inwoner/cms/plugins/views.py
pi-sigma: Why is this `try/except` necessary? We should never get to the `ZakenPluginContentView` if the plugin is not configured. The `title` on the plugin config is not nullable and has "Mijn Zaken" as default.

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,186 @@
...
+  private static pluginRegistry: WebComponentPlugin[] = [
+    silentErrorPlugin,
+    skeletonPlugin,
+    // @ts-expect-error
+    ...(window.IS_DEV ? [performancePlugin] : []),
+  ];
+
+  /**
+   * Find all unique web component names on the current page
+   * @returns a unique array of strings from the founded component names.
```
swrichards:
```suggestion
   * @returns a unique array of strings from the found component names.
```

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,186 @@
...
+    // Remove i18n from options before passing to preact-custom-element
+    const { i18n, ...registerOptions } = options;
+
+    // If i18n option is enabled, wrap the component with IntlProvider
+    const ComponentToRegister = i18n ? withIntlWc(Component) : Component;
+
+    return register(ComponentToRegister, tagName, propNames, registerOptions);
+  }
+
+  public static async importWC(name: WebComponentTagName): Promise<void> {
```
swrichards: Generally, I wonder whether short names really help much compared to just reading the noun fully (e.g. optimize for reading, not writing):

```suggestion
  public static async importWebComponent(name: WebComponentTagName): Promise<void> {
```

src/open_inwoner/react/lib/web-component/loader.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  /**
+   * Register the webcomponents on the current page
+   * @returns
+   */
+  static async registerWebComponents() {
+    try {
+      // Look for all web components on the current page
+      const founded = WebComponentLoader.findWebComponentsOnPage();
```
swrichards:
```suggestion
      const foundComponents = WebComponentLoader.findWebComponentsOnPage();
```

swrichards: Maybe also instantiate `WebComponentLoader`? Now everything in the class is static, but there's no guarantee it'll stay that way, and the overhead is minimal.


```suggestion
      const founded = WebComponentLoader().findWebComponentsOnPage();
```

src/open_inwoner/react/main.ts
```diff
@@ -1,4 +1,8 @@
-import ModuleLoader from '@react/lib/loader';
+import htmx from 'htmx.org';
+import { WebComponentLoader } from './lib/web-component/loader';
 
-// Load the module.
-ModuleLoader.load();
+WebComponentLoader.registerWebComponents();
+
+htmx.on('htmx:afterSwap', () => {
+  WebComponentLoader.registerWebComponents();
```
swrichards, line 10: As above, maybe create an instance.

A comment about the why might be good here too.

swrichards (changes requested):
@stefrado Props for the big push here, I think it looks very good, just a few scattered comments and questions but nothing major.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.stories.tsx
jiromaykin: Commented out stuff needs to be removed then.

src/open_inwoner/cms/plugins/views.py
pi-sigma: I thought about it: if an admin deleted the plugin in the midst of an HTMX request triggered by user activity (after the render method from the zaken plugin has already returned) , this would trigger an error. Unlikely, but doesn't hurt to be defensive.

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
stefrado: renderAsString could be a boolean or string: (`true` | `false` | `"true"` | `"false"`).
Because web-component attrs can only be string, but preact components pass actual booleans. 

String converts a boolean into a boolean string `false` -> `"false"`.

This way we can check `"false" !== "false"` (which is false), `false !== "false"` (which is true)

> TLDR: String converts any value into a string.

src/open_inwoner/react/components/Spinner/Spinner.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+type Story = StoryObj<SpinnerProps>;
+
+export const Default: Story = {
+  args: {
+    loadingText: 'Laden...',
+    iconName: 'rotate_right',
+  },
+};
+
+export const CustomIcon: Story = {
```
jiromaykin, line 41: This is fun but not really functional since we do not give any option to change the spinner icon, so this would be a bit misleading.

src/open_inwoner/react/components/HomePluginCardItem/HomePluginCardItem.tsx
stefrado: Went from `71.5K (gzipped: 25.3k)` to `5.8K (gzipped: 2.5k)` 🥇

src/open_inwoner/react/components/HomepagePluginSection/HomepagePluginSection.tsx
```diff
@@ -0,0 +1,58 @@
+import { ComponentChildren, AnyComponent as FC, RenderableProps } from 'preact';
+import { MaterialIcon } from '../MaterialIcon';
+import './HomepagePluginSection.scss';
```
jiromaykin: General comment about this directory - the item component in other dir. is named "HomePlugin" so without the word 'page for both the directory and the files. 
Not sure if that was a conscious choice, but you could remove the 'page' and all name it HomePluginSection, so it's a bit shorter and maybe less prone tot errors.

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
stefrado: In react `FC` is a general convention to write components, this is mostly done for readablity. 
In Preact this is called AnyComponent.

under the hood:
```ts
export type AnyComponent<P = {}, S = {}> =
	| FunctionComponent<P>
	| ComponentConstructor<P, S>;
```

> TLDR I'll change the as from `FC` to `AC` which suits better.

src/open_inwoner/react/i18n/i18n.tsx
stefrado: Yes `en` can be used, this is the implementation of the language:

`const locale = lang || document.documentElement.lang || 'nl';`

So if lang is passed, for a controlled language (which is not the case), lang is used -> otherwise we use the HTML lang value -> and if even that is not set, we use "nl" as fallback

src/open_inwoner/react/lib/decorators/test-lib-decorators.spec.tsx
jiromaykin: And error-boundery.ts uses hyphen.

jiromaykin (changes requested):

src/open_inwoner/cms/plugins/tests/test_appointments.py
jiromaykin: I think we need to make list of stuff we'd like to add and improve but goes out of scope for this PR. Like this mock and some front-end testing.

src/open_inwoner/react/components/Spinner/Spinner.stories.tsx
stefrado: We do have that option:

```ts
export interface ILoadingSpinnerProps {
  loadingText?: string;
  iconName?: string;
}
```

src/open_inwoner/react/lib/web-component/loader.tsx
stefrado: 🚾

src/open_inwoner/react/lib/decorators/test-lib-decorators.spec.tsx
stefrado: Yes y'all are correct.

file names should have hyphens, not undescore

src/open_inwoner/react/components/Spinner/Spinner.stories.tsx
jiromaykin: In the props yes, but I mean: not optional for municipality humans.

src/open_inwoner/react/lib/web-component/plugins/error-boundary.ts
stefrado: I've added a todo inside the code :)

src/open_inwoner/react/lib/web-component/plugins/error-boundary.ts
stefrado: There actually is already a console.error inside the try/catch

src/open_inwoner/react/components/KVKBranchSelector/KVKBranchSelector.tsx
jiromaykin: 👍 Good to know new syntax.

src/open_inwoner/react/components/HomepagePluginSection/HomepagePluginSection.tsx
stefrado: I did not name these components, they where from you PR. So yes i'll remove page from the oip-homepage-plugin-section component

stefrado: Renamed the component to `HomePluginSection`

CHANGELOG.rst
```diff
@@ -44,6 +44,8 @@ Deployment aandachtspunten
   de primaire kleur (CSS-variabelen ``--color-primary`` en ``--color-secondary``) is hier nu een
   stijl verandering te zien. Call-to-action elementen gebruiken nu altijd de primaire kleur. Hierdoor
   worden de knoppen in tegels niet langer in een afwijkende secundaire kleur weergegeven.
+* De nieuwe frontend web-componenten moeten goed gecheckt worden of het werkt. Het zou kunnen dat de
```
jiromaykin: Enigszins kromme zin :-) dus laatste change:

* Controleer of de nieuwe frontend web-componenten correct functioneren. In sommige gevallen wordt de frontend-bundle mogelijk niet volledig overschreven.

jiromaykin (changes requested):
Almost there!

src/open_inwoner/react/components/Spinner/Spinner.stories.tsx
stefrado: Yes, but the storybook is a development tool. Each story should explain a part of the component (props or interactions)

CHANGELOG.rst
```diff
@@ -93,8 +95,10 @@ Nieuwe features
   belangrijk kan zijn voor security audits (zoals PENTests) van bijvoorbeeld DigiD.
   Standaard blijven de hints ingeschakeld om het huidige gedrag te behouden.
 * [:taiga-us:`3580`, :pr:`2033`]: De zoekfunctie kan nu worden uitgeschakeld in de Algemene Configuratie.
-* [:taiga-is`3600`, :pr:`2046`]: 'Signing requets' toegevoegd aan admin. Beheerders kunnen nu een CSR
+* [:taiga-is:`3600`, :pr:`2046`]: 'Signing requets' toegevoegd aan admin. Beheerders kunnen nu een CSR
```
jiromaykin:
```suggestion
* [:taiga-is:`3600`, :pr:`2046`]: 'Signing requests' toegevoegd aan admin. Beheerders kunnen nu een CSR
```

Someone else made a typing error here in the word 'requests' (missing S) - not your fault, but you can correct that here too.

src/open_inwoner/react/lib/web-component/loader.tsx
stefrado: For now i'll leave it like this (static), we can always refactor it later

CHANGELOG.rst
pi-sigma: My bad, thanks for spotting this!

pi-sigma (approved):
Looking good! Two comments/questions:

- I wasn't able to get the plugin to be displayed because the webcomponent was not registered unless I manually installed `@esbuild/linux-x64` (somehow it was not included by just running `npm install` and then `npm run build`. Double-check that this is not missing.
- The `silentErrorPlugin` currently hides errors from the users in case of client-side JS errors. Is this what we want (instead of showing the user a generic error message "Something went wrong; your zaken cannot be displayed", like we do on the backend)

src/open_inwoner/cms/plugins/views.py
swrichards: Fair enough, we'll keep it as-is

CHANGELOG.rst
jiromaykin: @stefrado I don't see this change yet in the latest commit?

jiromaykin (approved):
Incredible work, nice to continue with web-components in future.
Any loose ends from this that need to be solved separately may be put in our 'loose ends' story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3615

swrichards (approved):

PR #2049, opened 2025-12-03
[#3579] Store not just successful syncs, but also failures and filters with reasons

swrichards:
## Summary

First PR of a two-parter: this PR mainly ensures all API fields that have a counterpart in our ZGW models are in fact synced (especially during updates), and also that we give more comprehensive feedback regarding the sync outcome (including dropped entities, e.g. due to filters or errors).

## Issue References

- Taiga User Story: [#3579](https://taiga.maykinmedia.nl/project/open-inwoner/us/3579)
- Taiga Task: [#3601](https://taiga.maykinmedia.nl/project/open-inwoner/task/3601)

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1066 @@
...
 
 logger = structlog.stdlib.get_logger(__name__)
 
 
-def filter_zaaktypes(case_types: list[ZaakType]) -> list[ZaakType]:
-    return [c for c in case_types if c.indicatie_intern_of_extern == "extern"]
+class ExclusionReason(str, Enum):
+    """Reasons why an object was excluded from import"""
+
+    FILTERED_INTERNAL = _("Object uitgefilterd opmdat deze als 'intern' is aangemerkt")
```
pi-sigma:
```suggestion
    FILTERED_INTERNAL = _("Object uitgefilterd omdat deze als 'intern' is aangemerkt")
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1066 @@
...
+    API_ERROR = _(
+        "Object uitgefilterd omdat deze door een API fout niet kon worden opgevraagd"
+    )
+    DATABASE_ERROR = _(
+        "Object uitgefilterd omdat deze door een database fout niet kon worden opgeslagen"
+    )
+    MISSING_CATALOGUS = _("Object verwijst naar een niet-bestaande catalogus")
+    NO_CLIENT = _("OIbject uitgefilderd omdat er geen valide client beschikbaar was")
```
pi-sigma:
```suggestion
    NO_CLIENT = _("Object uitgefilderd omdat er geen valide client beschikbaar was")
```

CHANGELOG.rst
```diff
@@ -93,6 +93,10 @@ Nieuwe features
   belangrijk kan zijn voor security audits (zoals PENTests) van bijvoorbeeld DigiD.
   Standaard blijven de hints ingeschakeld om het huidige gedrag te behouden.
 * [:taiga-us:`3580`, :pr:`2033`]: De zoekfunctie kan nu worden uitgeschakeld in de Algemene Configuratie.
+* [:taiga-us:`3579`, :pr:`2049`]: Het commanda om de ZGW Catalogus te importeren geeft
```
pi-sigma:
```suggestion
* [:taiga-us:`3579`, :pr:`2049`]: Het commando om de ZGW Catalogus te importeren geeft
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1066 @@
...
+                    updated = False
+
+                    # Update all overlapping fields
+                    if ztstc.omschrijving != status_type.omschrijving:
+                        ztstc.omschrijving = status_type.omschrijving
+                        updated = True
+
+                    if ztstc.statustekst != status_type.statustekst:
+                        ztstc.statustekst = status_type.statustekst
+                        updated = True
```
pi-sigma, line 854: What about `omschrijving_generiek` and `is_eindstatus`? Should they be updated as well?

`Statustype` also has `volgnummer` and `informeren`, but these are not supported by eSuite so we can exclude them here?

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1066 @@
...
-                    # we got a record for this, see if we got data to update
+
+                if ztrtc := existing_map.get(resultaat_type.url):
+                    # Update existing
+                    updated = False
+
+                    # Update omschrijving if changed
+                    if ztrtc.omschrijving != resultaat_type.omschrijving:
+                        ztrtc.omschrijving = resultaat_type.omschrijving
+                        updated = True
```
pi-sigma, line 989: What about other fields?
- `archiefnominatie`
- `archiefactietermijn`
- `brondatum_archiefprocedure`
- `esuite_compat_naam`

src/open_inwoner/openzaak/management/commands/zgw_import_data.py
```diff
@@ -2,79 +2,28 @@
...
+        importers = [
+            ZGWCatalogusImporter(api_group)
+            for api_group in ZGWApiGroupConfig.objects.all()
+        ]
 
-        self.stdout.write("")
+        output = ""
+        for importer in importers:
+            result = importer.import_all()
+            output = output + result.pretty_print() + "\n"
```
pi-sigma:
```suggestion
        outputs = []
        for importer in importers:
            result = importer.import_all()
            outputs.append(result.pretty_print())
        self.stdout.write("\n".join(outputs))     
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1066 @@
...
+
+        # Import zaaktypen
+        result.zaaktypen = self.import_zaaktype_configs()
+
+        # Import related types for each zaaktype
+        for ztc in ZaakTypeConfig.objects.filter(
+            catalogus__service=self.zgw_api_group.ztc_service
+        ).order_by("catalogus__domein", "identificatie"):
+            result.informatieobjecttypen.append(
+                self.import_informatieobjecttype_configs_for_zaaktype(ztc)
```
pi-sigma, line 389: The method call includes a call to `fetch_case_types_by_identification_no_cache` for each `ztc` object, even though case types are already fetched via `import_zaaktype_configs`. This will slow down the import for installations with many zaaktypen - though I don't know how big of a problem that will be.

Perhaps we can store the zaaktypen in a cache (a mapping from zaaktype_identificatie to zaaktype) when we first fetch them, then later retrieve them from the cache when needed again. Not sure if it's worth it; it depends on the performance impact.

pi-sigma (changes requested):

src/open_inwoner/openzaak/zgw_imports.py
swrichards: They're not on our models, but I think you're right that they should be. I'll see if I can add that.

src/open_inwoner/openzaak/zgw_imports.py
swrichards: We don't use them anywhere, so I've left them out. But I think a useful (additional) feature would be something like `Fetch from backend` which would open in a new tab and show the whole object.

src/open_inwoner/openzaak/zgw_imports.py
swrichards: I tweaked this a bit, combining two of the client endpoints for the zaaktypen (just one list endpoint call, with optional params).

I thought about the caching too, but to be fair, given all the issues we've had here, I'd prefer to have it a bit slower but easier to follow. Adding caching introduces other complexity if e.g. you call some of the methods out of order/with time in between.

Given that this is supposed to be a long running job, and will run as a Celery task, I'd prefer slow-but-linear-and-transparent.

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1049 @@
...
-    # grab actual ZaakTypes for this identificatie
-    zaak_types: list[ZaakType] = get_configurable_zaaktypes_by_identification(
-        client, ztc.identificatie, ztc.catalogus_url
-    )
-    if not zaak_types:
-        return []
-
-    create = []
-    update = []
+        zaak_types = zaak_types = self.get_api_zaaktypen_for_saved_ztc(ztc)
```
pi-sigma:
```suggestion
        zaak_types = self.get_api_zaaktypen_for_saved_ztc(ztc)
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1049 @@
...
+                 - {{ item }}
+            {%- endfor %}
+            {%- endif %}
+            {%- if excluded_items %}
+              ⚠️  Excluded ({{ excluded_items|length }}):
+            {%- for exc_item in excluded_items %}
+                 - {{ exc_item.label }}
+                   Reason: {{ exc_item.reason }}
+            {%- if exc_item.error %}
+                   Error: {{ exc_item.error }}
```
pi-sigma: If we're going to use emojis for `created`, `updated`, `excluded`, we should use one here too for uniformity. Something like:

```suggestion
                   	🔴 Error: {{ exc_item.error }}
```

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1,436 +1,1049 @@
...
+                            logger.exception(
+                                "Failed to update informatieobjecttype config",
+                                informatieobjecttype_url=info_type.url,
+                                exc_info=exc,
+                            )
+                            result.excluded.append(
+                                ExcludedObject(
+                                    object_type="InformatieObjectType",
+                                    url=info_type.url,
+                                    reason=ExclusionReason.DATABASE_ERROR,
```
pi-sigma, line 740: We're catching the broad `Exception` but report the more specific `DATABASE_ERROR`. If we know that it's a database error, would it be enough to catch this at l. 730? It's a minor thing though.

pi-sigma (changes requested):
Nice refactor! I think it's almost good to go, just a few small remarks.

pi-sigma:
@swrichards The failing test passes locally for me, maybe a fluke?

swrichards:
> @swrichards The failing test passes locally for me, maybe a fluke?

I'll take a look, some kind of hidden dependency on run order I think.

src/open_inwoner/openzaak/zgw_imports.py
swrichards: That's fair, but the only other lines it could fail on are more or less covered by the typing, and the more narrow try/except would make it a bit harder to read. Given that we also stringify the error and log it, I think we should be safe.

pi-sigma (approved):

PR #2051, opened 2025-12-08
[#3603] Use different action + status texts for zaken/formulieren

pi-sigma:
## Summary

- 'Zaken' and 'Openstaande formulieren' receive different texts for the status as well as the action button/link. Also introduced an Enum to handle the case type ('Zaak' vs 'OpenSubmission').

- Refactored some classes, methods etc. to use Dutch terms referencing the domain (`submission` -> `formulier`, `SubmissionWithApiGroup` -> `FormulierWithApiGroup` etc.)

## Issue References

- Taiga User Story: https://taiga.maykinmedia.nl/project/open-inwoner/us/3603

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -104,6 +104,8 @@ Nieuwe features
   uit de API correct worden overgenomen in de lokale versie van de catalogus.
 * [:taiga-us:`3578`, :pr:`2043`]: Nieuw Front-end ontwerp voor CMS plug-in 'Balie Afspraken' gebouwd met
   design-tokens volgens het NLDS principe.
+* [:taiga-us:`3606`, :pr:`2051`]: Zaken en open formulieren hebben nu verschillende teksten voor de status
+  en actieknop/link.
 
 Bugfixes
 --------
--- a/src/open_inwoner/openzaak/api_models.py
+++ b/src/open_inwoner/openzaak/api_models.py
@@ -142,7 +142,6 @@ class Zaak(ZGWModel):
             "result": self.result_text,
             "zaaktype_config": getattr(self, "zaaktype_config", None),
             "statustype_config": getattr(self, "statustype_config", None),
-            "case_type": "Zaak",
         }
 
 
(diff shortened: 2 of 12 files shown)


src/open_inwoner/openzaak/api_models.py
```diff
@@ -142,7 +148,7 @@ def process_data(self) -> dict:
             "result": self.result_text,
             "zaaktype_config": getattr(self, "zaaktype_config", None),
             "statustype_config": getattr(self, "statustype_config", None),
-            "case_type": "Zaak",
+            "case_type": TypeAanvraag.ZAAK.value,
```
swrichards: Wouldn't it make more sense to do this at the level of the service? We enrich the core objects with other fields too there, e.g. the originating API group.

swrichards (commented):
I think this is a useful addition, I'd just add the enum at a higher level: in the case list service.

src/open_inwoner/openzaak/api_models.py
pi-sigma: Makes sense, didn't think of this option.

CHANGELOG.rst
```diff
@@ -93,12 +93,14 @@ Nieuwe features
   belangrijk kan zijn voor security audits (zoals PENTests) van bijvoorbeeld DigiD.
   Standaard blijven de hints ingeschakeld om het huidige gedrag te behouden.
 * [:taiga-us:`3580`, :pr:`2033`]: De zoekfunctie kan nu worden uitgeschakeld in de Algemene Configuratie.
-* [:taiga-is`3600`, :pr:`2046`]: 'Signing requets' toegevoegd aan admin. Beheerders kunnen nu een CSR
+* [:taiga-is:`3600`, :pr:`2046`]: 'Signing requets' toegevoegd aan admin. Beheerders kunnen nu een CSR
```
swrichards: Also: 
```suggestion
* [:taiga-is:`3600`, :pr:`2046`]: 'Signing requests' toegevoegd aan admin. Beheerders kunnen nu een CSR
```

swrichards (approved):

PR #2054, opened 2025-12-11
[#3619] Remove Webpack and use Vite

stefrado:
## Summary

- Remove Webpack and use Vite

closes #2035

## Issue References

Purge webpack -> https://taiga.maykinmedia.nl/project/open-inwoner/task/3619

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [x] CHANGELOG.rst updated
- [ ] Documentation updated


src/open_inwoner/scss/components/Typography/H3.scss
```diff
@@ -1,5 +1,4 @@
-@import '~@utrecht/heading-3-css/dist/index.css';
-@import '~microscope-sass/lib/responsive';
+@use '@utrecht/heading-3-css/dist/index.css';
```
jiromaykin, line 1: We still have a lot of vanilla CSS media queries instead of microscope-sass so I'm wondering wether we should fully use microscope-sass or not at. all.

src/open_inwoner/scss/pdf/overwrites/_index.scss
```diff
@@ -1,2 +1,2 @@
-@import './Base.scss';
-@import './Link.scss';
+@use './Base.scss';
```
jiromaykin, line 1: Good to know (using @forward to limit the use of these styles within PDF's only).

src/open_inwoner/scss/components/Form/Submit.scss
```diff
@@ -10,7 +11,7 @@
     background-color 300ms;
 
   &:hover {
-    background-color: darken(#248641, 5);
+    background-color: color.adjust(#248641, $lightness: -5%);
```
jiromaykin, line 14: I don't think the .submit class is actually used anywhere (?). We do have Type=submit for the buttons, but they seem to all use `--color-primary-darker` as hover background. But i guess it's interesting to see the modern syntax in action.

src/open_inwoner/scss/screen.scss
```diff
@@ -1,15 +1,15 @@
 // External
-// @import 'material-icons/iconfont/material-icons-outlined.woff2';
-@import 'material-icons/iconfont/material-icons.css';
-@import '~@fortawesome/fontawesome-free/scss/fontawesome.scss';
-@import '~@fortawesome/fontawesome-free/scss/regular.scss';
-@import '~@fortawesome/fontawesome-free/scss/brands.scss';
-@import '~leaflet/dist/leaflet.css';
-@import '~flatpickr/dist/flatpickr.min.css';
+// @use 'material-icons/iconfont/material-icons-outlined.woff2';
+@use 'material-icons/iconfont/material-icons.css';
```
jiromaykin, line 2: When bundling Vite now warns about "`Deprecation Warning [import]: Sass @import rules`" of these third party packages, so I'm wondering if it's possible to make Vite suppress those, but perhaps not?

vite.config.js
```diff
@@ -37,25 +37,32 @@ export default defineConfig({
...
+        admin_overrides: `${__dirname}/${paths.scssSrcDir}/admin/admin_overrides.scss`,
+        'pdf-p': `${__dirname}/${paths.scssSrcDir}/pdf/pdf_portrait.scss`,
+        'django-admin': `${__dirname}/${paths.jsSrcDir}/django-admin.js`,
       },
       output: {
         entryFileNames: '[name].js',
-        chunkFileNames: '[name].bundle.js',
+        chunkFileNames: '[name]-[hash].bundle.js',
+        // Remove hash from asset file name.
+        assetFileNames: '[name].[ext]',
```
jiromaykin, line 61: But why should the hash be removed from the filename? Isn't it better to have unique filenames so they won't get cached?

jiromaykin (approved):
The Purge is upon us :-) I actually don't see anything to improve here, just a bunch of questions.

vite.config.js
swrichards: There's no way to consistently refer to them from our Django templates' `{% static "js/assets.WHATSTHEHASH?.js %}`. This should be solved at the Django collectstatic level, which can also insert checksums into filenames, but plays nice with the static files finder.

package.json
```diff
@@ -9,10 +9,8 @@
   "scripts": {
     "build": "npm run collect && npm run bundle",
     "build-storybook": "npx storybook build",
-    "bundle": "conc npm:bundle:*",
-    "bundle:webpack": "webpack",
-    "bundle:react": "vite build && npm run makemessages && npm run compilemessages",
-    "collect": "cp -r node_modules/leaflet/dist/images src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff src/open_inwoner/static/bundles && cp -r node_modules/material-icons/iconfont/*.woff2 src/open_inwoner/static/bundles && cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
+    "bundle": "vite build && npm run makemessages && npm run compilemessages",
+    "collect": "cp -r node_modules/@fortawesome/fontawesome-free/webfonts src/open_inwoner/static/",
```
swrichards: Double check that the fonts still end up in the Docker container when building: the convoluted invocation of `cp` was I believe intended to also work from a fresh Docker build, before you have existing folder structures in place as you might have during local dev.

swrichards (changes requested):
LGTM, but one thing to check re Docker. This was a lot of pain to get right at the time, hence my caution, but if all is good consider this an approve.

src/open_inwoner/scss/components/Typography/H3.scss
stefrado: In my opinion we should go fully into it. Keeps every mq breakpoint the same.

package.json
stefrado: Ok in that case i might revert this change to make sure i do not mess up the Docker configuration. What i know is that vite also bundles the fonts to the dist folder. That is why I removed these lines. Maybe we can check tomorrow if it still works as expected with Docker.

src/open_inwoner/scss/screen.scss
stefrado: `quietDeps: true,` did the trick to silence the warnings.

package-lock.json
```diff
@@ -105,8 +105,7 @@
         "prettier-stylelint": "^0.4.2",
         "raw-loader": "^4.0.2",
         "regenerator-runtime": "^0.14.1",
-        "sass": "latest",
-        "sass-loader": "latest",
+        "sass-embedded": "lastest",
```
jiromaykin: "laStest"?? ` :-D  should be "latest",
but we're probably not using this package anymore?
Though that's out of scope perhaps here.

jiromaykin (changes requested):

package-lock.json
stefrado: It is actually recommended to use `sass-embedded`, instead of `sass`

swrichards (approved):

src/open_inwoner/scss/screen.scss
```diff
@@ -1,15 +1,15 @@
 // External
-// @import 'material-icons/iconfont/material-icons-outlined.woff2';
-@import 'material-icons/iconfont/material-icons.css';
-@import '~@fortawesome/fontawesome-free/scss/fontawesome.scss';
-@import '~@fortawesome/fontawesome-free/scss/regular.scss';
-@import '~@fortawesome/fontawesome-free/scss/brands.scss';
-@import '~leaflet/dist/leaflet.css';
-@import '~flatpickr/dist/flatpickr.min.css';
+// @use 'material-icons/iconfont/material-icons-outlined.woff2';
```
swrichards: Can we remove this line?

jiromaykin: The old line was from 2022 so I think we can safely remove this yes...

jiromaykin (changes requested):

jiromaykin (approved):

PR #2058, opened 2025-12-15
Resolve bug where HTML comment breaks accessibility speak functionality

stefrado:
## Summary

This PR fixes a bug where HTML comment breaks accessibility speak functionality

This bug occurred since we have the new ZakenHomePlugin (not yet in production).

### Deep dive

Error in browser console:
```
Uncaught TypeError: node.getAttribute is not a function
    at ReadOut.getText (
```

What happens is that js tries to use `getAttribute()` method from a HTML comment - which does not exists.

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/templates/cms/plugins/zaken/zaken.html
+++ b/src/open_inwoner/templates/cms/plugins/zaken/zaken.html
@@ -1,6 +1,5 @@
 {% load i18n icon_tags button_tags django_htmx %}
 
-<!-- start zaken plugin -->
 {% if show_zaken_plugin %}
     {% if request.htmx %}
         <oip-home-plugin-section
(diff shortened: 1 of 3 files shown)


stefrado:
@swrichards found this bug where the 'lees voor' functionality is broken due to the HTML comments in `zaken.html`. This new feature is not yet in production. Should I mention something about this bug inside the changelog? I don't think that it matters describing a bug that no user ever could have seen.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
swrichards, line 1: @stefrado dan mogen deze comments helemaal weg, ze waren alleen nuttig als placeholder in de dom.

swrichards (changes requested):
@stefrado no new entry in the changelog, but add this PR to the existing entry

swrichards (approved):

PR #2061, opened 2025-12-15
[#3637] Refactor sitemap

pi-sigma:
## Summary

- Removed duplicate contact form link from sitemap
- Added check for CMS page publication status in profile + sitemap. The  profile and sitemap failed to check for CMS page publication status when reversing URL's, which led to errors when a page was not published but we tried to display it anyways.
- Renamed the 'Platform' section for sitemap to 'General' and moved CMS pages not associated with 'Mijn Profiel' to this general section
- Restructured template: general CMS pages are now placed under `Algemeen -> Homepage`, unless they are explicitly configured to be displayed on `Mijn Profiel`.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3637, https://taiga.maykinmedia.nl/project/open-inwoner/issue/3321

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -185,6 +185,8 @@ Bugfixes
   wordt nu de correcte styling getoond.
 * [:taiga-ta:`3636`, :pr:`2068`]: Automatisch toegevoegen van een extra veelgestelde vraag in
   product admin is uitgezet.
+* [:taiga-is:`3537`, :taiga-is:`3321`, :pr:`2061`]: Sitemap bijgewerkt: duplicaten verwijderd,
+  structuur vereenvoudigd.
 
 Onderhoud
 ---------
--- a/src/open_inwoner/templates/pages/sitemap/sitemap.html
+++ b/src/open_inwoner/templates/pages/sitemap/sitemap.html
@@ -5,103 +5,41 @@
     <h1 class="utrecht-heading-1">{% trans "Links naar pagina's op " %}{{ site_name }}</h1>
 
     <div class="sitemap">
-        {# pages with information about OIP #}
+        {# All pages - flat list, alphabetically sorted #}
         <section class="sitemap__section">
-            <h2 class="utrecht-heading-2">{% trans "Platform" %}</h2>
             <ul class="utrecht-unordered-list">
-                {% for page in platform_pages %}
-                    {% if page.url_name %}
-                        <li class="utrecht-unordered-list__item">
+                {% for page in sitemap_pages %}
+                    <li class="utrecht-unordered-list__item">
+                        {% if page.url_name %}
                             <a href="{% url page.url_name %}" class="link--secondary">
                                 {% icon icon="chevron_right" icon_position="before" primary=True outlined=True %}
-                                <span class="link__text">{% trans page.link_text %}</span>
+                                <span class="link__text">{{ page.link_text }}</span>
                             </a>
-                        </li>
-                    {% elif page.url %}
-                        <li class="utrecht-unordered-list__item">
+                        {% elif page.url %}
                             <a href="{{ page.url }}" class="link--secondary">
                                 {% icon icon="chevron_right" icon_position="before" primary=True outlined=True %}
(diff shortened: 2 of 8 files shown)


src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -5,35 +5,57 @@
     <h1 class="utrecht-heading-1">{% trans "Links naar pagina's op " %}{{ site_name }}</h1>
 
     <div class="sitemap">
-        {# pages with information about OIP #}
+        {# General links #}
         <section class="sitemap__section">
-            <h2 class="utrecht-heading-2">{% trans "Platform" %}</h2>
+            <h2 class="utrecht-heading-2">{% trans "Algemeen" %}</h2>
```
jiromaykin: 1. More correct would be: to just change 'Algemeen' into "Home" and turn it into a link to the Home page,
just like Onderwerpen is link.

In that case you would not need an indent anymore, since everything falls under Home

2. in the issue https://taiga.maykinmedia.nl/project/open-inwoner/issue/3321 ?
They specifically want Uitkeringen to also fall under Home with an indent.

3. I'm using https://mijn.overheid.nl/sitemap/ as an example.
That one also contains a link to the Sitemap page itself which is probably how it should be done.

CHANGELOG.rst
```diff
@@ -178,6 +178,9 @@ Bugfixes
   toegoevegd, en de logica voor alle login types is opgeschoond.
 * [:taiga-is:`3599`, :pr:`2045`]: Bij het bekijken van de bron code van componenten in Storybook
   wordt nu de correcte styling getoond.
+* [:taiga-is:`3537`, :taiga-is:`3321`, :pr:`2061`]: Sitemap bijgewerkt: duplicaten verwijderd,
+  koptekst ‘Platform’ hernoemd naar ‘Algemeen’, weergave van niet-gepubliceerde CMS-pagina's
```
jiromaykin: So I'd put it all under 'Home' instead of 'Algemeen'.

jiromaykin (changes requested):

CHANGELOG.rst
jiromaykin: Change this changelog according to your latest changes,. so not using 'Algemeen'. Also you'll need to update your test.

src/open_inwoner/templates/pages/sitemap/sitemap.html
```diff
@@ -5,68 +5,71 @@
...
-                        </li>
-                    {% elif page.url %}
-                        <li class="utrecht-unordered-list__item">
-                            <a href="{{ page.url }}" class="link--secondary">
-                                {% icon icon="chevron_right" icon_position="before" primary=True outlined=True %}
-                                <span class="link__text">{% trans page.link_text %}</span>
-                            </a>
-                        </li>
-                    {% endif %}
+                {# Homepage #}
```
jiromaykin: Visually this looks like this for me since the last commit (see image). I see double pagelinks still.
I see the 3 pages "Zaken/ Samenwerkingen / Berichten" even when logged out. 
Also I will probably need to do some front-end work here. 

It could all then just be 1 UL with a 'column count' so there would only be 1 tree. This may make it easier for you also, so you'll just need to generate 1 tree for the total hierarchy, and you can remove the Sections.
Another example for this: https://www.utrecht.nl/over-deze-website/sitemap

So you could leave the styling up to the front-end.

Image

jiromaykin (changes requested):

src/open_inwoner/templates/pages/sitemap/sitemap.html
pi-sigma: I've simplified the structure. Everything except the links for categories/onderwerpen is a flat list now, sorted alphabetically.

jiromaykin (approved):
Looks much cleaner and better to me, 
the only thing I'd change: put the Homepage all the way at the top so it's the first page.

swrichards (approved):

PR #2064, opened 2025-12-16
Mark orphaned ZGW objects during sync with backend

swrichards:
## Summary

Over time, synced ZGW objects grow stale because they're no longer in the target backend, or appear to be duplicated, because the backend has a similarly named, but newer, version of a ZGW object. Auto-deleting objects we can't find in the backend would be too invasive, but this PR strikes a middle ground by adding a diagnostic flag to help users figure out if the object they're editing is, in fact, current according to the underlying API.


src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -626,6 +728,19 @@ def import_zaaktype_configs(self) -> ImportResult[ZaakTypeConfig]:
...
+        for key, config in existing_ztc.items():
+            if key not in zaaktype_keys_seen:
+                # This zaaktype no longer exists in the API
+                logger.info(
+                    "ZaakType config exists in database but not found in API",
+                    identificatie=config.identificatie,
+                    catalogus_url=config.catalogus.url if config.catalogus else None,
+                )
+                config.found_in_api = False
+                config.save()
```
pi-sigma: Is bulk-updating an option here?

```
orphaned_configs = []
...
orphaned_configs.append(config)
...
if orphaned_configs:
    ZaakTypeConfig.objects.filter(
        id__in=[c.id for c in orphaned_configs]
    ).update(found_in_api=False)
```

Similarly for the other config types

pi-sigma (approved):

src/open_inwoner/openzaak/zgw_imports.py
swrichards: Good point

PR #2065, opened 2025-12-16
[#3642] Remove 'Zaaknummer' text for Formulieren on Home

jiromaykin:
## Summary

'Zaaknummer' does not apply well to Openstaande Formulieren, 
where the 'identificatie' is both the name as well as a number.
So we will keep that empty but keep the stylistic spacing.

## Issue References

- Taiga Issue: https://taiga.maykinmedia.nl/project/open-inwoner/task/3642

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -106,8 +106,9 @@ Nieuwe features
   vlag zodat verweesde configuratie makkelijk kan worden geïdentificeerd.
 * [:taiga-us:`3578`, :pr:`2043`]: Nieuw Front-end ontwerp voor CMS plug-in 'Balie Afspraken' gebouwd met
   design-tokens volgens het NLDS principe.
-* [:taiga-us:`3606`, :pr:`2051`]: Zaken en open formulieren hebben nu verschillende teksten voor de status
-  en actieknop/link.
+* [:taiga-us:`3615`, :pr:`2051`, :pr:`2065`]: Zaken en open formulieren hebben nu
+  verschillende teksten voor de status en actieknop/link en op de Home pagina. Het
+  zaaknummer wordt alleen getoond voor reguliere zaken, niet voor formulieren.
 * [:taiga-us:`3606`]: Experimentele client geïmplementeerd met mock data ter ondersteuning van het ontwikkelen
   van de 'Mijn Afval' module.
 * [:taiga-us:`3607`, :pr:`2075`]: Basisapp ‘Mijn Afval’ geïmplementeerd en geïntegreerd met Django CMS.
--- a/src/open_inwoner/cms/plugins/views.py
+++ b/src/open_inwoner/cms/plugins/views.py
@@ -15,6 +15,7 @@ from open_inwoner.cms.cases.views.services import CaseListService
 from open_inwoner.cms.plugins.models import CMSZakenPluginConfig
 from open_inwoner.cms.plugins.models.zaken import MAX_CASES_DEFAULT, MIN_CASES
 from open_inwoner.htmx.mixins import RequiresHtmxMixin
+from open_inwoner.openzaak.constants import TypeAanvraag
 from open_inwoner.openzaak.types import UniformCase
 
 logger = structlog.stdlib.get_logger(__name__)
@@ -24,6 +25,19 @@ class ZakenPluginContentView(RequiresHtmxMixin, CaseLogMixin, View):
     """
     HTMX endpoint that fetches and returns content for the `CMSZakenPlugin`.
     """
+
+    def _get_detail_label(self, zaak: dict) -> str:
+        """
+        Generate the detail label to display under the card title.
+        """
+
+        is_formulier = zaak.get("type_aanvraag") == TypeAanvraag.FORMULIER.value
+        identification = zaak.get("identification", "")
+
+        if is_formulier or not identification:
+            return ""
+
+        return _("Zaaknummer: %(identification)s") % {"identification": identification}
 
     def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
         case_service = CaseListService(request)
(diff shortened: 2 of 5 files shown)


jiromaykin:
@pi-sigma @swrichards Not sure if this is the best solution, since the Openstaande Formulieren show a combined name/number, but it's still not a 'Zaaknummer' so perhaps a 'Formulier' - see also screenshot in Taiga link.

src/open_inwoner/cms/plugins/views.py
```diff
@@ -112,6 +112,7 @@ def get(self, request: HttpRequest, plugin_id: int) -> HttpResponse:
                     # for 'formulieren':
                     # "naam" from API -> "description" for web component
                     "description": zaak.get("naam", zaak.get("description", "")),
+                    "is_formulier": zaak.get("type_aanvraag") == "Formulier",
```
pi-sigma: Use the constant:

Top of file:
`from open_inwoner.openzaak.constants import TypeAanvraag`

Then:
```suggestion
                    "is_formulier": zaak.get("type_aanvraag") == TypeAanvraag.FORMULIER.value,
```

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -13,10 +13,15 @@
             {% for zaak in zaken %}
                 <oip-home-plugin-card
                     title="{{ zaak.description }}"
-                    description="{% if zaak.identification %}{% trans "Zaaknummer: " %}{{zaak.identification}}{% endif %}"
+                    description="{% if zaak.identification %}
+                                    {% if zaak.is_formulier %}
+                                        {% trans "Formulier: " %}
+                                    {% else %}{% trans "Zaaknummer: " %}
+                                    {% endif %}
+                                {{ zaak.identification }}{% endif %}"
```
pi-sigma: Should we use a default value for `description`? Just a thought.

pi-sigma (changes requested):

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
jiromaykin: Yes, it could be a good idea to not show Description for a Formulier and only show a hard coded text "Formulier afronden" which doesn't need to be part of the Button (there is no button text in the design)

swrichards: Happy to try out the "Formulier afronden" approach. To be clear, then we would also not sure `zaak.identification`, because that's already in the title, correct?

swrichards (commented):
Let's see what the proposed change in the text for formulieren looks like and re-review.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
jiromaykin: Yes, but i am not sure if ALL cases have a description? If they have none, that same line will be empty and the card will shrink,
i've put the question up for the designer to think about.

jiromaykin: As requested by Joost and in Taiga: we are keeping this empty for Forms but keep stylistic spacing: 
https://taiga.maykinmedia.nl/project/open-inwoner/task/3642

swrichards (approved):

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -13,10 +13,10 @@
             {% for zaak in zaken %}
                 <oip-home-plugin-card
                     title="{{ zaak.description }}"
-                    description="{% if zaak.identification %}{% trans "Zaaknummer: " %}{{zaak.identification}}{% endif %}"
+                    description="{% if not zaak.is_formulier and zaak.identification %}{% trans "Zaaknummer: " %}{{ zaak.identification }}{% endif %}"
```
pi-sigma: Template logic is a bit complex and could be moved to the view (I've added `detail_label` and the helper method):

```
# In the `get` method of `ZakenPluginContentView`:
case_data = {
    "uuid": zaak["uuid"],
    "url": f_url.url,
    "identification": zaak.get("identification", ""),
    "description": zaak.get("naam", zaak.get("description", "")),
    "is_formulier": zaak.get("type_aanvraag") == TypeAanvraag.FORMULIER.value,
    "detail_label": self._get_detail_label(zaak),
}

# Below:
def _get_detail_label(self, zaak: dict) -> str:
    """
    Generate the detail label to display under the card title.
    """

    is_formulier = zaak.get("type_aanvraag") == TypeAanvraag.FORMULIER.value
    identification = zaak.get("identification", "")
    
    if is_formulier or not identification:
        return ""
    
    return _("Zaaknummer: %(identification)s") % {"identification": identification}
```

Then in the template:
```
description="{{ zaak.detail_label }}"
```

CHANGELOG.rst
```diff
@@ -106,8 +106,8 @@ Nieuwe features
   vlag zodat verweesde configuratie makkelijk kan worden geïdentificeerd.
 * [:taiga-us:`3578`, :pr:`2043`]: Nieuw Front-end ontwerp voor CMS plug-in 'Balie Afspraken' gebouwd met
   design-tokens volgens het NLDS principe.
-* [:taiga-us:`3606`, :pr:`2051`]: Zaken en open formulieren hebben nu verschillende teksten voor de status
-  en actieknop/link.
+* [:taiga-us:`3606`, :pr:`2051`, :pr:`2065`]: Zaken en open formulieren hebben nu verschillende teksten
+  voor de status en actieknop/link en op de Home pagina.
```
pi-sigma: You could add that "Het zaaknummer wordt alleen getoond 
  voor reguliere zaken, niet voor formulieren." since this is the main point of the PR

pi-sigma (approved):
Looking good, Two small suggestions, but they are cosmetic

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
jiromaykin: That'd be great!

CHANGELOG.rst
jiromaykin: But the other PR's mentioned here are not just this 1 change, it's more than that.

swrichards: I added both.

PR #2086, opened 2026-01-05
Document the Github-based workflow

swrichards:
## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables

----
📚 Documentation preview 📚: https://open-inwoner--2086.org.readthedocs.build/en/2086/


docs/development/workflow.rst
```diff
@@ -0,0 +1,182 @@
...
+
+**Issue Types** *(set at creation/refinement)*
+  * **Feature**: User-facing functionality, generally mapping to features/user stories
+    from Jira
+  * **Bug**: Verified defects requiring fixes
+  * **Task**: Developer-driven chores, cleanups, and technical improvements
+  * **Epic**: Currently not in use
+
+**Issue Attributes** *(set at creation/refinement)*
+  * **Milestone**: Issues should be assigned to one of the active milestones, virtually
```
jiromaykin, line 130: We haven't really made use of milestones ourselves yet I think? So then milestones would need to b created first, etc, etc, I'm also sort of hoping for a 'shadow workflow' for developers to use as steps that we'd have to do as a team :-) 
Also: our naming of PR branches has become a bit of a mess (_issue/task/feat/feature/fix_blabla) and I'd prefer them to be consistent with these 'feature/bug/tas' names but that's just a personal hang-up.

swrichards: We've only used them implicitly (the milestone was always the next release!), but they're fairly lightweight to create, and I doubt we'd ever have more than 1-3 at any given time. In fact, we could say that we only use milestones for the sprints, so we can easily see what was done in what sprint. I don't think it makes sense to use them for much else really, I'll amend the docs to say that.

I think there's a lot of space for "shadow" workflows: the board allows for many filters, and you can build different views for different purposes (e.g. based on a different tagging schema).

PR naming: agreed, but that's a different topic :) (related also to commit conventions). We'll turn to that soon.

docs/development/workflow.rst
```diff
@@ -0,0 +1,182 @@
...
+Issues come from three primary channels:
+
+**Taiga**
+  Customer support and incident response. Triaged by product owners and Maykin support
+  staff. Most issues will be reported to the main open inwoner project
+  (https://taiga.maykinmedia.nl/project/open-inwoner), though some issues might be
+  reported in client-specific environments as well.
+
+**Jira (Dimpact)**
+  Feature specifications created and refined by product owners.
```
jiromaykin, line 23: Here I would expect explicit mention of User Stories instead of 'issues'.

docs/development/workflow.rst
```diff
@@ -0,0 +1,182 @@
...
+
+        TD --> IP
+        IP --> IR
+        IR --> DN
+
+        GH -.->|External Issue URL| H
+        GH -.->|External Issue URL| I
+
+**No Status**
+  Issues can be added here by anyone for discussion during refinement or weekly
```
jiromaykin, line 81: These 'no status' issues can only be added to an active Github sprint I think? 
(same for the chapter "Issue Creation" >> 'Developer') 
I am a bit unsure on what to do with our long list of free floating issues that do not come from sprints that are in 'danger' of never being looked at again: https://taiga.maykinmedia.nl/project/open-inwoner/issues?order_by=status

PR #2090, opened 2026-01-05
chore: release 2.0

swrichards:


swrichards:
Released as tag, and merged with #2093

PR #2092, opened 2026-01-05
[#2091] Allow OIDC logout via GET request for staff users

### LINKED ISSUE #2091: Admin OIDC logout
Taiga: https://taiga.maykinmedia.nl/project/open-inwoner/task/3635

pi-sigma:
## Summary

Whenn staff users log in via OIDC (eIDAS), they get `login_type=LoginTypeChoices.oidc`. The `get_logout_url()` method returns `reverse("oidc_logout")` which mapped to mozilla_django_oidc's `OIDCLogoutView`. This view only accepts POST requests by default. A `GenericOIDCLogoutView` is added to allow OIDC users to log out by following a link (GET request).

## Issue References

Closes #2091 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,8 @@ Nieuwe features
 Bugfixes
 --------
 
+* [:gh-issue:`2091`]: Beheerders toestaan om uit te loggen via frontend.
+
 Onderhoud
 ---------
 
--- a/src/open_inwoner/accounts/tests/test_oidc_views.py
+++ b/src/open_inwoner/accounts/tests/test_oidc_views.py
@@ -3166,3 +3166,106 @@ class EIDASOIDCFlowTests(WebTest):
             "duplicate-pseudo-id",
             msg="User's eIDAS pseudo_id must remain unchanged",
         )
+
+
+@override_settings(ROOT_URLCONF="open_inwoner.cms.tests.urls")
+class GenericOIDCLogoutViewTests(WebTest):
+    """Tests for the generic OIDC logout view (for staff/admin users)."""
+
+    def test_generic_oidc_logout_accepts_get_request_for_oidc_users(self):
+        """
+        Test that the generic OIDC logout view accepts GET requests for users
+        with login_type=oidc.
+
+        Needed because staff users who log in via OIDC need to be able to log
+        out using a simple link (GET request), not a form POST.
+        """
+        user = UserFactory.create(login_type=LoginTypeChoices.oidc)
+        self.client.force_login(user)
+
+        logout_url = reverse("oidc_logout")
+
+        # Verify user is logged in
+        self.assertTrue(self.client.session.get("_auth_user_id"))
+
+        # Perform logout via GET request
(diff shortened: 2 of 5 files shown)


src/open_inwoner/urls.py
```diff
@@ -118,6 +119,7 @@
         "sessions/",
         include("open_inwoner.extended_sessions.urls", namespace="sessions"),
     ),
+    path("oidc/logout/", generic_oidc_logout, name="oidc_logout"),
```
swrichards, line 122: For my understanding: what guarantees that this link will not be used for DigiD/eHerkenning users? This link is only actually used in the admin panel?

pi-sigma: Good point. The frontend routes users based on their login type:
- DigiD users: `login_type=digid` -> `/digid-oidc/logout`
- eHerkenning users: `login_type=eherkenning` -> `/eherkenning-oidc/logout/`

The `login_type` is set by the authentication backend (e.g. `DigiDOIDCBackend`). However, the generic logout view currently does not prevent a DigiD/eHerkenning/Eidas users from manually navigating to `oidc/logout/`, which means their session at the identity provider would remain active.

I've added a check for this, the view now redirects users to their proper logout route.

src/open_inwoner/accounts/tests/test_oidc_views.py
```diff
@@ -3166,3 +3166,74 @@ def test_duplicate_pseudo_id_prevents_user_creation(
...
+        response = self.client.get(reverse("oidc_logout"))
+
+        # Should redirect to the correct eHerkenning logout URL
+        expected_logout_url = user.get_logout_url()
+        self.assertRedirects(
+            response, expected_logout_url, fetch_redirect_response=False
+        )
+
+        # User should still be logged in (redirect happened before logout)
+        self.assertTrue(self.client.session.get("_auth_user_id"))
```
swrichards: This pattern can be generalized as subtests to loop over _all_ LoginType values _except_ the generic oidc value.

swrichards (changes requested):

src/open_inwoner/accounts/tests/test_oidc_views.py
pi-sigma: Yes, but it's slightly annoying because the login type must be coordinated with the right user factory. That's why I went for the simple copy/paste approach. I'll take another look though.

CHANGELOG.rst
```diff
@@ -216,6 +216,7 @@ Bugfixes
   product admin is uitgezet.
 * [:taiga-is:`3537`, :taiga-is:`3321`, :pr:`2061`]: Sitemap bijgewerkt: duplicaten verwijderd,
   structuur vereenvoudigd.
+* [:taiga-is:`3653`, :pr:`2092`]: Beheerders toestaan om uit te loggen via frontend.
```
swrichards: This guy needs to move up top

swrichards (approved):

swrichards (approved):

PR #2095, opened 2026-01-06
Add pyproject toml

### LINKED ISSUE #2100: Add a pyproject.toml manifest

swrichards:
Part of general efforts to make the repo more compliant with contemporary Python standards.


pi-sigma (approved):
We could move `.prettierrc` , `.babelrc`, and `chromatic.config.json` to `package.json` to consolidate our configuration, but that can be done separately

swrichards:
> We could move `.prettierrc` , `.babelrc`, and `chromatic.config.json` to `package.json` to consolidate our configuration, but that can be done separately

Indeed, and perhaps `.babelrc` (like `karma.conf.js`) is not even needed in the new frontend infra. To investigate @stefrado @jiromaykin

jiromaykin:
> Indeed, and perhaps `.babelrc` (like `karma.conf.js`) is not even needed in the new frontend infra...

Yes, i thought we had already removed a lot of babel, but need to investigate more.

PR #2106, opened 2026-01-07
[#2105] fix: use proper kvk lookup in eHerkenning SAML backend

### LINKED ISSUE #2105: eHerkenning SAML backend uses incorrect claim getter for KvK number
### Describe the bug

[Our eHerkenning SAML backend uses `get_kvk_number` to fetch the user's KvK](https://github.com/maykinmedia/open-inwoner/blob/v2.0.0/src/eherkenning/backends.py#L32-34), when this should be `get_legal_subject_kvk`.

`get_kvk_number` looks up the `urn:etoegang:1.11:attribute-represented:KvKnr` (which is optional and used for delegation/representation purposes) whereas what we need is [urn:etoegang:1.9:EntityConcernedID:KvKnr](https://afsprakenstelsel.etoegang.nl/Startpagina/as/urn-etoegang-1-9-entityconcernedid-kvknr) which is returned by `get_legal_subject_kvk`.

Image

The fix is to replace `get_kvk_number` with `get_legal_subject_kvk`.

swrichards:


pi-sigma (approved):

alextreme (approved):
Verified based on SAML attributes discussed on 21 may 2021 with BD "HandelendePartijnaam en KVK-nummer", the structure of the saml_attributes is an exact match for these

PR #2113, opened 2026-01-08
✨ Create afval chart

### LINKED ISSUE #2128: Web Component Chart for mijn afval

stefrado:
## Summary

Implement a Chart using `chart.js` for mijn afval.

## Issue References

- Github Issue: #2128

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Nieuwe features
   in Docker/Kubernetes.
 * [:gh:`2107`]: Productie logs worden weggeschreven als JSON voor gebruik in log analyse
   tools.
+* [:gh:`2113`]: Diagram (chart.js) toegevoegd ten behoeve van de ‘Mijn Afval’ app.
 
 Bugfixes
 --------
--- a/package.json
+++ b/package.json
@@ -60,7 +60,9 @@
     "@utrecht/web-component-library-stencil": "^3.5.0",
     "bem.js": "^1.0.10",
     "caniuse-lite": "^1.0.30001707",
+    "chart.js": "^4.5.1",
     "clsx": "^2.1.1",
+    "date-fns": "^4.1.0",
     "emojibase-data": "^7.0.1",
     "escape-html": "^1.0.3",
     "flatpickr": "^4.6.9",
@@ -144,7 +146,8 @@
   },
   "overrides": {
     "form-data": ">=2.5.4 || <3.0.0 || >=3.0.4 || <4.0.0 || >=4.0.4",
-    "playwright": ">=1.57.1"
+    "playwright": ">=1.57.1",
+    "date-fns": "^4.1.0"
   },
   "optionalDependencies": {
     "@esbuild/linux-x64": "*",
--- a/src/open_inwoner/react/components/Chart/Chart.scss
+++ b/src/open_inwoner/react/components/Chart/Chart.scss
@@ -1,0 +1,9 @@
+.afval-chart {
+  border: 1px solid var(--color-gray);
+  border-radius: 2px;
+
+  &__container {
+    min-height: 320px;
+    max-height: 400px;
+  }
+}
--- a/src/open_inwoner/react/components/Chart/Chart.tsx
+++ b/src/open_inwoner/react/components/Chart/Chart.tsx
@@ -1,0 +1,67 @@
+import { Chart, registerables } from 'chart.js';
+import { useEffect, useRef } from 'preact/hooks';
+import './Chart.scss';
+import { AfvalData, ChartDataPoint, ChartPeriod } from './types';
+import { useChart } from './useChart';
+import { AnyComponent as AC } from 'preact';
+import { FormattedMessage } from 'react-intl';
+
+Chart.register(...registerables);
+
+export interface IChartProps {
+  data?: AfvalData;
+  period?: ChartPeriod;
+  dataId?: string;
+}
+
+const OIPChart: AC<IChartProps> = ({ data, period = 'month', dataId }) => {
+  const canvasRef = useRef<HTMLCanvasElement>(null);
+  const chartRef = useRef<Chart<'bar', ChartDataPoint[]> | null>(null);
+
+  // Generate chart configuration using custom hook
+  const config = useChart(data, dataId, period);
+
+  // Create or update chart when config changes
+  useEffect(() => {
+    if (!config || !canvasRef.current) return;
+
+    if (!chartRef.current) {
+      // Create chart if it doesn't exist yet
+      chartRef.current = new Chart(canvasRef.current, config);
+    } else {
+      // Update existing chart
+      chartRef.current.data = config.data;
+      if (config.options) {
+        chartRef.current.options = config.options;
+      }
+      chartRef.current.update();
+    }
+  }, [config]);
+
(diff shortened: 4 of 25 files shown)


package.json
```diff
@@ -74,6 +75,7 @@
     "leaflet": "^1.7.1",
     "material-icons": "^1.13.12",
     "microscope-sass": "latest",
+    "moment": "^2.30.1",
```
swrichards: I assume you've looked at this, but dropping it here just to be sure: https://momentjs.com/docs/#/-project-status/

stefrado: I changed the usage of `moment` to `date-fns`. This package is used to get the current week number.

CHANGELOG.rst
```diff
@@ -18,6 +18,7 @@ Nieuwe features
   gebruik in Docker/Kubernetes.
 * [:gh-issue:`2107`, :pr:`2117`]: Productie logs worden weggeschreven als JSON voor
   gebruik in log analyse tools.
+* [:gh-issue:`2113`]: Chart component (chart.js) toegevoegd ten behoeve van de ‘Mijn Afval’ app.
```
jiromaykin: "Chart" is wellicht niet zo duidelijk voor de lezers? Beter: "diagram".

src/open_inwoner/react/components/Chart/Chart.tsx
```diff
@@ -0,0 +1,56 @@
...
+      }
+    };
+  }, [config]);
+
+  return (
+    <div style={{ minHeight: '320px', position: 'relative' }}>
+      <canvas
+        ref={canvasRef}
+        aria-label="Chart containing Mijn Afval data"
+        role="img"
```
jiromaykin: The `role="img"` approach would require a very detailed aria-label or even ALT to convey all the chart information meaningfully. 

But since users can already access the complete data through the accessible Tables below, it would better to hide the entire chart for screenreaders like so:

`<oip-chart data-id="afval_data" aria-hidden="true">`

So both `aria-label` and `role` should be removed from the canvas.

The Tables are a LOT to read, but one day they will be filtered :-) and: if they are real tables they will be structured data that is much easier for screenreaders to understand, and for users to 'skip' through.

src/open_inwoner/react/components/Chart/config.ts
```diff
@@ -0,0 +1,103 @@
...
+    title: 20,
+    legend: 14,
+    axis: 12,
+  },
+  fontWeights: {
+    regular: 400,
+    bold: 700,
+  },
+  colors: {
+    gft: '#22c55e',
```
jiromaykin: Fun: are these the actual colors of the waste baskets?
If these are static it would solve design-token variable issues 😄

src/open_inwoner/react/components/Chart/config.ts
```diff
@@ -0,0 +1,103 @@
...
+    align: 'start',
+  },
+  legend: {
+    position: 'bottom',
+    align: 'center',
+  },
+} as const;
+
+export const chartOptions = (title: string = `Chart`): ChartOptions<'bar'> => ({
+  maintainAspectRatio: false,
```
jiromaykin, line 57: This makes the chart a bit narrow on mobile, but I'm assuming the mobile design is not completely thought out yet.

We do not support screens smaller than 360px wide, maybe we want the Canvas to scroll horizontally on small screens?

But that is more of a design question.

src/open_inwoner/react/components/Chart/Chart.scss
```diff
@@ -0,0 +1,4 @@
+canvas {
```
jiromaykin: Can you add a class to the canvas? 
Styling tags will push this style to everything-everywhere-all-at-once.

src/open_inwoner/react/components/Chart/Chart.tsx
```diff
@@ -0,0 +1,56 @@
+import { Chart, registerables } from 'chart.js';
+import { useEffect, useRef } from 'preact/hooks';
+import './Chart.scss';
+import { AfvalData, ChartDataPoint, ChartPeriod } from './types';
+import { useChart } from './useChart';
+import { AC } from 'preact';
```
jiromaykin: `AC`? Should all of these instead be `import { AnyComponent as AC } from 'preact';` for Preact?

src/open_inwoner/react/components/Chart/Chart.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }, [config]);
+
+  return (
+    <div style={{ minHeight: '320px', position: 'relative' }}>
+      <canvas
+        ref={canvasRef}
+        aria-label="Chart containing Mijn Afval data"
+        role="img"
+      >
+        <p>Chart containing Mijn Afval data</p>
```
jiromaykin: This text will only render if Canvas is not supported by the browser, so this is not text for Accessibility 😊

Usually you would put some kind of fallback text about 'please use modern browser' or something.

Or in this case:

`<p>De grafiek kan niet worden weergegeven. Zie de tabellen hieronder voor de gegevens.</p>`

src/open_inwoner/react/components/Chart/Chart.tsx
```diff
@@ -0,0 +1,56 @@
...
+  period?: ChartPeriod;
+  dataId?: string;
+}
+
+const OIPChart: AC<IChartProps> = ({ data, period = 'month', dataId }) => {
+  const canvasRef = useRef<HTMLCanvasElement>(null);
+  const chartRef = useRef<Chart<'bar', ChartDataPoint[]> | null>(null);
+
+  // Generate chart configuration using custom hook
+  const config = useChart(data, dataId, period);
```
jiromaykin, line 22: chart gets destroyed and recreated on every render?

jiromaykin (changes requested):
Looking gorgeous.

src/open_inwoner/react/components/Chart/Chart.tsx
stefrado: I don't get it, but the TS types of preact keep on changing in my editor and now `AC` is valid and `AnyComponent as AC` is not.

src/open_inwoner/react/components/Chart/Chart.tsx
stefrado: Yes i'll fix this

src/open_inwoner/react/components/Chart/config.ts
stefrado: I found this: 

https://afvaleducatie.nl/images/bibliotheek/Richtlijnen_voor_gebruik_pictogrammen_afvalscheiding.pdf

src/open_inwoner/react/components/Chart/config.ts
stefrado: In my opinion - it looks better with `maintainAspectRatio: false`

src/open_inwoner/react/components/Chart/config.ts
jiromaykin: Awesome!!

jiromaykin (approved):
Looks great (and my screenreader properly hops towards the Table), wondering how this will all work together with the filters maybe in future.

swrichards (approved):

PR #2135, opened 2026-01-13
fix: [#2099] concurrency control for Storybook should only be on deploy

### LINKED ISSUE #2099: Publish Storybook to GitHub Pages
Apparently no Chromactic 'project' is needed, so no financial billing that could affect the use for Open-Formulieren (it would be joint billing in Chromatic)

Github pages are free

first create Github page  `https://<github-org>.github.io/<repository>/`

Then likely just literally take over the Storybook & Deploy jobs from https://github.com/open-formulieren/open-forms/blob/master/.github/workflows/sb.yml

swrichards:
The concurrency constraint applied to all Storybook builds, not just
deploys, which meant that you would sometimes get failing storybook
CI on feature branches, where the build step is required to succeed.

### DIFF
--- a/.github/workflows/storybook.yml
+++ b/.github/workflows/storybook.yml
@@ -15,11 +15,6 @@ on:
       - ready_for_review
   workflow_dispatch:
 
-# Allow one concurrent deployment
-concurrency:
-  group: 'pages-${{ github.ref_name }}' # unique builds for branch/tag name
-  cancel-in-progress: true
-
 jobs:
   storybook:
     name: Create storybook build
@@ -47,6 +42,11 @@ jobs:
       contents: read
       pages: write
       id-token: write
+
+    # Allow one concurrent deployment
+    concurrency:
+      group: 'pages-${{ github.ref_name }}' # unique builds for branch/tag name
+      cancel-in-progress: true
 
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/develop' && github.event_name == 'push' # Exclude PRs


.github/workflows/storybook.yml
```diff
@@ -48,6 +43,11 @@ jobs:
       pages: write
       id-token: write
 
+    # Allow one concurrent deployment
+    concurrency:
+      group: 'pages-${{ github.ref_name }}' # unique builds for branch/tag name
+      cancel-in-progress: true
```
jiromaykin, line 49: What would be the expected behaviour on Github?
Your own PR 'skipped' the storybook check :-) 
So you will likely still get notifications for this

Image

.github/workflows/storybook.yml
```diff
@@ -48,6 +43,11 @@ jobs:
       pages: write
       id-token: write
 
+    # Allow one concurrent deployment
+    concurrency:
```
stefrado, line 47: Just looked at the code of open-forms and this code is added for chromatic since they added the chromatic deployment. Because we don't have this in our CI, we might be able to remove this code completely. The deploy CI is already cancelled for PR's.

The only thing where i am not sure about is what will happen if we remove this code and have two commit's (to develop) after each other deploy the new SB. Do we then want to cancel the first deployment and only leave the last one? If that is the case then we might want to keep your changes.

.github/workflows/storybook.yml
swrichards: It didn't skip the build step though, only the deploy step, which is what we want!

image

The behavior should be: the build storybook job is _always_ run, for each PR and run on `develop`. In contrast, de `deploy` task is run only on `develop`, and if another merge occurs on `develop` before the previous `deploy` run is complete, the older `deploy` run is cancelled in favour of the new one.

.github/workflows/storybook.yml
swrichards: >The only thing where i am not sure about is what will happen if we remove this code and have two commit's (to develop) after each other deploy the new SB. Do we then want to cancel the first deployment and only leave the last one? If that is the case then we might want to keep your changes.

I don't think it really matters if we deploy to Github pages more frequently -- I suspect the Chromatic concurrency check is because the subscription meters on the number of deploys.

I say we just merge it and see if it works, if it does what we think (cancel in-progress deploys if a new one is triggered), that behaviour is fine by me.

PR #2136, opened 2026-01-15
ci: detect package lock drift in frontend ci

swrichards:


swrichards:
Should already be done by `npm ci`

PR #2151, opened 2026-01-20
[#2119] Create config model and client for Mijn Afval

### LINKED ISSUE #2119: Create OpenAfval API client + configuration

### LINKED ISSUE #2182: Mijn Afval CMS page configuration
### Describe the bug

When configuring the 'Mijn Afval' CMS page via the admin, we currently have to select one out of potentially many apphook configurations, even if there is only one. This is different from the CMS page configuration for 'Mijn profiel'; both should work the same way (the behavior of 'Mijn profiel' is the expected one).

pi-sigma:
## Summary

Created config model for 'Mijn Afval' and basic client for interacting with OpenAfval API

## Issue References

Closes #2119 

Fixes #2182 

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/mijn_afval/views.py
```diff
@@ -216,19 +287,33 @@ def dispatch(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        client = AfvalApiClient(base_url="")
-        data = client.fetch_bag_objects_for_bsn(bsn=self.request.user.bsn)
+        api_client = build_zgw_client(
+            service=MijnAfvalConfig.get_solo().openafval_service,
+            client_factory=OpenAfvalAPIClient,
+        )
```
swrichards, line 314: I think we should handle the "no service configured case" with a `message.error(self.request, "This module is not yet configured")` or some such (should be an edge case, but you never known).

src/open_inwoner/mijn_afval/clients.py
```diff
@@ -1,26 +1,33 @@
...
+            logger.exception("Network error when fetching Afval profiel")
+            raise MijnAfvalException from exc
+
+        try:
+            response_json = response.json()
+        except requests.exceptions.JSONDecodeError as exc:
+            logger.exception("Invalid JSON from OpenAfval API")
+            raise MijnAfvalException from exc
+        else:
+            return AfvalProfiel.model_validate(response_json)
```
swrichards, line 35: I think the `else` is redundant: let's also catch the pydantic validation error and re-raise as domain exception to avoid hard 500 pages

swrichards (changes requested):
LGTM, just small things

CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ Nieuwe features
 * [:gh:`2107`]: Productie logs worden weggeschreven als JSON voor gebruik in log analyse
   tools.
 * [:gh:`2113`]: Diagram (chart.js) toegevoegd ten behoeve van de ‘Mijn Afval’ app.
+* [:gh:`2199`]: API-client en configuratie voor 'Mijn Afval' aangemaakt.
```
swrichards: Somehow failing: https://github.com/maykinmedia/open-inwoner/issues/2199

pi-sigma: Broken link due to wrong issue number

src/open_inwoner/mijn_afval/views.py
```diff
@@ -216,19 +287,43 @@ def dispatch(self, request, *args, **kwargs):
...
+        if not self.request.user.bsn:
+            return context
+
+        # fetch data from API
+        afval_config = MijnAfvalConfig.get_solo()
+
+        if not (openafval_service := afval_config.openafval_service):
+            messages.error(_("This module is not yet configured"))
+            return context
+
```
swrichards, line 310: Maybe one more guard here: `if not self.request.user.bsn` message "Only available to DigiD users"

swrichards (approved):
One final nitpick, then good to merge

swrichards (approved):

PR #2195, opened 2026-02-03
[#2192] Implement 'Mijn afval' filters

### LINKED ISSUE #2192: Open Afval Filter Backend - Frontend
To add a new feature request please use this template.

### Theme

[E.g. Frontend, Admin, Design]

### Description

A description of the idea or need, possibly supplemented with a specific
solution direction

### Added value

The added value for business operations or services

### Additional context

Any other context or screenshots about the feature request

pi-sigma:
## Summary

Added suuport for filtering ledigingen and containers by address, waste type, and date range.

## Issue References

Closes #2192 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/mijn_afval/api_models.py
+++ b/src/open_inwoner/mijn_afval/api_models.py
@@ -21,15 +21,6 @@ class Periode(PydanticModel):
     laatste_lediging: datetime
 
 
-class Summary(PydanticModel):
-    totaal_gewicht: float
-    totaal_gewicht_per_afval_type: dict[str, float]
-    aantal_ledigingen: int
-    aantal_containers: int
-    aantal_container_locaties: int
-    periode: Periode
-
-
 class AfvalContainer(PydanticModel):
     id: str
     afval_type: str
@@ -55,7 +46,6 @@ class AfvalLediging(PydanticModel):
 
 class AfvalProfiel(PydanticModel):
     klant: Klant
-    summary: Summary
     containers: list[AfvalContainer]
     container_locaties: list[AfvalContainerLocatie]
     ledigingen: list[AfvalLediging]
--- a/src/open_inwoner/mijn_afval/clients.py
+++ b/src/open_inwoner/mijn_afval/clients.py
@@ -10,9 +10,42 @@ logger = structlog.stdlib.get_logger(__name__)
 
 
 class OpenAfvalAPIClient(APIClient):
-    def get_afval_profiel(self, bsn: str) -> AfvalProfiel:
+    def get_afval_profiel(
+        self,
+        bsn: str,
+        adressen: list[str] | None = None,
+        afval_type: str | None = None,
+        startdatum: str | None = None,
+        einddatum: str | None = None,
+    ) -> AfvalProfiel:
+        """
+        Fetch AfvalProfiel for a BSN with optional filters.
+
+        Args:
+            bsn: BSN of the customer
+            adressen: Filter container locations by address (multiple allowed)
+            afval_type: Filter containers by waste type (e.g., 'gft', 'restafval')
+            startdatum: Filter ledigingen from this date (YYYY-MM-DD)
+            einddatum: Filter ledigingen until this date (YYYY-MM-DD)
+
+        Returns:
+            AfvalProfiel model instance
+
+        Raises:
+            MijnAfvalException: On any API or validation error
+        """
+        params = {}
+        if adressen:
+            params["adres"] = adressen
+        if afval_type:
+            params["afval-type"] = afval_type
+        if startdatum:
+            params["startdatum"] = startdatum
+        if einddatum:
+            params["einddatum"] = einddatum
+
         try:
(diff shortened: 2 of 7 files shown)


src/open_inwoner/mijn_afval/clients.py
```diff
@@ -10,9 +10,42 @@
 
 
 class OpenAfvalAPIClient(APIClient):
-    def get_afval_profiel(self, bsn: str) -> AfvalProfiel:
+    def get_afval_profiel(
+        self,
+        bsn: str,
+        adres: str | None = None,
+        afval_type: str | None = None,
```
stefrado: These two can be list[str] and should be treated like this.

src/open_inwoner/mijn_afval/views.py
```diff
@@ -21,57 +21,121 @@
...
+    LoginRequiredMixin, BaseBreadcrumbMixin, AppConfigMixin, TemplateView
+):
+    template_name = "pages/mijn_afval/afval-profiel.html"
 
-    decimal_places = 0 if isinstance(value, int) else 1
-    return number_format(value, decimal_pos=decimal_places, force_grouping=False)
+    @cached_property
+    def crumbs(self):
+        current_page = self.request.current_page
+        title = current_page.get_title() if current_page else _("Mijn Afval")
```
swrichards: Just remind me, why is this needed? Because the CMS config might have overridden the title?

src/open_inwoner/mijn_afval/clients.py
```diff
@@ -10,9 +10,42 @@
...
+        if afval_type:
+            params["afval-type"] = afval_type
+        if startdatum:
+            params["startdatum"] = startdatum
+        if einddatum:
+            params["einddatum"] = einddatum
+
         try:
-            response = self.get(f"afval-profiel/{bsn}")
+            response = self.get(f"afval-profiel/{bsn}", params=params)
```
swrichards:
```suggestion
            response = self.get(f"afval-profiel/{bsn}/", params=params)
```

src/open_inwoner/mijn_afval/clients.py
```diff
@@ -10,9 +10,42 @@
 
 
 class OpenAfvalAPIClient(APIClient):
-    def get_afval_profiel(self, bsn: str) -> AfvalProfiel:
+    def get_afval_profiel(
+        self,
+        bsn: str,
+        adres: str | None = None,
+        afval_type: str | None = None,
+        startdatum: str | None = None,
```
swrichards, line 18: The filter indeed allows for selecting multiple periods: @stefrado @lidwienl what do you think, should we support selecting multiple periods (e.g. 2023 and 2025), or should we modify the filter to accept one period at a time?

src/open_inwoner/mijn_afval/api_models.py
swrichards, line 29:
```suggestion
    totaal_gewicht: float | None
```

But maybe we should fix that in the API spec.

src/open_inwoner/mijn_afval/views.py
```diff
@@ -21,57 +21,121 @@
...
+                )
+                afval_data = _format_afval_profiel(filtered_profiel)
+            else:
+                # No filters applied, reuse unfiltered data
+                afval_data = _format_afval_profiel(unfiltered_profiel)
 
-    # Return original if pattern doesn't match
-    return address
+        except MijnAfvalException:
+            messages.error(
```
swrichards, line 120: Let's get some logging here too (`logger.warning` with `exc_info=True`)

src/open_inwoner/mijn_afval/clients.py
swrichards, line 1: This might be an Open Afval API issue rather than here, but it seems to me the total weight for a container (in the accordion header) does not vary with the filters. My expectation was that the graph and the accordion are based on identical data (and that totals move along with it).

pi-sigma: It's fixed in the API

src/open_inwoner/mijn_afval/clients.py
pi-sigma: Discussed during standup: we only support filtering by single period for the time being

src/open_inwoner/mijn_afval/views.py
pi-sigma: The fallback is not needed since the section is only accessible via CMS. Removed it.

src/open_inwoner/mijn_afval/api_models.py
pi-sigma: `totaal_gewicht` is always annotated in the API before the serializer runs. Perhaps you referred to the earlier filter implementation, now this should be correct as-is.

src/open_inwoner/mijn_afval/clients.py
pi-sigma: Fixed for the adres param, but we only send a single `afval_type` to the API (since we only support two, selecting two = not filtering at all)

swrichards (approved):

PR #2205, opened 2026-02-09
[#2157] Select newest from multiple OpenKlant answers

### LINKED ISSUE #2157: Show openklant2 answers from Contact in OIP
### Description
Als gebruiker wil ik het definitieve antwoord zien op mijn portaal

### Acceptatie Criteria
* Het definitieve antwoord is degene die het laatst is aangemaakt in openklant, met hetzelfde kanaal (latere antwoorden vervangen eerdere antwoorden)
  * Dit moet ook mogelijk zijn wanneer er meer tussenliggende contactmomenten zijn geregistreerd op een ander kanaal
* alle tussenliggende contactmomenten worden genegeerd
* Status van de interne taak word niet uitgelezen
  * Deze is of onbeantwoord of beantwoord (afhaneklijk van of er een antwoord is gegeven)

pi-sigma:
## Summary

Instead of displaying the first answer attached to a question, we now check for all answers and display the newest.

## References

Depends on #2021 (we're moving from our vendored openklant2 client to the packaged version)

Closes #2157

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -44,7 +44,7 @@ from openklant_client.types.resources.partij import (
     PartijListParams,
 )
 from openklant_client.types.resources.partij_identificator import PartijIdentificator
-from pydantic import BaseModel, ConfigDict, TypeAdapter
+from pydantic import BaseModel, ConfigDict, TypeAdapter, field_validator
 from requests.exceptions import RequestException
 from typing_extensions import TypedDict
 from zgw_consumers.api_models.base import factory
@@ -897,7 +897,6 @@ class OpenKlant2Answer:
         )
 
 
-@dataclass(frozen=True)
 class OpenKlant2Question(BaseModel):
     url: str
     question: str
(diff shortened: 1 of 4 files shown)


src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -146,16 +146,184 @@ def test_logs_warning_when_more_than_two_phone_numbers(self, mock_client_class):
         service = OpenKlant2Service(config=self.config)
         user = UserFactory()
 
+<<<<<<< HEAD
```
swrichards: Oops!

src/open_inwoner/openklant/services.py
```diff
@@ -929,10 +930,15 @@ def from_klantcontact_and_answer(
...
             ),
-            answer=answer,
+            answers=answers or [],
             url=klantcontact["url"],
         )
 
+    @property
+    def answer(self) -> OpenKlant2Answer | None:
+        """Return the newest/latest answer (first in the sorted list), or None if no answers."""
+        return self.answers[0] if self.answers else None
```
swrichards, line 960: Can we be sure they're ordered by datetime desc?

swrichards (changes requested):
LGTM, but I'd like to have a simple coverage test that shows we do in fact pick the last question.

src/open_inwoner/openklant/services.py
pi-sigma: Good point, we need to ensure the answers are sorted independently of how they're created. I've moved the sorting to a field validator in the `OpenKlant2Question` class. Also discovered a bug in the process: `OpenKlant2Question` was decorated with `@dataclass(frozen=True)` which shouldn't be mixed with pydantic models. The field validator actually failed with this.

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -321,3 +326,89 @@ def test_create_persoon_without_identificatoren_when_no_bsn(
             call_data,
             "Should not include identificatoren when user has no BSN",
         )
+
+
+class OpenKlant2QuestionAnswerTestCase(TestCase):
```
swrichards, line 332: Sorry to hard on about the testing, but I am still a bit concerned: when we phased out the VCR code, we also lost a lot of coverage, in particular coverage for fetching the answers from the API and unspooling them into a Question-Answer structure. Codecov shows that most of that logic is now untested. Given that this logic has been a bit notorious in complexity and production issues, I think it'd be important to have some tests that also actually test the fetching of the answers and the onderwerp object logic by mocking the client methods.

swrichards (changes requested):

swrichards (approved):

PR #2216, opened 2026-02-12
:bug: FIX bug where link markup is shown in footer menu item plugin

### LINKED ISSUE #2208: DRT 408-1 - Footer link rendering
### Describe the bug

> https://dimpact.atlassian.net/browse/DRT-408

Original issue was resolved with 2.0.2 and confirmed. We'll reuse this issue for an installation-specific issue when publishing the Homepage changes by Groningen

stefrado:
## Summary

Fix issue where link markup is shown in footer menu item plugin

## Issue References

- Github Issue: #2208 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/cms/plugins/cms_plugins/links.py
+++ b/src/open_inwoner/cms/plugins/cms_plugins/links.py
@@ -57,6 +57,17 @@ class LinkPlugin(OriginalLinkPlugin):
     model = ExtendedCMSLink
     form = CustomLinkForm
     allow_children = False
+    render_template = "cms/plugins/links/link.html"
+
+    def get_render_template(self, context, instance, placeholder):
+        """
+        Override to use our custom template that supports ProsemirrorModelField
+
+        Note: setting the template via the render_template attribute does not work;
+        template resolution seems to be broken by inheritance. Revisit when upgrading to
+        Django CMS 4
+        """
+        return self.render_template
 
     fieldsets = [
         (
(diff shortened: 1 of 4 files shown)


src/open_inwoner/templates/cms/plugins/links/link.html
```diff
@@ -0,0 +1,21 @@
...
+        {{ instance.anchor }}
+        {{ instance.mailto }}
+        {{ instance.phone }}
+    {{ instance.target }}
+    {{ instance.attributes_str }}
+{% endcomment %}
+
+{% comment %}
+# Modified from djangocms_link/default/link.html to support ProsemirrorModelField
+# Changed {{ instance.name }} to {{ instance.name.html|safe }}
```
pi-sigma, line 20: You're actually using `instance.name|safe` above; both seem to work, though `instance.name.html|safe` is more explicit about what we're dealing with. Either way, make the  code and comment consistent.

src/open_inwoner/cms/plugins/cms_plugins/links.py
```diff
@@ -57,6 +57,11 @@ class LinkPlugin(OriginalLinkPlugin):
     model = ExtendedCMSLink
     form = CustomLinkForm
     allow_children = False
+    render_template = "cms/plugins/links/link.html"
+
+    def get_render_template(self, context, instance, placeholder):
+        """Override to use our custom template that supports ProsemirrorModelField"""
+        return self.render_template
```
pi-sigma, line 70: You don't normally need both attribute and method, the attribute should suffice, though I tested this and realize it doesn't work. This seems to be a bug in Django CMS (at least of our version), where template resolution is somehow broken by inheritance. Consider adding a note to the docstring about this (e.g. "Note: setting the template via the `render_template` attribute does not work; template resolution seems to be broken by inheritance" or similar) because it's something we could revisit and perhaps clean up when upgrading to Django CMS 4

pi-sigma (changes requested):

pi-sigma (approved):

PR #2226, opened 2026-02-16
Merge in changelog for 2.0.2 and remove duplicates

pi-sigma:
## Summary

- Mergen in changelog for 2.0.2
- Removed duplicate entries from changelog

## Checklist

- [x] CHANGELOG.rst updated


swrichards (changes requested):
As part of this, could you also merge [in the changelog of 2.0.2](https://docs.openinwoner.nl/en/v2.0.2/changelog.html#id1)? That will give us a full overview of all entries so we can avoid dupes.

swrichards (approved):

PR #2246, opened 2026-02-24
:label: [#2203] Add NLDS --nl prefix candidate Paragraph as web component

### LINKED ISSUE #2203: PP-324 Update Utrecht Paragraphs to NL prefix components and make web-component
Update Utrecht Paragraphs to NL suffix components from NL Design System.

They are candidate: https://www.nldesignsystem.nl/handboek/estafettemodel/componenten/definition-of-done/

JIRA refined issue: https://dimpact.atlassian.net/browse/PP-324
 
- [x] first make nl-paragraph a web component and implement it in the Home page (= multiple plugins).
- [x] add -nl design tokens to our package and publish on NPM.
- [x] all templates: all utrecht-paragraph have to be converted to nl-paragraph...maybe in some cases a web component is not possible, but then the HTML version can be used.
- [ ] then see if we can remove the modifiers (compact, centered, small, muted) and make them part of the desigjn-tokens as new --oip-nl modifiers so that they may be used inside the shadow DOM.
- [ ] see if all of these web-components need new storybook/tests.
- [x] Update design-tokens in our package https://github.com/maykinmedia/open-inwoner-design-tokens

Note: utrecht small paragraph no longer exists, so needs to become an open-inwoner/oip modifier

jiromaykin:
## Summary

Adding the new NLDS --nl prefix candidate Paragraph as a web component,
and then implementing that in all of the plugins on the Home page, if possible.
If web component is not possible, the HTML version is rendered.

- keeping Utrecht as  'legacy' styles for now, to be removed later once the 'utrecht-paragraph-small' can be removed and given a new oip-nl modifier.
- so: `-small`, `-compact` and -`centered` etc. are still mostly utrecht here
- but `-muted` is added here as an extra class (without it being needed) just as an example since we need it for other templates
- there is no muted style in the Home plugin cards in the Figma design, so that's been removed there
- if there is no HTML template available to add the web component, only Python tags, then we just add the HTML version of the `p.nl-paragraph` component, like in `src/open_inwoner/components/templatetags/string_tags.py`

## Issue References

- depends on design-token package PR: https://github.com/maykinmedia/open-inwoner-design-tokens/pull/41
- and on https://github.com/maykinmedia/open-inwoner-design-tokens/pull/42

## Checklist

- [x] CHANGELOG.rst updated
- [ ] Documentation updated 
      

src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @nl-design-system-candidate/paragraph-react is React-only and incompatible with
+ * the Preact + JSDOM test environment. Rendering it directly causes:
+ *
+ *   InvalidCharacterError: "[object Object]" did not match the QName production
+ *
+ * Mock it with a semantic HTML equivalent so unit tests can verify OIP wrapper logic.
+ * Actual rendering is covered by Django integration tests and Storybook.
+ */
+vi.mock('@nl-design-system-candidate/paragraph-react', () => ({
```
stefrado, line 17: This is something that i'll fix in my coming (`shadowDOM`) PR. I think i have resolved this by using playwright with the open-formulieren/admin-ui testing set-up.

You might want to add a comment here with TODO so we don't forget that we'll need to fix this.

src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
```diff
@@ -0,0 +1,81 @@
...
+  Paragraph: ({ children, className, purpose }: any) => (
+    <p
+      class={[
+        'nl-paragraph',
+        purpose === 'lead' ? 'nl-paragraph--lead' : '',
+        className,
+      ]
+        .filter(Boolean)
+        .join(' ')}
```
stefrado:
```suggestion
      class={clsx(
        'nl-paragraph',
        purpose === 'lead' ? 'nl-paragraph--lead' : '',
        className,
      )}
```

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,23 @@
...
+  children?: ComponentChildren;
+}
+
+const Paragraph: AC<IParagraphProps> = ({
+  children,
+  lead = false,
+  extraClasses,
+}) => {
+  return (
+    <NLDSParagraph purpose={lead ? 'lead' : undefined} className={extraClasses}>
```
stefrado: Wouldn't it be better if we have the same Paragraph interface as `NLDSParagraph`? This would make more sense to me:

```suggestion
    <NLDSParagraph {...props}>
```

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  parameters: {
+    layout: 'padded',
+  },
+};
+
+export default meta;
+
+export const Default: Story = {
+  render: () => (
+    <Paragraph>Pa's wijze lynx bezag vroom het fikse aquaduct.</Paragraph>
```
stefrado: 👴🧠🐈👀💧🚇

src/open_inwoner/scss/components/Typography/P.scss
```diff
@@ -1,4 +1,6 @@
 @use '@utrecht/paragraph-css/dist/index.css';
+// When shadow DOM is on, we need to import these styles for paragraphs that are not a web component:
+@use '@nl-design-system-candidate/paragraph-css/paragraph.scss';
```
stefrado, line 3: Are we going to use any paragraphs that are not `shadowDOM`?
And can't we get the same stylesheet from `@nl-design-system-candidate/paragraph-react`?

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ Deployment aandachtspunten
 
 Nieuwe features
 ---------------
+* [:gh:`2203`, :oip-nlds:`41`]: Nieuwe --nl NLDS prefix paragraph toegevoegd als web-component zodat deze hergebruikt kan worden.
```
stefrado:
```suggestion
* [:gh:`2203`, :oip-nlds:`41`]: Nieuwe --nl NLDS prefix paragraph toegevoegd als web-component, zodat deze hergebruikt kan worden.
```

src/open_inwoner/scss/components/Typography/P.scss
```diff
@@ -72,3 +74,11 @@ body [class^='utrecht-heading'] + .utrecht-paragraph,
     }
   }
 }
+
+// Context styles for elements that directly follow the web component paragraph
+
+// Home page plug-ins
+nl-paragraph + .card-container,
+nl-paragraph + .map {
+  margin-top: var(--spacing-large);
```
stefrado:
```suggestion
  margin-block-start: var(--spacing-large);
```

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.scss
stefrado, line 45: Still needed?

src/open_inwoner/cms/banner/tests/test_plugin_banner.py
```diff
@@ -48,4 +48,4 @@ def test_banner_text_with_description_is_rendered_in_plugin(self):
         )
 
         self.assertIn(f'<h1 class="utrecht-heading-1">{title} </h1>', html)
-        self.assertIn(f'<p class="utrecht-paragraph">{description}</p>', html)
+        self.assertIn(f"<nl-paragraph>{description}</nl-paragraph>", html)
```
stefrado, line 51: Maybe it would be nice if we could check if this renders a `<p>` in the `shadowDOM`

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
```diff
@@ -38,7 +38,7 @@ const HomePluginCard: AC<HomepageCardTypes> = ({
             )}
           </Heading>
           <p
-            class={clsx('utrecht-paragraph', 'utrecht-paragraph-muted')}
+            class={clsx('nl-paragraph')}
```
stefrado:
```suggestion
            class="nl-paragraph"
```

Besides that why are you not using the Preact component, but is it recreated here?

src/open_inwoner/react/components/Paragraph/Paragraph.scss
```diff
@@ -0,0 +1,49 @@
...
+      color: var(--oip-nl-paragraph-muted-color);
+    }
+
+    &-centered {
+      text-align: var(--oip-utrecht-paragraph-centered-text-align);
+    }
+
+    &-compact {
+      margin-top: var(--oip-utrecht-paragraph-compact-margin-top);
+      margin-bottom: var(--oip-utrecht-paragraph-compact-margin-bottom);
```
stefrado:
```suggestion
      margin-block-start: var(--oip-utrecht-paragraph-compact-margin-top);
      margin-block-end: var(--oip-utrecht-paragraph-compact-margin-bottom);
```

However the variables do not align these CSS, the suggestion is a bit better

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
```diff
@@ -38,7 +38,7 @@ const HomePluginCard: AC<HomepageCardTypes> = ({
...
             id={addressLineFirst ? id : undefined}
```
stefrado: Id should also be a passable prop to the `Paragraph` right?

stefrado (changes requested):

src/open_inwoner/scss/components/Typography/P.scss
jiromaykin: 1. Well yes: we had at least 1 python file in this pull-request that creates a paragraph in Python, I was not sure could actually be turned int oa web-component: `src/open_inwoner/components/templatetags/string_tags.py`

Maybe there are more like these.

Also I think there are a bunch of documentation pages with paragraphs that may not need web components: I am not sure every paragraph should always be a web component, even if it's not part of a front-end that needs to be tested? 

But making everythign the same type of web component would nicely limit the amount of 'context' styles, or else I'd have to duplicate them for both the surrounding `nl-paragraph` web component and the HTML equivalent.

2. There is no true CSS stylesheet inside "@nl-design-system-candidate/paragraph-react", it does have a dist folder for CSS but that actually points to "import '@nl-design-system-candidate/paragraph-css/paragraph.css';"

src/open_inwoner/cms/banner/tests/test_plugin_banner.py
jiromaykin: Yes but Django's test renderer has no JavaScript so the shadow DOM is never mounted . The <nl-paragraph> assertion is what the server actually renders, right? 🤔

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.scss
jiromaykin: Yes, still. Read the description of this PR, that explains a lot. This PR is just a start of everything-everywhere-all-at-once.

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
jiromaykin: Ah yeah, it's not conditional anymore,
And about 'recreating' the paragraph: all i did was just replacing the code that was already there without refactoring, so i will do that too.

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
jiromaykin: ~~Well I would have to add it to Paragraph.tsx. The OIP paragraph wrapper is its own interface which I don't think will automatically pass through standard HTML attributes.~~

Edit: or i should add ...props!

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
jiromaykin: ah okay, so then id does not need to be explicitly added.

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
jiromaykin: I was looking for alternatives to "The quick brown fox" :-)

src/open_inwoner/react/components/Paragraph/Paragraph.scss
jiromaykin: Ooops, i should have known; this actually means i'll have to split up all the margin shorthands.

src/open_inwoner/cms/banner/tests/test_plugin_banner.py
stefrado: Yes, that is the most important. It would have been nice tho...

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.scss
stefrado: Oops, this was one of the first comments i added to the PR... of course are 23 file changes not all our paragraphs 😆

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
```diff
@@ -37,12 +38,9 @@ const HomePluginCard: AC<HomepageCardTypes> = ({
               title
             )}
           </Heading>
-          <p
-            class={clsx('utrecht-paragraph', 'utrecht-paragraph-muted')}
```
stefrado, line 41: We do have this `nl-paragraph--oip-muted` modifier right, why not using it?

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,31 @@
...
+  type ParagraphProps,
+} from '@nl-design-system-candidate/paragraph-react';
+import { AnyComponent as AC, ComponentChildren } from 'preact';
+import './Paragraph.scss';
+
+export interface IParagraphProps extends ParagraphProps {
+  lead?: boolean;
+  extraClasses?: string;
+  children?: ComponentChildren;
+}
```
stefrado: We could just use their interface, since this includes exactly what we need.
```suggestion
export interface IParagraphProps extends ParagraphProps {}
```

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,31 @@
...
+const Paragraph: AC<IParagraphProps> = ({
+  children,
+  lead = false,
+  extraClasses,
+  ...props
+}) => {
+  return (
+    <NLDSParagraph
+      purpose={lead ? 'lead' : undefined}
+      className={extraClasses}
```
stefrado: These two can be passed just like they are used in `NLDSParagraph`

```
purpose?: 'lead';
className?: string;
```

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
```diff
@@ -0,0 +1,31 @@
...
+      {...props}
+    >
+      {children}
```
stefrado:
```suggestion
      <slot />
```

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
stefrado: Yeah very nice, i like this one 🏆

src/open_inwoner/scss/components/Typography/P.scss
```diff
@@ -17,17 +19,9 @@
...
     }
-
-    &:empty {
-      margin: 0 !important;
-    }
-  }
-
-  &:empty {
-    margin: 0 !important;
   }
```
stefrado, line 31: I don't see where these things have been moved to, are they not needed anymore?

src/open_inwoner/scss/components/Typography/P.scss
```diff
@@ -17,17 +19,9 @@
     }
 
     &-compact {
-      margin-top: var(--oip-utrecht-paragraph-compact-margin-top);
-      margin-bottom: var(--oip-utrecht-paragraph-compact-margin-bottom);
+      margin-top: var(--oip-utrecht-paragraph-compact-margin-block-start);
+      margin-bottom: var(--oip-utrecht-paragraph-compact-margin-block-end);
```
stefrado:
```suggestion
      margin-block-start: var(--oip-utrecht-paragraph-compact-margin-block-start);
      margin-block-end: var(--oip-utrecht-paragraph-compact-margin-block-end);
```

src/open_inwoner/scss/components/Typography/P.scss
stefrado: Alright clear!

at least 1 python file in this pull-request that creates a paragraph in Python

This python function returns a `<nl-paragraph />`, which means that this has the `adoptedStyleSheet` including the required CSS.

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { withLoader } from '@react/lib/decorators/storybook';
+
+import { PARAGRAPH_DEFINITION, type IParagraphProps } from '.';
+import Paragraph from './Paragraph';
+
+type Story = StoryObj<IParagraphProps>;
+
+const meta: Meta<IParagraphProps> = {
+  title: 'Components/Paragraph',
```
stefrado: I think that this component should become a `web-component only` story. Look at #2310 `Accordion.stories.tsx`. Then we have full round-trip coverage.

stefrado (changes requested):

src/open_inwoner/react/components/HomePluginCard/HomePluginCard.tsx
jiromaykin: Because in the Figma design these cards do not contain any muted texts.
Maybe the designer made a mistake and removed the lighter color... but muting would be bad for readability anyway + the NLDS cards never have a text that is lighter than the body text,

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
jiromaykin: Will do!

src/open_inwoner/scss/components/Typography/P.scss
jiromaykin: Not needed, since modern browsers do not give default spacing to empty Paragraphs anymore, but it's also helped by having zero margins as the default value in our tokens now.

src/open_inwoner/scss/components/Typography/P.scss
jiromaykin: note: These 'utrecht' legacy styles and legacy designtokens will be deleted anyway in next PR.

src/open_inwoner/react/components/Paragraph/Paragraph.tsx
jiromaykin: Oh my, it's getting real.

src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
jiromaykin: Also this setup is unable to query the content of Slots, so that's another TODO.

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
jiromaykin: @stefrado I have created the web component-only stories but I also added a new TODO to this file, because in this PR I do not have the new decorators set-up for the entire project + I do not have the play functions/the entire @react/lib/testing directory here, so maybe I should not add all of those yet (?)
(I've already added the createStyelSheet things which may conflict).

stefrado (approved):

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+  args: {
+    text: "Pa's wijze lynx bezag vroom het fikse aquaduct.",
+  },
+};
+
+export default meta;
+
+export const Default: Story = {
+  decorators: [withLoader(PARAGRAPH_DEFINITION.tagName)],
```
stefrado: One small comment:

You can also add this `decorators: [withLoader(PARAGRAPH_DEFINITION.tagName)],` to the meta. This way you do not have to write it 3 times

stefrado (approved):

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
jiromaykin: Ah okay, i'll try.

PR #2248, opened 2026-02-26
feat: validate file length in zgw document download view

### LINKED ISSUE #2289: Zaak document download does not validate length
Sometimes the ZGW document download object is behind a gateway. If there is an error in the gateway, you might get e.g. an XML error response in the response body of a document download, rather than the file bytes. This tends to result in weird gateway errors, because we set Content-Length based on the filesize retrieved from the document _object_ endpoint (not the blob endpoint). This length will differ from the error response length, leading to confusing errors. We should add a check with this.

This happened on the Almelo test server.

swrichards:
If the ZGW document backend is misbehaving, for instance because
an intermediate service gateway is broken and returns some kind
of XML message, the user will see a broken download message,
because the announced content-length will not in fact match the
content. In those cases, it's better UX to log and display an
error message.

### DIFF
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -828,7 +828,6 @@ class CaseDocumentDownloadView(CaseLogMixin, CaseAccessMixin, View):
             raise PermissionDenied()
 
         # retrieve and stream content
-        content_stream = None
         content_stream = api_group.documenten_client.download_document(
             info_object.inhoud
         )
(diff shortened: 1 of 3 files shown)


src/open_inwoner/cms/cases/views/status.py
```diff
@@ -848,6 +848,34 @@ def get(self, request, *args, **kwargs):
         if not content_stream:
             raise Http404
 
+        # Validate the actual content length matches the expected size
+        actual_content_length = content_stream.headers.get("Content-Length")
+        if (
+            actual_content_length
+            and int(actual_content_length) != info_object.bestandsomvang
```
pi-sigma, line 855: - Could raise an exception if `actual_content_length` is malformed, perhaps `try/except` the `int` part
- is `bestandsomvang` required? Otherwise `int(actual_content_length) != None` will raise a false positive

src/open_inwoner/openzaak/tests/test_documents.py
```diff
@@ -518,6 +518,29 @@ def test_no_data_is_retrieved_when_document_download_data_http_500(self, m):
 
         self.app.get(self.informatie_object_file.url, user=self.user, status=404)
 
+    def test_document_download_with_size_mismatch_shows_error(self, m):
```
pi-sigma, line 521: The test name `_shows_error` suggests you're testing something about the error (e.g. status, error message), but you only verify the redirect. Either add check for the error message or change test name

src/open_inwoner/openzaak/tests/test_documents.py
```diff
@@ -518,6 +518,29 @@ def test_no_data_is_retrieved_when_document_download_data_http_500(self, m):
...
+        self._setUpAccessMocks(m)
+        m.get(self.informatie_object["url"], json=self.informatie_object)
+        # Mock the download endpoint with mismatched content length
+        wrong_content = b"wrong content"
```
pi-sigma, line 525: This works because `wrong_content` has the wrong length, but it's implicit and not exactly obvious. Perhaps add comment or change variable name to something like `content_wrong_length = b"wrong_content"`

pi-sigma (changes requested):
The commit message seems to have a double space (not visible on GitHub for some reason but shows up in CLI)

src/open_inwoner/cms/cases/views/status.py
swrichards: Good catches both.

src/open_inwoner/openzaak/tests/test_documents.py
```diff
@@ -522,6 +522,64 @@ def test_no_data_is_retrieved_when_document_download_data_http_500(self, m):
...
+            status=302,
+            auto_follow=False,
+        )
+
+        expected_redirect = f"/cases/{self.api_group.id}/{self.zaak['uuid']}/status/"
+        self.assertEqual(response.location, expected_redirect)
+
+        # Follow the redirect to access messages
+        response_redirect = response.follow()
+        messages = [str(m) for m in response_redirect.context["messages"]]
```
pi-sigma: `m` shadows the mock parameter. Works for now, but still clearer to use `msg` or similar

src/open_inwoner/openzaak/tests/test_documents.py
pi-sigma, line 1: There's a test for malformed content length header, but none for missing content length header. Could be added as subtest.

src/open_inwoner/cms/cases/views/status.py
pi-sigma, line 1: I noted in passing (LL 831-832):
```
content_stream = None
content_stream = api_group.documenten_client.download_document(...)
```
First assignment is useless because immediately overridden. Old code, but could be cleaned up.

pi-sigma (approved):
A few nice-to-have's, but otherwise good to go

Issue #2306, opened 2026-03-11
POC: use config-checks to do a ZGW Api Group zaken-fetch check


swrichards:
- The basic idea would be to use https://github.com/maykinmedia/config-checks to help developers _and admins_ verify whether the system is properly configured
- The difference between the current config-checks and our requirements:
  - We need to parametrize tests (e.g. `Run this check for BSN 111222333`)
  - It should be possible to provide the parameters in the Django admin

- Outcome of the POC might be to contribute code back the config-checks library
  - That would involve a renaming: `ConfigCheck != HealthCheck` (confusion with the latter in https://maykin-django-common.readthedocs.io/en/stable/health_checks.html) 

A few implementation ideas:
- We could define a Django Form with a config check, which holds the parameters. We can have some boilerplate which generates an admin view, captures the params, and passed them to the health check
- As part of _a different admin_, we could have a new Django admin field which links to a custom admin page to provide the parameters and then run them

Example:

- Say we want to add a "Fetch Cases for User Health Check", the purpose of which is to fetch cases for a specific user and report: did we get a response from the server? Could we resolve all the references statussen/rollen/zaaktypen/enz?

<img width="1293" height="674" alt="Image" src="https://github.com/user-attachments/assets/9ced5a0d-4932-408b-b701-44b0bba52232" />

- Now imagine we add a field to this view, which would be something like:

```
class FetchCasesConfigCheckParams(forms.Form):
    bsn = models.CharField(max_length, validate_11_proef)

@admin.register(ZgwAPIGroupConfig)
class ZgwAPIGroupConfigAdmin(admin.ModelAdmin):
    list_display = [..., "initiate_fetch_cases_for_user_config_check"]

    @admin.display(description="Verify the configuration for this API Group")
    def initiate_fetch_cases_for_user_config_check(self, obj):
          link_to_page = config_check_intiate_link(FetchCasesConfigCheckConfigCheck, FetchCasesConfigCheckParams(api_group_id=obj.pk)
# render the link
```

# Permissions and security

- As a rule, check reports should not be information-rich, e.g. "BSN 1234 has N cases", rather than actually saying which cases
- For now, we assume if you can access the admin page where the link is contained, you can access the link
   - But: maybe we could optionally specific a Django permission which is required to run the check?
   - Or: there is a generic Django permission `can_run_config_checks_in_admin`
   - Or: check level permissions 
   - **Idea**: specifiy required permissions? `is_superuser | has_perm(x)`
[...]

PR #2316, opened 2026-03-16
:sparkles: [#2306] Add configuration check to test fetching cases by BSN

OlhaZahoruiko:
## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables

### DIFF
--- a/src/open_inwoner/config_checks/forms.py
+++ b/src/open_inwoner/config_checks/forms.py
@@ -1,0 +1,17 @@
+from django import forms
+
+from open_inwoner.openzaak.models import ZGWApiGroupConfig
+
+
+class FetchCasesConfigCheckParams(forms.Form):
+    api_group = forms.ModelChoiceField(
+        queryset=ZGWApiGroupConfig.objects.all(),
+        label="API group",
+        help_text="Select the API group to run the check against",
+        empty_label=None,
+    )
+    bsn = forms.CharField(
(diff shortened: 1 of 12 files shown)


src/open_inwoner/config_checks/forms.py
```diff
@@ -0,0 +1,9 @@
+from django import forms
+
+
+class FetchCasesConfigCheckParams(forms.Form):
+    bsn = forms.CharField(
```
swrichards, line 13: I would like to add the API group as a field here, and see if we can get it prefilled when the user clicks the button under a specific API group. This will help running the check independently later (e.g. from the management command/API or tests).

swrichards (changes requested):
@OlhaZahoruiko as discussed, this looks good. Just one comment I'd like you to look at regarding the form class.

I've changed the base to `poc-interactive-config-checks` which is where we'll merge this work to for the coming period.

After this I would like to implement two more parallel checks, which will give us a sense of how to organize the interfaces before seeing if we can come up with a way to introduce this back into the config checks library. I'll follow up on those with issues.

swrichards (approved):

Issue #2305, opened 2026-03-10
Zaken detail page spins forever on errors


swrichards:
<img width="1174" height="498" alt="Image" src="https://github.com/user-attachments/assets/f76ec44f-bb1e-40fe-aa58-6c618b4e2470" />

If there are any errors in loading the zaak detail page, the spinner will simply show forever. We should handle the error response gracefully and display a message to the user.

PR #2350, opened 2026-03-24
fixup! :sparkles: [#2305] Make sure the zaken cases spinner handles e…

swrichards:
Adding a playwright test


swrichards:
Merged into #2317

PR #2351, opened 2026-03-25
tests: persist Playwright traces in e2e and expose them as CI artifacts

swrichards:
The general effort here is to start expanding our Playwright footprint a bit, what with all the webcomponent and frontend changes in the pipeline.

This ensures the Playwright traces are published in CI, see for instance: https://github.com/maykinmedia/open-inwoner/actions/runs/23550996777#artifacts

To see why this is useful, download one of the `e2e-test-results-*` files, unpack, and upload one of the zip files to:
https://trace.playwright.dev/

### DIFF
--- a/src/open_inwoner/utils/tests/playwright.py
+++ b/src/open_inwoner/utils/tests/playwright.py
@@ -1,17 +1,23 @@
 import os
 from collections.abc import Callable
+from pathlib import Path
 
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 from django.core.files.base import ContentFile
 from django.urls import reverse
 
 from furl import furl
-from playwright.sync_api import Browser, Playwright, sync_playwright
+from playwright.sync_api import Browser, BrowserContext, Playwright, sync_playwright
 
 from open_inwoner.accounts.models import User
 from open_inwoner.configurations.choices import CustomFontName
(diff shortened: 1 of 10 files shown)


src/open_inwoner/utils/tests/playwright.py
```diff
@@ -119,6 +130,34 @@ def tearDownClass(cls):
...
+        self._traced_contexts.append(context)
+        return context
+
+    def tearDown(self):
+        test_id = f"{self.__class__.__name__}.{self._testMethodName}"
+
+        for i, context in enumerate(self._traced_contexts):
+            suffix = f"_{i}" if i > 0 else ""
+            trace_path = os.path.join(PLAYWRIGHT_TRACE_DIR, f"{test_id}{suffix}.zip")
+            context.tracing.stop(path=trace_path)
```
jiromaykin, line 157: Wouldn't this also need some kind of "finally:....context.close()"?

jiromaykin (approved):
Just 1 question, but approving so it doesn't get blocked.

src/open_inwoner/utils/tests/playwright.py
swrichards: `tearDown` will always run (even if the test raises), and I believe the context does not have to be explicitly closed, so we should be good.

PR #2354, opened 2026-03-25
feat: [#2353] strip file extension from document title on upload

### LINKED ISSUE #2353: When uploading documents to the ZGW backend, strip extension from the informatieobject title
In eSuite, all documents will if possible be converted to PDF/A, and some gemeenten will configure their OIP client to also return this PDF via the ZGW API, if present. Because we set the title of uploaded documents to the filename with extension (`foobar.docx`), you get the counterintuitive situation that this will download as a PDF.

The solution is to strip the extension from the title. No relevant information is lost (the filename is stored separately, which will remain docx for users that do not wish to use the eSuite feature, and will have been changed to foobar.pdf in the relevantly configured eSuite installs).

swrichards:
In eSuite, uploaded documents can be converted to PDF/A, and some
gemeenten configure OIP to return this converted PDF via the ZGW API.
Because the document title was set to the full filename including
extension (e.g. foobar.docx), this caused a counterintuitive situation
where the document would download as a PDF despite the title suggesting
otherwise.

The fix strips the extension from the title on upload. No information is
lost: the original filename is stored separately on the document and
reflects the true source format. In eSuite installs with PDF/A
conversion enabled, the filename will already have been updated to
foobar.pdf, making the title-without-extension the correct neutral
representation in both cases.

This commit also takes the opportunity to generally clean up
the use of the file list/file tags. It moves the logic for building
the expected DTO objects in the views themselves, making the tag
logic simpler and dedicated to display concerns. Given the different
sources of file (Filer, Django File Field, ZGW API), it was quite
hard to follow what the expected types were and what edge cases
we had to consider. This change uses an explicit DTO with explicit
side-effect less conversion functions to make clear both the expected
shape and the point of conversion.

Closes: https://github.com/maykinmedia/open-inwoner/issues/2353

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,13 @@ Nieuwe features
   wordt de verbinding met Elasticsearch beveiligd met een gebruikersnaam en wachtwoord.
   De Docker Compose- en CI-configuratie zijn bijgewerkt om ``xpack.security.enabled``
   in te schakelen.
+* [:gh:`2353`]: Bij het uploaden van een document naar een zaak wordt de bestandsextensie
+  nu verwijderd uit de documenttitel (bijv. ``rapport.pdf`` krijgt als titel ``rapport``).
+  De bestandsextensie in de documentenlijst wordt nu afgeleid van het ``formaat``-veld
+  (en als fallback ``bestandsnaam``) in plaats van de titel, zodat de extensie correct wordt
+  weergegeven ook wanneer eSuite het document heeft omgezet naar PDF/A.
+  Daarnaast is de interne weergave van bestanden in templates gerefactored om de
+  conversielogica te centraliseren en de templatetags eenvoudiger te maken.
 * ...
 * [:gh:`2268`]: WCAG toegankelijkheidsverbeteringen uit rapport gemeente Enschede toegepast.
 
--- a/src/open_inwoner/cms/cases/views/status.py
+++ b/src/open_inwoner/cms/cases/views/status.py
@@ -1,7 +1,6 @@
-import dataclasses
 import datetime as dt
+import os
 from collections import defaultdict
-from datetime import datetime
 from typing import Iterable, Protocol, cast
 
 from django.conf import settings
@@ -32,6 +31,7 @@ from zgw_consumers.api_models.constants import RolOmschrijving, RolTypes
 
 from open_inwoner.accounts.models import User
 from open_inwoner.cms.cases.forms import CaseContactForm, CaseUploadForm
+from open_inwoner.components.file_item import FileItem
 from open_inwoner.mail.service import send_contact_confirmation_mail
 from open_inwoner.openklant.constants import KlantenServiceType
 from open_inwoner.openklant.models import (
@@ -68,14 +68,6 @@ from open_inwoner.utils.views import CommonPageMixin
 from .mixins import CaseAccessMixin, CaseLogMixin, OuterCaseAccessMixin
 
 logger = structlog.stdlib.get_logger(__name__)
-
-
-@dataclasses.dataclass
-class SimpleFile:
-    name: str
-    size: int
-    url: str
-    created: datetime | None = None
 
 
 class VragenService(Protocol):
@@ -737,7 +729,7 @@ class InnerCaseDetailView(
     @staticmethod
     def get_case_document_files(
         zaak: Zaak, api_group: ZGWApiGroupConfig
-    ) -> list[SimpleFile]:
+    ) -> list[FileItem]:
         client = api_group.zaken_client
         case_info_objects = client.fetch_zaak_information_objects(zaak.url)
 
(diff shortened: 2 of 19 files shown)


CHANGELOG.rst
```diff
@@ -16,6 +16,11 @@ Nieuwe features
   wordt de verbinding met Elasticsearch beveiligd met een gebruikersnaam en wachtwoord.
   De Docker Compose- en CI-configuratie zijn bijgewerkt om ``xpack.security.enabled``
   in te schakelen.
+* [:gh:`2353`]: Bij het uploaden van een document naar een zaak wordt de bestandsextensie
+  nu verwijderd uit de documenttitel (bijv. ``rapport.pdf`` krijgt als titel ``rapport``).
+  De bestandsextensie in de documentenlijst wordt nu afgeleid van het ``bestandsnaam``-veld
+  (en als fallback ``formaat``) in plaats van de titel, zodat de extensie correct wordt
```
pi-sigma: Do I misunderstand this or is the order described here the reverse of what the code does?

In `file_tags.py`:

```
guessed = mimetypes.guess_extension(file.formaat) if file.formaat else None
```

pi-sigma: Do I misunderstand this or is the order described here the reverse of what the code does?

In `file_tags.py`:

```
guessed = mimetypes.guess_extension(file.formaat) if file.formaat else None
```

src/open_inwoner/components/templatetags/file_tags.py
```diff
@@ -110,36 +112,50 @@ def file(file, **kwargs):
...
-    else:
-        try:
+            if not kwargs.get("name"):
+                kwargs.update(
+                    name=file.name
+                    if file.name
+                    else file.label.replace(f".{file.extension}", ""),
+                )
+        case SimpleFile():
+            guessed = mimetypes.guess_extension(file.formaat) if file.formaat else None
```
pi-sigma: I think this is platform-dependent. `mimetypes.guess_extension("application/pdf")` will return `pdf` on most systems but may return something else (`.kpdf` for example).

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -907,7 +912,7 @@ def handle_document_upload(self, request, form):
         created_documents = []
 
         for file in files:
-            title = file.name
+            title = os.path.splitext(file.name)[0] or file.name
```
pi-sigma, line 898: In `file_tags.py` you use `pathlib` instead. Could be used here as well:

```
title = pathlib.Path(file.name).stem or file.name
```

src/open_inwoner/components/templatetags/file_tags.py
```diff
@@ -110,36 +112,50 @@ def file(file, **kwargs):
...
-        kwargs.update(
-            is_image=file.file_type == "Image",
-            description=file.description,
-            extension=file.extension,
-            size=file.size,
-            url=file.url,
-        )
-        if not kwargs.get("name"):
+    if file is None:
+        return {**kwargs}
```
pi-sigma: The old code provided defaults for all file-related fields (`is_image=False, extension="", size=0, url="", name=str(file)`. This passes a `dict` to the template that misses these fields. Is this going to work?

pi-sigma (changes requested):

src/open_inwoner/components/templatetags/file_tags.py
swrichards: Good point: I actually think we should raise here, we expect a file, not providing it is an error.

swrichards: Redundant, became a big refactor.

CHANGELOG.rst
swrichards: Sharp eye.

src/open_inwoner/components/templatetags/file_tags.py
swrichards: I think we should be safe given that it's running on a vanilla Linux docker container (for the purposes of this library), and is not dependent on the _user_ OS.

src/open_inwoner/components/file_item.py
```diff
@@ -0,0 +1,94 @@
...
+    size: int
+    url: str
+    is_image: bool
+    created: datetime | None = None
+    description: str | None = None
+
+    @classmethod
+    def from_filer_file(cls, file: FilerFile, name: str | None = None) -> "FileItem":
+        resolved_name = name or (
+            file.name if file.name else file.label.replace(f".{file.extension}", "")
```
pi-sigma: You could use `pathlib.Path(file.label).stem`, similar to how you do it below.

src/open_inwoner/components/templatetags/file_tags.py
```diff
@@ -89,70 +47,26 @@ def file_table(files, **kwargs):
 
 
 @register.inclusion_tag("components/File/File.html")
-def file(file, **kwargs):
+def file_item(file: FileItem, **kwargs):
```
pi-sigma, line 50: Consider raising `TypeError` in case `file` is not a `FileItem`

pi-sigma (approved):

src/open_inwoner/components/templatetags/file_tags.py
swrichards: `dataclasses.asdict` would've done this, but an explicit raise is more readable, I'll add it.

PR #2356, opened 2026-03-25
fix: [#2355] enforce model change permission on ZGW config import views

### LINKED ISSUE #2355: ZGW import admin page should check model permissions in addition to staff
The ZGW catalogus import pages (for catalogus and zaaktype) verify that a user is staff, but not that the user has change permissions. The latter would be preferable for the rare cases where a user is staff but has more scoped permissions.

swrichards:
The import views for CatalogusConfig and ZaakTypeConfig were registered via admin_site.admin_view(), which only requires staff status. Any staff user could therefore reach the import endpoint regardless of whether they had change permission on the relevant model.

Both views now raise PermissionDenied when the requesting user lacks change permission. The shared logic has also been extracted into ZGWImportExportMixin to avoid duplication between the two admin classes.

Closes: #2355

### DIFF
--- a/src/open_inwoner/openzaak/admin.py
+++ b/src/open_inwoner/openzaak/admin.py
@@ -1,7 +1,7 @@
 import datetime
 
 from django.contrib import admin, messages
-from django.core.exceptions import ValidationError
+from django.core.exceptions import PermissionDenied, ValidationError
 from django.db.models import BooleanField, Count, ExpressionWrapper, Q
 from django.forms import Form, ModelForm, Textarea
 from django.forms.models import BaseInlineFormSet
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/admin.py
```diff
@@ -250,6 +199,63 @@ def process_file_view(self, request):
...
+    ]
+    ordering = ("domein", "rsin")
+    list_filter = (
+        "service",
+        "found_in_api",
+    )
+
+    actions = ["export_catalogus_configs"]
+
+    def get_urls(self):
```
pi-sigma: This method is duplicated in `ZaakTypeConfigAdmin`. Could be extracted to the mixin, with `import_path` and `import_name` as class attributes.

pi-sigma (approved):

PR #2359, opened 2026-03-26
:sparkles: [#2352] Consolidate interactive config checks in a declara…

### LINKED ISSUE #2352: Consolidate interactive config checks in a declarative API
Now that we have a few examples of interactive config checks in #2334 we can look towards consolidating the API for implementing them.

The basic idea is that:

- You only have to write your interactive config check according to the specification
- You declare you want to use the check in a given admin, and you'll 'automatically' have a read-only field available to renders links to the various checks.

So [something like we see here](https://gist.github.com/swrichards/2aa94dbbffeb211639f1dd32373c5fed). The goal specifically is to:

- Define the `InteractiveConfigCheck` protocol which the concrete implementations will inherit from and;
- Define a mechanism to attach these to a `ModelAdmin` making available some kind of `config_check_links` field the user can insert into the field declarations where they please

Once this is done, we'll have an end-to-end POC we can briefly discuss with a broader group before upstreaming the new API back into the config checks library.

OlhaZahoruiko:
…tive API

## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables


src/open_inwoner/config_checks/api.py
```diff
@@ -0,0 +1,67 @@
...
+class InteractiveConfigCheck(Protocol[TParamsForm, TModel]):
+    identifier: ClassVar[str]
+    label: ClassVar[str]
+    form_class: ClassVar[type[TParamsForm]]
+
+    def __init__(self, form: TParamsForm): ...
+    def run(self, obj: TModel) -> GenericHealthCheckResult: ...
+
+
+def with_config_checks(*checks):
```
swrichards, line 24:
```suggestion
def with_config_checks(*checks: InteractiveConfigCheck):
```

src/open_inwoner/config_checks/registry.py
```diff
@@ -0,0 +1,15 @@
+from collections import defaultdict
+
+
+class ConfigCheckRegistry:
+    def __init__(self):
+        self._registry = defaultdict(list)
+
+    def register(self, model, check):
+        self._registry[model].append(check)
```
swrichards, line 9: I think we should probably use `identifier` as a key and raise on duplicates, because then passing the identifier is ambiguous as to which check is being requested.

src/open_inwoner/config_checks/views.py
```diff
@@ -0,0 +1,42 @@
...
+def run_config_check(request, app_label, model_name, pk, check_id):
+    model = apps.get_model(app_label, model_name)
+    obj = get_object_or_404(model, pk=pk)
+
+    checks = registry.get_checks(model)
+
+    check_class = next(
+        (c for c in checks if c.identifier == check_id),
+        None,
+    )
```
swrichards, line 19: See comment in the registry: we should ensure check_id is unique in the registry.

src/open_inwoner/config_checks/views.py
```diff
@@ -0,0 +1,42 @@
+from django.apps import apps
+from django.contrib.admin.views.decorators import staff_member_required
+from django.http import Http404
+from django.shortcuts import get_object_or_404, render
+
+from .registry import registry
+
+
+@staff_member_required
+def run_config_check(request, app_label, model_name, pk, check_id):
```
swrichards, line 10: I think it would be a good idea to expose some signals and trigger those here, so we can e.g. attach logging to run/failed checks (given they will often have access to sensitive data). I would suggest we start with `interactive_config_check_triggered` passing all the parameters to this view (or perhaps the resolved check class and resolved object).

src/open_inwoner/config_checks/views.py
```diff
@@ -0,0 +1,42 @@
...
+    model = apps.get_model(app_label, model_name)
+    obj = get_object_or_404(model, pk=pk)
```
swrichards, line 12: I think we should make the error handling a bit more explicit, perhaps adding a custom error page where we explain exactly why no permission was given / which permission is required (we can keep staff member required, but we might add extra permissions at a later stage).

src/open_inwoner/config_checks/registry.py
```diff
@@ -0,0 +1,15 @@
+from collections import defaultdict
+
+
+class ConfigCheckRegistry:
```
swrichards, line 4: Let's include some type hints here.

src/open_inwoner/config_checks/fetch_cases.py
```diff
@@ -1,70 +1,71 @@
...
-    identifier = "fetch_cases_for_bsn"
-    verbose_name = "Fetch cases for BSN"
 
-    def __init__(self, api_group, bsn):
-        self.api_group = api_group
-        self.bsn = bsn
+class FetchCasesCheck:
+    identifier = "fetch_cases"
+    label = "Fetch cases for BSN"
+    form_class = FetchCasesConfigCheckParams
```
swrichards, line 13:
```suggestion
from config_checks.permissions import IsSuperuser, HasPermission

class FetchCasesCheck:
    identifier = "fetch_cases"
    label = "Fetch cases for BSN"
    form_class = FetchCasesConfigCheckParams
    
    required_permissions = [IsSuperuser(), HasPermission("zgw_api_group_edit")]

```

swrichards: Might be possible to do something like: `HasModelRead/Write`

src/open_inwoner/config_checks/forms.py
```diff
@@ -27,9 +18,4 @@ class FetchBRPConfigCheckParams(forms.Form):
 
 
 class FetchUserfeedConfigCheckParams(forms.Form):
-    user = forms.ModelChoiceField(
-        queryset=User.objects.all(),
-        label="User",
-        help_text="User to fetch feed for",
-        empty_label=None,
-    )
+    pass
```
swrichards, line 21: Can we use the Django Forms API to pre-fill certain field base don the `obj` passed?

src/open_inwoner/config_checks/admin_urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path
+
+from open_inwoner.config_checks.views import run_config_check
+
+urlpatterns = [
+    path(
```
swrichards, line 6: Do we need an extra path without app_label/model_name/pk, with only check_id?

src/open_inwoner/config_checks/api.py
```diff
@@ -0,0 +1,67 @@
...
+TModel = TypeVar("TModel")
+
+
+class InteractiveConfigCheck(Protocol[TParamsForm, TModel]):
+    identifier: ClassVar[str]
+    label: ClassVar[str]
+    form_class: ClassVar[type[TParamsForm]]
+
+    def __init__(self, form: TParamsForm): ...
+    def run(self, obj: TModel) -> GenericHealthCheckResult: ...
```
swrichards, line 21:
```suggestion
    def run(self, obj: TModel | None) -> GenericHealthCheckResult: ...
```

src/open_inwoner/config_checks/forms.py
swrichards: Maybe we can somehow pass the obj, if present, to https://docs.djangoproject.com/en/6.0/ref/forms/api/#initial-form-values

src/open_inwoner/config_checks/api.py
swrichards, line 24: Double-check it still works on `InlineAdmin` (and others?).

swrichards (approved):
@OlhaZahoruiko I'll merge this, and then you can handle the permissions and other issues in a followup PR, so we can just focus on the changes.

PR #2380, opened 2026-03-31
:sparkles: [#2375] Config check on InlineAdmin

### LINKED ISSUE #2375: Implement permission checks and stand-alone/pre-filled config checks
As discussed on 31/3, the last step:

- Support specifying permissions
  - staff is a baseline, but we support other flavours too: `IsSuperuser()`, perhaps something related to the admin model/content type permissions, generic permission ('can_run_cases_config_check').
 - Support for both standalone and pre-filled runs
   - The config check can be triggered as now on a _specific object_ in the admin. But this flow should _pre-fill_ data from the object where necessary (maybe through form initial fields)
   - We should be able to navigate directly to a config check, without being attached to a specific instance/object (And thus no pre-fill).

OlhaZahoruiko:
## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables


src/open_inwoner/config_checks/api.py
```diff
@@ -21,15 +22,19 @@ def __init__(self, form: TParamsForm): ...
     def run(self, obj: TModel) -> GenericHealthCheckResult: ...
 
 
-def with_config_checks(*checks):
+def with_config_checks(*checks: InteractiveConfigCheck):
```
swrichards: Another thing I forgot on the earlier PR: let's see if we can add back a kwarg to customize the field name (with a default):


```suggestion
def with_config_checks(*checks: InteractiveConfigCheck, target_field="config_check_links"):
```

src/open_inwoner/config_checks/forms.py
```diff
@@ -18,4 +26,8 @@ class FetchBRPConfigCheckParams(forms.Form):
 
 
 class FetchUserfeedConfigCheckParams(forms.Form):
-    pass
+    api_group = forms.ModelChoiceField(
```
swrichards:
```suggestion
    user = forms.ModelChoiceField(
```

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,49 @@
...
+from django.db.models import Model
+from django.http import HttpRequest
+
+
+class BasePermission:
+    def has_permission(self, request: HttpRequest, obj: Model) -> bool:
+        return request.user.is_active and request.user.is_staff
+
+    def get_error_message(self, obj: Model) -> str:
+        return "You must be an active staff member to perform this check."
```
swrichards: We can make this an `abc.ABC` which just defines the signatures:

```suggestion
class BasePermission(ABC):
    def has_permission(self, request: HttpRequest, obj: Model) -> bool:
        pass

    def get_error_message(self, obj: Model) -> str:
        pass

class IsActiveAdminUser(BasePermission):        
    def has_permission(self, request: HttpRequest, obj: Model) -> bool:
        return request.user.is_active and request.user.is_staff

    def get_error_message(self, obj: Model) -> str:
        return "You must be an active staff member to perform this check."
```

We can then make `IsActiveAdminUser` the default permission class if none are specified.

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,49 @@
...
+class HasModelPermission(BasePermission):
+    def __init__(self, action: str):
+        self.action = action
+
+    def _get_perm_string(self, obj: Model) -> str:
+        opts = obj._meta
+        codename = get_permission_codename(self.action, opts)
+        return f"{opts.app_label}.{codename}"
+
+    def has_permission(self, request: HttpRequest, obj: Model) -> bool:
```
swrichards: These should probably also have to handle the case where there is no instance (e.g. when we navigate to a check directly), so these should probably accept a model class as well as an instance ([this is also supported in the ModelAdmin implementations](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.has_view_permission), which can check permissions on the type of model, not just the instance, we can see how they solve the problem there).

src/open_inwoner/config_checks/receivers.py
```diff
@@ -0,0 +1,21 @@
...
+
+logger = structlog.get_logger(__name__)
+
+
+@receiver(interactive_config_check_triggered)
+def log_interactive_check(sender, request, check_class, obj, result, **kwargs):
+    user = request.user
+    status = "success" if result.success else "failure"
+
+    logger.info(
```
swrichards: Have a look at `from open_inwoner.utils.logentry import user_action`, in our code base it would be good if this is a `user_action`.

src/open_inwoner/config_checks/signals.py
```diff
@@ -0,0 +1,3 @@
+from django.dispatch import Signal
+
+interactive_config_check_triggered = Signal()
```
swrichards: Let's follow the common pre/post pattern here (the current receiver will bind to `interactive_config_check_post_run`):

```suggestion
interactive_config_check_pre_run = Signal()  # <-- will receive the config check instance, but not the result
interactive_config_check_post_run = Signal() # <-- also receives the result, as currently
```

src/open_inwoner/config_checks/views.py
```diff
@@ -2,41 +2,118 @@
...
+        if not perm.has_permission(request, obj)
+    ]
+
+    opts = obj._meta
+    # TODO: doublecheck this
+    try:
+        back_url = reverse(
+            f"admin:{opts.app_label}_{opts.model_name}_change", args=[obj.pk]
+        )
+    except NoReverseMatch:
```
swrichards: Assumes we have an object, but we still have the case of one-off run directly from the config check

src/open_inwoner/config_checks/admin_urls.py
```diff
@@ -8,4 +8,9 @@
         run_config_check,
         name="run_config_check",
     ),
+    # path(
+    #     "admin/config-check/<str:check_id>/",
+    #     run_config_check_standalone,
+    #     name="run_config_check_standalone",
+    # ),
 ]
```
swrichards, line 16: I think we still need this one, to test the direct navigation route.

To test this. let's add a new top-level "Configuratie checks" menu root in `src/open_inwoner/conf/fixtures/django-admin-index.json` and expose direct links to the different checks there.

src/open_inwoner/config_checks/views.py
```diff
@@ -2,41 +2,118 @@
...
+                "failed_reasons": failed_reasons,
+                "back_url": back_url,
+            },
+            status=403,
+        )
+
+    initial = {}
+    if obj:
+        initial["api_group"] = obj
+    form = check_class.form_class(request.POST or None, initial=initial)
```
swrichards: This will somehow have to be done on the config checks, becuase we don't need what initial field to set. We could expose a `get_form_kwargs(self, obj: Model | None)` method on the config check, analogous to how this is done on views, to allow for this, and then use that hook to set `initial` here.

swrichards (changes requested):
Looking good, two things to look at beyond the small comments I left:

- I'd like added here is to explicitly allow for navigating to a check form _without_ first going through a link on a model instance, but straight to the form
- We'll need to expose some kind of `get_form_kwargs` on the config check interface to properly set the initial fields

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_error_message(
+        self,
+        obj: Model | Type[Model] | None,
+    ) -> str:
+        pass
+
+
+class IsActiveAdminUser(BasePermission):
+    def has_permission(self, request: HttpRequest, obj=None) -> bool:
+        return request.user.is_active and request.user.is_staff
```
swrichards: Needs a `request.is_authenticated` guard as well, I am not sure `AnonymousUser` has the `is_active`/`is_staff` fields.

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def _get_perm_string(self, obj: Model | Type[Model]) -> str:
+        opts = obj._meta
+        codename = get_permission_codename(self.action, opts)
+        return f"{opts.app_label}.{codename}"
+
+    def has_permission(self, request: HttpRequest, obj) -> bool:
+        if not request.user.is_active or not request.user.is_staff:
+            return False
+
```
swrichards: Technically not part of this permission class: permissions can be combined, so I think this class should be more narrow and just focus on the model permission

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,82 @@
...
+        obj: Model | Type[Model] | None,
+    ) -> bool:
+        pass
+
+    @abstractmethod
+    def get_error_message(
+        self,
+        obj: Model | Type[Model] | None,
+    ) -> str:
+        pass
```
swrichards, line 30: I was looking at https://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions for inspiration, and I think we should follow an adapted version of their base pattern by explicitly separating general permissions and object permissions. So we'd get:

```suggestion
class BasePermission(ABC):
    @abstractmethod
    def has_permission(
        self,
        request: HttpRequest,
    ) -> bool:
        pass

    @abstractmethod
    def has_object_permission(
        self,
        request: HttpRequest,
        obj: Model,
    ) -> bool:
        pass
        
    @abstractmethod
    def get_error_message(
        self,
        obj: Model | Type[Model] | None,
    ) -> str:
        pass
```

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,82 @@
...
+
+
```
swrichards, line 32: We also need an `IsSuperUser`

src/open_inwoner/config_checks/permissions.py
```diff
@@ -0,0 +1,82 @@
...
+class IsActiveAdminUser(BasePermission):
+    def has_permission(self, request: HttpRequest, obj=None) -> bool:
+        return request.user.is_active and request.user.is_staff
+
+    def get_error_message(self, obj=None) -> str:
+        return "You must be an active staff member to perform this check."
+
+
+class HasModelPermission(BasePermission):
+    def __init__(self, action: str):
```
swrichards: To handle the case above, where we split `has_permission` and `has_object_permission`, we should pass the model class to `__init__` so you can define it on the class with:

```
class ConfigCheck:
    permission_classes = (IsSuperUser(), HasModelPermission(FooModel, action="create"),)
```

src/open_inwoner/config_checks/fetch_cases.py
```diff
@@ -4,25 +4,61 @@
...
+from .permissions import HasModelWrite
 
 
 class FetchCasesCheck:
     identifier = "fetch_cases"
     label = "Fetch cases for BSN"
     form_class = FetchCasesConfigCheckParams
+    required_permissions = [HasModelWrite()]
 
     def __init__(self, form: FetchCasesConfigCheckParams):
```
swrichards: One more API change: I don't actually think passing the `form` to `__init__` helps much, and just creates extra boilerplate. I think the `run` signature should be:

```
# form is injected to run, both methods accept `obj` as nullable in case the check is run directly
@classmethod
def get_form_kwargs(cls, obj: TModel | None):
   ...
def run(self, form, obj: TModel | None = None):
    ...
```

swrichards (changes requested):
Final API tweaks for now, let me know if anything is unclear. Then we'll move on to your other PR.

src/open_inwoner/config_checks/views.py
```diff
@@ -2,41 +2,136 @@
...
 
+    check_class = registry.get_check_by_identifier(model, check_id)
     if not check_class:
         raise Http404()
 
-    form = check_class.form_class(request.POST or None)
+    permissions = getattr(
+        check_class,
+        "required_permissions",
+        (IsActiveAdminUser(),),
```
swrichards: I would set this in `api.py` as `get_default_permissions_classes` (we'll use a method to avoid accidental mutatability), and let's be a bit more defensive and use `IsSuperUser()` by default.

src/open_inwoner/config_checks/views.py
```diff
@@ -2,41 +2,136 @@
...
+#
+#     return render(
+#         request,
+#         "admin/run_config_check.html",
+#         {
+#             "title": f"{check_class.label}",
+#             "form": form,
+#             "result": result,
+#         },
+#     )
```
swrichards: Legacy?

src/open_inwoner/config_checks/views.py
```diff
@@ -2,41 +2,136 @@
...
 from django.http import Http404
 from django.shortcuts import get_object_or_404, render
+from django.urls import NoReverseMatch, reverse
 
+from .permissions import IsActiveAdminUser
 from .registry import registry
+from .signals import interactive_config_check_post_run, interactive_config_check_pre_run
 
 
 @staff_member_required
```
swrichards: This is a bit odd, because it means we either have two places defining default permissions classes (this decorator and the fallback for `required_permissions`), and our `IsAdminUser` is a bit redundant because it's enforced by this decorator. Let's remove this for now and a be a bit stricter in our check of the permission classes below.

src/open_inwoner/config_checks/views.py
swrichards: In light of the decorator discussion above, I would suggest we:

1. First get `required_permissions` with `get_default_permissions_classes` as fallback
2. Check for a falsy value, raise with "You must define at least one permission class" to guard against e.g. an empty list
3. Validate that we have an iterable of `BasePermission` classes, raise if not ("All permission classes must inherit from BasePermission").

swrichards (changes requested):

src/open_inwoner/config_checks/api.py
```diff
@@ -65,3 +86,30 @@ def config_check_links(self, obj):
             ],
         ),
     )
+
+
+def resolve_permissions(check_class) -> tuple[BasePermission, ...]:
```
swrichards, line 91: This is nice and explicit, the issue I'd say is that this does change the API to effectively make `required_permissions` mandatory, because there's no fallback value (my comment was mostly to avoid people explicitly _emptying_ the attribute, rather than omitting it, which might be justified). That's fine for the moment, it's really a discussion for the documentation PR with the other team members whether it makes sense to have a fallback default or require people to provide the class. Let's use this for now, just an observation that this changes the interface somewhat.

src/open_inwoner/config_checks/views.py
swrichards: @OlhaZahoruiko I don't think we need this now, given that your recent changes make setting permission classes mandatory. It could lead to unexpected results for users creating a `IsAnyUser`, and then still getting a 403.

swrichards (changes requested):
Almost there!

swrichards (approved):

PR #2383, opened 2026-04-02
:accessibility: [#2270] Add live region for HTMX content, update spin…

### LINKED ISSUE #2270: Dat wat je ziet, moet je ook kunnen horen – dat kan met een statusbericht. Dus in dit geval bij ‘gegevens laden’.

stefrado:
## Summary

Add live region for HTMX content, update spinners

> This PR closes #2349 

## Issue References

- Github Issue: #2270 

## Checklist

- [x] CHANGELOG.rst updated (updated in main PR for #2268)


src/open_inwoner/js/components/cases/spinner.js
```diff
@@ -1,12 +1,29 @@
+const notifyLoading = (e) => {
+  // Generic: announce loading for any HTMX element with data-spinner-live-region
```
jiromaykin: Now that this Spinner needs to become generic, it would need to be taken outside of the cases vanilla components and become its own component,
also perhaps move all functionality to _src/open_inwoner/react/components/Spinner/Spinner.tsx_ instead.

Feels maybe a little bit out of scope, but if we have these as a functional web-component is will become easier to re-use.

src/open_inwoner/js/components/cases/spinner.js
jiromaykin: These make the spinner show/hide logi hardcoded to `cases-content`, `submissions-content`, and `spinner-container`. So that makes it less generic - but doing that may be out of scope.

src/open_inwoner/templates/pages/cases/status_outer.html
```diff
@@ -2,16 +2,17 @@
 {% load i18n icon_tags %}
 
 {% block notifications %}
+    <div id="spinner-live-region" class="sr-only" role="status" aria-live="polite" aria-atomic="true"></div>
     <div id="cases-status-messages-oob"></div>
 {% endblock notifications %}
 
 {% block content %}
     <div class="case-detail" id="cases-detail-content"><div>
```
jiromaykin: Old mystery issue: 
this Div is not closed correctly, yet when one corrects the `</div>` the spinner seems to not swap correctly anymore. This problem is outside of the scope of this issue but you may be able to take a look at it here as well.

src/open_inwoner/templates/pages/cases/status_inner.html
```diff
@@ -1,5 +1,8 @@
 {% load i18n utils anchor_menu_tags card_tags file_tags form_tags grid_tags solo_tags link_tags button_tags icon_tags notification_tags render_tags %}
 
+{# Screenreader live region - responds to Spinner js - maybe not necessary #}
```
jiromaykin: This change is not needed; the live region belongs in _status_outer.html,_ which is correct there.

jiromaykin (changes requested):

src/open_inwoner/js/components/cases/spinner.js
stefrado: I spend some time fixing this and i think it is worth it.

stefrado:
@jiromaykin based on your comment i thought about the current implementation and refactored some of the spinner logic. This also solved the mysterious div issue.

src/open_inwoner/js/components/spinner/index.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (!cached) return;
+
+    const spinners = cached.spinners ?? this.getSpinners(targetId);
+
+    // Hide spinners (error occured).
+    spinners.forEach((spinner) =>
+      spinner.classList.add('loader-container--hide')
+    );
+
+    // Update error text.
```
jiromaykin, line 174: Before this (the anyError part is only visible for sighted users), the error still needs to actually be 'sent' as an update to the Live regions, or else the failure won't be read to screenreaders immediately:
so add something like 

```
// Announce that loading failed (Accessibility)
    const errorText = this.getSpinnerText(
      spinners,
      'error-text',
      'Er is iets misgegaan bij het ophalen van uw zaken. Ververs de pagina of probeer het later opnieuw.'
    );
    cached.liveRegions.forEach((region) => (region.textContent = errorText));
```

but then the type for getSpinnerText starts yelling , so `error-text`will need to be added to the 'attr'.

Or in a different way:

```
cached.liveRegions.forEach((region) => (region.textContent =
      anyError?.dataset.errorMessage ??
      'Er is iets misgegaan bij het ophalen van uw zaken. Ververs de pagina of probeer het later opnieuw.'
    ));
```

just as long as it 'updates' the Live region.

jiromaykin (changes requested):

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
```diff
@@ -21,18 +21,25 @@
...
                 title="{{ plugin_title }}"
                 next-url="{{ mijn_zaken_url }}"
                 show-indicator="false"
                 next-url-label="{% trans 'Bekijk alle zaken' %}"
                 columns="1"
             >
-                <oip-loading-spinner loading-text="{% trans 'Zaken laden...' %}"></oip-loading-spinner>
+                <oip-loading-spinner
+                    loading-text="{% trans 'Zaken laden...' %}"
+                    loaded-text="{% trans 'Zaken zijn geladen.' %}"
```
jiromaykin, line 35: Shouldn't this also have the compact prop?

jiromaykin (approved):
Just 1 question about the compact prop.

src/open_inwoner/templates/cms/plugins/zaken/zaken.html
stefrado: Yes, this looks way better!

PR #2399, opened 2026-04-09
fix: [#3298] use standard logout URL for DigiD when OIDC is disabled

swrichards:
DigiD SAML single logout has been deprecated, so the rp-initiated logout URL is no longer valid. Fall back to the standard logout URL instead of the (non-existent) DigiD SAML logout endpoint.

Closes: #3298


swrichards:
#2400

PR #2401, opened 2026-04-09
Refactor BRP/Haalcentraal clients to be vendor agnostic

### LINKED ISSUE #2362: PP-357: Add enable U headings for BRP

### LINKED ISSUE #2411: SSL Verification is disabled for BRP client requests
No good reason I can see.

swrichards:
Closes: #2362 

----
📚 Documentation preview 📚: https://open-inwoner--2401.org.readthedocs.build/en/2401/


src/open_inwoner/haalcentraal/validators.py
```diff
@@ -1,6 +1,81 @@
...
+    if conflicts:
+        raise ValidationError(
+            _(
+                "The following header names conflict with standard HTTP headers "
+                "and cannot be used: {headers}"
+            ).format(headers=", ".join(conflicts))
+        )
+
 
 def validate_verwerking_header(value: str) -> None:
```
pi-sigma, line 80: It looks like this function is no longer used since the field has been removed, but is needed for the squashed migrations? If that's correct, add a comment/docstring for this.

src/open_inwoner/haalcentraal/clients.py
```diff
@@ -9,23 +9,42 @@
...
 
-    def __init__(self):
-        self.config = HaalCentraalConfig.get_solo()
-        if not self.config.service:
+    def __init__(self, client, extra_headers: dict | None = None):
+        self.client = client
+        self.extra_headers = extra_headers or {}
+
+    @classmethod
+    def from_config(cls) -> "BRPClient":
```
pi-sigma, line 21: The method doesn't use `cls` but returns one of the hard-coded clients. So `BRPClient_1_3.from_config()` could return a `BRPClient_2_0`, depending on the config. Perhaps a module-level `create_brp_client` would be more transparent.

src/open_inwoner/haalcentraal/clients.py
pi-sigma, line 1: `parse_data` has `if not data["personen"]`, which could raise a `KeyError`. Not your change, but worth adding a guard anyways

pi-sigma (changes requested):
Do we need the migrations 0009 and 0019 separately? Could be moved into 0008 I think. Minor issue though.

swrichards:
> Do we need the migrations 0009 and 0019 separately? Could be moved into 0008 I think. Minor issue though.

I wanted to keep the split commits, and given there's some complexity in the migrations, there's value in testing them separately. I squashed some old ones though to compensate :) but I'd like to leave this as-is for readability.

src/open_inwoner/haalcentraal/validators.py
swrichards: Good point, added a comment and updated the docs.

src/open_inwoner/haalcentraal/clients.py
swrichards: I think it's fine as is, the the concrete implementations implement the `BRPClient` interface, and I think the class factory has better discoverability than module level methods.

src/open_inwoner/haalcentraal/clients.py
swrichards: This comment triggered a whole range of further refactors, cleanups and moving things around (in this case, specifically, explicit error handling and no exception swallowing at the bottom of the stack). Apologies, but we don't often get a chance to do this right, so I went ahead and fixed all this.

src/open_inwoner/haalcentraal/clients.py
```diff
@@ -0,0 +1,174 @@
...
+                "type": "RaadpleegMetBurgerservicenummer",
+                "burgerservicenummer": [user_bsn],
+            },
+            headers=headers,
+        )
+        return response
+
+    def fetch_brp_data_for_bsn(self, user_bsn: str) -> BRPData | None:
+        data = get_json_response(self.make_request(user_bsn))
+        if not data or not data["personen"]:
```
pi-sigma:
```suggestion
        if not data or not data.get("personen"):
```

src/open_inwoner/accounts/signals.py
```diff
@@ -30,6 +31,35 @@
 logger = structlog.stdlib.get_logger(__name__)
 
 
+def _update_user_from_brp(user: User, initial=True):
```
pi-sigma:
```suggestion
def _update_user_from_brp(user: User):
```

the argument is not used inside the function

src/open_inwoner/accounts/signals.py
```diff
@@ -66,7 +96,7 @@ def update_user_on_login(sender, user, request, *args, **kwargs):
 
     # update brp fields when login with digid and brp is configured
     if user.is_bsn_user and HaalCentraalConfig.get_solo().service:
```
pi-sigma:
```suggestion
    if user.is_bsn_user:
```

`_update_user_from_brp` already calls `BRPClient.from_config()` and catches `ImproperlyConfigured` for missing service.

src/open_inwoner/accounts/signals.py
```diff
@@ -30,6 +31,35 @@
...
+        logger.warning("no service configured for Haal Centraal")
+        return
+
+    try:
+        brp = client.fetch_brp_data_for_bsn(user.bsn)
+    except Exception:
+        logger.exception("unexpected error fetching BRP data")
+        return
+
+    if not brp:
```
pi-sigma, line 54: Why pass silently? Could use `logger.warning("missing BRP")`

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -26,17 +26,13 @@ class BRPData:
...
-        user.infix = self.infix
-        user.last_name = self.last_name
-
-        user.street = self.street
-        user.housenumber = self.get_housenumber()
-
-        user.city = self.city
-        user.birthday = self.birthday
+    @classmethod
+    def parse_date(cls, data, path, default=None):
```
pi-sigma:
```suggestion
    def parse_date(cls, data: dict, path: str, default: date | None = None) -> date | None:
```

src/open_inwoner/accounts/models.py
```diff
@@ -766,6 +769,21 @@ def is_bsn_user_with_brp(self) -> bool:
...
+        self.last_name = brp_data.last_name
+
+        self.street = brp_data.street
+        self.housenumber = brp_data.get_housenumber()
+
+        self.city = brp_data.city
+        self.birthday = brp_data.birthday
+
+        self.is_prepopulated = True
+        self.save()
```
pi-sigma: Instead of bare `.save()` you could pass the `update_fields`.

Incidentally, the code for updating KVK users (in `accounts/signals.py`) is inconsistent in this regard: for company_name we have `user.save(update_fields=["company_name"])` but for branch_name we have bare `user.save()`.

pi-sigma (changes requested):
A few more small things

src/open_inwoner/accounts/signals.py
swrichards: You're right, and even stronger: I am just going to raise in the client. The days of swallowing errors are _over_.

swrichards: Mind you: this could just mean user is not known to gemeente (`None` is a valid return), but I'll nevertheless raise in the client to distinguish that case from unknown.

swrichards: @pi-sigma made a few more changed triggered by this: stopped the "swallow the error in the client" pattern, and moved the DTO stuff to the accounts module. Haalcentraal client is a client: it fetches, serializes, and raises if anything goes wrong. Converting the data to the user model and error handling happens in the accounts module.

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -1,52 +1,84 @@
...
+        parts = [str(self.huisnummer)] if self.huisnummer is not None else []
+        if self.huisletter:
+            parts.append(self.huisletter)
+        if self.huisnummertoevoeging:
+            parts.extend([" ", self.huisnummertoevoeging])
         return "".join(parts)
+
+
+# BRP 1.3 (flat verblijfplaats)
+class BRPVerblijfplaats13(BRPAdres):
```
pi-sigma: `class BRP13Verblijfplaats` makes more sense to me (the version is of BRP after all)

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -1,52 +1,84 @@
...
+    straat: str = ""
+    land: Waardetabel = Field(default_factory=Waardetabel)
+
+
+class BRPPersoon13(BaseModel):
```
pi-sigma: `BRP13Persoon` - same as previous

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -1,52 +1,84 @@
...
+# BRP 2.x (nested verblijfadres)
+class BRPVerblijfadres(BRPAdres):
+    officieleStraatnaam: str = ""
+
+
+class BRPVerblijfplaats2x(BaseModel):
+    verblijfadres: BRPVerblijfadres = Field(default_factory=BRPVerblijfadres)
+
+
+class BRPPersoon2x(BaseModel):
```
pi-sigma: `BRP2xPersoon`

src/open_inwoner/haalcentraal/api_models.py
```diff
@@ -1,52 +1,84 @@
...
+    """Common house-number fields shared by both API versions."""
+
+    huisnummer: int | None = None
+    huisletter: str = ""
+    huisnummertoevoeging: str = ""
+    woonplaats: str = ""
+    postcode: str = ""
+
+    @property
+    def housenumber_string(self) -> str:
```
pi-sigma: Dead code, remove

src/open_inwoner/accounts/brp.py
```diff
@@ -0,0 +1,83 @@
...
+            infix=persoon.naam.voorvoegsel,
+            last_name=persoon.naam.geslachtsnaam,
+            initials=persoon.naam.voorletters,
+            birthday=persoon.geboorte.datum.datum,
+            birth_place=persoon.geboorte.plaats.omschrijving,
+            gender=gender,
+            street=street,
+            housenumber=housenumber,
+            houseletter=houseletter,
+            housenumbersuffix=housenumbersuffix,
```
pi-sigma, line 79: The individual housenumber fiels are not used anywhere, as far as I can see. We always call `get_housenumber` to re-assemble them. Could just store the assembled house number directly here. OTOH, we might need them in the future. Your call.

src/open_inwoner/accounts/models.py
```diff
@@ -27,7 +27,9 @@
 from privates.storages import PrivateMediaFileSystemStorage
 from timeline_logger.models import TimelineLog
 
+from open_inwoner.accounts.brp import BRPData
 from open_inwoner.configurations.models import SiteConfiguration
+from open_inwoner.haalcentraal.api_models import BRPPersoon2x, BRPPersoon13
```
pi-sigma: These are only imported for type checking. Consider moving them to the bottom:

```
from typing import TYPE_CHECKING
...
if TYPE_CHECKING:
    import ...
```

pi-sigma (approved):
Mostly good to go. Suggestions only.

src/open_inwoner/accounts/brp.py
swrichards: As this is a DTO I'll leave it in for now.

Issue #2409, opened 2026-04-13
Refactor web-component logic and reimplement `preact-custom-element`.


stefrado:
- Creating a fork of `preact-custom-element` and use this, which allows `ElementInternals`
- Implement the first `shadow: true` web components.
- Use `vitest/browser` and `playwright`

PR #2310, opened 2026-03-12
🏗️ Refactor preact components so they make use of the Shadow DOM

stefrado:
## Summary

> De web component-architectuur is gerefactored om shadow DOM te ondersteunen. Componenten (``Accordion``, ``Action``, ``ActionList``, ``Chart``) beschikken nu over factory-functies die een web component registreren. De testopzet is herschreven: gedeelde test-utilities zijn ondergebracht in ``react/lib/testing/``, Storybook-decorators en web component-decorators zijn gesplitst in afzonderlijke modules, en de Vitest-configuratie is bijgewerkt. Daarnaast zijn de CI-workflow en Storybook-configuratie samengevoegd en gemoderniseerd.

This PR contains the required changes we need to make sure we can use `ElementInternals` and the usage of a `shadowDOM` in out web components.

### Notes for the reviewer.

This PR got a bit out of hand and most of the changes are dependent on each other, this makes it quite complicated to make sure the changes are solid (since a lot of changes are needed). I did the following things:

#### (1) I recreated the package `preact-custom-elements`

I did this to make sure we can use `AttachInternals` and have more flexibility overal. But i encountered a problem with our current test set-up, which is not actually testing in a real browser and has not all `shadowDOM` attributes present (`serializable` option). This caused me to use a `playwright` test set-up (`open-formulieren/admin-ui` as a reference).

The recreation of the `preact-custom-elements` went besides that very smooth.

Changing our test set-up meant that we could test storybook better in the CI, so i copied more of the OF/admin-ui code to extend this testing. This is actually what is causing most of the changes.

> [!NOTE]
> Changing this test set-up is not really a breaking change since it is not touching the actual code, but is dev-only.
> The focus should lie on the renderer for this step.

#### (2) I created shadow web-components from the `action-list` (now `oip-action` and `oip-action-list`) and `oip-accordion`.

Not really big changes, but this is the first `shadowDOM` implementation (for this PR).

## Issue References

- Github Issue:

## Checklist

- [x] CHANGELOG.rst updated
- [x] Codecov report reviewed for untested lines
- [x] Storybook component updated/created
- [x] Vitest tests added/updated 
- [x] Updated the `docker-compose.yml`

PR #2410, opened 2026-04-13
Vitest/impelement coverage

stefrado:
## Summary

Implement coverage reports for vitest testing.

```zsh
npm run vitest:coverage
```

## Issue References

- Github Issue: #2409 

## Checklist

- [x] CHANGELOG.rst updated (same changelog as #2310)

> [!WARNING]
> Merging to branch should be updated !


package.json
```diff
@@ -21,6 +21,7 @@
     "test:vitest": "vitest --run --project=unit",
     "vitest:ui": "vitest --ui",
     "vitest:watch": "vitest --watch",
+    "vitest:coverage": "vitest --watch --coverage",
```
jiromaykin: why not:

```
"vitest:watch": "vitest --watch",
"vitest:coverage": "vitest run --coverage",
"vitest:coverage:watch": "vitest --watch --coverage"
```

so we can split the --watch option if we want to?

jiromaykin (changes requested):

package.json
stefrado: Good point! I added --coverage to UI as well since this just adds a button to see the report in the UI:

image

jiromaykin (approved):

PR #2415, opened 2026-04-13
[#2414] Add ClamAV antivirus scanning of uploads

### LINKED ISSUE #2414: Add support for scanning files for malware with ClamAV
[Validator along these lines](https://github.com/open-formulieren/open-forms/blob/b3da37df24c9d309047d258b4ed3022e705fbc81/src/openforms/formio/api/validators.py#L154-L198) to be added to the ZGW document / Samenwerking file upload.

- [ ] Update config with ClamAV connections
- [ ] Update docker compose
- [ ] Add validator
- [ ] Add e2e test

swrichards:
Closes #2414

---

To test:

1. Ensure clamav is running: `docker compose up -d clamav`
2. Enable the flag under "Antivirus" on the global config page
3. Navigate to a case or plan with a file upload
4. Upload the EICAR test file (ping @swrichards for this).


src/open_inwoner/configurations/models.py
```diff
@@ -710,6 +710,33 @@ class SiteConfiguration(SingletonModel):
         ),
     )
 
+    # Virus scan (ClamAV)
+    enable_virus_scan = models.BooleanField(
+        verbose_name=_("Enable virus scan"),
+        default=False,
+        help_text=_(
+            "When enabled, uploaded files are scanned for viruses using ClamAV. "
+            "Requires a reachable ClamAV daemon at the configured host and port."
```
pi-sigma, line 719:
```suggestion
            "Requires a ClamAV service reachable at the configured host and port."
```

Simpler language for admins

src/open_inwoner/utils/validators.py
```diff
@@ -210,3 +215,69 @@ def validate_array_contents_non_empty(list_: list) -> None:
...
+        scanner = clamd.ClamdNetworkSocket(
+            host=config.clamav_host,
+            port=config.clamav_port,
+            timeout=config.clamav_timeout,
+        )
+
+        uploaded_file.seek(0)
+
+        try:
+            result = scanner.instream(uploaded_file)
```
pi-sigma, line 254: You're reading the file without resetting the file pointer to 0 (in the `OK` branch below). You have a test for this, but it only checks `seek` at the beginning of the scan, without advancing the pointer to simulate file reading during the scan.

src/open_inwoner/configurations/models.py
```diff
@@ -710,6 +710,33 @@ class SiteConfiguration(SingletonModel):
...
+        ),
+    )
+    clamav_host = models.CharField(
+        verbose_name=_("ClamAV host"),
+        max_length=255,
+        default="clamav",
+        blank=True,
+        help_text=_("Hostname or IP address of the ClamAV daemon."),
+    )
+    clamav_port = models.PositiveIntegerField(
```
pi-sigma, line 729: Add `MaxValueValidator` to exclude invalid large numbers (port number is an unsigned 16-bit int, so max value is 65535).

src/open_inwoner/configurations/models.py
pi-sigma, line 1: the class could use cross-field validation on `enable_virus_scan`, `clamav_host`, and `clamav_host` to ensure that if `enable_virus_scan` is `True`, the other fields are non-empty.

pi-sigma (changes requested):

src/open_inwoner/configurations/models.py
swrichards: Excellent idea.

src/open_inwoner/utils/validators.py
swrichards: Well spotted

pi-sigma (approved):

PR #2429, opened 2026-04-16
PP-324 #2427 Update paragraph stories and tests

### LINKED ISSUE #2427: PP-324 Update Paragraph tests/storybook with shadow DOM
Now that shadow DOM PR has been merged, update the paragraph storybook/tests.

jiromaykin:
## Summary

Now that shadow DOM PR has been merged, update the paragraph component stories and tests.

## Issue References

- sub issue for #2427 

## Checklist

- [x] CHANGELOG.rst updated
      
### DIFF
--- a/src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
+++ b/src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
@@ -1,71 +1,35 @@
-import { describe, expect, it, vi } from 'vitest';
 import { render } from '@testing-library/preact';
-import '@testing-library/jest-dom';
-import clsx from 'clsx';
-import Paragraph from './Paragraph';
-
-/**
- * @nl-design-system-candidate/paragraph-react is React-only and incompatible with
- * the Preact + JSDOM test environment.
- *
- * TODO: Refactor this once shadow DOM PR #2310 is merged
- *
- * For now: Paragraph.tsx renders <slot /> instead of {children} so slotted content is
- * never present in the JSDOM tree, therefore these tests query by class and DOM structure
- * instead of by text content.
- */
-vi.mock('@nl-design-system-candidate/paragraph-react', () => ({
-  Paragraph: ({ children, className, purpose }: any) => (
-    <p
-      class={clsx(
-        'nl-paragraph',
-        purpose === 'lead' ? 'nl-paragraph--lead' : '',
-        className
-      )}
-    >
-      {purpose === 'lead' ? (
-        <b class="nl-paragraph__lead">{children}</b>
-      ) : (
-        children
-      )}
-    </p>
-  ),
-}));
+import { describe, expect, it, beforeAll, afterEach } from 'vitest';
+import { Paragraph, PARAGRAPH_DEFINITION } from '.';
+import { WebComponentLoader } from '@react/lib/web-component';
 
 describe('Paragraph', () => {
-  it('renders a paragraph element without crashing', () => {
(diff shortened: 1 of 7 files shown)


src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
```diff
@@ -1,60 +1,24 @@
...
-        children
-      )}
-    </p>
-  ),
-}));
+import { describe, expect, it, beforeAll, afterEach } from 'vitest';
+import { Paragraph, PARAGRAPH_DEFINITION } from '.';
+import { WebComponentLoader } from '@react/lib/web-component';
 
 describe('Paragraph', () => {
```
stefrado, line 6: I think the new spec tests should contain two test parts:

1. validate constants.
  - These tests contain checks to see if we expect the correct data for a component. 
2. validate web component rendering in spec
  - Just a simple test to `assertEqual` the complete `nl-paragraph` (not split up in multiple smaller tests).

> Our storybook set-up should assert the different component states in a `play` test.

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
```diff
@@ -32,13 +30,29 @@ export default meta;
 
 export const Default: Story = {
   render: ({ text }) => <nl-paragraph>{text}</nl-paragraph>,
```
stefrado: Can't we use the `children` prop for this and use that to render the content inside these stories. Then `type Story` can just be `StoryObj<typeof meta>`.

Also i found out that the default `<slot>` inside `Paragraph` should be changed to `{children}`... This is required to pass a context. Which is not needed for this component, but we should set the standard here.

stefrado (changes requested):

src/open_inwoner/react/components/Paragraph/Paragraph.stories.tsx
jiromaykin: Oh my, okay I will look into it.

jiromaykin: New note: in the updated code `<slot>` is rendered internally by `@nl-design-system-candidate/paragraph-react`, not by our `Paragraph.tsx` anymore. So that's what we want?

src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
```diff
@@ -4,33 +4,17 @@ import { Paragraph, PARAGRAPH_DEFINITION } from '.';
...
-  it('renders lead paragraph with nl-paragraph--lead class', () => {
-    const { container } = render(<Paragraph purpose="lead" />);
-    expect(container.querySelector('p')).toHaveClass('nl-paragraph--lead');
-  });
-
-  it('does not render lead class when purpose is not set', () => {
-    const { container } = render(<Paragraph />);
-    expect(container.querySelector('p')).not.toHaveClass('nl-paragraph--lead');
+  describe('constants', () => {
+    it('has the correct definition', () => {
```
stefrado: Nitpick.. `has the expected definition`

stefrado (approved):
It is a bit hard to follow what is exactly changed in this PR, but the feedback seems to be implemented and works fine 🔥

src/open_inwoner/react/components/Paragraph/Paragraph.spec.tsx
jiromaykin: Nitpicking is good!

PR #2441, opened 2026-04-21
:sparkles: [#2440] Add config checks framework

### LINKED ISSUE #2440: Move Config Checks from PoC to Production Usage
The current interactive config-check implementation works as a proof of concept. The next step is to turn it into production-ready functionality focused on real operational checks.

- [x] 1. Review the current config_checks implementation and copy/adapt the useful parts into the new PR.
- [x] 2. Restrict default access to superusers only.
- [x] 3. Extend FetchCasesCheck from basic demo behavior into a real operational validation check.
- [x] 4. Ensure FetchCasesCheck performs actual functional validation and handles expected failure scenarios gracefully.
- [x] 5. Add tests.

OlhaZahoruiko:
## Issue References

- Taiga User Story:
- Taiga Issue:
- Taiga Dimpact Issue:
- Github Issue:

## Checklist

- [ ] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables


src/open_inwoner/openzaak/config_checks/fetch_cases.py
swrichards, line 1: Let's have this in the `openzaak` model, to follow the pattern `{django_app}/config_checks.py` (or `django_app/config_checks/fetch_cases.py` if we want to separate them out)

src/open_inwoner/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,27 @@
...
+from .results import CheckResult
+
+
+class FetchCasesForm(forms.Form):
+    bsn = forms.CharField(max_length=9, label="BSN")
+
+
+class FetchCasesCheck:
+    identifier = "fetch_cases"
+    label = "Fetch cases for BSN"
```
swrichards: Wrap in `gettext as _`

src/open_inwoner/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,27 @@
...
+        self.form = form
+
+    def run(self, obj):
+        bsn = self.form.cleaned_data["bsn"]
+
+        return CheckResult(
+            success=True,
+            identifier=self.identifier,
+            verbose_name=self.label,
+            message=f"Check executed for BSN {bsn}",
```
swrichards: Translate

src/open_inwoner/openzaak/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,27 @@
...
+
+class FetchCasesForm(forms.Form):
+    bsn = forms.CharField(max_length=9, label="BSN")
+
+
+class FetchCasesCheck:
+    identifier = "fetch_cases"
+    label = "Fetch cases for BSN"
+    form_class = FetchCasesForm
+
```
swrichards, line 37: Needs `IsSuperuser()`

swrichards (commented):
@OlhaZahoruiko will continue this later when https://github.com/maykinmedia/open-inwoner/pull/2380 is done -- started here and realized the other should be done first.

src/maykin_config_checks/admin_urls.py
swrichards, line 1: All the actual library changes can be applied to `src/config_checks`: that's the main purpose of vendoring the library, so we can continue its development here but in the right place/module structure, which will allow us to extract it later. 

All these changes can be added there, the only thing that lives in `src/open_inwoner` is the actual config check.

src/maykin_config_checks/api/api.py
```diff
@@ -0,0 +1,47 @@
+from django.urls import reverse
+from django.utils.html import format_html
+
+from .registry import registry
+
+
+def with_config_checks(*checks):
```
swrichards: Let's keep the kwarg for the target link here, it's a quick fix and will give us some freedom in how we organize the admin.

src/maykin_config_checks/permissions.py
```diff
@@ -0,0 +1,6 @@
+class IsSuperUser:
+    def has_permission(self, request, obj=None):
+        return request.user.is_active and request.user.is_superuser
+
+    def get_error_message(self, obj=None):
+        return "Only superusers may run configuration checks."
```
swrichards: Let's wrap in gettext

src/maykin_config_checks/views.py
```diff
@@ -0,0 +1,68 @@
...
+from django.urls import NoReverseMatch, reverse
+
+from .registry import registry
+
+
+def run_config_check(request, app_label, model_name, check_id, pk=None):
+    model = apps.get_model(app_label, model_name)
+
+    if model is None:
+        raise Http404("Model not found")
```
swrichards:
```suggestion
        raise Http404(f"Model {app_label}_{model_name} not found")
```

src/maykin_config_checks/views.py
```diff
@@ -0,0 +1,68 @@
...
+
+    obj = None
+    if pk is not None:
+        obj = get_object_or_404(model, pk=pk)
+
+    check_class = registry.get_check(model, check_id)
+
+    if check_class is None:
+        raise Http404("Check not found")
```
swrichards:
```suggestion
        raise Http404("No check registered for model with id={check_id}")
```

src/maykin_config_checks/registry.py
```diff
@@ -0,0 +1,16 @@
+class ConfigCheckRegistry:
+    def __init__(self):
+        self._registry = {}
+
+    def register(self, model, check):
```
swrichards: I only really noticed that now, but I am not sure we actually have to register checks per model, and we explicitly do not want to do that for the case that (1) the check can be executed outside of a model instance and (2) is explicitly multi-model, such as the user feed check.

src/maykin_config_checks/admin_urls.py
```diff
@@ -0,0 +1,16 @@
+from django.urls import path
+
+from .views import run_config_check
+
+urlpatterns = [
+    path(
```
swrichards, line 6: As per my other comment, I don't think the stand-alone runner should be tied to a specific model (this is also in line with Sergei's comments on the PR). It should just accept the check id, and nothing else, in addition to the other URL entry which _does_ accept a model ID.

src/maykin_config_checks/views.py
```diff
@@ -0,0 +1,68 @@
+from django.apps import apps
+from django.http import Http404, HttpResponseForbidden
+from django.shortcuts import get_object_or_404, render
+from django.urls import NoReverseMatch, reverse
+
+from .registry import registry
+
+
+def run_config_check(request, app_label, model_name, check_id, pk=None):
+    model = apps.get_model(app_label, model_name)
```
swrichards: As noted in other comments: the stand-alone is not connected to a model. `app_label_ and `model_name` are optional, we should only validate them if they exist (because then it's being executed on a model instance).

src/open_inwoner/openzaak/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,160 @@
...
+    identifier = "fetch_cases"
+    label = _("Fetch cases for BSN")
+    form_class = FetchCasesForm
+
+    required_permissions = (IsSuperUser(),)
+
+    def get_target_object(self, form, obj):
+        return form.cleaned_data.get("api_group") or obj
+
+    def run(self, form, obj=None):
```
swrichards: I find myself convinced by Sergei's suggestion at https://github.com/maykinmedia/config-checks/pull/12#discussion_r3123840268 -- we should inject the cleaned data here, not the form, because that keeps open the future use-case of running the checks from somewhere other than the Django admin.

src/maykin_config_checks/api/api.py
```diff
@@ -0,0 +1,47 @@
+from django.urls import reverse
+from django.utils.html import format_html
+
+from .registry import registry
+
+
```
swrichards, line 8: I think we should define the base protocol for the interactive config checks here. Based on the comments in the thread, I think we can maybe go with this:

```python
from typing import Any, Generic, Optional, Protocol, TypeVar

from django.db.models import Model
from django.forms import Form

from maykin_config_checks import GenericConfigCheckResult

TCleanedData = TypeVar("TCleanedData", bound=dict[str, Any], contravariant=True)
TModel = TypeVar("TModel", bound=Model)


class InteractiveConfigCheck(Protocol, Generic[TCleanedData, TModel]):

    check_id: str
    label: str
    form_class: type[Form]

    @classmethod
    def get_form_kwargs(cls, instance: Optional[TModel] = None) -> dict[str, Any]:
        """
        Get keyword arguments for form instantiation.

        Args:
            instance: Optional Django model instance to use for form initialization

        Returns:
            Dictionary of keyword arguments to pass to the form class
        """
        ...

    def run(
        self,
        data: TCleanedData,
        instance: Optional[TModel] = None,
    ) -> GenericConfigCheckResult:
        ...
```

The big difference here is that we don't pass the form to `run` but the cleaned data, so we can run the check outside of a form as well.

This means we would do something like:

```
class FetchCasesCheckForm(forms.Form):
    api_group = forms.ModelChoiceField(
        queryset=ZGWApiGroupConfig.objects.all(),
        required=False,
        help_text=_("Select API group (required if not running from object page)"),
    )
    bsn = forms.CharField(max_length=9, label="BSN")

class FetchCasesCheckParams(TypedDict):
    api_group: ZGWApiGroupConfig
    bsn: str

class FetchCasesCheck(InteractiveConfigCheck[FetchCasesCheckParams, ZGWApiGroupConfig]):
    identifier = "fetch_cases"
    label = _("Fetch cases for BSN")
    form_class = FetchCasesForm
```

src/maykin_config_checks/templates/admin/run_config_check.html
swrichards, line 1: I think this should be in `src/maykin_config_checks` no?

src/open_inwoner/openzaak/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,193 @@
...
+    )
+    bsn = forms.CharField(max_length=9, label="BSN")
+
+
+class FetchCasesCheckParams(TypedDict):
+    api_group: ZGWApiGroupConfig | None
+    bsn: str
+
+
+class FetchCasesCheck(
```
swrichards, line 28: Do we need to register this with the registry?

src/open_inwoner/openzaak/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,193 @@
...
+                verbose_name=self.label,
+                message=_("No API group selected"),
+                extra={
+                    "hint": _("Select an API group or run this from an object page")
+                },
+            )
+
+        try:
+            client = obj.zaken_client
+            zaken = client.fetch_zaken(user_bsn=bsn)
```
swrichards: We just merged some changes in the signature for this function, you'll need this:

```
diff --git a/src/open_inwoner/openzaak/config_checks/fetch_cases.py b/src/open_inwoner/openzaak/config_checks/fetch_cases.py
index f2610fb59..28517db3b 100644
--- a/src/open_inwoner/openzaak/config_checks/fetch_cases.py
+++ b/src/open_inwoner/openzaak/config_checks/fetch_cases.py
@@ -9,7 +9,7 @@ from maykin_config_checks import GenericConfigCheckResult
 from maykin_config_checks.permissions import IsSuperUser
 from maykin_config_checks.protocols import InteractiveConfigCheck
 from open_inwoner.openzaak.models import OpenZaakConfig, ZGWApiGroupConfig
-
+from open_inwoner.accounts.user_identification import BSNIdentification

 class FetchCasesForm(forms.Form):
     api_group = forms.ModelChoiceField(
@@ -87,7 +87,7 @@ class FetchCasesCheck(

         try:
             client = obj.zaken_client
-            zaken = client.fetch_zaken(user_bsn=bsn)
+            zaken = client.fetch_zaken(user_identification=BSNIdentification(bsn=bsn))
             count = len(zaken)

```

src/maykin_config_checks/templates/admin/run_config_check.html
```diff
@@ -0,0 +1,35 @@
...
+      {% if result.success %}
+        ✅ {{ result.message }}
+      {% else %}
+        ❌ {{ result.message }}
+      {% endif %}
+    </p>
+
+    {% if result.extra %}
+      <pre>{{ result.extra }}</pre>
+    {% endif %}
```
swrichards, line 32:
```
  <pre>{{ result.extra|pprint }}</pre>
```

src/open_inwoner/openzaak/tests/test_fetch_cases.py
swrichards, line 1: Should be in `src/open_inwoner/openzaak/tests`

src/maykin_config_checks/tests/test_fetch_cases.py
```diff
@@ -0,0 +1,219 @@
...
+        self.client = Client()
+
+        self.superuser = User.objects.create_superuser(
+            email="admin@test.com",
+            password="pass",
+        )
+        self.user = User.objects.create_user(
+            email="user@test.com",
+            password="pass",
+        )
```
swrichards: You can use `UserFactory` here.

src/maykin_config_checks/views.py
```diff
@@ -0,0 +1,82 @@
+from django.apps import apps
+from django.http import Http404, HttpResponseForbidden
+from django.shortcuts import get_object_or_404, render
+from django.urls import NoReverseMatch, reverse
+
+from maykin_config_checks.registry import registry
+
+
+def run_config_check(
```
swrichards, line 14: Let's add some type hints here.

src/maykin_config_checks/protocols.py
```diff
@@ -0,0 +1,40 @@
...
+    def get_form_kwargs(
+        cls,
+        instance: Optional[TModel] = None,
+    ) -> dict[str, Any]:
+        ...
+
+    def run(
+        self,
+        data: TCleanedData,
+        instance: Optional[TModel] = None,
```
swrichards, line 39: I'd like to add a `request: Optional[HttpRequest] | None = None` to the signature.

The reason is explained below in the comment about adding a `system_action`.

src/open_inwoner/openzaak/config_checks/fetch_cases.py
```diff
@@ -0,0 +1,193 @@
...
+        instance: Optional[ZGWApiGroupConfig],
+    ) -> Optional[ZGWApiGroupConfig]:
+        return data.get("api_group") or instance
+
+    def run(
+        self,
+        data: FetchCasesCheckParams,
+        instance: Optional[ZGWApiGroupConfig] = None,
+    ) -> GenericConfigCheckResult:
+        bsn = data["bsn"]
```
swrichards, line 60: As noted above, I think we should inject the `request` here too as an optional var, so we can add some logging. Initially we had thought about using signals for this, but I think Sergei is right that signals should generally be avoided.

However, we can just add it to `run`:

```
from open_inwoner.utils.logentry import system_action, user_action

#...
def run(...):
       log_message = f"fetch cases check run for bsn {cleaned_data["bsn"][:3]}******"

       if request:
            user_action(request, user, log_message )
       else:
           system_action(log_message)
```

Given that these _are_ sensitive checks, it's important to have some visibility around it.

swrichards (changes requested):
Almost there, one more iteration I think and then hopefully we can move towards merging soon.

src/maykin_config_checks/tests/test_fetch_cases.py
```diff
@@ -0,0 +1,219 @@
...
+            "run_config_check_standalone",
+            args=["fetch_cases"],
+        )
+
+        response = self.client.post(url, {"bsn": "123456789"})
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "No API group selected")
+
+class ConfigCheckRegistryTests(TestCase):
```
swrichards: Belongs in `maykin_config_checks/tests`

swrichards (approved):

PR #2446, opened 2026-04-22
PP-400: Fix corrupted ProseMirror contents

### LINKED ISSUE #2445: PP-400: Prosemirror contents corrupted
Creating/publishing CMS pages leads to JSON errors due to corrupted ProseMirror contents

pi-sigma:
## Summary

The CMS page creation wizard used a plain Textarea widget for its content field (since `djangocms_text_ckeditor` was removed). When a user typed into the wizard, the raw string was passed directly to `add_plugin(body=<text>`), bypassing `ProsemirrorFormField` entirely. The string was stored as a JSON string in DB and read back as a Python `str`, causing `json.loads` to crash whenever the plugin content was rendered.

  Fixes:

  - ~`get_product_rendered_content` guard: added the same try/except that get_rendered_content already had, so malformed body data doesn't produce a `500` when rendering product pages.~
  - Data migration: converts existing Text plugins whose body is a plain string: empty strings become null, legacy HTML/text is converted to Prosemirror JSON via `html_to_doc`, failures fall back to null. Cleans up data already corrupted
  before the wizard fix.
  - Disable CMS wizard body field: sets `CMS_PAGE_WIZARD_CONTENT_PLUGIN_BODY = ""` so the wizard no longer pre-populates the Text plugin body. Prevents new corruption; users add content through the Prosemirror editor after page creation.

Note: the changes to `get_product_rendered_content` are no longer needed; they are fixed in the library.

## Issue References

Fixes #2445 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/cms/plugins/migrations/0015_fix_legacy_text_plugin_body.py
+++ b/src/open_inwoner/cms/plugins/migrations/0015_fix_legacy_text_plugin_body.py
@@ -1,0 +1,130 @@
+import json
+
+from django.db import migrations
+
+import structlog
+from django_prosemirror.config import ProsemirrorConfig
+from django_prosemirror.constants import EMPTY_DOC
+from django_prosemirror.schema import MarkType, NodeType
+from django_prosemirror.serde import html_to_doc
+
+logger = structlog.stdlib.get_logger(__name__)
+
+
+def fix_legacy_text_plugin_body(apps, schema_editor):
+    """
+    Fix Text plugin body fields that contain legacy or invalid data.
+
+    After replacing djangocms_text_ckeditor with the Prosemirror-based TextPlugin,
+    the body column may contain JSON strings (not dicts) from the old plugin:
+
+    - Empty string JSON values ("") -> set to EMPTY_DOC (body column is NOT NULL)
+    - Non-empty legacy HTML strings -> convert to Prosemirror JSON via html_to_doc;
+      fall back to EMPTY_DOC if conversion fails
+
+    Raw SQL is required because the ProsemirrorModelField descriptor raises
+    ValidationError when Django tries to instantiate models with string body values,
+    making it impossible to load the corrupt rows via the ORM.
+    """
+
+    # Mirror the schema from 0008_add_text_plugin
+    config = ProsemirrorConfig(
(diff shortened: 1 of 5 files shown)


src/open_inwoner/cms/plugins/migrations/0015_fix_legacy_text_plugin_body.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if raw_value is None or isinstance(raw_value, dict):
+            continue
+
+        if raw_value == "":
+            logger.info(
+                "Fixing empty string in Text.body",
+                plugin_id=text.pk,
+            )
+            text.body = None
```
swrichards: @pi-sigma I added an extra guard to the library, precisely to prevent something like the underlying problem again manifesting in the future. Unfortunately, this means we'll have to do some tweaking here. I think it should work if we do this:


```suggestion
            text.body._raw_data = None
```

src/open_inwoner/cms/plugins/migrations/0015_fix_legacy_text_plugin_body.py
```diff
@@ -0,0 +1,113 @@
...
+    - Non-empty legacy HTML strings -> convert to Prosemirror JSON via html_to_doc;
+      fall back to None if conversion fails
+
+    ProsemirrorModelField is a JSONField. Django's from_db_value returns the raw
+    string when the stored value is a JSON string (e.g. "asdfasdfsdf" or ""), which
+    then causes json.loads to fail in doc_to_html.
+    """
+    from django_prosemirror.config import ProsemirrorConfig
+    from django_prosemirror.schema import MarkType, NodeType
+    from django_prosemirror.serde import html_to_doc
```
swrichards: I don't think these have to be in-function?

swrichards (approved):

src/open_inwoner/cms/plugins/migrations/0015_fix_legacy_text_plugin_body.py
pi-sigma: Nope, can be imported at module level.

swrichards (approved):

PR #2457, opened 2026-04-29
[#2454] Copy ZGW object configs for duplicates with the same omschrijving

### LINKED ISSUE #2454: Copy over ZGW object configs for newly created entities that match existing entries on natural keys
## Summary

Updating the status types for a case type (label, text, send notification flag) creates new status types, resulting in duplicate case statuses that must be reconfigured each time. This clutters the list and increases maintenance.

Duplication of existing cases should be avoided, or at least the configuration of existing cases should be persisted.

## Issue references 

This is similar in spirit to a [Dimpact issue PP-378](https://dimpact.atlassian.net/browse/PP-378), though that is mostly to underline that in future we will want a more user-friendly solution here.

pi-sigma:
## Summary

The configuration for existing ZGW objects (status type, resultaat type) is copied to newly created (by OpenZaak) entries with the same 'omschrijving', which are duplicates from a user-perspective.

## Issue References

Closes #2454 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_zgw_imports.py
+++ b/src/open_inwoner/openzaak/tests/test_zgw_imports.py
@@ -1252,6 +1252,98 @@ class ZGWImportTest(ClearCachesMixin, TestCase):
         self.assertIn("Database lock timeout", excluded.error_message)
         self.assertEqual(excluded.extra_context["zaaktype_identificatie"], "ZAAK-001")
 
+    def test_import_informatieobjecttype_configs_copies_config_from_duplicate(self, m):
+        """
+        When OpenZaak creates a new informatieobjecttype with the same omschrijving,
+        OIP config fields are copied from the existing entry to the new one.
+        """
+        root = self.roots[0]
+        api_group = self.api_groups[0]
+
+        catalogus_url = f"{root}catalogussen/1234-5678"
+        catalogus = CatalogusConfigFactory.create(
+            url=catalogus_url,
+            domein="TEST",
+            rsin="12345",
+            service=api_group.ztc_service,
+        )
+        zaaktype_config = ZaakTypeConfig.objects.create(
+            identificatie="ZAAK-001",
+            omschrijving="Test Zaaktype",
+            catalogus=catalogus,
+            urls=[f"{root}zaaktypen/zt-001"],
+        )
+
+        # Existing config with configured OIP fields and the old informatieobjecttype URL
+        old_iot_url = f"{root}informatieobjecttypen/iot-old"
+        ZaakTypeInformatieObjectTypeConfig.objects.create(
+            zaaktype_config=zaaktype_config,
+            informatieobjecttype_url=old_iot_url,
+            omschrijving="Besluit",
+            zaaktype_uuids=[],
+            document_upload_enabled=True,
+            document_notification_enabled=True,
+        )
+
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/tests/test_zgw_imports.py
```diff
@@ -1736,6 +1736,113 @@ def save_with_error(instance, *args, **kwargs):
...
+            omschrijving="In behandeling",
+            statustekst="Uw zaak wordt behandeld",
+            zaaktype_uuids=[],
+            status_indicator="warning",
+            status_indicator_text="Let op",
+            notify_status_change=False,
+            action_required=True,
+            document_upload_enabled=False,
+            description=None,
+            document_upload_description=None,
```
swrichards: Can you set this to a known value? It's a prosemirror field, and I want to make sure that copying prosemirror fields from one model to the next works, just in case there's more corner cases we've missed in the prosemirror library.

src/open_inwoner/openzaak/tests/test_zgw_imports.py
```diff
@@ -1736,6 +1736,113 @@ def save_with_error(instance, *args, **kwargs):
...
+        # OIP config fields copied from the existing entry
+        self.assertEqual(new_config.status_indicator, "warning")
+        self.assertEqual(new_config.status_indicator_text, "Let op")
+        self.assertFalse(new_config.notify_status_change)
+        self.assertTrue(new_config.action_required)
+        self.assertFalse(new_config.document_upload_enabled)
+        self.assertEqual(new_config.call_to_action_url, "https://example.com")
+        self.assertEqual(new_config.call_to_action_text, "Actie vereist")
+        self.assertEqual(new_config.case_link_text, "Bekijk zaak")
+
```
swrichards, line 1959: Further to comment above, also assert against the (html) content of `document_upload_description`

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -1052,6 +1063,29 @@ def import_statustype_configs_for_zaaktype(
                         zaaktype_uuids=[zt.uuid for zt in using_zaak_types],
                         found_in_api=True,
                     )
+                    if config_source is not None:
```
swrichards, line 1090: Could walrus operator this with `l.1055` to keep it tightly together.

src/open_inwoner/openzaak/zgw_imports.py
```diff
@@ -952,10 +952,16 @@ def import_statustype_configs_for_zaaktype(
             for url in zaak_type.statustypen:
                 status_queue[url].append(zaak_type)
```
swrichards, line 977: I am aware it was not in the original ticket, but could we apply the same pattern to the informatie object types, for the fields `document_upload_enabled` and `document_notification_enabled`?

swrichards (changes requested):

src/open_inwoner/openzaak/zgw_imports.py
pi-sigma: I moved the related lines together but decided against walrus. I find it less readable when it spreads across multiple lines including parentheses etc (case in point, I made a small mistake here and had to think about why the test suddenly fails).

swrichards (approved):
👍 just needs a rebase

PR #2459, opened 2026-04-30
[#2450] Amend ProseMirror migrations

### LINKED ISSUE #2450: Amend legacy prosemirror migrations to avoid direct assignment
The latest prosemirror version has becomes stricter, and now raises if you try to assign non-JSON content directly to the field. We will have to adapt some of our recent PM migrations to adapt so they continue to work (not every instance has run those migrations yet).

See the pattern in https://github.com/maykinmedia/open-inwoner/pull/2446 where we write/read to `.raw_data` for integrity checks.

pi-sigma:
## Summary

Amend ProseMirror migrations to avoid direct assignment of non-JSON content

## Issue References

Closes #2450

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
+++ b/src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
@@ -4,8 +4,30 @@ from django.db import migrations
 from django.utils.text import slugify
 
 from cms.api import add_plugin, create_page
+from django_prosemirror.config import ProsemirrorConfig
+from django_prosemirror.schema import MarkType, NodeType
+from django_prosemirror.serde import html_to_doc
 
 from open_inwoner.cms.footer.cms_plugins import CMSFlatPagePlugin
+
+
+def _html_to_pm_doc(html):
+    if not html or not html.strip():
+        return None
+
+    config = ProsemirrorConfig(
+        allowed_node_types=[NodeType.HARD_BREAK, NodeType.PARAGRAPH, NodeType.HEADING],
+        allowed_mark_types=[
+            MarkType.STRONG,
+            MarkType.ITALIC,
+            MarkType.UNDERLINE,
+            MarkType.LINK,
+        ],
+    )
+    try:
+        return html_to_doc(html, schema=config.schema)
+    except Exception:
+        return None
 
 
 def create_cms_pages(apps, schema_editor):
@@ -35,7 +57,7 @@ def create_cms_pages(apps, schema_editor):
             plugin_type=CMSFlatPagePlugin,
             language=language,
             title=flatpage.title,
-            content=flatpage.content,
+            content=_html_to_pm_doc(flatpage.content),
         )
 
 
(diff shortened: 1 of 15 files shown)


swrichards (commented):
@pi-sigma this is good, but I was a bit concerned that DPM didn't properly handle these cases in the library and that the code for these fixes is a bit fragmented. On Monday I'd like to look at this https://github.com/maykinmedia/django-prosemirror/pull/48 with you (you can have a look at the README, there's a section for migration fixes) and see if we can test those changes in this PR, so we've got a solid basis for any future situations this might occur.

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
```diff
@@ -4,10 +4,32 @@
...
+            MarkType.STRONG,
+            MarkType.ITALIC,
+            MarkType.UNDERLINE,
+            MarkType.LINK,
+        ],
+    )
+    try:
+        return html_to_doc(html, schema=config.schema)
+    except Exception:
+        return None
```
swrichards, line 30: Double checking: we're sure we're only using this on nullable fields?

src/open_inwoner/utils/migration_operations.py
```diff
@@ -163,87 +165,69 @@ def migrate_to_prosemirror_field(
...
 
-    # Get model
     Model = apps.get_model(app_label, model_name)
+    tmp_col = f"{field_name}_tmp"
 
-    # Get instances to process
+    qs = Model.objects.values("pk", tmp_col)
     if use_singleton:
-        instances = [Model.objects.first()] if Model.objects.exists() else []
-    else:
```
swrichards, line 178: This could possibly raise (even though Singletons probably already exist, can't hurt to be defensive):

```suggestion
    qs = Model.objects.values("pk", tmp_col)
if use_singleton and qs.exists():
        qs = qs[:1]
```

src/open_inwoner/openzaak/migrations/0078_fix_empty_string_prosemirror_fields.py
```diff
@@ -1,59 +1,62 @@
...
-                    logger.info(
-                        "Fixing empty string literal in ProseMirror field",
-                        model=model.__name__,
-                        field=field_name,
-                        config_id=row["pk"],
-                        omschrijving=row["omschrijving"],
+        for pk_value, omschrijving, raw_text in rows:
+            if raw_text is None:
+                continue
+            if json.loads(raw_text) == "":
```
swrichards: Can this raise?

swrichards (changes requested):
Nice improvements, just some sanity checks to look at it, might all be irrelevant.

src/open_inwoner/cms/footer/migrations/0002_migrate_flatpages_content_to_cms.py
pi-sigma: Yes, content is `null=True, blank=True` so returning `None` is safe here

src/open_inwoner/openzaak/migrations/0078_fix_empty_string_prosemirror_fields.py
pi-sigma: No, `raw_text` is the result of `JSONB::text` cast (an instance of `{field_name}::text` above) so this would be safe. However, the migration still uses raw SQL. This was on oversight and should be re-written.

PR #2466, opened 2026-05-04
[#2142] Improve error handling for API clients/services

### LINKED ISSUE #2142: Error handling in get_json_response() API utility
## Summary

The `get_json_response()` function in `src/open_inwoner/utils/api.py` has overly broad exception handling that makes error debugging difficult and can silently lose error information.

### Overly Broad Exception Catching

```python
try:
    response_json = response.json()
except Exception:  # Catches everything including SystemExit, KeyboardInterrupt, etc.
    response_json = None
```

This catches all exceptions (not just JSON-related errors), which can mask unexpected failures.

### Silent Loss of Error Information

When JSON parsing fails, `response_json = None` is set silently. If the HTTP status is `4xx`, this raises `ClientError(None)`, losing the actual error response from the API that might contain valuable debugging information.

### Confusing Exception Pattern

The recommended usage is except (RequestException, ClientError), but:

- RequestException is the base for all requests errors (including HTTPError)
- ClientError only handles `4xx` errors (raised by `get_json_response`)
- `5xx` errors are re-raised as HTTPError (which IS a `RequestException`)
- This creates confusion about which errors are caught where

### Inconsistent Server Error Handling

```python
if response.status_code >= 500:
    raise  # Re-raises as HTTPError
raise ClientError(response_json) from exc  # 4xx becomes ClientError
```

Server errors (`5xx`) and client errors (`4xx`) use different exception types, making consistent handling harder.

## Impact

- Used in 51 locations across the codebase
- Makes debugging API errors difficult
- Could mask unexpected failures
- Inconsistent error handling across different API clients

## Proposal

1. Create specific exception types: `ClientError`, `ServerError`, `InvalidJSONError`
2. Only catch JSON-specific exceptions (`ValueError`, `JSONDecodeError`)
3. Provide consistent error hierarchy under a single `APIError` base class
4. Update exception handling to use except `APIError` instead of `except (RequestException, ClientError)` (debatable: perhaps a more granular approach is preferable; the API clients would `except ServerError`, `except ClientError` etc. separately, providing flexibility for logging and defining user-facing error messages).

pi-sigma:
## Summary

An error hierarchy for interacting with API's is introduced, with `APIError` as base and `ServerError`, `ClientError` etc. inheriting from it. Instead of suppressing errors and returning a sentinel value (empty list or `None`, the current pattern) clients and services now raise our own typed exceptions, which need to be handed by the call-sites (e.g. the views).

At the client layer: clients no longer hide failures -- they raise typed exceptions
At the service layer: intentional catches remain, not suppression. The three new except blocks in services are all at the business-logic level:

  1. `search_zaken` -- continue: if a zaak's type can't be resolved, skip that result rather than crashing the whole search. This was already the intent; the refactor just makes it visible and logged instead of relying on None propagation.
  2. `get_documents` -- continue: if one document's metadata can't be fetched, skip it and show the rest. Same pattern.
  3. `get_zaak_for_klantcontactmoment` -- `ocm = None:` if the object-contactmoment link can't be retrieved, the contact moment detail still loads without the zaak association. The failure is logged and the caller gets (question, None) instead
  of crashing.

## Issue References

Closes #2142 

## Checklist

- [x] CHANGELOG.rst updated


src/open_inwoner/openklant/clients.py
swrichards: One upshot of this is that this should never be `None` I think?

src/open_inwoner/utils/api.py
```diff
@@ -8,10 +8,43 @@
 Object = dict[str, Any]
 
 
-class ClientError(Exception):
+class APIError(Exception):
```
swrichards, line 13: I think the hierarchy is correct, though I'd still say we should duplicate these per client for scope and possible divergence.

src/open_inwoner/utils/api.py
```diff
@@ -8,10 +8,43 @@
...
+
+class NetworkError(APIError):
+    pass
+
+
+class InvalidJSONError(APIError):
+    pass
+
+
+def raise_for_status(response: requests.Response) -> None:
```
swrichards: Possible we can use request's own `raise_for_status` and handle _their_ exceptions and remap them to ours.

src/open_inwoner/openklant/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+    except RequestException as exc:
+        raise KlantAPINetworkError(str(exc)) from exc
+    except ServerError as exc:
+        raise KlantAPIServerError(str(exc)) from exc
+    except ClientError as exc:
+        raise KlantAPIClientError(str(exc)) from exc
+    except InvalidJSONError as exc:
+        raise KlantAPIInvalidJSONError(str(exc)) from exc
+    except (TypeError, ValueError, KeyError) as exc:
+        raise KlantAPIDataError(str(exc)) from exc
```
swrichards: This is a nice pattern!

src/open_inwoner/openklant/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+class KlantAPIInvalidJSONError(KlantAPIError, InvalidJSONError):
+    pass
+
+
+class KlantAPIDataError(KlantAPIError, DataError):
+    pass
+
+
+@contextmanager
+def klant_api_errors():
```
swrichards:
```suggestion
def remap_klant_api_exceptions():
```

src/open_inwoner/openklant/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+        yield
+    except RequestException as exc:
+        raise KlantAPINetworkError(str(exc)) from exc
+    except ServerError as exc:
+        raise KlantAPIServerError(str(exc)) from exc
+    except ClientError as exc:
+        raise KlantAPIClientError(str(exc)) from exc
+    except InvalidJSONError as exc:
+        raise KlantAPIInvalidJSONError(str(exc)) from exc
+    except DataError as exc:
```
swrichards: `DataError` also needs an in-module definition

src/open_inwoner/openklant/exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+from requests.exceptions import RequestException
+
+from open_inwoner.utils.api import (
+    APIError,
+    ClientError,
+    DataError,
+    InvalidJSONError,
+    NetworkError,
+    ServerError,
+)
```
swrichards: I'm leaning against doing this: I am not sure there is a use-case for "Ignore all exceptions of type X for clients XYZ". Where you explicitly want this, I think having to import and list all those (root) exceptions is less likely to lead to surprises, rather than suppressing something you might care about (the naming also makes this confusion, you could think you are narrow-targeting but in fact you're catching the broad base version of an exception).

src/open_inwoner/openklant/clients.py
```diff
@@ -46,20 +52,15 @@ def retrieve_klant(
...
-        except (RequestException, ClientError):
-            logger.exception("Error fetching Klant for DigiD user")
-            return []
-
-        klanten = factory(Klant, all_data)
-
-        return klanten
+            raise_for_status(response)
+            raw_json_data = parse_json(response)
+            paginated_data = list(pagination_helper(self, raw_json_data))
```
swrichards: One thing to note is that `pagination_helper` does some potential raising of its own, though I think they're mostly the same exceptions we're already catching.

This did remind me that we're not exposing `max_requests` at all. Not an issue here, but for longer endpoints (e.g. with questions) we might want extra handling from the pagination helper, e.g. if there were _more_ pages and we didn't fetch them, we would _also_ want to show this to the user somehow. This is not a theoretical concerns: for Rotterdam with the case list, this is already happening (some businesses have upwards of 1k cases).

Still, I think the issue there might be something along the lines of https://github.com/maykinmedia/open-inwoner/pull/2313 (returning a response object with metadata, and a flag if we had to truncate). Let's discuss Wednesday, I think this should be part of the broader solution set.

swrichards: To be clear, don't do anything right now, but it also triggered me that we might want to move all these kinds of settings to the Openzaak/API group model config, away from the env vars, so users can play with the optimal settings for their use-case.

swrichards (changes requested):

src/open_inwoner/openklant/exceptions.py
pi-sigma: It already has

src/open_inwoner/openklant/exceptions.py
pi-sigma: Sounds right, this is over-engineered

src/open_inwoner/utils/api.py
pi-sigma: Our `raise_for_status` is a wrapper around request's `response.raise_for_status`. Did you have something else in mind?

src/open_inwoner/openklant/exceptions.py
swrichards: I think I wasn't clear enough: but my proposal would be to _entirely_ remove the exceptions in `open_inwoner.utils.api` -- I don't think they have a good use-case, and are more likely to lead to symbol confusion. Shall we remove them until we have a better use cases? @pi-sigma

swrichards (commented):
Before reviewing further, take a look at the comment in https://github.com/maykinmedia/open-inwoner/pull/2466#discussion_r3225520087

src/open_inwoner/openklant/exceptions.py
pi-sigma: They have three purposes:
- `raise_for_status` encapsulates the 4xx/5xx split. Without `ClientError`/`ServerError`, the remap context managers would need to catch `HTTPError` and branch on status code. This would be duplicated for the ZGW errors (and for any other API module)
- `DataError`: without it, the ctx managers would need `except (TypeError, ValueError, KeyError)`, which would catch programming bugs in the calling code that uses the context manager. Unlikely, since the complexity of the calling code should be very low, but still...
- `APIError` as a cross-module base. Code that doesn't care whether the error is from Klant or ZGW can catch `APIError`.

src/open_inwoner/utils/api.py
```diff
@@ -4,14 +4,59 @@
...
+    try:
+        response.raise_for_status()
+    except requests.HTTPError as exc:
+        if response.status_code >= 500:
+            raise ServerError(str(exc)) from exc
+        raise ClientError(str(exc)) from exc
```
swrichards:
```suggestion
def raise_for_status(response: requests.Response, server_error_exc: type[Exception], client_error_exc: type[Exception]) -> None:
    try:
        response.raise_for_status()
    except requests.HTTPError as exc:
        if response.status_code >= 500:
            raise server_error_exc(str(exc)) from exc
        raise client_error_exc(str(exc)) from exc
```

swrichards: Use `functools.partial` in the concrete module to pre-fill the excception classes.

src/open_inwoner/utils/api.py
```diff
@@ -4,14 +4,59 @@
...
+
+
+def factory(model, data):
```
swrichards: Also inject `data_error_exc` here, use `functools.partial` within the module.

swrichards (commented):
Discussed, make utils generic by injecting exc dependencies, remove generic exception hierarchy in `utils`

src/open_inwoner/openzaak/notifications.py
```diff
@@ -80,7 +82,11 @@ def handle_zaken_notification(notification: Notification):
     service = ZGWService(use_cache=False)
 
     # check if we have users that need to be informed about this case
-    if not (roles := service.fetch_zaak_roles(zaak_url, api_group)):
+    try:
+        roles = service.fetch_zaak_roles(zaak_url, api_group)
+    except (ZGWAPIError, RequestException):
+        roles = []
+    if not roles:
```
swrichards: We can just return here.

```suggestion
```

src/open_inwoner/utils/api.py
```diff
@@ -4,31 +4,43 @@
 
 import requests
 from pydantic_core import Url
+from zgw_consumers.api_models.base import factory as _factory
 
 Object = dict[str, Any]
 
 
-class ClientError(Exception):
+class APIError(Exception):
```
swrichards, line 13: Let's call it `GenericAPIError` to distinguish it from others.

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -140,27 +142,36 @@ def get_context_data(self, **kwargs):
...
+            if not service:
+                continue
+
+            try:
+                service_questions = service.list_questions(
+                    self.get_fetch_params(service),
+                    user=self.request.user,
+                )
+                questions.extend(service_questions)
+            except (KlantAPIError, requests.RequestException):
```
swrichards: About this pattern in general: the only real behavioral changes between the different except blocks is the error message.. I wonder whether a catch-all wouldn't be fine here. The only real difference in behavior might be the _user facing_ error message, to distinguish between "try again later" versus "It's broken, we're sorry". But otherwise fewer exception blocks might be more readable.

src/open_inwoner/accounts/views/profile.py
```diff
@@ -363,15 +365,19 @@ def update_klant_via_esuite(self, user_form_data: dict, user: User) -> bool:
             return False
 
         if fetch_params := service.get_fetch_parameters(user):
```
swrichards, line 367: `get_fetch_parameters` is still a bit of an eye-sore, would be nice to use `UserIdentification` there. For another PR.

src/open_inwoner/accounts/views/registration.py
```diff
@@ -266,7 +274,10 @@ def update_klant_via_esuite(self, form: NecessaryUserForm):
         if "case_notification_channel" in form.cleaned_data:
             update_fields.append("toestemmingZaakNotificatiesAlleenDigitaal")
 
-        service.update_klant_from_user(klant, user, update_fields=update_fields)
+        try:
+            service.update_klant_from_user(klant, user, update_fields=update_fields)
+        except (KlantAPIError, requests.RequestException):
+            logger.error("Error updating klant during post-registration sync")
```
swrichards: Could arguably all be in the same try/except, happy to consolidate the error message (The exception traceback should tell us who exactly raised).

src/open_inwoner/templates/pages/cases/service_unavailable.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n link_tags %}
+
+<h1 class="utrecht-heading-1">{% trans "De service is tijdelijk niet beschikbaar" %}</h1>
+<p class="utrecht-paragraph">{% trans "Er is een fout opgetreden bij het ophalen van uw zaakgegevens. Probeer het later opnieuw." %}</p>
+<p class="utrecht-paragraph">{% link href="pages-root" text=_("Ga naar de home pagina") %}</p>
```
swrichards, line 5: @stefrado Need your 👀 on this -- `nl-paragraph`? Is there a good reason not to use that in error pages?

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -140,13 +142,17 @@ def dispatch(self, request, *args, **kwargs):
...
-                # the refactoring of error-handling is complete (GH #2142)
-                pass
+            except ZGWAPIClientError as exc:
+                raise Http404 from exc
+            except (ZGWAPIServerError, RequestException):
+                return TemplateResponse(
+                    request,
+                    "pages/cases/service_unavailable.html",
+                    status=503,
+                )
```
swrichards: This is useful, and we'll have this case for many pages. Maybe wrap it in a `ServiceUnavailableResponse`? Even nicer might be to have a middleware that returns this response when a `raise ServiceUnavailable from exc` is done, that would keep the call sites cleaner (like the `raise Http404 from exc`

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -140,13 +142,17 @@ def dispatch(self, request, *args, **kwargs):
...
             else:
                 if result is None:
-                    return self.handle_no_permission()
+                    raise Http404
```
swrichards, line 165: Let's add a logging call here at level warning (in principle people should hit this rarely, want it to stand out in logs, but not to spam Sentry).

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -647,16 +646,23 @@ def get(self, request, *args, **kwargs):
...
+        except (ZGWAPIError, RequestException) as exc:
+            raise Http404 from exc
 
         # check if this info_object belongs to this zaak
-        if not api_group.zaken_client.fetch_zaak_information_objects_for_zaak_and_info(
-            self.zaak.url, info_object.url
-        ):
+        try:
+            zaak_info_objects = (
+                api_group.zaken_client.fetch_zaak_information_objects_for_zaak_and_info(
```
swrichards: I note these are still direct calls to the clients -- we should really consolidate this in the servers.

src/open_inwoner/laposta/client.py
```diff
@@ -29,17 +46,15 @@ class LapostaClient(APIClient):
     def get_lists(self) -> list[LapostaList]:
         try:
             response = self.get("list")
-            data = get_json_response(response)
-        except (RequestException, ClientError):
+            raise_for_status(response)
+            data = parse_json(response)
+        except (LapostaAPIError, RequestException):
             logger.exception("error while fetching newsletters with Laposta")
             return []
```
swrichards: Guess we're stil suppressing here: any reason not to change this at the call-site? I have no idea how this is used, so there's an argument for keeping existing behavior intact.

src/open_inwoner/laposta/client.py
```diff
@@ -29,17 +46,15 @@ class LapostaClient(APIClient):
...
 
-        if not data:
+        if not isinstance(data, dict):
             return []
```
swrichards: Same as above: feels like an odd guard, it's tantamount to "This server has broken the contract". But if the goal here is to keep it.

src/open_inwoner/openklant/views/contactform.py
```diff
@@ -243,14 +245,18 @@ def _register_via_esuite(
         if klant:
             self._update_klant_with_form_data(klant, form.cleaned_data)
 
-        contactmoment = self._create_contactmoment(
-            form.cleaned_data, esuite_config, klant
-        )
+        try:
```
swrichards, line 248: This view is still very complicated due to error handling and all the possible ways questions can be persisted. I feell it wouldn't be a bad idea to make the behavior configurable (e.g. if API registration is enabled, should we always report an error if downstream fails?). Not for now, but might be for our backlog and to pitch to Dimpact.

swrichards: https://github.com/maykinmedia/open-inwoner/issues/2516

src/open_inwoner/openklant/services.py
```diff
@@ -215,23 +218,14 @@ def retrieve_klant(
...
-        klanten = factory(Klant, all_data)
-
-        return klanten
+        response = self.client.get(
+            "klanten",
+            params={"subjectNatuurlijkPersoon__inpBsn": user_bsn},
+        )
+        raise_for_status(response)
+        data = parse_json(response)
+        all_data = list(pagination_helper(self.client, data))
```
swrichards, line 221: One more thought: do we need to wrap `pagination_helper` as well to ensure we get the same exception mapping, given that it also fires additional requests?

src/open_inwoner/openklant/services.py
```diff
@@ -472,56 +433,36 @@ def __init__(self, config: ESuiteKlantConfig | None = None):
...
-            return
-
-        contactmoment = factory(ContactMoment, data)
+    ) -> ContactMoment:
+        response = self.client.post("contactmomenten", json=data)
+        raise_for_status(response)
+        raw_data = parse_json(response)
+        contactmoment = factory(ContactMoment, raw_data)
 
         if klant:
```
swrichards, line 450: In a perfect world, we would trigger a Celery task to eventually delete the created contactmoment in case of a broken transaction (contactmoment works, klantcontactmoment failed). Given that eSuite is on the out, not a priority.

src/open_inwoner/openzaak/services.py
swrichards, line 490: I think we might be missing some kind of `wait` call here.  The logic here is that we want either all futures to complete, or timeout, so I think we would need something like

````
concurrent.futures.wait(zaak.futures.values(), timeout=self._timeouts[...)
````

Then the check below will discard futures that are not done by checking `f.done()`.

Happy to do this in the followup where we do pagination, as this logic is likely to change a bit there.

**Review by swrichards — CHANGES_REQUESTED** [review]
Great to finally see us shaking off this dodgy pattern.

Apart from random issues, I think the bigger picture issues are:

- Should we wrap `requests.RequestException` in the exception hierarchy?
- We need to look into `pagination_helper` -- might need a custom utility so it has the same exception raising properties as the main calls

Let's discuss before the next iteration.

File: src/open_inwoner/openzaak/services.py
Comment (swrichards) [reply src/open_inwoner/openzaak/services.py:490]: Happy to consolidate the three resolvers into a single callable.

File: src/open_inwoner/laposta/client.py (hunk 19 above)
Comment (pi-sigma) [reply src/open_inwoner/laposta/client.py]: Moved exception handling to the form

File: src/open_inwoner/laposta/client.py (hunk 20 above)
Comment (pi-sigma) [reply src/open_inwoner/laposta/client.py]: Not a lot of trust in the API here. I'm going to keep the guard since I have no experience with the API either, but raise `LapostaAPIError` instead of swallowing

File: src/open_inwoner/openklant/services.py (hunk 22 above)
Comment (pi-sigma) [reply src/open_inwoner/openklant/services.py:221]: `pagination_helper` is covered by the new approach

File: src/open_inwoner/utils/api.py (hunk 11 above)
Comment (pi-sigma) [reply src/open_inwoner/utils/api.py:13]: I don't know about this. The other exceptions inherit from `APIError`, and it is used in type annotations. I find `APIError` more readable. Compare: `Exception` is also not called `GenericException`.

File: src/open_inwoner/accounts/views/profile.py (hunk 13 above)
Comment (pi-sigma) [reply src/open_inwoner/accounts/views/profile.py:367]: Agreed, let's do this in a follow up

File: src/open_inwoner/templates/pages/cases/service_unavailable.html (hunk 15 above)
Comment (pi-sigma) [reply src/open_inwoner/templates/pages/cases/service_unavailable.html:5]: We still use `utrecht-paragraph` everywhere, I didn't see any uses of `nl-paragraph`. Have we migrated yet?

Comment (swrichards) [reply src/open_inwoner/templates/pages/cases/service_unavailable.html:5]: Yes good point, that's all still in the feature branch. We keep this for now, it will be part of that rebase to fix this.

File: src/open_inwoner/utils/api.py [hunk 24]
```diff
@@ -4,31 +4,81 @@
...
-    except Exception:
-        if strict:
-            raise
+class APIClientMixin:
+    """
+    Mixin for APIClient subclasses that maps requests exceptions and HTTP errors
+    to domain-specific exceptions configured on the class.
+    """
+
+    network_error_type: type[APIError] = APIError
```
swrichards: I thhink the `type[APIError]` suggests the exceptions must be of that type or inherit from it. I think we need this:


```suggestion
    network_error_type: type[Exception] = APIError
```

src/open_inwoner/utils/api.py
```diff
@@ -4,31 +4,81 @@
...
+        raise exc_type(str(exc)) from exc
+
+
+def parse_json(
+    response: requests.Response, *, exc_type: type[Exception] = APIError
+) -> dict:
+    try:
+        return response.json()
+    except requests.exceptions.JSONDecodeError as exc:
+        raise exc_type(str(exc)) from exc
```
swrichards: Is there value in having these as separate functions, over inlining them in the mixin?

src/open_inwoner/utils/api.py
```diff
@@ -4,31 +4,81 @@
...
     pass
 
 
-def get_json_response(response: requests.Response, strict=False) -> dict | None:
-    try:
-        response_json = response.json()
-    except Exception:
-        if strict:
-            raise
+class APIClientMixin:
```
swrichards: I wonder if we need a mixin, or whethher we should simply inherit from `ape_pie.APIClient`  and make that the base class for everytthing. I suspect we're always going to use the mixin on classes that iherit from `APIClient`, so maybe it's cleaner just to have an explicit base class: `class CustomAPIClient` (not an imaginative name, but I'd like something explicitly different from the generic `APIClient`. What do you think?

src/open_inwoner/accounts/views/contactmoments.py
swrichards: I still think `RequestException` does not add much here and it represents "something we didn't anticipate', makes more sense to me to let it fall through to the `Exception` cach below.

pi-sigma: My bad, this should have been removed.

swrichards (approved):
Looks good, only nitpick is the extra catching of `requests.RequestException` -- I'd say that's "Miscelaneous error we didn't account for" and can be dealt with by the fallback `except Exception` blocs.

swrichards (approved):

PR #2467, opened 2026-05-04
:bug: [#2465] Reimplement CKEditor configuration

### LINKED ISSUE #2465: PP-431 Opmaak ontbreekt in 'Actie nodig' mail
https://dimpact.atlassian.net/browse/PP-431

stefrado:
## Summary
CKEditor has a feature called Advanced Content Filter (ACF) that is enabled by default. When ACF is active, CKEditor silently strips any HTML elements, attributes, or CSS classes it doesn't recognise as "safe" on every save. Setting `allowedContent: True` disables ACF entirely and tells CKEditor to preserve whatever HTML is present verbatim.

 When CKEditor was added without `CKEDITOR_CONFIGS["mail_editor"]` in commit [69b176f44d0333271fc8abf1f7f825b0b6f69101](https://github.com/maykinmedia/open-inwoner/commit/69b176f44d0333271fc8abf1f7f825b0b6f69101) the widget fell back to CKEditor's default config - ACF enabled, with restrictive allowlist.

Mail templates saved after 2026-03-04, the date of release `2.1.0`, would have lost their styling do to this ACF mechanism.

The PR adds `CKEDITOR_CONFIGS` back. A management command `find_modified_mail_templates` is added to find and report mail templates which might have been corrupted.

## Issue References
Fixes #2465 

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables

### DIFF
--- a/src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
+++ b/src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
@@ -79,6 +79,8 @@ class Command(BaseCommand):
             if template.language:
                 label += f" ({template.language})"
 
+            path = reverse("admin:mail_editor_mailtemplate_change", args=[template.pk])
+
             self.stdout.write(self.style.SUCCESS(label))
             self.stdout.write(f"  Subject : {template.subject}")
             self.stdout.write(
@@ -86,9 +88,10 @@ class Command(BaseCommand):
                 f"by {last_save.user}"
             )
             self.stdout.write(f"  Saves since cutoff : {len(saves)}")
+            self.stdout.write(f"  Admin URL : {path}")
             self.stdout.write("")
 
-            blocks.append((label, template))
+            blocks.append((label, template, path))
 
         email_address = options["email"]
         if not email_address:
@@ -98,8 +101,7 @@ class Command(BaseCommand):
         lines = []
         list_items = []
 
-        for label, template in blocks:
-            path = reverse("admin:mail_editor_mailtemplate_change", args=[template.pk])
+        for label, template, path in blocks:
             url = f"https://{domain}{path}"
             lines.append(f"{label}: {url}")
             list_items.append(f'<li><a href="{url}">{label}</a></li>')
--- a/src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
+++ b/src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
@@ -1,0 +1,123 @@
+from datetime import date
+
+from django.contrib.admin.models import ADDITION, CHANGE, LogEntry
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.sites.shortcuts import get_current_site
+from django.core.mail import EmailMultiAlternatives
+from django.core.management.base import BaseCommand, CommandError
+from django.urls import reverse
+
+from mail_editor.models import MailTemplate
+
+DEFAULT_CUTOFF = date(2026, 3, 4)
+
+
+class Command(BaseCommand):
+    help = (
+        "List mail templates saved on or after a given date (default: 2026-03-04), "
+        "when CKEditor was re-added without CKEDITOR_CONFIGS, which may have stripped "
+        "CSS classes from the body. Pass --email to receive the full template bodies "
+        "at that address."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--since",
+            type=date.fromisoformat,
+            default=DEFAULT_CUTOFF,
+            metavar="YYYY-MM-DD",
+            help=f"Only report templates saved on or after this date (default: {DEFAULT_CUTOFF}).",
+        )
+        parser.add_argument(
+            "--email",
+            help="Send the affected templates to this address.",
+        )
+
+    def handle(self, *args, **options):
+        since = options["since"]
+        ct = ContentType.objects.get_for_model(MailTemplate)
+
+        log_entries = (
+            LogEntry.objects.filter(
+                content_type=ct,
+                action_time__date__gte=since,
+                action_flag__in=[ADDITION, CHANGE],
+            )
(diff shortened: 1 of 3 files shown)


stefrado:
This is the missing configuration for which caused the issue @pi-sigma. Let me know if you need more information.

pi-sigma (approved):

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--since",
+            type=date.fromisoformat,
+            default=DEFAULT_CUTOFF,
+            metavar="YYYY-MM-DD",
+            help=f"Only report templates saved on or after this date (default: {DEFAULT_CUTOFF}).",
+        )
+        parser.add_argument(
```
swrichards, line 31: Can't remember the syntax, but since there's no default, is this optional?

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
```diff
@@ -0,0 +1,124 @@
...
+        if not log_entries.exists():
+            self.stdout.write(f"No mail templates were saved on or after {since}.")
+            return
+
+        affected_pks = [
+            int(pk) for pk in log_entries.values_list("object_id", flat=True).distinct()
+        ]
+        templates = MailTemplate.objects.filter(pk__in=affected_pks).order_by(
+            "template_type", "language"
+        )
```
swrichards, line 60: Nitpick, but if we need an extra fixup, I think you can merge this:


```suggestion
        templates = MailTemplate.objects.filter(pk__in=log_entries.values_list("object_id", flat=True).distinct()).order_by(
            "template_type", "language"
        )
```

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
```diff
@@ -0,0 +1,124 @@
...
+
+        for label, template in blocks:
+            plain_parts.append(
+                f"=== {label} ===\nSubject: {template.subject}\n\n{template.body}\n"
+            )
+            html_parts.append(
+                f"<h2>{label}</h2>"
+                f"<p><strong>Subject:</strong> {template.subject}</p>"
+                f"<hr>{template.body}"
+            )
```
swrichards: I think it would be better just to add a list of URLs to the template (i.e. to the admin detail page)

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
```diff
@@ -0,0 +1,124 @@
...
+                f"  Last saved : {last_save.action_time:%Y-%m-%d %H:%M} "
+                f"by {last_save.user}"
+            )
+            self.stdout.write(f"  Saves since cutoff : {len(saves)}")
+            self.stdout.write("")
+
+            blocks.append((label, template))
+
+        email_address = options["email"]
+        if not email_address:
```
swrichards, line 97: Also display the output.

swrichards (changes requested):

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
pi-sigma: Yes it works without `email` argument

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
pi-sigma: Sorry I was too quick. I had something like this earlier, but we need the `int` cast because `LogEntry.object_id` is a `CharField`, so the subquery passes text values to a Postgres integer column

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
pi-sigma: Should already be displayed with or without email. Do you not see any output in the terminal?

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
swrichards: Fair enough, we could cast it with an annotated field but too much effort for too little pay-off.

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
swrichards: Misread, my bad.

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
```diff
@@ -0,0 +1,123 @@
...
+            return
+
+        domain = get_current_site(None).domain
+        lines = []
+        list_items = []
+
+        for label, template in blocks:
+            path = reverse("admin:mail_editor_mailtemplate_change", args=[template.pk])
```
swrichards: I would also like to include this in the stdout output

swrichards: And I'd be happy to simplify both email and stdout to note show theh whole template, I am not sure it'll make the output much easier to follow. I think label, ID, link, and last save date/user is fine.

(I thought I added this as an earlier comment but apparently not, my bad , apologies :/ )

swrichards (changes requested):

src/open_inwoner/configurations/management/commands/find_modified_mail_templates.py
pi-sigma: I've added the path. I think the whole template wouldn't add much since it wouldn't tell you what part of it was changed.

swrichards (approved):

PR #2482, opened 2026-05-07
ci: [#2481] add zizmor GitHub Actions security analysis

### LINKED ISSUE #2481: Use zizmor for GHA security

swrichards:
Closes: #2481

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,8 @@ on:
 env:
   IMAGE_NAME: maykinmedia/open-inwoner
 
+permissions: {}
+
 jobs:
   #
   # Unit and integration tests
@@ -21,6 +23,8 @@ jobs:
     name: Run the Django test suite
     runs-on:
       ${{ matrix.test-type == 'main' && 'maykin-runner' || 'ubuntu-latest' }}
+    permissions:
+      contents: read
     strategy:
       matrix:
         test-type: [main, elastic, migrations]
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -47,14 +51,16 @@ jobs:
           - 9200:9200
           - 9300:9300
       redis:
-        image: redis
+        image: redis:6
```
pi-sigma, line 54: Redis 6 is EOL. Is there a reason for this version?

[https://redis.io/docs/latest/operate/rs/installing-upgrading/product-lifecycle/](https://redis.io/docs/latest/operate/rs/installing-upgrading/product-lifecycle/)

.github/workflows/ci.yml
```diff
@@ -109,18 +115,22 @@ jobs:
...
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version-file: 'pyproject.toml'
       - name: Install coverage.py
         # We only need coverage, but we need to match the version that was used
         # to generate the coverage files. Grab it from the dependencies.
         run: grep "^coverage" requirements/ci.txt | xargs -r pip install
-      - uses: actions/download-artifact@v4
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```
pi-sigma: We previsouly used the same major version for `actions/download-artifact` and `actions/upload-artifact`, now they come apart (`7.0.1` vs `8.0.1`). Is this going to work or are they meant to be upgraded in tandem?

zizmor.yaml
```diff
@@ -0,0 +1,5 @@
+rules:
+  unpinned-uses:
+    config:
+      policies:
+        github/codeql-action/*: ref-pin
```
pi-sigma, line 8: Could use a quick comment why we have this exception.

pi-sigma (changes requested):

.github/workflows/ci.yml
swrichards: It's what is pinned in docker compose. I am going to leave it for now, but fix this in a separate PR (to also update docker compose).

.github/workflows/ci.yml
swrichards: Good catch, will pin both to 7

zizmor.yaml
swrichards: Done.

pi-sigma (approved):

PR #2492, opened 2026-05-12
[#2514] Cache ZGW objects with Celery on user login

### LINKED ISSUE #2514: ZGW caching on user login

pi-sigma:
## Summary

The ZGW cache is preloaded in the background via a Celery task when a user logs in, so that the case list and case detail pages are served from the cache on the first visit.

## Issue References

Closes #2514 

## Checklist

- [x] CHANGELOG.rst updated

### DIFF
--- a/src/open_inwoner/conf/base.py
+++ b/src/open_inwoner/conf/base.py
@@ -126,7 +126,7 @@ SOLO_CACHE = "local"  # Avoid Redis overhead
 
 # ZGW API caches
 CACHE_ZGW_CATALOGI_TIMEOUT = config("CACHE_ZGW_CATALOGI_TIMEOUT", default=60 * 60 * 24)
-CACHE_ZGW_ZAKEN_TIMEOUT = config("CACHE_ZGW_ZAKEN_TIMEOUT", default=60 * 1)
+CACHE_ZGW_ZAKEN_TIMEOUT = config("CACHE_ZGW_ZAKEN_TIMEOUT", default=60 * 5)
 
 # Maximum number of pagination requests to follow when fetching zaken from ZGW APIs
 ZGW_MAX_REQUESTS = config("ZGW_MAX_REQUESTS", default=8)
@@ -882,6 +882,7 @@ CELERY_BEAT_SCHEDULE = {
 # *should* have the same effect...
 CELERY_WORKER_PREFETCH_MULTIPLIER = 1
 
+
 #
 # SENTRY - error monitoring
 #
@@ -1056,6 +1057,11 @@ ZGW_CASE_LIST_NUM_WORKERS = (
 # cases on the "Mijn Zaken" page. Should be set to slightly less than the overall
 # timeout.
 ZGW_CASE_LIST_FETCH_TIMEOUT = config("ZGW_CASE_LIST_FETCH_TIMEOUT", default=25)
+
+# Timeout in seconds for the login cache warm-up task per API group.
+# Needs to be longer than ZGW_CASE_LIST_FETCH_TIMEOUT because the warm-up fetches
+# status history, roles, and documents on top of what the list view resolves.
+ZGW_CACHE_WARMUP_TIMEOUT = config("ZGW_CACHE_WARMUP_TIMEOUT", default=120)
 
 # notifications
 ZGW_LIMIT_NOTIFICATIONS_FREQUENCY = config(
--- a/src/open_inwoner/openzaak/services.py
+++ b/src/open_inwoner/openzaak/services.py
@@ -510,6 +510,7 @@ class ZGWService:
 
         resolved_zaken.sort(
             key=lambda c: (
+                # negate ordinal for descending order: date has no __neg__.
                 -c.zaak.startdatum.toordinal(),
                 all_api_groups.index(c.api_group),
             )
--- a/src/open_inwoner/openzaak/services.py
+++ b/src/open_inwoner/openzaak/services.py
@@ -148,13 +148,19 @@ class ZGWService:
         )
 
     @staticmethod
-    def _zaken_client_factory(group: ZGWApiGroupConfig) -> ZakenClient:
+    def _zaken_client_factory(
+        group: ZGWApiGroupConfig, config: OpenZaakConfig | None = None
+    ) -> ZakenClient:
+        if config is None:
+            config = OpenZaakConfig.get_solo()
         return cast(
             ZakenClient,
             build_zgw_client_from_service(
                 group.zrc_service,
                 use_openzaak_120_params=group.fetch_eherkenning_zaken_with_openzaak_120_params,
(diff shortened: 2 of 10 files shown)


src/open_inwoner/openzaak/tasks.py
swrichards, line 1: @pi-sigma this is nice, my sense locally is that this does indeed give a non-trivial speedup.

I think we can move quickly ahead with this PoC, it would simple require a small refactor of `CaseListService` so that it does not require a `HttpRequest` -- we might have to do that consolidated identity part of the refactor here, and add a `from_request / from_user` classmethod factory, to build that identity from a request.

Then in the task we can simple call the service methods within a task and don't have to fragment the logic here.

swrichards (commented):
Good approach! As I said in my other comment, here:

1. Let's already start the refactor by ensuring `CaseListService` does not need a `HttpRequest` in its constructor. Instead, we can implement the sealed identity part of the refactor plan, and add various classmethod factories for different call-sites, e.g. `from_request`, `from_user`. That'll allow us to use the service anywhere, not just in a request context (e.g. later on also in the notification context).
2. In this task, let's then actually call e.g. `service.get_zaken()`, which will always match the implementation and uses the threadpools under the hood. That avoids fragmentation of the code.

swrichards:
> Good approach! As I said in my other comment, here:
> 
> 1. Let's already start the refactor by ensuring `CaseListService` does not need a `HttpRequest` in its constructor. Instead, we can implement the sealed identity part of the refactor plan, and add various classmethod factories for different call-sites, e.g. `from_request`, `from_user`. That'll allow us to use the service anywhere, not just in a request context (e.g. later on also in the notification context).
> 2. In this task, let's then actually call e.g. `service.get_zaken()`, which will always match the implementation and uses the threadpools under the hood. That avoids fragmentation of the code.

Post discussion 13/5:

- Move CaseListService to `openzaak.services` as `ZGWService`
- Add `UserIdentity` logic, also implement that in the klanten module
- Rewrite `__init__` and method signatures to have user identities injected, update call sites
- Cache warming task calls `ZGWService.get_zaken_for_user(user)`

Why not error handling first? Because current calls to ZGWCliennt are too coarse: hard to distinguish between "any error should cause us to bail", and "partial fails might be ok if we flag it". Requires better methods with better result return types, to allow the call-site to distinguish what went wrong.

src/open_inwoner/openzaak/tasks.py
```diff
@@ -30,3 +39,101 @@ def process_zaken_notification(notification_data: dict):
     logger.info("Started process_zaken_notification() task")
     notification = factory(Notification, notification_data)
     handle_zaken_notification(notification)
+
+
+@app.task(base=QueueOnce, once={"keys": ["user_bsn", "user_kvk"], "graceful": True})
```
swrichards: I think all args should be keys? kvk/rsin are possibly dupes, but vestigingsnummer could be meaningfully different. I would add all of them.

src/open_inwoner/conf/base.py
```diff
@@ -882,6 +882,10 @@
 # *should* have the same effect...
 CELERY_WORKER_PREFETCH_MULTIPLIER = 1
 
+CELERY_TASK_ROUTES = {
+    "open_inwoner.openzaak.tasks.warm_cache_for_user": {"queue": "zgw_cache_warmup"},
+}
+
```
swrichards: I would leave this out for now for a follow-up, where we also tackle the Celery worker invocation. I think thhis might cause the tasks not to be picked up, because the worker does not know about this queue name.

```suggestion
```

src/open_inwoner/openzaak/tasks.py
```diff
@@ -30,3 +41,115 @@ def process_zaken_notification(notification_data: dict):
     logger.info("Started process_zaken_notification() task")
     notification = factory(Notification, notification_data)
     handle_zaken_notification(notification)
+
+
+@app.task(
+    base=QueueOnce,
+    once={"keys": ["user_bsn", "user_kvk", "user_vestigingsnummer"], "graceful": True},
```
swrichards: Let's be completist here:

```suggestion
    once={"keys": ["user_bsn", "user_kvk", "user_rsin", "user_vestigingsnummer"], "graceful": True},
```

src/open_inwoner/openzaak/services.py
```diff
@@ -449,14 +470,30 @@ def get_zaken(
...
 
+        # Build clients in the main thread so spawned threads make no DB queries.
+        clients_per_group = {
+            group.pk: (
+                self._zaken_client_factory(group),
+                self._catalogi_client_factory(group),
+                group.fetch_eherkenning_zaken_with_rsin,
+            )
+            for group in all_api_groups
+        }
```
swrichards: This will end up in threads sharing clients and thus state in the `requests.Session` from which the client inherits. I would prefer creating these anew for each thread.

swrichards (changes requested):
LGTM. We might want to revisit how to do the cache warming after the pagination list is in, but otherwise just a few small points.

src/open_inwoner/openzaak/services.py
pi-sigma: What is the benefit? The clients are immutable, their attributes only read by the service methods. Also, building them new for each thread would bring back `OpenZaakConfig.get_solo()` into the thread, which we wanted to avoid, no?

swrichards: The issue is that our API clients are instances of `requests.Session`, and there is [some discussion about their thread-safety](https://github.com/psf/requests/issues/2766). The thrust seems to be that it works most of the time, but the authors seem reluctant to declare it's officially supported. Might be best to err on the side of safety.

src/open_inwoner/openzaak/tasks.py
```diff
@@ -30,3 +41,115 @@ def process_zaken_notification(notification_data: dict):
...
+    with parallel(max_workers=10) as executor:
+        futures = {
+            executor.submit(
+                _warm_single_zaak,
+                zaak_with_group.zaak,
+                *clients_per_group[zaak_with_group.api_group.pk],
+                identification,
+            ): zaak_with_group
+            for zaak_with_group in raw_zaken
+        }
```
swrichards, line 103:
```suggestion
        for zaak_with_group in raw_zaken:
            zaken_client = ZGWService._zaken_client_factory(group)
            catalogi_client = ZGWService._catalogi_client_factory(group)
            future = executor.submit(
                _warm_single_zaak,
                zaak_with_group.zaak,
                zaken_client,
                catalogi_client
              )
        }
```

src/open_inwoner/openzaak/services.py
pi-sigma: Misunderstanding on my part

swrichards (approved):
🎊

PR #2493, opened 2026-05-12
[#2449] Set bestandsformaat to upload content type when uploaded to a zaak

### LINKED ISSUE #2449: PP-385: Set bestandsformaat to upload content type when uploaded to a zaak

pi-sigma:
## Summary

Without bestandsformaat (mime type), external services could not determine whether a preview could be rendered for a given file.

## Issue References

Fixes #2449

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables

### DIFF
--- a/src/open_inwoner/openzaak/clients.py
+++ b/src/open_inwoner/openzaak/clients.py
@@ -751,6 +751,8 @@ class DocumentenClient(ZgwAPIClient):
             "taal": "dut",
             "informatieobjecttype": informatieobjecttype_url,
         }
+        if file.content_type:
+            document_body["formaat"] = file.content_type
 
         try:
             response = self.post("enkelvoudiginformatieobjecten", json=document_body)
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openzaak/clients.py
```diff
@@ -746,6 +746,7 @@ def upload_document(
             "inhoud": base64.b64encode(file.read()).decode("utf-8"),
             "bestandsomvang": file.size,
             "bestandsnaam": file.name,
+            "bestandsformaat": file.content_type,
```
swrichards: LGTM, just double-checking if this is nullable (it seems to be technically from the Django source, but it seems all code paths for uploads fill it. If this is nullable in the API we don't have to do anything, otherwise we might only want to add this conditionally.

swrichards (approved):
One question, otherwise LGTM.

src/open_inwoner/openzaak/clients.py
pi-sigma: Goog point. You mean the other way around I think: if the field is nullable, add it conditionally? Also, I used the wrong field name.

swrichards (approved):

PR #2496, opened 2026-05-12
[#2365] Updating requirements to support Python 3.13

alextreme:
## Summary

Currently Python 3.12 is being used for open-inwoner, which is currently on security-support only (https://endoflife.date/python).

I've updated the requirements as a first step to support Python 3.13, I did encounter problems with zope when running management commands so this PR is mainly a first step to go to 3.13


swrichards:
Picked up in #2500

PR #2503, opened 2026-05-13
[#2501] Refactor ZGW business logic

### LINKED ISSUE #2501: ZGW Service Centralisation
All ZGW-related business logic should be centralized in a `class ZGWService` which lives in `openzaak/services.py`

pi-sigma:
## Summary

All ZGW-related business logic is centralized in a `class ZGWService` which lives in `openzaak/services.py`. A linting rule ensures that `MultiZgwClientProxy` is only imported in openzaak modules that need the class. Other sites should use the new `ZGWService` instead.

- [x] Move CaseListService -> ZGWService
- [x] Migrate CaseAccessMixin to ZGWService.check_zaak_access()
- [x] Migrate zaak detail fetching into ZGWService.get_zaak_detail()
- [x] Migrate search to ZGWService.search_zaken()
- [x] Migrate notifications to ZGWService
- [x] Migrate PDC to ZGWService.get_raw_zaken()
- [x] Migrate logic for retrieval of questions in Klant service methods to ZGWService

## Issue References

Closes #2501

Also fixes [Dimpact PP-410](https://dimpact.atlassian.net/browse/PP-410). Previously, the RSIN was incorrectly sent instead of KVK for companies with vestigingsnummer because the `user_rsin` flag was absent from `get_user_fetch_parameters`. The ZGWService methods handle this correctly.

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables


src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+
+
+class CaseFilterFormOption(enum.Enum):
+    FORMULIER = _("Openstaande formulieren")
+    ZAAK_OPEN = _("Lopende aanvragen")
+    ZAAK_AFGEROND = _("Afgeronde aanvragen")
+
+
+@dataclass(frozen=True)
+class DigiDIdentity:
```
swrichards: Let's not to it to digid/eherkenning (even though we've done that before -- it mixes concept, the attribute with the retrieval mechanism, and eidas can also yield these). `BsnIdentity` and `KvkIdentity` (encompasses rsin and vestigingsnummer, which are assigned by kvk).

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+        self._zaak_update_lock = threading.RLock()
+
+        logger.debug(
+            "Configured ZGWService",
+            timeouts=self._timeouts,
+            max_workers=self._max_workers,
+        )
+
+    @staticmethod
+    def get_identity_for_request(request: HttpRequest) -> UserIdentity | None:
```
swrichards: I think this might actually  be cleaner as a factory classmethod on `UserIdentity` itself (though we might have to make `UserIdentity` a baseclass for that). Then you get:

`UserIdentity.from_request(request) -> KvKIdentity | BsnIdentity`

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+    initiator: str
+
+
+class Timeouts(TypedDict):
+    fetch_raw_zaken: int | float
+    resolve_zaken: int | float
+    fetch_formulieren: int | float
+
+
+def _identity_to_fetch_params(identity: UserIdentity, use_rsin: bool = True) -> dict:
```
swrichards: This translation layer is also in OpenKlant2 and it seems a bit cumbersome to go from User => Identity DTO => Fetch Dto => final call. Happy to keep this now, but I would opt for our ZGW client calls accepting the `UserIdentity` DTOs directly, but we can do that in a followup PR.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+                    zaak.url, identity.vestigingsnummer
+                )
+            else:
+                if api_group.fetch_eherkenning_zaken_with_rsin and identity.rsin:
+                    kvk_or_rsin = identity.rsin
+                else:
+                    kvk_or_rsin = identity.kvk
+                rollen = zaken_client.fetch_roles_for_zaak_and_kvk_or_rsin(
+                    zaak.url, kvk_or_rsin
+                )
```
swrichards: Another sign that the client calls should probably accept the identity DTOs immediately.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -306,7 +280,7 @@ def get_context_data(self, **kwargs):
                 "questions": questions,
             }
```
swrichards: I wonder if this entire direct except for `questions` has added value. My instinct would be to make sure we migrate all this display logic to the service, so we can return a single object with a clear type signature here.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+        self,
+        object_id: str,
+        identity: UserIdentity,
+        api_group: ZGWApiGroupConfig,
+    ) -> tuple[Zaak, ZGWApiGroupConfig] | None:
+        zaken_client = self._zaken_client_factory(api_group)
+
+        zaak = zaken_client.fetch_single_zaak(object_id)
+        if not zaak:
+            raise ZaakNotFound(f"Zaak {object_id!r} not found or API unavailable")
```
swrichards: I imagine this is a temporary exception because of the error suppression? If so let's add an explicit TODO to later differentiate between legitimate not found and API errors.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+
+        config = OpenZaakConfig.get_solo()
+        limit_access_to_role = config.limit_user_visible_cases_to_role
+
+        if isinstance(identity, DigiDIdentity):
```
swrichards: Let's use `match` to be more specific and have an exception if we're passed a type we don't recognize in the `case _`.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+            rollen = zaken_client.fetch_roles_for_zaak_and_bsn(zaak.url, identity.bsn)
+        else:
+            if identity.vestigingsnummer:
```
swrichards: As noted elsewhere, what if we move this translation layer to a new method on `ZakenClient` with something like `fetch_zaken_for_user(identity)` and encapsulate this translation there as en intermediate step?

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+        self,
+        zaken: Iterable[ZaakWithApiGroup],
+        formulieren: Iterable[FormulierWithApiGroup],
+    ) -> dict[CaseFilterFormOption, int]:
+        zaak_statuses = [self.get_zaak_filter_status(zaak.zaak) for zaak in zaken]
+        zaak_statuses += [CaseFilterFormOption.FORMULIER for _ in formulieren]
+        return {
+            status: zaak_statuses.count(status) for status in list(CaseFilterFormOption)
+        }
+
```
swrichards: Here I actually might think it does belong on the case detail view: it manages UI state for that page, and it's closely coupled to the filter widget setup.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,771 @@
...
+                try:
+                    fetched_zaken.extend(task.result())
+                except BaseException:
+                    logger.exception("Error fetching raw zaken for group")
+        except concurrent.futures.TimeoutError:
+            logger.warning("Timed out fetching raw zaken", exc_info=True)
+
+        return fetched_zaken
+
+    def get_zaken(self, identity: UserIdentity | None) -> list[ZaakWithApiGroup]:
```
swrichards:
```suggestion
    def get_zaken(self, identity: UserIdentity) -> list[ZaakWithApiGroup]:
```

swrichards (commented):
Good start here, cleaner already. I think the big questions around the user identity DTO (where to put it, and the intermediate layer, should we do this in the ZGWClient?) and some of the case detail display logic (what lives in the service, what lives in the view). We can have a chat Monday, but we're getting there.

src/open_inwoner/cms/cases/views/status.py
pi-sigma: This is presentation logic, implementing this in the service creates the wrong coupling IMO. However, I'll extract this into a separate method on the view with a  dataclass to make `get_context_data` more readable and improve type safety.

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -129,86 +136,18 @@ def dispatch(self, request, *args, **kwargs):
...
-                if not is_zaak_visible(self.zaak):
-                    logger.info(
-                        "CaseAccessMixin - permission denied: case  is not visible",
-                        zaak_url=self.zaak.url,
-                    )
+            try:
+                result = ZGWService().check_zaak_access(object_id, identity, api_group)
+            except ZaakNotFound:
+                # API unavailable or zaak removed; let the detail view render gracefully
+                pass
```
swrichards, line 146: Why shouldn't this also raise, defaulting to 404/permission denied?

src/open_inwoner/cms/cases/views/mixins.py
```diff
@@ -129,86 +136,18 @@ def dispatch(self, request, *args, **kwargs):
...
+            else:
+                if result is None:
                     return self.handle_no_permission()
+                zaak, api_group = result
+                request.case = zaak
+                request.case_api_group = api_group
+                self.zaak = zaak  # backward compat for detail views not yet handled by ZGWService
```
swrichards: Are there any?

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -70,6 +66,74 @@
...
+            description=zaak.description,
+            statuses=statuses,
+            end_statustype_data=end_statustype_data,
+            second_status_preview=second_status_preview,
+            documents=documents,
+            allowed_file_extensions=sorted(openzaak_config.allowed_file_extensions),
+            new_docs=has_new_elements(
+                documents,
+                "created",
+                dt.timedelta(days=settings.DOCUMENT_RECENT_DAYS),
```
swrichards, line 129: Not for this PR, but this is a good candidate of a config setting that does not have to be settings/env configurable, could be on `OpenZaakConfig`.

src/open_inwoner/cms/cases/views/status.py
```diff
@@ -237,76 +287,54 @@ def get_context_data(self, **kwargs):
...
-                self.resulttype_config_mapping,
-                zaken_client,
-                catalogi_client,
+                status_types_mapping=zaak_detail.status_types_mapping,
+                end_statustype=self.handle_end_statustype(
+                    zaak_detail.statuses, zaak_detail.statustypen
+                ),
             )
 
+            documents = [
```
swrichards, line 305: Do we need to reproduce the original sorting logic?

src/open_inwoner/openzaak/clients.py
```diff
@@ -76,39 +76,37 @@ def __init__(
...
-        raise ValueError("You must supply either a bsn or kvk/rsin/vestigingsnummer")
+    ) -> list[Zaak]:
+        match user_identity:
+            case BSNIdentity():
+                return self.fetch_zaken_by_bsn(
+                    user_identity.bsn,
+                    max_requests=max_requests or settings.ZGW_MAX_REQUESTS,
+                    identificatie=identificatie,
+                )
+            case KVKIdentity():
```
swrichards: This makes me wonder whether an additional `RSINIdentity` isn't cleaner, avoids the conditionals in this leg.

swrichards: Actually never mind, I guess that logic is now in the client for now.

src/open_inwoner/openzaak/identity.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from django.http import HttpRequest
+
+
+class UserIdentity:
```
swrichards: Nitpick (could be a followup commit): what about calling this `UserIdentification` (more precise), move it to auth, and add it as a `@property` on the user model, so we can start using this as the common way of doing things, also e.g. at some point in the open klant service.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+            RolTypes.vestiging,
+        )
+        roles = []
+        for betrokkene_type in available_user_betrokkene_types:
+            roles += client.fetch_zaak_roles(zaak_url, betrokkene_type=betrokkene_type)
+        return roles
+
+    def check_zaak_access(
+        self,
+        object_id: str,
```
swrichards: Confusing name: `zaak_id`?

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+                    fetched_zaken.extend(task.result())
+                except BaseException:
+                    logger.exception("Error searching zaken in group")
+        except concurrent.futures.TimeoutError:
+            logger.warning("Timed out searching zaken", exc_info=True)
+
+        visible_zaken = []
+        for zaak_with_group in fetched_zaken:
+            try:
+                self._resolve_zaak_type(zaak_with_group)
```
swrichards, line 427: Add a TODO here (or maybe in the ticket) that this could be done in parallel too -- it potentially includes API calls, and given it's called for each zaak, it could be a quick-win speed-wise.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+                try:
+                    fetched_zaken.extend(task.result())
+                except BaseException:
+                    logger.exception("Error fetching raw zaken for group")
+        except concurrent.futures.TimeoutError:
+            logger.warning("Timed out fetching raw zaken", exc_info=True)
+
+        return fetched_zaken
+
+    def search_zaken(
```
swrichards, line 417: Up to you whether it belongs in this PR, but there's a fair bit of DRY here, the main difference is the `identificatie` parameter. Maybe we could add that as an optional kwarg to `get_raw_zaken` and re-use that?

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+                    "Culling zaak %s because not all resolutions completed successfully",
+                    zaak.identification,
+                )
+                continue
+
+            zaak_with_resolved_zgw_refs = self._replace_catalogus_api_with_model_refs(
+                zaak
+            )
+
+            if is_zaak_visible(zaak_with_resolved_zgw_refs.zaak):
```
swrichards: Could we already move `is_zaak_visible` here and make it a private method, as further defence in depth to avoid other call-sites doing their own visibility logic?

swrichards: Or at least from `utils` into this module as `_is_zaak_visible`

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+        for zaak_with_group in fetched_zaken:
+            try:
+                self._resolve_zaak_type(zaak_with_group)
+            except ResolveCaseException:
+                logger.warning(
+                    "Unable to resolve zaaktype for search result",
+                    zaak_url=zaak_with_group.zaak.url,
+                )
+                continue
+            if is_zaak_visible(zaak_with_group.zaak):
```
swrichards: As above, I think we don't want to duplicate this logic if we can.

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+                    all_futures.append(future)
+                    zaak_futures[zaak_with_group].append(future)
+
+        try:
+            concurrent.futures.wait(
+                all_futures,
+                timeout=self._timeouts["resolve_zaken"],
+            )
+        except concurrent.futures.TimeoutError:
+            logger.warning("Timed out resolving zaken", exc_info=True)
```
swrichards: For the todo list: this is a great place for an OTEL metric, so we can track/alert in production how often this happens (which might mean having to tweak number of workers, timeouts, that sort of thing).

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+
+    def _get_initiator_display(
+        self,
+        zaak: Zaak,
+        api_group: ZGWApiGroupConfig,
+        user_identity: UserIdentity,
+    ) -> str:
+        zaken_client = api_group.zaken_client
+        if api_group.fetch_rollen_with_betrokkene_type:
+            if isinstance(user_identity, BSNIdentity):
```
swrichards: `match`!

src/open_inwoner/openzaak/services.py
```diff
@@ -0,0 +1,774 @@
...
+
+@dataclass
+class ZaakDetailData:
+    """All ZGW API data fetched for the zaak detail page."""
+
+    statuses: list[Status]
+    statustypen: list[StatusType]
+    status_types_mapping: dict[str, StatusType]
+    documents: list[ZaakDocumentData]
+    result: dict
```
swrichards, line 82: Cleaner as `dict | None` for explicitness, given in the function it might be empty?

swrichards (changes requested):
Looking really good here. A few things here and there, for most of that: feel free to shunt it off to an upcoming PR if you feel that's more appropriate.

src/open_inwoner/openzaak/tests/test_services.py
```diff
@@ -0,0 +1,509 @@
...
+        bad_future = concurrent.futures.Future()
+        bad_future.set_exception(ConnectionError("forms api unreachable"))
+
+        with patch("concurrent.futures.as_completed", return_value=iter([bad_future])):
+            result = self.service.get_formulieren(_USER_IDENTITY)
+
+        self.assertEqual(result, [])
+
+
+class TimeoutHandlingTests(ClearCachesMixin, TestCase):
```
swrichards, line 14: I think this is the main test to keep (and a good one)). I am not sure asserting against the log is the most interesting condition to assert: what we would really want to assert in these cases is that the unfinished futures are discarded but the finished ones carry through. That's a more elaborate test, but we can maybe defer it to the error handling PR.

As for the other tests: they feel a bit too white-box for my tate, and the error handling PR will change their behavior significantly, but happy to keep them in if that helps the later refactor.

src/open_inwoner/cms/cases/views/mixins.py
pi-sigma: This is also temporary, since we currently don't distinguish between temporary API failures and genuine `404`. Will be updated when tackling #2142. I've added a `TODO`

src/open_inwoner/cms/cases/views/mixins.py
pi-sigma: Comment removed. Also, I removed `request.case = zaak` and `request.case_api_group = api_group` as these are not needed. The zaak is view-level state and belongs on the view, same as `api_group`.

src/open_inwoner/cms/cases/views/status.py
pi-sigma: The sorting is already in `_fetch_zaak_documents` in `ZGWService`.

src/open_inwoner/openzaak/services.py
pi-sigma: Good idea!

src/open_inwoner/openzaak/services.py
pi-sigma: What do you mean, you prefer the old annotation?

swrichards: The calling code does `if zaak.result`, thought it might cleaner if it was `None` on the empty case to show the intent.

src/open_inwoner/openzaak/identity.py
pi-sigma: Good idea!

src/open_inwoner/openzaak/tests/test_services.py
pi-sigma: Fair enough, I've removed the other tests and merged the commit with the one that fixes the timeout handling. Less code and fewer commits.

src/open_inwoner/openzaak/services.py
pi-sigma: Got it. But in that case every call site needs an `if zaak_detail.result else ""` guard against `None`. We only have two, but still...

swrichards (approved):

PR #2509, opened 2026-05-19
fix: [#2508] normalize phone/email before digitale adressen deduplica…

### LINKED ISSUE #2508: PP-420: Syncing phone/email to openklant should not create new address if one exists
Mainly tricky due to reliable comparisons of phone numbers with their varying format. How would we normalize this?

swrichards:
…tion

Digital addresses, especially phone numbers, would be duplicated due to missing normalization (e.g. spaces, parentheses). This commit ensure in both reading and writing that emails and and phone numbers are normalized.

Closes: #2508

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 import uuid
 from dataclasses import dataclass
 from datetime import timedelta
@@ -31,6 +32,7 @@ from openklant_client.types.resources.betrokkene import (
 from openklant_client.types.resources.digitaal_adres import (
     DigitaalAdres,
     DigitaalAdresCreateData,
+    DigitaalAdresPartialUpdateData,
 )
 from openklant_client.types.resources.interne_taak import InterneTaak
 from openklant_client.types.resources.klant_contact import (
@@ -85,6 +87,18 @@ from open_inwoner.utils.time import instance_is_new
 from open_inwoner.utils.views import LogMixin
 
 logger = structlog.stdlib.get_logger(__name__)
+
+
+def _normalize_phone(phone: str) -> str:
+    """Strip formatting; convert leading '+' to '00' (API requires 00-prefix, not +)."""
+    stripped = re.sub(r"[^0-9+]", "", phone.strip())
+    if stripped.startswith("+"):
+        stripped = "00" + stripped[1:]
+    return stripped
+
+
+def _normalize_email(email: str) -> str:
+    return email.strip().lower()
 
 
 class BsnFetchParam(TypedDict):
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1254,17 +1268,29 @@ def _get_or_create_digitaal_adres(
...
             soort_digital_adres=soort_adres,
             adressen=adressen,
         )
         for digitaal_adres in digitale_adressen:
-            if (
-                digitaal_adres["adres"] == adres
-                and digitaal_adres["isStandaardAdres"] == is_standaard_adres
-            ):
+            if normalize(digitaal_adres["adres"]) == adres:
+                if is_standaard_adres and not digitaal_adres["isStandaardAdres"]:
```
pi-sigma: When a caller is looking for an address with `is_standard_adres=False` but the matching address has `is_standard_adres=True`, we're going to return the standard address. This means there is no way to distinguish between "found matching standard address" and "found matching non-standard address". To be fair, the previous behavior (creating a new one) was worse, and I don't have a suggestion to improve this. Worth noting though.

src/open_inwoner/openklant/services.py
```diff
@@ -87,6 +89,18 @@
 logger = structlog.stdlib.get_logger(__name__)
 
 
+def _normalize_phone(phone: str) -> str:
+    """Strip formatting; convert leading '+' to '00' (API requires 00-prefix, not +)."""
+    stripped = re.sub(r"[^0-9+]", "", phone.strip())
```
pi-sigma, line 94: This will leave occurrences of '+' in the middle of the number. Since this is for normalization not cleaning, I guess this is acceptable.

pi-sigma (approved):

src/open_inwoner/openklant/services.py
swrichards: Good point. I made `None` the standard so you can explicitly choose to upgrade (there's no symmetric downgrade if it's `False', but I think that's fine for now).

src/open_inwoner/openklant/services.py
swrichards: Agreed, and the + at the start is the only element we're confident making assumptions about.

src/open_inwoner/openklant/tests/data.py
```diff
@@ -363,17 +353,27 @@ def install_mocks_openklant(self, m) -> "MockAPIReadPatchData":
             },
         )
 
-        # Mock for creating new digital addresses
+        # Mock for creating new digital addresses: echo back the request body so the
+        # returned soortDigitaalAdres matches what was sent (prevents false cache hits).
+        partij_uuid = self.partij_uuid
```
pi-sigma: `partij_uuid` is only used once in the mock below; might as well use `self.partij_uuid` directly there (unless you cannot use it inside the closure for some reason)

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -870,6 +872,219 @@ def test_questions_for_partij_missing_inhoud(self, mock_client_class):
         self.assertEqual(q3.answers[0].answer, "Second answer to Q3")
 
 
+class NormalizeHelpersTestCase(TestCase):
```
pi-sigma, line 875: What do we gain from testing the implementation, when we already test the interface below with `GetOrCreateDigitaalAdresNormalizationTestCase`?

pi-sigma (approved):

src/open_inwoner/openklant/tests/test_openklant2_service.py
swrichards: Mostly just a helpful visual overview of the expected transformations.

PR #2522, opened 2026-05-26
[#2513] Paginate case list view

### LINKED ISSUE #2513: Pagination of zaken list

pi-sigma:
## Summary

Previously every request to the zaken list view or CMS plugin fully resolved all of the user's zaken (status, statustype, resultaat, resultaattype) before any pagination was applied, even though only a small slice is displayed.

Changes:

  - services.py: Replaces `get_zaken()` with two new methods:
    - `get_visible_zaken()` - resolves only zaaktype for all zaken (cheap; cached). Uses the existing _is_zaak_visible check to filter, returning the full visible list for pagination metadata.
    - `fully_resolve_zaken()` - resolves status/statustype and resultaat/resultaattype for a given slice only.
- cms/cases/views/cases.py (`InnerCaseListView`): Computes page boundaries against the full visible list, then passes only the page slice to fully_resolve_zaken. Formulieren fill the first slots and are accounted for in the combined offset
   math.
  - cms/plugins/views.py (`ZakenPluginContentView`): Same pattern - `get_visible_zaken()` for the full list, `fully_resolve_zaken` on only as many zaken as `num_zaken` requires.
  - utils/mixins.py: Adds `PaginationMixin.paginate_preloaded()` for building Django pagination context from a pre-sliced list with a separately-known total count.
 - Fixes a pre-existing bug in `_resolve_status_and_status_type` that always culled null-status zaken, making `show_cases_without_status` useless.

## Issue References

Closes #2513 

## Checklist

- [x] CHANGELOG.rst updated
  - [ ] Deployment notes added
        
- [ ] Documentation updated 
      
        environment variables

### DIFF
--- a/src/open_inwoner/cms/cases/views/cases.py
+++ b/src/open_inwoner/cms/cases/views/cases.py
@@ -166,7 +166,7 @@ class InnerCaseListView(
         combined_total = formulieren_count + len(all_visible_zaken)
 
         paginator_dict = self.paginate_preloaded(
-            page_items, combined_total, page_number
+            page_items, combined_total, page_number, self.paginate_by
         )
         zaken_dicts = [case.process_data() for case in paginator_dict["object_list"]]
 
--- a/src/open_inwoner/cms/cases/views/cases.py
+++ b/src/open_inwoner/cms/cases/views/cases.py
@@ -16,7 +16,7 @@ from open_inwoner.openzaak.api_models import Zaak
 from open_inwoner.openzaak.models import OpenZaakConfig
 from open_inwoner.openzaak.services import (
     FormulierWithApiGroup,
-    ZaakWithApiGroup,
+    ZaakWithApiGroupZaakTypeResolved,
     ZGWService,
 )
 from open_inwoner.openzaak.types import UniformCase
@@ -41,7 +41,7 @@ def _get_zaak_filter_status(zaak: Zaak) -> CaseFilterFormOption:
 
 
 def _get_zaak_status_frequencies(
-    zaken: Iterable[ZaakWithApiGroup],
+    zaken: Iterable[ZaakWithApiGroupZaakTypeResolved],
     formulieren: Iterable[FormulierWithApiGroup],
 ) -> dict[CaseFilterFormOption, int]:
     zaak_statuses = [_get_zaak_filter_status(zaak.zaak) for zaak in zaken]
@@ -107,26 +107,26 @@ class InnerCaseListView(
         case_service = ZGWService()
         context["filter_form_enabled"] = config.zaken_filter_enabled
 
-        # update ctx with formulieren and cases (possibly filtered)
+        try:
+            page_number = int(self.request.GET.get(self.page_kwarg, 1))
+        except (TypeError, ValueError):
+            page_number = 1
+
         formulieren: Sequence[UniformCase] = case_service.get_formulieren(
             user_identification
         )
-        preprocessed_zaken: Sequence[UniformCase] = case_service.get_zaken(
-            user_identification
-        )
+        all_visible_zaken = case_service.get_visible_zaken(user_identification)
 
         if config.zaken_filter_enabled:
             case_status_frequencies = _get_zaak_status_frequencies(
-                zaken=preprocessed_zaken,
+                zaken=all_visible_zaken,
                 formulieren=formulieren,
             )
-            # Separate frequency data from statusname
             context["status_freqs"] = [
                 (status.value, frequency)
                 for status, frequency in case_status_frequencies.items()
             ]
 
-            # Validate statuses are valid according to the options enum
             statuses: list[CaseFilterFormOption] = []
             for status in self.request.GET.getlist("status"):
                 try:
@@ -138,24 +138,40 @@ class InnerCaseListView(
                         user=self.request.user,
                     )
 
-            # Actually filter the formulieren
             if statuses:
                 formulieren = (
                     formulieren if CaseFilterFormOption.FORMULIER in statuses else []
                 )
-                preprocessed_zaken = [
+                all_visible_zaken = [
                     zaak
-                    for zaak in preprocessed_zaken
+                    for zaak in all_visible_zaken
                     if _get_zaak_filter_status(zaak.zaak) in statuses
                 ]
 
-        paginator_dict = self.paginate_with_context([*formulieren, *preprocessed_zaken])
+        # Formulieren fill the first slots; zaken follow.
+        # max(0, ...) prevents negative indices when the page falls entirely within
+        # the formulieren block, in which case zaak_start == zaak_end == 0 and
+        # all_visible_zaken[0:0] == [] so no zaken are resolved.
+        formulieren_count = len(formulieren)
+        combined_start = (page_number - 1) * self.paginate_by
+        combined_end = page_number * self.paginate_by
+        formulieren_page = formulieren[combined_start:combined_end]
+        zaak_start = max(0, combined_start - formulieren_count)
+        zaak_end = max(0, combined_end - formulieren_count)
+        zaak_page = case_service.fully_resolve_zaken(
+            all_visible_zaken[zaak_start:zaak_end]
+        )
+
+        page_items = [*formulieren_page, *zaak_page]
+        combined_total = formulieren_count + len(all_visible_zaken)
+
+        paginator_dict = self.paginate_preloaded(
+            page_items, combined_total, page_number
+        )
         zaken_dicts = [case.process_data() for case in paginator_dict["object_list"]]
 
         context["zaken"] = zaken_dicts
         context.update(paginator_dict)
-
-        # other data
         context["hxget"] = reverse("cases:cases_content")
         context["title_text"] = config.title_text
 
--- a/src/open_inwoner/cms/cases/views/cases.py
+++ b/src/open_inwoner/cms/cases/views/cases.py
@@ -115,7 +115,7 @@ class InnerCaseListView(
         formulieren: Sequence[UniformCase] = case_service.get_formulieren(
             user_identification
         )
-        all_visible_zaken = case_service.get_visible_zaken(user_identification)
+        all_visible_zaken = case_service.get_visible_zaken(user_identification).zaken
 
         if config.zaken_filter_enabled:
             case_status_frequencies = _get_zaak_status_frequencies(
@@ -160,7 +160,7 @@ class InnerCaseListView(
         zaak_end = max(0, combined_end - formulieren_count)
         zaak_page = case_service.fully_resolve_zaken(
             all_visible_zaken[zaak_start:zaak_end]
-        )
+        ).zaken
 
         page_items = [*formulieren_page, *zaak_page]
         combined_total = formulieren_count + len(all_visible_zaken)
--- a/src/open_inwoner/openzaak/services.py
+++ b/src/open_inwoner/openzaak/services.py
@@ -99,6 +99,14 @@ class ZaakWithApiGroup:
         return hash((self.identification, self.api_group.pk))
 
 
+class ZaakWithApiGroupZaakTypeResolved(ZaakWithApiGroup):
+    """ZaakWithApiGroup whose zaaktype has been resolved from URL to ZaakType."""
+
+
+class ZaakWithApiGroupFullyResolved(ZaakWithApiGroupZaakTypeResolved):
+    """ZaakWithApiGroup with zaaktype, status, and resultaat all resolved."""
+
+
 @dataclass(frozen=True)
 class FormulierWithApiGroup:
     formulier: Formulier
@@ -121,9 +129,10 @@ class FormulierWithApiGroup:
 
 
 class Timeouts(TypedDict):
-    fetch_raw_zaken: int | float
-    resolve_zaken: int | float
-    fetch_formulieren: int | float
+    get_raw_zaken: int | float
+    get_visible_zaken: int | float
+    fully_resolve_zaken: int | float
+    get_formulieren: int | float
 
 
 class ZGWService:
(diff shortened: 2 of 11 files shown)


src/open_inwoner/openzaak/services.py
```diff
@@ -466,55 +478,86 @@ def get_zaken(
...
-        resolved_zaken.sort(
-            key=lambda c: (
-                -c.zaak.startdatum.toordinal(),
-                all_api_groups.index(c.api_group),
-            )
+        return visible_zaken
+
+    def fully_resolve_zaken(
+        self, zaken: list[ZaakWithApiGroup]
+    ) -> list[ZaakWithApiGroup]:
```
swrichards: I think one of the nice things about slicing it up this way is that we can do separate types for each method: raw zaak -> raw zaak with result zaaktype -> fully resolved zaak.

Maybe too much for this PR, but if we did we could for instance here enforce that this is only called on a raw zaak that has a resolved zaaktype.

src/open_inwoner/openzaak/services.py
```diff
@@ -434,9 +440,15 @@ def search_zaken(
...
         self, user_identification: UserIdentification
     ) -> list[ZaakWithApiGroup]:
+        """
+        Fetch all visible zaken with only zaaktype resolved (status/resultaat
+        left as raw URLs). This is cheap because zaaktype lookups are cached and
+        status presence is checked from raw data. Use this for pagination and
+        total-count; pass the page slice to fully_resolve_zaken for display.
+        """
         if not user_identification:
             return []
```
swrichards: I think we can probably do without this?

src/open_inwoner/openzaak/services.py
```diff
@@ -466,55 +478,86 @@ def get_zaken(
...
+                zaaktype_futures[zaak_with_group] = executor.submit(
+                    self._resolve_zaak_type, zaak_with_group
+                )
 
-        # TODO add wait call or refactor: we want either all futures to complete or timeout
-        resolved_zaken: list[ZaakWithApiGroup] = []
-        for zaak, futures in zaak_futures.items():
-            if not all(f.done() and not f.exception() for f in futures):
+        visible_zaken: list[ZaakWithApiGroup] = []
+        for zaak_with_group in fetched_zaken:
```
swrichards: I think we could perhaps use futures `as_completed` here with a timeout, and reverse the mapping? (future is the key, zaak is the value)

src/open_inwoner/openzaak/services.py
```diff
@@ -466,55 +478,86 @@ def get_zaken(
...
+        self, zaken: list[ZaakWithApiGroup]
+    ) -> list[ZaakWithApiGroup]:
+        """
+        Fully resolve status+statustype and resultaat+resultaattype
+        for a page slice. Input zaken must already have zaaktype resolved
+        (as returned by get_visible_zaken). Mutates zaak objects in
+        place; zaken that fail resolution are culled from the result.
+        """
+        if not zaken:
+            return zaken
```
swrichards: I am not sure this adds much: for empty lists, all we do is construct the executor which I don't think is a very expensive operation.

src/open_inwoner/openzaak/services.py
```diff
@@ -466,55 +478,86 @@ def get_zaken(
...
+                )
+                if isinstance(zaak_with_group.zaak.status, str):
+                    zaak_futures[zaak_with_group].append(
+                        executor.submit(
+                            self._resolve_status_and_status_type, zaak_with_group
+                        )
+                    )
+
+        fully_resolved: list[ZaakWithApiGroup] = []
+        for zaak_with_group in zaken:
```
swrichards, line 539: As above, let's add timeouts and `as_completed`.

src/open_inwoner/openzaak/services.py
swrichards: (the timeout in any case is important)

src/open_inwoner/openzaak/services.py
```diff
@@ -572,6 +615,9 @@ def _resolve_status_and_status_type(
         zaken_client = ZGWService._zaken_client_factory(zaak_with_group.api_group)
         catalogi_client = ZGWService._catalogi_client_factory(zaak_with_group.api_group)
 
+        if zaak_with_group.zaak.status is None:
+            return
```
swrichards, line 586: Nitpick, but can move above client destruction.

src/open_inwoner/openzaak/services.py
swrichards, line 592: Can also move above client destruction.

src/open_inwoner/utils/mixins.py
```diff
@@ -160,6 +160,43 @@ def paginate_with_context(
...
+    ) -> PaginationContext:
+        """
+        Build pagination context for a pre-sliced page of items.
+
+        Use when the caller has already determined the correct items
+        for the current page (e.g. after server-side filtering and slicing).
+        The paginator uses total_count for metadata (num_pages, has_next,
+        etc.) without re-slicing the list.
+        """
+        page_size = page_size or self.paginate_by
```
swrichards: If it's pre-sliced, the page size should also be fixed as the pre-slicing will have made assumptions about that. I'd remove the `or` and make the argument mandatory.

src/open_inwoner/cms/cases/views/cases.py
```diff
@@ -138,24 +138,37 @@ def get_context_data(self, **kwargs):
...
 
-        paginator_dict = self.paginate_with_context([*formulieren, *preprocessed_zaken])
+        # Formulieren fill the first slots; zaken follow. Compute each type's page slice.
+        formulieren_count = len(formulieren)
+        combined_start = (page_number - 1) * self.paginate_by
+        combined_end = page_number * self.paginate_by
+        formulieren_page = formulieren[combined_start:combined_end]
+        zaak_start = max(0, combined_start - formulieren_count)
+        zaak_end = max(0, combined_end - formulieren_count)
+        zaak_page = case_service.fully_resolve_zaken(
```
swrichards, line 161: Had to read this a few times! Might be nice to add a comment for the slow reader like me, e.g. 

```
zaken not within the combined page will end up reading as all_visibible_zaken[0:0] and thus nit resolve
```

or some such

swrichards (changes requested):
Great stuff, the breakdown of the service methods feels a lot cleaner now. Just need some explicit timeout/await logic.

src/open_inwoner/openzaak/services.py
swrichards: Note also we define the timeouts on the service (for now), they might need re-organizing to reflect the new breakdown, the timeout keys assumed a different method breakdown.

src/open_inwoner/openzaak/services.py
pi-sigma: This is also called from the CMS zaken plugin, where a non-authenticated user has `user_identification = None`. We return early to avoid calling the API

src/open_inwoner/openzaak/services.py
pi-sigma: Good idea!

src/open_inwoner/openzaak/services.py
```diff
@@ -130,10 +161,15 @@ class ZGWService:
...
+    # Setup
+    # -------------------------------------------------------------------------
+
     def __init__(self, use_cache: bool = True):
         self._use_cache = use_cache
         self._timeouts = {
             "fetch_raw_zaken": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.3,
+            "fetch_zaaktype": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.3,
             "resolve_zaken": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.5,
             "fetch_formulieren": settings.ZGW_CASE_LIST_FETCH_TIMEOUT * 0.2,
```
swrichards: Technically the multiplication factors should sum to 1, the idea was to distribute the sub-timeouts over the total available timeouts.

But this is actually a good nudge that, now that this is a more generic service, these timeouts should probably be configurable in the constructor with sane defaults, and they should only use this proportional logic within the case list view.

The strict distribution only matters in a synchronous view, less so in Celery tasks where the (default) timeouts can be more generous.

src/open_inwoner/openzaak/services.py
swrichards: As per other comment, might want to make these timeouts match the method names
```suggestion
                    timeout=self._timeouts["fetch_visisble_zaken"],
```

src/open_inwoner/openzaak/services.py
```diff
@@ -466,91 +369,135 @@ def get_zaken(
...
+                        skipped.append(
+                            SkippedZaak(
+                                zaak_url=zaak_with_group.zaak.url,
+                                reason=SkipReason.ZAAKTYPE_RESOLUTION_FAILED,
+                            )
+                        )
+                        continue
+                    skip_reason = self._get_zaak_skip_reason(zaak_with_group.zaak)
+                    if skip_reason is None:
+                        visible_ids.add(id(zaak_with_group))
```
swrichards, line 450: Is the `id` needed? I think it'll use `__hash__` for this and that should be unique by default. (See also line 430 below).

swrichards: Ah it's because we (I) overwrote `__hash__` on `ZaakWithApiGroup`. Let's keep `id` then, I would prefer to remove `__hash__` (I suspect I just cared about equality, which means the that logic could move to `__eq__`), but it's for a follow-up cleanup, I'll make a note of it.

src/open_inwoner/openzaak/services.py
```diff
@@ -466,91 +369,135 @@ def get_zaken(
...
+                for future in concurrent.futures.as_completed(
+                    future_to_zaak,
+                    timeout=self._timeouts["resolve_zaken"],
+                ):
+                    zaak_with_group = future_to_zaak[future]
+                    remaining[id(zaak_with_group)] -= 1
+                    try:
+                        future.result()
+                    except (ZgwAPIError, ResolveCaseException):
+                        failed_ids.add(id(zaak_with_group))
```
swrichards, line 533: I think we should catch all exceptions here, otherwise one failing result (for reasons other than these exceptions) can bring down the entire method.

src/open_inwoner/openzaak/services.py
```diff
@@ -466,91 +369,135 @@ def get_zaken(
...
+            try:
+                for future in concurrent.futures.as_completed(
+                    future_to_zaak,
+                    timeout=self._timeouts["fetch_zaaktype"],
+                ):
+                    zaak_with_group = future_to_zaak[future]
+                    processed_ids.add(id(zaak_with_group))
+                    try:
+                        future.result()
+                    except (ZgwAPIError, ResolveCaseException):
```
swrichards: Broaden this, see below: don't want this to bring down the whole thing.

src/open_inwoner/openzaak/services.py
swrichards, line 533: This is also a skip reason.

swrichards (changes requested):
LGTM! The only real thing of note: we don't have a container with skipped results for the fully resolve method. I think it would be useful to have that: later on we'll want it for the config check, and we might also want to show some user feedback for certain skip reasons, i.e. if any are due to a timeout or resolution failure, we should probably display a "Could not display all your cases" messages (for a later PR).

src/open_inwoner/openzaak/services.py
swrichards: I was ambiguous: for now just make sure they balance to 1, I'll add moving the timeouts to a running cleanup list.

src/open_inwoner/openzaak/services.py
pi-sigma: I deliberately left those out because they are skipped due to technical/infra concerns, not business logic. But since we don't always have easy access to prod logs, it makes sense to blur this line so we are able to inspect all skipped zaken from the admin

src/open_inwoner/openzaak/services.py
```diff
@@ -173,189 +213,43 @@ def _documenten_client_factory(group: ZGWApiGroupConfig) -> DocumentenClient:
         return cast(DocumentenClient, build_zgw_client_from_service(group.drc_service))
 
     @staticmethod
-    def _is_zaak_visible(zaak: Zaak) -> bool:
+    def _get_zaak_skip_reason(zaak: Zaak) -> SkipReason | None:
```
swrichards: The semantics of this are slightly odd to me: get the skip reason, but None means it was not skipped, is also a proxy for "is visible": what about `_is_zaak_visible(zaak: Zaak) -> tuple[bool, SkipReason | None] ` where the bool indicates visibility, and the second element the optional reason?

src/open_inwoner/openzaak/services.py
```diff
@@ -489,125 +393,179 @@ def get_zaken(
...
+        )
+
+    def fully_resolve_zaken(
+        self, zaken: list[ZaakWithApiGroupZaakTypeResolved]
+    ) -> ZakenResult[ZaakWithApiGroupFullyResolved]:
+        """
+        Fully resolve status+statustype and resultaat+resultaattype
+        for a page slice. Input zaken must already have zaaktype resolved
+        (as returned by get_visible_zaken). Mutates zaak objects in
+        place; zaken that fail resolution are culled from the result.
```
swrichards, line 488: We see mutates in place, which is true, but we still return a new result. It looks like the order of `zaken` is preserved, but that's a bit implicit. What abot just adding 

```
fetched_zaken.sort(
            key=lambda c: (
                # negate ordinal for descending order: date has no __neg__.
                -c.zaak.startdatum.toordinal(),
                all_api_groups.index(c.api_group),
            )
        )
```

To be extra sure. A bit slower, but probably not meaningfully so.

swrichards (approved):
LGTM! 🎊 

Just two minor nitpicks, but happy to pick those up myself tomorrow so we can get this merged in and you can enjoy some well-deserved time off!

src/open_inwoner/openzaak/services.py
pi-sigma: I like it!

src/open_inwoner/openzaak/services.py
pi-sigma: I don't think this adds anything. In `get_visible_zaken`, the sorting is needed because the futures can be completed in any order. In `fully_resolve_zaken`, on the other hand, the `future_to_zaak` dict is only used to determine which future belongs to which zaak; the output list is constructed from the original zaken list, so the order is preserved by construction. I'd be surprised to see sorting here because I'd assume there must be something that destroys the order of the input list. I've amended the method docstring.

swrichards (approved):

src/open_inwoner/openzaak/services.py
swrichards: Fair enough.

PR #2526, opened 2026-05-27
feat: [#2507] Make case detail metric labels configurable via OpenZaa…

### LINKED ISSUE #2507: PP-411: Configureerbare labels bij zaak besluit/einddatum
Configuratie optie met drie teksten:

- Label voor “Zaak ingediend: {datum}”
- Label voor zaak afgerond
- Label voor zaak nog niet afgerond

swrichards:
…kConfig

Add four new fields to OpenZaakConfig (zaak_identificatie_label, zaak_start_date_label, zaak_end_date_label, zaak_expected_end_date_label) so administrators can customise the labels shown in the case detail dashboard without code changes.

The view now reads all four labels from the config object instead of hardcoded translation strings. A new "Zaakdetail dashboard labels" fieldset is added to the admin. Defaults preserve the existing Dutch strings so there is no visible change for existing deployments.

Closes: #2507


src/open_inwoner/openzaak/models.py
```diff
@@ -470,6 +470,41 @@ def form_services(self):
         help_text=_("Give users the option to filter zaken by status"),
     )
 
+    # Case detail dashboard metric labels
+    zaak_identificatie_label = models.CharField(
+        verbose_name=_("Label: zaaknummer"),
+        max_length=100,
+        default="Zaaknummer:",
```
pi-sigma: Translate
```suggestion
        default=_("Zaaknummer:"),
```
though the colon should not be part of the label

src/open_inwoner/openzaak/models.py
```diff
@@ -470,6 +470,41 @@ def form_services(self):
...
+        help_text=_(
+            "Label shown next to the case identification number on the case detail page."
+        ),
+    )
+    zaak_start_date_label = models.CharField(
+        verbose_name=_("Label: zaak ingediend op"),
+        max_length=100,
+        default="Zaak ingediend op:",
```
pi-sigma: Translate

pi-sigma (changes requested):
- Defaults for labels should be translated
- I found some places which still have hard-coded labels, e.g. `cms/plugins/views.py` has `return _("Zaaknummer: %(identification)s") % {"identification": identification}`. Also, the notifications templates (e.g. `case_document_notification.html` has 'Zaaknummer'). Double-check if these should be amended.

swrichards:
> * Defaults for labels should be translated
> * I found some places which still have hard-coded labels, e.g. `cms/plugins/views.py` has `return _("Zaaknummer: %(identification)s") % {"identification": identification}`. Also, the notifications templates (e.g. `case_document_notification.html` has 'Zaaknummer'). Double-check if these should be amended.

Very good point, bit too much context switching last week. Added the various other call sites.

pi-sigma (approved):

PR #2559, opened 2026-06-02
fix: [#2520] show non 200 response error on case detail page

### LINKED ISSUE #2520: htmx logic on case detail page does not display a user-facing error on non 200 responses
The HTMX loader for the case detail page does not show any user-facing errors if the case cannot be fetched:

Image

Needs a basic "We couldn't fetch data, please try again later" type of message.

stefrado:
## Summary

htmx logic on case detail page did not display a user-facing error on non 200 res ponses

## Issue References

- Github Issue: #2520

### DIFF
--- a/src/open_inwoner/templates/pages/cases/status_outer.html
+++ b/src/open_inwoner/templates/pages/cases/status_outer.html
@@ -7,6 +7,8 @@
 {% endblock notifications %}
 
 {% block content %}
+    <div class="case-detail" id="any-error" data-error-message="{% trans 'We kunnen op dit moment de gegevens niet ophalen, probeer het later opnieuw.' %}"></div>
+
     <div class="case-detail" id="cases-detail-content">
         <oip-loading-spinner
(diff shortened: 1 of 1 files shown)


src/open_inwoner/templates/pages/cases/status_outer.html
```diff
@@ -7,6 +7,8 @@
 {% endblock notifications %}
 
 {% block content %}
+    <div class="case-detail" id="any-error" data-error-message="{% trans 'We kunnen op dit moment de zaak niet ophalen, probeer het later opnieuw.' %}"></div>
```
swrichards: Let's try and avoid zaak for the moment:

```suggestion
    <div class="case-detail" id="any-error" data-error-message="{% trans 'We kunnen op dit moment de gegevens niet ophalen, probeer het later opnieuw.' %}"></div>
```

swrichards (changes requested):

swrichards (approved):

pi-sigma (approved):

PR #2564, opened 2026-06-02
chore: [#2563] update vitest to 4.1.8 and storybook to 10.4.1

### LINKED ISSUE #2563: Dependabot warnings from 02-06-2026
Today its only vitest: https://github.com/vitest-dev/vitest/security/advisories/GHSA-g8mr-85jm-7xhm

stefrado:
## Summary

update vitest to 4.1.8 and storybook to 10.4.1

## Issue References

- Github Issue: #2563 

## Additional info

For this PR we changed the `min-release-age` to `1`, but this should be reverted back to `2`.


.npmrc
```diff
@@ -1 +1 @@
-min-release-age=2
+min-release-age=1
```
swrichards: What happens if we set this back to `2` after the bump? Ideally, like here, we'd want to be able to override for a single package, I see this more as a protection against bulk-upgrades.

stefrado: All CI's of today will probably yell that some packages can't be installed. 

`.npmrc` does not support per-package `min-release-age` syntax (e.g. `min-release-age[vitest]=1` doesn't work). This is the cleanest workaround, but requires a second commit which would reset the `min-release-age`.

Another solution is to merge this PR tomorrow, we are not really affected by this CVE since we do not expose vitest on the netwok

swrichards (approved):

PR #2566, opened 2026-06-02
[#2548] Use maykin-common to document environment variables

### LINKED ISSUE #2548: Document environment variables using maykin-common helper
Replace our legacy `config` helper with that in https://maykin-django-common.readthedocs.io/en/latest/env_docs_helpers.html -- and add documentation to the env vars and include the directive in our docs.

pi-sigma:
## Summary

- updated dependencies:
  - bumped `django-health-check` to 4.4.1
  - bumped `django-maykin-common` to 0.19.0
- legacy `config()` utility replaced with utility from `maykin-common`
- use `maykin-common` to document environment variables

## Issue References

Closes #2548 

----
📚 Documentation preview 📚: https://open-inwoner--2566.org.readthedocs.build/en/2566/


src/open_inwoner/conf/base.py
```diff
@@ -36,25 +34,83 @@
...
+        ),
+        group="Security",
+    ),
+)
 
 # = domains we're running on
-ALLOWED_HOSTS = config("ALLOWED_HOSTS", default="", split=True)
+ALLOWED_HOSTS = config(
+    "ALLOWED_HOSTS",
+    default=[],
```
swrichards, line 91: Are mutable defaults an issue here (e.g. should it be `default=list`)? I suppose not given it's not a function definition.

swrichards (approved):

src/open_inwoner/conf/base.py
pi-sigma: Not an issue, only for function definitions

PR #2570, opened 2026-06-02
fix: [#2569] guard against ZGW API errors in per-zaaktype related-typ…

### LINKED ISSUE #2569: Missing ZGW exception handling in catalogus sync code
Some ZGW calls on the catalogus syncing code were missing a try/except, meaning that any server error would bring down the whole syncing code and break the import. This is blocking for users and should be backported to 2.3.0.

swrichards:
…e imports

A failing fetch_zaaktypes_no_cache call inside get_api_zaaktypen_for_saved_ztc would propagate unhandled through import_all, aborting the entire import run.

Add try/except around the get_api_zaaktypen_for_saved_ztc call in each of the three per-zaaktype import methods (informatieobjecttype, statustype, resultaattype) so an API error for one zaaktype records an ExcludedObject and returns cleanly instead of bubbling up.

Cover each branch with a test that mocks the zaaktypen endpoint to raise a ConnectionError and asserts the method returns a result with one excluded entry.

Closes: #2569

### DIFF
--- a/src/open_inwoner/openzaak/tests/test_zgw_imports.py
+++ b/src/open_inwoner/openzaak/tests/test_zgw_imports.py
@@ -1072,6 +1072,46 @@ class ZGWImportTest(ClearCachesMixin, TestCase):
         self.assertIn("Failed to fetch informatieobjecttype", excluded.error_message)
         self.assertEqual(excluded.extra_context["zaaktype_identificatie"], "ZAAK-001")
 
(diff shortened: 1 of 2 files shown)


pi-sigma (changes requested):
The `ExcludedObject` entries in the inner loop of `zgw_imports.py` (ll. 891 ff.) have `extra_context={"zaaktype_identificatie": ztc.identificatie}`. The new entries don't have this. Should be added for consistency.

PR #2571, opened 2026-06-03
[#2476] Make ZGW connection params model configurable

### LINKED ISSUE #2476: Make ZGW connection params model configurable
Various ZGW connection params are env var configurable only, for no good reason. Let's add them the relevant models (per API group or global Openzaak config, depending on the settings):

- `ZGW_CASE_LIST_NUM_WORKERS` (Global config, depends on system resources)
- `ZGW_CASE_LIST_FETCH_TIMEOUT` (Per group, depends partly on responsiveness of backend)
- `CACHE_ZGW_CATALOGI_TIMEOUT` (Per group, nullable to disable cache)
- `CACHE_ZGW_ZAKEN_TIMEOUT` (Per group, nullable to disable cache)

Add a data migration to set the defaults from the current django setting (which takes it from the environment), and remove the Django settings in a subsequent commit.

pi-sigma:
## Summary

`ZGW_CASE_LIST_NUM_WORKERS`, `ZGW_CASE_LIST_FETCH_TIMEOUT`, `CACHE_ZGW_ZAKEN_TIMEOUT`, and `CACHE_ZGW_CATALOGI_TIMEOUT` were only configurable via environment variables. This requires a restart to change and cannot be tuned per API group.

The `num_workers` and `fetch_timeout settings` move to `OpenZaakConfig` (global, because the case-list pipeline runs all API groups under a single concurrent timeout budget). The two cache timeouts move to `ZGWApiGroupConfig` (per group, so each backend can be tuned independently).

## Issue References

Closes #2476


src/open_inwoner/openzaak/services.py
```diff
@@ -162,30 +159,34 @@ class Timeouts(TypedDict):
...
+        logger.debug("Configured ZGWService", max_workers=self._max_workers)
+
+    @staticmethod
+    def _stage_timeouts(config: OpenZaakConfig) -> Timeouts:
+        """Split the case-list time budget across pipeline stages.
+
+        The total budget is a global setting because all groups participate in every
+        stage under a single concurrent timeout. The fractions must sum to 1.
+        """
+        t = config.case_list_fetch_timeout
```
swrichards, line 185: Could be set in `__init__`, making this a `@property`

src/open_inwoner/utils/decorators.py
```diff
@@ -65,7 +65,7 @@ def cache(
     key: str,
     alias: str = "default",
     *,
-    timeout: int = _DEFAULT_CACHE_TIMEOUT,
+    timeout: int | Callable[..., int] = _DEFAULT_CACHE_TIMEOUT,
```
swrichards, line 68: Could do with a new test (there's a file for cache tests) just to validate that the callable is properly honoured.

src/open_inwoner/utils/decorators.py
```diff
@@ -101,6 +102,8 @@ def decorator(func: Callable[..., RT]) -> Callable[..., RT]:
 
         @wraps(func)
         def wrapped(*args, **kwargs) -> RT:
+            resolved_timeout = timeout(args[0]) if callable(timeout) else timeout
```
swrichards: Need to guard against `args[0]` raising an `IndexError`, which would happen on bare methods as you point out in the docstring. I think the behavior we want here is to still call `timeout` but with no arguments.


```suggestion
            match timeout:
              case int():
                  resolved_timeout = timeout
              case _ if callable(timeout):
                  resolved_timeout = timeout(*args[:1])
```

src/open_inwoner/openzaak/admin.py
swrichards, line 1: Create  a todo on the ticket to remind us to update the documentation for this.

src/open_inwoner/utils/decorators.py
swrichards: Also for the `timeout=0` behavior (does not cache).

swrichards: Also for the `timeout=0` behavior (does not cache).

swrichards (changes requested):

src/open_inwoner/openzaak/services.py
pi-sigma: I was tempted, but on second thought I found it more desirable to keep the service stateless. `django-solo` caches the singleton, and the pipeline methods guard against additional DB hits. This makes several `OpenZaakConfig.get_solo()` calls rather cheap.

src/open_inwoner/utils/tests/test_cache.py
```diff
@@ -272,6 +272,53 @@ def func(a: str, b: str | None = None, *, c: str | None = None):
         self.assertEqual(result4, "foo:bar:None")
         self.assertEqual(result5, "foo:bar:baz")
 
+    @freezegun.freeze_time("2024-05-31 12:00:00", as_kwarg="frozen_time")
+    def test_callable_timeout_is_resolved_per_instance(self, frozen_time):
+        m = mock.Mock(side_effect=lambda x: x)
+
+        class TestClass:
```
swrichards, line 279: Can we also have a test with the callable on a plain function (which should receive no arguments)?

swrichards (changes requested):

swrichards (approved):

PR #2572, opened 2026-06-03
[#2552] Bump log-outgoing-requests to 0.9.1

### LINKED ISSUE #2552: Bump django-log-outgoing-requests to latest

pi-sigma:
## Issue References

Closes #2552


src/open_inwoner/utils/logging.py
swrichards: @pi-sigma drive-by comment which might be useful: we should also replace our home-grown logger here with the [library version](https://django-log-outgoing-requests.readthedocs.io/en/stable/structlog.html).

swrichards (approved):

PR #2394, opened 2026-04-07
[#2203] Replace all static utrecht-paragraphs with --nl web components

jiromaykin:
## Summary

To keep the number of file-changes limited, this PR only changes all of the static utrecht-paragraphs with the new `--nl` candidate web component.
So not yet making changes for the muted/compact/small variants (!)

### Notable things

- Some Django tests won't be able to test rendered javascript and can only test for the existence of the surrounding `<nl-paragraph>` web component ,so won't be able to detect the `<p class="nl-paragraph">` element inside.
- In `src/open_inwoner/utils/html.py` we need to change the classes for WYSIWYG elements from rich text editors, those are not web components, just P-tags that need a class.
- most changes in this PR are an easy find-all-and-replace-all change where all utrecht paragraphs are replaced with the NL web component, so not yet changing all the paragraphs with modifiers (will be a separate PR because it required tokens changing)
- no changes in the Paragraph.tsx here, but once the Shadow DOM PR #2310 has been merged I will add a new issue for updating the paragraph component accordingly for Storybook testing etc.

## Issue References

No changes in design tokens package needed for this PR.

## Checklist

- [x] CHANGELOG.rst updated

PR #2576, opened 2026-06-04
fix: [#2573] missing prosemirror filter on some paragraphs

### LINKED ISSUE #2573: Bug: ProseMirror P/NLDS class tag for Admin Configuration fields not applied
### Describe the bug

Pages that contain rich text and are configured in the Admin, do not get the correct NLDS class.
This is a bug in the new paragraphs that were built in PP-324/#2203

### To Reproduce

In front-end this can be seen in the warning banner and Search 'no results' fields and more

Image

### Expected behavior

Any ProseMirror paragraph needs to get the `.nl-paragraph` class.

jiromaykin:
add prosemirror filter so nlds css class is applied to paragraphs

some pages were missing correct render_tags and cms_tags

## Issue References

Additional bugfix for issue #2203 and #2394 that were closed. 

## Checklist

- [ ]

### DIFF
--- a/src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
+++ b/src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
@@ -10,7 +10,7 @@
                 <h4>{% trans "Question:" %}</h4>
                 <nl-paragraph>{{ step.question }}</nl-paragraph>
                 <h4>{% trans "Answer:" %}</h4>
-                <nl-paragraph>{{ step.answer }}</nl-paragraph>
+                <p>{{ step.answer }}</p>
             </li>
             {% endfor %}
         </ol>
@@ -18,14 +18,14 @@
 
     <div class="qa_results">
         <h4 class="utrecht-heading-4">{{ last_step.question }}</h4>
-        <nl-paragraph>{{ last_step.content.html|safe}}</nl-paragraph>
+        <p>{{ last_step.content.html|safe}}</p>
         {% if related_products.exists %}
             <h4>{% trans "Related products:" %}</h4>
             <ol>
                 {% for product in related_products %}
                 <li>
                     <h4>{{ product.name }}</h4>
-                    <nl-paragraph>{{ product.summary }}</nl-paragraph>
+                    <p>{{ product.summary }}</p>
                 </li>
                 {% endfor %}
             </ol>
--- a/src/open_inwoner/accounts/templates/registration/login.html
+++ b/src/open_inwoner/accounts/templates/registration/login.html
@@ -1,5 +1,5 @@
 {% extends 'master.html' %}
-{% load i18n static utils grid_tags card_tags form_tags link_tags button_tags icon_tags solo_tags cms_tags djangocms_alias_tags %}
+{% load i18n static utils grid_tags card_tags form_tags link_tags button_tags icon_tags solo_tags render_tags cms_tags djangocms_alias_tags %}
 
 
 {% block header_image %}
@@ -9,7 +9,7 @@
 {% block content %}
     <h1 class="utrecht-heading-1">{% trans 'Welkom' %}</h1>
     {% if login_text.html %}
-        <div class="wysiwyg">{{ login_text.html|safe }}</div>
+        <div class="wysiwyg">{{ login_text.html|prosemirror_content|safe }}</div>
     {% endif %}
 
     {# Tab panels Start #}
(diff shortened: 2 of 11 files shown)


jiromaykin:
@stefrado @pi-sigma We need to check much more rigorously whether the rich text paragraphs are _actually_ rendered with the proper class. I think these are the ones that were missing, but please let me know if there are more!
Also: I think we no not need any CSS classes for the PDF export files?
(the ones in src/open_inwoner/templates/export/... ?
Or: at least those would never work with web-components I assume. 

This PR can be considered as part of the larger Paragraph modification issue.

src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
```diff
@@ -3,29 +3,29 @@
...
+        <h4>{{ last_step.question }}</h4>
+        <p>{{ last_step.content.html|safe}}</p>
         {% if related_products.exists %}
             <h4>{% trans "Related products:" %}</h4>
             <ol>
                 {% for product in related_products %}
                 <li>
                     <h4>{{ product.name }}</h4>
-                    <nl-paragraph>{{ product.summary }}</nl-paragraph>
+                    <p>{{ product.summary }}</p>
```
jiromaykin, line 28: I am assuming adding a web-component to a PDF makes no sense -  i am not sure about the other classes in this same file.

src/open_inwoner/templates/pages/ssd/reports_base.html
```diff
@@ -12,7 +12,7 @@ <h1 class="utrecht-heading-1" id="title">
                 {{ page_title }}
             </h1>
             <p>
-                {% blocktrans with mijn_uitkeringen_text=client.config.mijn_uitkeringen_text.html|safe %}
+                {% blocktrans with mijn_uitkeringen_text=client.config.mijn_uitkeringen_text.html|prosemirror_content|safe %}
```
pi-sigma: `SSDConfig.mijn_uitkeringen_text` is a plain text field.  `config.mijn_uitkeringen_text.html` was already wrong (the text field has no `.html` attribute and will return "" in the template), the filter adds nothing. The fix is to remove the `.html` access, but this should be done in a separate commit. For this PR, just skip this one.

pi-sigma (changes requested):
These places need the filter as well (`ExtendedCMSLink.name` is a Prosemirror field): 
- [external-links.html](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/templates/cms/plugins/links/external-links.html#L13) 
- [link.html](https://github.com/maykinmedia/open-inwoner/blob/develop/src/open_inwoner/templates/cms/plugins/links/link.html#L3)

src/open_inwoner/templates/pages/ssd/reports_base.html
jiromaykin: Will do. That's why we could use a more rigorous check on this.

src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
jiromaykin: What do you think about styling for PDF's here @pi-sigma  ?

src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
```diff
@@ -3,29 +3,29 @@
 
 {% block content %}
     <div class="qa">
-        <h1 class="utrecht-heading-1">{% trans "Zelfdiagnose results" %}</h1>
+        <h1>{% trans "Zelfdiagnose results" %}</h1>
```
pi-sigma: The text is much larger now without the class. Is that what we want?

src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
pi-sigma: Thanks for reminding me of this one. You're right, the `nl-paragaph` won't work here. I'm not sure if we need styling with classes for the paragraphs

src/open_inwoner/templates/export/questionnaire/questionnaire_export.html
jiromaykin: I'll check, so I guess it did take styling.

stefrado (approved):

pi-sigma (changes requested):
Please sqash the refactor commit, there's no reason for it to stand on its own.

jiromaykin:
> Please sqash the refactor commit, there's no reason for it to stand on its own.

@pi-sigma  I know - i was just showing what was changed after changing it for code review - i will always squash these.

pi-sigma (approved):

PR #2590, opened 2026-06-08
fix: [#2589] move ELASTICSEARCH_DSL_AUTOSYNC/AUTO_REFRESH to base set…

### LINKED ISSUE #2589: Autosync in django-elasticsearch-dsl breaks PDC model saves with missing elasticsearch
In instances where elasticsearch is not/misconfigured, trying to add products will trigger a 500 because django-elasticsearch-dsl attempts to auto-index on a model save signal, bringing down the whole page. We should just disable this behavior and rely on the Celery task for this.

swrichards:
…tings

Consolidates these settings into base.py as env-var-configurable flags (defaulting to False), removing the duplicates from dev.py and ci.py. Autosync is off by default to avoid 500s when Elasticsearch is absent; it can be enabled via env var when Elasticsearch is guaranteed available.

Closes: #2589


src/open_inwoner/conf/base.py
```diff
@@ -1027,6 +1027,12 @@
 ES_MAX_SIZE = 10000
 ES_SUGGEST_SIZE = 5
 
+# Enable to trigger indexing on each model save. Only enable if Elasticsearch is
+# guaranteed to be available and correctly configured, otherwise this can cause 500s on
+# admin operations. By default we rely on the Celery index rebuild task instead.
+ELASTICSEARCH_DSL_AUTO_REFRESH = config("ELASTICSEARCH_DSL_AUTO_REFRESH", default=False)
```
pi-sigma: As discussed, `default=True` preserves existing behavior for gemeentes who do use Elasticsearch. Those who want this feature off should turn the knob.

pi-sigma (changes requested):

pi-sigma (approved):

PR #2595, opened 2026-06-09
[WIP] CMS pages through setup configuration

swrichards:


docker-compose.yml
```diff
@@ -156,8 +161,21 @@ services:
...
       - ./docker/setup_configuration:/app/setup_configuration
+    # Override depends_on to prevent a circular dependency: web-init inherits
+    # *web-service which would otherwise include web-init itself.
+    depends_on:
+      db:
+        condition: service_started
+      redis:
+        condition: service_started
+      prepare-volumes:
+        condition: service_completed_successfully
```
swrichards, line 178: @pi-sigma would be good to extract this into a separate fix

docker/setup_configuration/data.yaml
```diff
@@ -212,6 +212,62 @@ sites_config:
     - domain: test-site
       name: The Django Sites default
 
+# CMS pages
+#
+# Pages are created once: re-running setup_configuration does not update pages
+# that already exist, so edits made here only affect fresh installations. Change
+# existing pages in the CMS admin instead.
+cms_pages_config_enable: true
```
swrichards, line 238: @pi-sigma is it feasible to also enable (some) plugins, so we get a landing page that has the Mijn zaken / Taken / etc. plugins pre-configured?

src/open_inwoner/conf/app/setup_configuration.py
```diff
@@ -9,1931 +7,3 @@
     "django_setup_configuration.contrib.sites.steps.SitesConfigurationStep",
     "open_inwoner.configurations.bootstrap.cms.CMSPagesConfigurationStep",
 ]
```
swrichards, line 14: This would technically be a breaking commit (removing functionality and public-facing env vars).

src/open_inwoner/configurations/tests/bootstrap/test_setup_site_config.py
swrichards, line 1: Can we add a specific test for `warning_banner_text` to ensure we handle django prosemirror fields properly?

src/open_inwoner/configurations/bootstrap/saml.py
swrichards, line 1: This commit (`refactor: [https://github.com/maykinmedia/open-inwoner/issues/2594] overwrite SAML settings on every run`) can be folded into f54277d

src/open_inwoner/configurations/bootstrap/zgw.py
swrichards, line 1: This commit (`refactor: [https://github.com/maykinmedia/open-inwoner/issues/2594] overwrite existing ZGWApiGroupConfig fields on re-run`) arguably also a breaking change we should note separately.

swrichards (commented):
LGTM, just some notes that the refactor commitments in fact change public APIs (the env vars), so we should mark them as such (even though those env vars have been silently broken for some time).

src/open_inwoner/conf/app/setup_configuration.py
pi-sigma: Added a note in the ticket

src/open_inwoner/configurations/bootstrap/zgw.py
pi-sigma: Added to the ticket

pi-sigma (approved):

PR #2603, opened 2026-06-10
[#1900] Small fixes to webhook admin screens

### LINKED ISSUE #1900: Improve admin observability of webhook handling

swrichards:
Closes: #1900

### DIFF
--- a/src/notifications/models.py
+++ b/src/notifications/models.py
@@ -1,7 +1,7 @@
 from django.contrib.postgres.fields import ArrayField
 from django.core.exceptions import ImproperlyConfigured
 from django.db import models
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext, gettext_lazy as _
 
 from ape_pie import APIClient
 from solo.models import SingletonModel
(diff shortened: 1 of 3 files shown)


src/notifications/models.py
```diff
@@ -52,8 +52,8 @@ def __str__(self):
             if self.notifications_api_service
             else _("no service configured")
         )
-        return _("Notifications API configuration ({api_root})").format(
-            api_root=api_root
+        return str(
```
pi-sigma: Why force evaluation of the lazy-translated string?

swrichards: It was the lazy gettext causing the issue, used the eager version in `__str__`

PR #2608, opened 2026-06-15
[#2604] Support multiple emails and phonenumbers in User datamodel

### LINKED ISSUE #2604: PP-427: Werkwijze OpenKlant verbeteren/uitbreiden
See #2593 for additional context.

swrichards:
Closes: #2604


src/open_inwoner/openklant/services.py
```diff
@@ -102,18 +102,8 @@ def _normalize_email(email: str) -> str:
 
 
 def _resolve_phonenumber_update(user: "User", update_data: dict) -> None:
```
pi-sigma, line 110: Can we not just delete it? There is one call in the `eSuiteKlantenService`, but that can also be removed if it's just a no-op.

src/open_inwoner/accounts/models.py
```diff
@@ -978,9 +980,71 @@ def save(
...
+        email_changed = (
+            sync_email and not is_new and self.email != self.__original_email
         )
+        phone_changed = (
+            sync_phone
+            and not is_new
+            and self.phonenumber != self.__original_phonenumber
+        )
+
+        super().save(force_insert, force_update, using, update_fields=update_fields)
```
pi-sigma, line 996: Wrap this and the following address writes `with transaction.atomic():`?

src/open_inwoner/accounts/models.py
```diff
@@ -978,9 +980,71 @@ def save(
...
+                        value=self.__original_email,
+                    ).update(value=self.email)
+
+            if phone_changed:
+                if self.__original_phonenumber:
+                    DigitalAddress.objects.filter(
+                        user=self,
+                        type=DigitalAddressType.phone,
+                        value=self.__original_phonenumber,
+                    ).update(value=self.phonenumber)
```
pi-sigma, line 1037: `""` is valid input for phone number fields (the validator only runs on non-empty input). If a user 'deletes' their number by setting it to `""`, this will overwrite the existing entry with an invalid number. Should be `.delete()` instead

src/open_inwoner/openklant/models.py
```diff
@@ -357,3 +357,22 @@ def has_api_service_configured(
                 return getattr(config, "service", None) is not None
             case _:
                 return False
+
+
+class DigitaalAdresKlant2Mapping(models.Model):
```
pi-sigma, line 362: `DigitaalAdresOpenKlantMapping`

Also below: I think we should stop using "OK2", "OpenKlant2" etc and just use "OpenKlant"

src/open_inwoner/accounts/tests/factories.py
```diff
@@ -6,7 +6,19 @@
...
+
+
+class DigitalAddressFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = "accounts.DigitalAddress"
+
+    user = factory.SubFactory("open_inwoner.accounts.tests.factories.UserFactory")
+    type = DigitalAddressType.email
+    value = factory.LazyAttribute(
+        lambda o: "%s%d@example.com" % (o.type, random.randint(0, 10000000))
```
pi-sigma, line 19: For `type=phone` this creates `phone1234@example.com`. We should use valid emails and numbers based on type.

Simple approach with random:

```
lambda o: (
          f"06{random.randint(10000000, 99999999)}"
          if o.type == DigitalAddressType.phone
          else f"{random.randint(0, 10000000)}@example.com"
      )
```
Or you can use a `factory.Faker()` with `numerify` (see https://faker.readthedocs.io/en/master/providers/baseprovider.html). I think the latter might be nicer (we use it elsewhere)

src/open_inwoner/openklant/services.py
```diff
@@ -1388,35 +1388,109 @@ def update_user_from_partij(self, partij_uuid: str, user: User):
     def update_partij_from_user_data(
         self,
         partij_uuid: str,
-        update_data: PartijUpdateData,
+        user: User,
     ) -> list[str]:
-        updated_fields: list[str] = []
-        adressen = self.retrieve_digitale_addressen_for_partij(partij_uuid)
+        from open_inwoner.accounts.choices import DigitalAddressType
```
pi-sigma, line 1393: Why lazy import?

src/open_inwoner/openklant/services.py
```diff
@@ -1351,42 +1355,109 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
+                local_addr, _ = user.digital_addresses.get_or_create(
+                    type=addr_type,
+                    value=adres_value,
+                    defaults={"login_type": user.login_type},
+                )
+                DigitaalAdresKlant2Mapping.objects.update_or_create(
+                    digital_address=local_addr,
+                    defaults={"ok2_uuid": remote_uuid},
+                )
+                existing_mappings[remote_uuid] = local_addr
```
pi-sigma, line 1406: Shouldn't we store the `DigitaalAdresKlant2Mapping` here instead of the `local_addr`?

src/open_inwoner/openklant/services.py
```diff
@@ -1388,35 +1388,109 @@ def update_user_from_partij(self, partij_uuid: str, user: User):
...
-            digital_adres, created = self.get_or_create_digitaal_adres_for_partij(
-                partij_uuid=partij_uuid,
-                soort_adres="telefoonnummer",
-                adres=phonenumber,
-                is_standaard_adres=True,
-                adressen=adressen,
+        updated_fields: list[str] = []
+        remote_adressen: list[DigitaalAdres] | None = None  # lazy-loaded for backfill
+
+        for address in user.digital_addresses.all():
```
pi-sigma, line 1399: `prefetch_related("openklant2_mapping")` to avoid hitting the DB when querying below

src/open_inwoner/openklant/services.py
```diff
@@ -320,7 +320,12 @@ def update_user_from_klant(self, klant: Klant, user: User):
         if klant.telefoonnummer and klant.telefoonnummer != user.phonenumber:
             update_data["phonenumber"] = klant.telefoonnummer
 
-        # TODO [#2604] inbound eSuite alternative phone → DigitalAddress
+        if klant.telefoonnummer_alternatief:
+            user.digital_addresses.update_or_create(
```
pi-sigma, line 324: We used to prevent creating alternative phone numbers identical to the primary. Is this still needed?

src/open_inwoner/openklant/services.py
```diff
@@ -1362,81 +1366,210 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
 
-        if update_data:
-            _resolve_phonenumber_update(user, update_data)
-            for attr, value in update_data.items():
+        # Delete local addresses whose mapping UUID is no longer in the remote
+        stale_mappings = DigitaalAdresKlant2Mapping.objects.filter(
+            digital_address__user=user
+        ).exclude(ok2_uuid__in=remote_uuids)
+        for mapping in stale_mappings:
+            mapping.digital_address.delete()
```
pi-sigma, line 1447: This seems risky, but perhaps that's because I don't fully grok the ways in which `DigitaalAdresKlant2Mapping` instances are created, updated, and tracked. Is it possible that we delete the `DigitalAddress` for a primary phone number or email because its UUID is removed from the `remote_uuids` by some other service?

pi-sigma (changes requested):

swrichards:
Deprecated in favor of #2618

PR #2610, opened 2026-06-15
feat: [#2611] add DigitalAddress model and User.preferred_address

### LINKED ISSUE #2604: PP-427: Werkwijze OpenKlant verbeteren/uitbreiden

### LINKED ISSUE #2611: Add DigitalAddress model

swrichards:
Closes: #2611


src/open_inwoner/accounts/tests/test_digital_address.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertIsNotNone(addr.pk)
+
+    def test_unique_constraint_allows_same_value_different_type(self):
+        existing = DigitalAddressFactory(
+            type=DigitalAddressType.email, value="test@example.com"
+        )
+        addr = DigitalAddressFactory(
+            user=existing.user,
+            type=DigitalAddressType.phone,
+            value="test@example.com",
```
pi-sigma:
```suggestion
            value="0612345678",
```
Test passes because it only checks the constraints, but it looks weird

src/open_inwoner/accounts/models.py
```diff
@@ -251,6 +253,83 @@ def generate_uuid_image_name(instance, filename):
...
+                name="unique_email_digital_address_for_regular_users",
+            ),
+            UniqueConstraint(
+                fields=["user", "type", "value"],
+                name="unique_digital_address_per_user_type_value",
+            ),
+            UniqueConstraint(
+                fields=["user", "type"],
+                condition=Q(is_standard_for_type=True),
+                name="unique_standard_digital_address_per_user_type",
```
pi-sigma, line 305: The constraint means that if we want to change which address is standard, we need to unset the old one first. There's no logic in the `clean()` . This will be added later?

pi-sigma (approved):

src/open_inwoner/accounts/models.py
swrichards: It will, next PR

PR #2614, opened 2026-06-15
[#2612] Sync digital addresses with OpenKlant

### LINKED ISSUE #2612: Sync digital addresses with OpenKlant

pi-sigma:
## Summary

- [x] Implement outbound sync (OIP -> OpenKlant): mapped addresses are `PATCH`ed via their stored UUID; unmapped addresses are backfilled by matching against the partij's existing digital addresses or created new, then mapped. A `404` on `PATCH` drops the stale mapping and falls back to backfill. `User.preferred_address` syncs to `Partij.voorkeursDigitaalAdres` when mapped.
- [x] Implement inbound sync (OpenKlant -> OIP): Rewrites `update_user_from_partij `to sync `DigitalAddress` rows rather than flat fields. OpenKlant is treated as the single source of truth: addresses present remotely are created or updated locally (tracked via `DigitaalAdresOpenKlantMapping` UUID), addresses no longer present remotely are deleted. Exception: if OpenKlant has no standard email, we write our standard email back. Key behaviour:
  - UUID known -> update local value in place
  - UUID unknown -> create `DigitalAddress` + mapping
  - UUID gone -> delete `DigitalAddress` (CASCADE removes the mapping)
  - `isStandaardAdres=True` -> `is_standard_for_type=True` locally (per type); when
    remote sends no standard for a type the local flag is left unchanged
  - `voorkeursDigitaalAdres` -> `user.preferred_address`
  - Standard email/phone synced to `user.email` / `user.phonenumber` flat fields;
    email skipped if already owned by another OIP user

## Issue References

Closes #2612


src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
-                    partij_uuid=partij_uuid,
-                )
+                if (
+                    digital_address_type == DigitalAddressType.email
+                    and User.objects.filter(email__iexact=value)
+                    .exclude(pk=user.pk)
+                    .exists()
+                ):
+                    # Another user already owns this email address; don't adopt it.
+                    continue
```
swrichards: Let's have some logging here, and I think an OTEL metric would be good (we haven't been keeping up on those, but it's worth getting back into the habit). The metric would check

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
+                    user_update_fields.add("preferred_address")
+            elif user.preferred_address_id is not None:
+                user.preferred_address = None
+                user_update_fields.add("preferred_address")
+
+            if user_update_fields:
+                user.save(update_fields=user_update_fields)
+
+            change_summary = []
+            if created_count:
```
swrichards: The summary is nice, to the point where I would like an `InboundOpenklantSyncResult` or some such (a bit like the ZGW import), which we can use to fill the system action logger, but would also give us options latter to e.g. return that from a Celery task, if we decide to make this async.

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -3,232 +3,367 @@
...
-        ) as logs:
-            service.update_user_from_partij("partij-uuid", user)
+        mock_client.digitaal_adres.create.assert_called_once()
+        mapping = local_email_da.openklant_mapping
+        self.assertEqual(str(mapping.ok_uuid), new_remote_uuid)
 
-        self.assertTrue(
-            any("More than two phone numbers found" in log for log in logs.output),
-            "Should log warning when more than 2 phone numbers exist",
+    def test_push_back_standard_email_when_remote_has_no_standard_email(
```
swrichards: Shouldn't this test also assert that the client methods are actually called?

src/open_inwoner/openklant/tests/test_openklant2_service.py
```diff
@@ -1178,161 +1313,240 @@ def _list_response(self, results):
...
+        user = DigidUserFactory()
 
         service = OpenKlant2Service(config=self.config)
         result = service.update_partij_from_user_data(
-            partij_uuid=self.PARTIJ_UUID,
-            update_data={},
+            partij_uuid=self.PARTIJ_UUID, user=user
         )
 
         self.assertEqual(result, [])
```
swrichards, line 1400: This is no longer the signature for this function, and not sure the test adds much.

src/open_inwoner/openklant/models.py
```diff
@@ -357,3 +357,22 @@ def has_api_service_configured(
...
+
+
+class DigitaalAdresOpenKlantMapping(models.Model):
+    digital_address = models.OneToOneField(
+        "accounts.DigitalAddress",
+        on_delete=models.CASCADE,
+        related_name="openklant_mapping",
+        verbose_name=_("Digital address"),
+    )
+    ok_uuid = models.UUIDField(
```
swrichards, line 369: Candidate for an index

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
+        with transaction.atomic():
+            local_by_uuid: dict[str, DigitalAddress] = {
+                str(mapping.ok_uuid): mapping.digital_address
+                for mapping in DigitaalAdresOpenKlantMapping.objects.filter(
+                    digital_address__user=user
+                ).select_related("digital_address")
+            }
 
-        adressen = self.retrieve_digitale_addressen_for_partij(partij_uuid)
+            # Phase 1: reconcile local DigitalAddress rows against the remote list.
```
swrichards, line 1403: I think some debug logging here would both be useful when debugging and help reading the individual cases

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
+                    ),
+                    None,
+                ),
+            }
 
-        if email := update_data.get("email"):
-            user.update_email(email)
+            # Phase 2: sync is_standard_for_type from remote isStandaardAdres.
+            # Only acts when remote explicitly designates a standard; if remote
+            # has no standard for a type the local flag is left unchanged.
```
swrichards, line 1490: We could theoretically have deleted our own standard address if it was (1) mapped (2) had the standard annotation remotely and (3) is deleted. This is the "They deleted the standard address" case.

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
...
+            ).first()
+            if local_standard_email:
+                remote_adres, created = self.get_or_create_digitaal_adres_for_partij(
+                    partij_uuid=partij_uuid,
+                    soort_adres="email",
+                    adres=local_standard_email.value,
+                    is_standaard_adres=True,
+                    adressen=remote_adressen,
+                )
+                with transaction.atomic():
```
swrichards: What does the bloc add here?

src/open_inwoner/openklant/services.py
swrichards: Reading Jacco's comment, I don't actually think it's a problem if remote has no standard address. I don't think we have to push our flag (that's also not what Jacco wants).

But: I _do_ think we want the invariant that `user.email|phonenumber` _always_ has at least one matching digital address, which is not addressed here.

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1361,268 @@ def get_or_create_digitaal_adres_for_betrokkene(
             adressen=adressen,
         )
 
-    @transaction.atomic
-    def update_user_from_partij(self, partij_uuid: str, user: User):
-        update_data = {}
+    def update_user_from_partij(self, partij: Partij, user: User) -> None:
```
swrichards: I only realized this now, but perhaps `update_user_from_partij` and `update_partij_from_user` are no longer sensible methods? In reality it might best be a single `sync_user_with_partij`, because with the mapping we now have, we can infer which addresses were created/deleted by remote (we have a mapping and it's disappeared, or we have no mapping, and thus we add it). And for the same reason, because of the mapping, we also know what we have to push from our end. What do you think about combining both in a single sync method?

swrichards: To add to this, I think we're missing the case where _we_ created an extra digital address (beyond `user.email`) that's not mapped to Openklant.

swrichards (changes requested):
A few issues we should probably discuss on monday, but otherwise on the right track, seeing it in code pushes the dge cases into view 👍 .

src/open_inwoner/openklant/models.py
pi-sigma: `ok_uuid` is only used as a key for lookup in the `DigitaalAdresOpenKlantMapping`, it's never used for querying (`.filter(ok_uuid=...`). Index wouldn't do anything (at least atm).

src/open_inwoner/openklant/services.py
pi-sigma: I've added a debug log when we skip emails due to being owned by another user. That's the only case not covered by `system_action`.

src/open_inwoner/openklant/services.py
pi-sigma: Good point about the invariant, I'll add it. Regarding the push-back: do we still want a narrow push-back: if OpenKlant does have addresses but none marked as standard?

src/open_inwoner/openklant/services.py
pi-sigma: Nothing, the single ORM call is already atomic. Removed it.

src/open_inwoner/openklant/services.py
pi-sigma: This makes sense conceptually. `update_user...` already includes some outbound logic with the push-back behavior; a bi-directional diff would be cleaner.

However, it would make things technically more complicated: (1) the two methods are triggered at different points: `update_user...` runs on login, `update_partij...` runs on user edit. Combining the methods could mess up the sync or require additional guards; (2) the methods (esp. `update_user...`) are alrady quite long. I think they're still reasonable (mostly flat and linear, no deep nesting of conditionals), but combining the two would not help to improve readability I think. Also, easier to test the behavior and our policies for syncing with two separate methods.

TLDR; I think it's better to keep them separate.

src/open_inwoner/openklant/tests/test_openklant2_service.py
pi-sigma: It already has `mock_client.digitaal_adres.create.assert_called_once()`. Something missing?

src/open_inwoner/openklant/models.py
swrichards: Indeed, we filter on user mainly, which is already an index because FK. Resolving.

src/open_inwoner/openklant/types.py
```diff
@@ -7,3 +8,20 @@ class PartijUpdateData(TypedDict):
     email: NotRequired[str]
     phonenumber: NotRequired[str]
     phonenumber_alternative: NotRequired[str]
+
+
+@dataclass
+class InboundOpenklantSyncResult:
```
swrichards, line 14: For most of our tests of the service, we could add an additional assert against the expected `InboundOpenklantSyncResult` that is returned.

src/open_inwoner/openklant/services.py
```diff
@@ -1358,104 +1363,337 @@ def get_or_create_digitaal_adres_for_betrokkene(
             adressen=adressen,
         )
 
-    @transaction.atomic
-    def update_user_from_partij(self, partij_uuid: str, user: User):
-        update_data = {}
+    def update_user_from_partij(
+        self, partij: Partij, user: User
```
swrichards, line 1367: Moving from `partij_uuid` to `partij` does couple this slightly more to the API objects.

swrichards (approved):
LGTM, just a minor point about adding the DTO to test asserts on the function signature

PR #2616, opened 2026-06-16
feat: [#2615] sync phonenumber and email with DigitalAddress

swrichards:
Remove phonenumber_alternative from User and replace it with a
non-standard phone DigitalAddress. Keep phonenumber and email as real
DB fields for auth and 2FA performance, but introduce update_email()
and update_phonenumber() model methods that write both the field and
the corresponding standard DigitalAddress atomically.

Add two data migrations: one backfills existing phone numbers into
DigitalAddress, another ensures every user's primary email and phone
have a matching standard DA. Update openklant services to use the new
sync methods.

Closes: #2615

### DIFF
--- a/src/open_inwoner/openklant/services.py
+++ b/src/open_inwoner/openklant/services.py
@@ -16,6 +16,7 @@ from typing import (
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.db import transaction
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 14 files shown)


src/open_inwoner/openklant/services.py
```diff
@@ -1361,6 +1358,7 @@ def get_or_create_digitaal_adres_for_betrokkene(
             adressen=adressen,
         )
 
+    @transaction.atomic
```
pi-sigma, line 1361: Do we want to keep the DB connection open for the entire duration of the HTTP request that fetches addresses from the API?

pi-sigma (approved):
`update_klant_via_openklant` in `profile.py` still loops over `["email", "phonenumber", "phonenumber_alternative"]` comparing `old_user`/`user` values. This is now dead code. Could be removed now or in a follow-up PR that reworks that path.

src/open_inwoner/openklant/services.py
swrichards: Good catch, and probably not -- but my goal here (also re your main comment) is simply to keep a passing CI/sane default), knowing this code will be replace with your syncing code. I'd say, let's get this in and rework it when we do syncing properly.

PR #2621, opened 2026-06-17
[#2613] PP-427: Sync digital addresses with eSuite

### LINKED ISSUE #2613: Sync digital addresses with eSuite

pi-sigma:
## Summary
Replace the old `update_or_create` logic for alternative phone in `update_user_from_klant` with a delete-and-recreate reconcile: compute desired_alternative (empty when eSuite sends none or the same value as primary), then clear all non-standard phone DAs and create a fresh one only when needed. This avoids `MultipleObjectsReturned` on repeated syncs and correctly handles removal.

## Issue References

Closes #2613

### DIFF
--- a/src/open_inwoner/openklant/tests/test_esuite_service.py
+++ b/src/open_inwoner/openklant/tests/test_esuite_service.py
@@ -1,10 +1,15 @@
-from dataclasses import asdict, dataclass
+from dataclasses import asdict
 
 from django.test import TestCase
 
 import requests_mock
 
-from open_inwoner.accounts.tests.factories import DigidUserFactory, UserFactory
+from open_inwoner.accounts.choices import DigitalAddressType
+from open_inwoner.accounts.tests.factories import (
+    DigidUserFactory,
+    DigitalAddressFactory,
+    UserFactory,
+)
 from open_inwoner.openklant.api_models import Klant
 from open_inwoner.openklant.services import eSuiteKlantenService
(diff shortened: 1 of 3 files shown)


src/open_inwoner/openklant/tests/test_esuite_service.py
```diff
@@ -242,6 +247,48 @@ class Klant:
...
+        @dataclass
+        class Klant:
+            bronorganisatie: str
+            klantnummer: str
+            subjectIdentificatie: str
+            url: str
+            emailadres: str
+            telefoonnummer: str
+            telefoonnummer_alternatief: str
+            toestemmingZaakNotificatiesAlleenDigitaal: str
```
swrichards: Why is this inline?

src/open_inwoner/openklant/tests/test_esuite_service.py
```diff
@@ -242,6 +247,48 @@ class Klant:
...
+
+        klant = Klant(
+            bronorganisatie="123456789",
+            klantnummer="12345678",
+            subjectIdentificatie={"inpBsn": "123456789"},
+            url=f"{KLANTEN_ROOT}klant/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
+            emailadres=user.email,
+            telefoonnummer="0611111111",
+            telefoonnummer_alternatief="0622222222",
```
swrichards, line 247: Shouldn't this be empty, or different from the existing alternative phonenumber?

swrichards (approved):

PR #2627, opened 2026-06-17
ci: enforce strict conventional commits, reject fixup/squash commits

swrichards:
Removes fixup!, squash!, and amend! from the commitizen allowed prefixes
and consolidates hook config into pyproject.toml so both local pre-commit
and CI share the same rules. Drops allowed_prefixes entirely since the
rebase-merge workflow means no auto-generated commit messages appear in
PR branches.

Also make sure we check the commits with reference to the PR's head commit, not the head of the merge commit (which has the default git "Merge X into Y" message, which would fail).

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -122,7 +122,7 @@ jobs:
         run: pip install commitizen==4.16.3
       - name: Check commit messages
         run: |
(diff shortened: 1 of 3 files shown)


swrichards:
Failing run: https://github.com/maykinmedia/open-inwoner/actions/runs/27699861845/job/81933020435

pi-sigma (approved):

PR #2629, opened 2026-06-18
feat: [#2628] add digital address formset to profile edit page

### LINKED ISSUE #2628: Amend profile edit form for multiple addresses

swrichards:
Closes: #2628

### DIFF
--- a/src/open_inwoner/accounts/views/profile.py
+++ b/src/open_inwoner/accounts/views/profile.py
@@ -5,6 +5,7 @@ from typing import Any
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.core.exceptions import ImproperlyConfigured
+from django.db import transaction
 from django.http import HttpResponseRedirect
 from django.shortcuts import redirect
 from django.urls import reverse, reverse_lazy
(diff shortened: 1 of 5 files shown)


swrichards:
@pi-sigma this is an intermediate PR so @stefrado can push on with the frontend, and so we have a proper hook for adding in your sync operations. I would say don't look at the syncing tests too closely, they'll be superseded by yours. Let's aim to merge this into the feature branch quickly so you both can rebase on this.

src/open_inwoner/accounts/views/profile.py
```diff
@@ -245,38 +249,117 @@ def crumbs(self):
...
+        if new_alt != old_phonenumber_alternative:
+            changed_data["phonenumber_alternative"] = new_alt
+
+        if not changed_data:
             messages.success(self.request, _("Uw wijzigingen zijn opgeslagen"))
             self.log_profile_modified(user)
             return HttpResponseRedirect(self.get_success_url())
 
-        # write changes to API's; abort saving if write fails
+        # Sync address changes to external APIs.
```
pi-sigma, line 339: The pre-existing invariant that DB and external APIs agree on email/phonenumber is undermined.

In the old `form_valid`, `form.save()` was only called after a successful API write for API-sensitive fields. If the API failed, the function returned early. Now the DA formsets are saved inside `transaction.atomic()` unconditionally, before the API calls. If eSuite or OpenKlant fails, the user sees "Please try again later", but the local data was already saved.

Do we want to switch to "save locally, best-effort sync" or keep the existing behavior?

swrichards: I suspect we'll want to do the latter, I think that makes more sense after all. The problem we tried to address is still there (divergence), and I think it might actually make sense, in the sync branch, to convert the sync to a Celery task with retry logic. But we can discuss.

This code here was really just to keep a functioning branch, I say we discuss this in your followup sync PR.

pi-sigma (approved):

PR #2634, opened 2026-06-23
fix: delete remote digital addresses by UUID

pi-sigma:
### Issue
When a user deletes a digital address via the profile formset, `_save_digital_address_formset` calls `form.instance.delete()`, which cascade-deletes the `DigitaalAdresOpenKlantMapping`. By the time the OpenKlant sync runs, the mapping UUID is gone, so the remote digitaal adres was never deleted from OpenKlant. On the next inbound sync, the remote address would be re-adopted locally, silently reversing the deletion.

A secondary issue: if a remote deletion API call were to fail, the exception would propagate outside the `try/except` blocks in form_valid, causing a `500` instead of a graceful rollback.

### Solution
Before `_save_digital_address_formset` runs, collect the OpenKlant mapping UUIDs for all addresses marked for deletion (while the mappings still exist in the DB). After the local formset saves, call the new `delete_remote_digitaal_adressen` service method with those UUIDs inside the existing OpenKlant `try/except` block. An API failure now follows the same path as any other sync error: set_rollback(True) rolls back the local deletion and the user sees the standard error message.

 The eSuite call guard is tightened to if changed_data only, since eSuite has no per-address remote deletion concept.

### DIFF
--- a/src/open_inwoner/accounts/tests/test_profile_views.py
+++ b/src/open_inwoner/accounts/tests/test_profile_views.py
@@ -29,7 +29,7 @@ from open_inwoner.accounts.forms import (
     PhoneDigitalAddressFormSet,
     UserForm,
 )
-from open_inwoner.accounts.models import User
+from open_inwoner.accounts.models import DigitalAddress, User
 from open_inwoner.cms.cases.cms_apps import CasesApphook
 from open_inwoner.cms.collaborate.cms_apps import CollaborateApphook
 from open_inwoner.cms.inbox.cms_apps import InboxApphook
@@ -42,7 +42,10 @@ from open_inwoner.haalcentraal.tests.mixins import HaalCentraalMixin
 from open_inwoner.laposta.models import LapostaConfig
 from open_inwoner.laposta.tests.factories import LapostaListFactory, MemberFactory
 from open_inwoner.openklant.constants import KlantenServiceType
-from open_inwoner.openklant.models import ESuiteKlantConfig
+from open_inwoner.openklant.models import (
+    DigitaalAdresOpenKlantMapping,
+    ESuiteKlantConfig,
+)
 from open_inwoner.openklant.tests.data import MockAPIReadPatchData
 from open_inwoner.openklant.tests.factories import DigitaalAdresOpenKlantMappingFactory
 from open_inwoner.pdc.tests.factories import CategoryFactory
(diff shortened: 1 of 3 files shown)


src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -972,6 +972,124 @@ def test_update_via_openklant_patches_when_mapping_exists(self, m):
...
+        self.client.force_login(data.digid_user)
+
+        post_data = {}
+        post_data.update(
+            self._da_formset_data(
+                "email_addresses",
+                [(email_addr.value, True)],
+                initial_pks=[email_addr.pk],
+            )
+        )
```
swrichards: Could be:

```suggestion
        post_data = self._da_formset_data(
                "email_addresses",
                [(email_addr.value, True)],
                initial_pks=[email_addr.pk],
            )
```

src/open_inwoner/accounts/tests/test_profile_views.py
```diff
@@ -972,6 +972,124 @@ def test_update_via_openklant_patches_when_mapping_exists(self, m):
...
+        post_data.update(self._da_formset_data("phone_addresses", []))
+
+        with patch(
+            "open_inwoner.openklant.services.OpenKlant2Service.delete_remote_digitaal_adressen",
+            side_effect=Exception("remote API unavailable"),
+        ):
+            response = self.client.post(self.url, data=post_data)
+
+        self.assertEqual(response.status_code, 302)
+        self.assertTrue(DigitalAddress.objects.filter(pk=email_addr.pk).exists())
```
swrichards, line 1056: Maybe also assert the mapping is still there too?

swrichards, line 1056: Maybe also assert the mapping is still there too?

src/open_inwoner/accounts/views/profile.py
```diff
@@ -354,7 +369,9 @@ def form_valid(self, form, email_formset, phone_formset):
                     KlantenServiceType.OPENKLANT2
                 ):
                     try:
-                        self.update_klant_via_openklant(changed_data, user)
+                        self.update_klant_via_openklant(
+                            changed_data, user, ok_uuids_to_delete
```
swrichards, line 373: What if `changed_data` is empty?

src/open_inwoner/accounts/views/profile.py
```diff
@@ -332,7 +345,7 @@ def form_valid(self, form, email_formset, phone_formset):
             if new_alt != old_phonenumber_alternative:
                 changed_data["phonenumber_alternative"] = new_alt
 
-            if changed_data:
+            if changed_data or ok_uuids_to_delete:
```
swrichards, line 348: Are there cases where `ok_uuids_to_delete` is truthy but `changed_data` is falsy?

src/open_inwoner/openklant/services.py
```diff
@@ -1724,15 +1724,11 @@ def update_partij_from_user_data(
...
-    def delete_remote_digitaal_adres_for_local(self, local_address) -> bool:
-        """Delete the remote digitaal adres corresponding to a local DigitalAddress."""
-        try:
-            mapping = local_address.openklant_mapping
-        except DigitaalAdresOpenKlantMapping.DoesNotExist:
-            return False
-        with contextlib.suppress(OK2NotFound):
-            self.client.digitaal_adres.delete(str(mapping.ok_uuid))
-        return True
+    def delete_remote_digitaal_adressen(self, ok_uuids: list) -> None:
```
```suggestion
    def delete_remote_digitaal_adressen(self, ok_uuids: Iter[str]) -> None:
```

src/open_inwoner/accounts/views/profile.py
swrichards: Right, we handle that below I now see to take care of both the data and uuid purge path.

swrichards (approved):
LGTM, just nitpicks

src/open_inwoner/accounts/views/profile.py
pi-sigma: `_save_digital_address_formset` never calls `user.update_email()` when addresses are removed because there's no "clear everything" path for emails. So `user.email` is unchanged and `changed_data` is empty.

PR #2635, opened 2026-06-23
ci: replace pre-commit/action with direct invocation

swrichards:
pre-commit/action is not in the org's allowed actions allowlist. Install pre-commit from the pinned version in requirements/dev.txt and invoke it directly instead.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -72,9 +72,12 @@ jobs:
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version-file: 'pyproject.toml'
+      - name: Install pre-commit
+        run: grep "^pre-commit" requirements/dev.txt | xargs pip install
       - name: npm install
         run: npm install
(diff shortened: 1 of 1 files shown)


.github/workflows/code-quality.yml
```diff
@@ -72,9 +72,12 @@ jobs:
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version-file: 'pyproject.toml'
+      - name: Install pre-commit
+        run: grep "^pre-commit" requirements/dev.txt | xargs pip install
```
pi-sigma, line 76: Why don't we just `pip install pre-commit==4.2.0`?

pi-sigma (approved):
Is adding the action to our allowlist not an option?

.github/workflows/code-quality.yml
swrichards: Because we always want this to run under whatever version we have pinned for local development, so you avoid "passes locally / fails remotely" issues.

swrichards:
> Is adding the action to our allowlist not an option?

Technically yes, but given how easy it is to install, we might as well avoid the dependency.

PR #2637, opened 2026-06-23
PP-427: Expose DigitalAddress as admin inline

pi-sigma:


src/open_inwoner/accounts/admin.py
```diff
@@ -79,6 +87,16 @@ def clean(self, *args, **kwargs):
                 raise ValidationError(_("The user with this email already exists."))
 
 
+class DigitalAddressInline(admin.TabularInline):
+    model = DigitalAddress
+    extra = 0
+    can_delete = False
+    readonly_fields = [field.name for field in DigitalAddress._meta.fields]
```
swrichards: Can't we just use `has_change_permission` and return False? (Same for `has_delete_permission` for good measure, in addition to `can_delete = False`).

swrichards (approved):

src/open_inwoner/accounts/admin.py
pi-sigma: Good idea, keep it simple :)

src/open_inwoner/accounts/admin.py
```diff
@@ -79,6 +87,20 @@ def clean(self, *args, **kwargs):
                 raise ValidationError(_("The user with this email already exists."))
 
 
+class DigitalAddressInline(admin.TabularInline):
+    model = DigitalAddress
+    extra = 0
```
swrichards, line 92: I think we want _both_ the `can_delete = True` _and_ `has_delete_permissions` function, otherwise the admin can still _try_ to delete, and will then get a server-side validation error., Better to just disable it at both levels.

swrichards (changes requested):

src/open_inwoner/accounts/admin.py
pi-sigma: Setting `has_delete_permission` to `False` also hides the delete button, so `can_delete` is not needed.

PR #2639, opened 2026-06-23
Add pre-seeded Docker services and setup configuration coverage for full testing setup

swrichards:


bin/stack.sh
```diff
@@ -0,0 +1,225 @@
...
+        bin/ensure_dev_network.sh
+
+        echo
+        echo "==> Starting satellite services (Keycloak, Open Zaak, Objects APIs, Haal Centraal BRP mock, Open Klant, Open Afval)"
+        echo "    web-init needs all of these reachable, so they come up -- and are waited on -- first."
+        "${COMPOSE[@]}" up -d --wait \
+            keycloak \
+            openzaak-web openzaak-celery \
+            objecttypes-web objects-web objects-celery \
+            personen-mock
```
pi-sigma: Most of these services (e.g. openzaak) do not define a healthchek; what are we waiting for here? `--wait` defaults to 'container is running', which is true immediately.

bin/stack.sh
```diff
@@ -0,0 +1,225 @@
...
+    -f docker/docker-compose.objects-apis.yml
+    -f docker/docker-compose.hc-brp-mock.yml
+    -f docker/docker-compose.openklant.yml
+    -f docker/docker-compose.openafval.yml
+    -f docker/docker-compose.observability.yml
+)
+
+wait_for_completion() {
+    # Waits for a one-shot container (openklant-seed/openafval-seed:
+    # `restart: 'no'`, no healthcheck) to exit, failing on a non-zero code.
```
pi-sigma: But `openklant-seed` has this:

```
    command: >
      sh -c "python src/manage.py loaddata --ignorenonexistent /fixtures/db.json
      || true" <- never fails?
```

does this ever return a non-zero code?

docker/setup_configuration/fixtures/openzaak_config.json
```diff
@@ -0,0 +1,406 @@
...
+      "zrc_service": 1,
+      "drc_service": 2,
+      "ztc_service": 3,
+      "form_service": null,
+      "fetch_eherkenning_zaken_with_rsin": true,
+      "fetch_eherkenning_zaken_with_openzaak_120_params": false,
+      "fetch_rollen_with_betrokkene_type": false,
+      "cache_zaken_timeout": 300,
+      "cache_catalogi_timeout": 86400,
+      "klant_backend": "esuite"
```
pi-sigma: `data.yml:243` has `primary_backend: openklant2`. Should be consistent

docker/setup_configuration/fixtures/openzaak_config.json
```diff
@@ -0,0 +1,406 @@
...
+    }
+  },
+  {
+    "model": "openzaak.catalogusconfig",
+    "pk": 1,
+    "fields": {
+      "url": "http://openzaak.internal:8000/catalogi/api/v1/catalogussen/989f65bb-4a6b-405c-a2c9-0859a0079734",
+      "domein": "SEEDD",
+      "rsin": "000000000",
+      "service": 3,
```
pi-sigma: This is fragile. We're relying on the order in which services are defined in `data.yml` and processed by `setup-configuration`. This works on an empty DB but: (a) it won't work on a DB with existing rows (in case we want to implement converging later on), (b) re-ordering or adding a service in `data.yml` will break this

docker/openklant/docker-init-openklant-db/0001-docker-init-openklant-db.sql
```diff
@@ -0,0 +1,5 @@
+CREATE USER openklant;
+CREATE DATABASE openklant;
```
pi-sigma: `docker-compose.openklant.yml` declares:

```
      - DB_NAME=open_klant
      - DB_USER=open_klant
      - DB_PASSWORD=open_klant
```

.github/workflows/ci.yml
```diff
@@ -257,6 +261,108 @@ jobs:
...
+
+          # Strip "v" prefix from tag name (if present at all)
+          [[ "${GITHUB_REF}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+
+          # Use Docker `latest` tag convention
+          [ "$VERSION" == "develop" ] && VERSION=latest
+
+          echo ::set-output name=tag::${VERSION}
+
+      # `bin/stack.sh up` always runs `docker compose build web`). We use
```
pi-sigma:
```suggestion
      # `bin/stack.sh up` always runs `docker compose build web`. We use
```

.github/workflows/ci.yml
```diff
@@ -292,49 +398,32 @@ jobs:
...
     runs-on: ubuntu-latest
     permissions:
       contents: read
 
+    env:
+      TAG: ${{ needs.full-app-stack-test.outputs.tag }}
+
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      # The image itself was already built (and proven healthy). so nothing
```
pi-sigma:
```suggestion
      # The image itself was already built (and proven healthy), so nothing
```

docs/installation/docker-compose.rst
```diff
@@ -43,29 +57,207 @@ Getting started
...
+Keycloak. If you need to switch users faster than that, log out at Keycloak
+(or use a private/incognito window) rather than relying on the timeout alone
+-- a still-active session is reused with no prompt at all.
 
 Testing OpenTelemetry Observability
 ===================================
 
-To test the OpenTelemetry setup and view metrics in Grafana:
+``bin/stack.sh up`` (see :ref:`installation_docker_compose_full_stack` above)
+already includes the observability stack and enables ``OTEL_SDK_DISABLED``
```
pi-sigma:
```suggestion
already includes the observability stack and sets ``OTEL_SDK_DISABLED`` to False
```

bin/setup_configuration.sh
```diff
@@ -12,5 +12,42 @@ SCRIPTPATH=$(dirname "$SCRIPT")
...
+    # populated by the `zgw_import_data` management command (also run daily
+    # via Celery beat, see CELERY_BEAT_SCHEDULE in conf/base.py), which
+    # discovers catalog data from the ZGW APIs live and upserts by
+    # url/identificatie. We load a fixture instead of running that command
+    # here, because it also pins fields the importer never touches --
+    # notify_status_changes, document_upload_enabled, status_indicator, etc.
+    # -- to specific values useful for visually testing case-visibility
+    # rules. This fixture is a snapshot: it was produced by running
+    # `zgw_import_data` for real against a stack seeded with
+    # seed_openzaak_fixtures.py (see repository root), then hand-editing in
```
pi-sigma: I cannot find `seed_openzaak_fixtures.py`. Was this included in an earlier version of the PR?

bin/stack.sh
swrichards: Good point, and as written it clearly doesn't. I will remove the fallback, we want it to fail loudly.

bin/stack.sh
swrichards: Right, adding them where they're missing (using the fallback /admin check, web only, should suffice for our purposes).

swrichards: Right, adding them where they're missing (using the fallback /admin check, web only, should suffice for our purposes).

bin/setup_configuration.sh
swrichards: It was, removed

docker/setup_configuration/fixtures/openzaak_config.json
swrichards: Used natural keys for the rows we don't create (e.g. the services).

Issue #2631, opened 2026-06-22
Include costs per lediging into the full Open Afval to Mijn Afval chain


swrichards:
For each lediging, we currently only include the weight, but not the costs, at both the row and totals level. For this feature, we will include a cost column throughout the chain:

- within the Open Afval data model
  - In the Lediging resource
  - in the afval profiel resources
- within Open Inwoner
  - In our client's API models
  - In the DTO exposed to the template
  - In the template where the extra column will be added...
  - ... and summed in the summary rows
  
Design Mijn Afval: https://www.figma.com/design/iKGhWhstaLIlFSaND2q7cE/OIP---Designs--new-?node-id=10880-25893&t=PXRJYPpaPfrUq8UM-1

<img width="732" height="843" alt="Image" src="https://github.com/user-attachments/assets/c5a7c568-46ba-44dd-b855-52b0641132d4" />

swrichards:
@pi-sigma https://sentry.maykinmedia.nl/organizations/maykin-media/issues/484113/?environment=maykin-test&project=427&query=is%3Aunresolved&referrer=issue-stream&stream_index=1

We either need to add the total costs at the klant level of Afval Profile in Open Afval, or remove it from the OIP API model spec (I'd say, let's go for the former).

PR #2642, opened 2026-06-24
feat: [#2631] Include lediging kosten in Mijn Afval

### LINKED ISSUE #2632: Include `cost` in API models and context vars for the Mijn Afval page

### LINKED ISSUE #2633: Expose `cost` column to Mijn Afval frontend
Only in the tables (cost column and total for that column). Graph is currently out of scope.

pi-sigma:
Closes #2632 and #2633

### DIFF
--- a/src/open_inwoner/mijn_afval/views.py
+++ b/src/open_inwoner/mijn_afval/views.py
@@ -140,6 +140,7 @@ class _LedigingData(TypedDict):
     tijdstip_tijd: str  # Time formatted as "HH:MM"
     tijdstip_dag: str  # Localized day of the week, e.g., "maandag"
     gewicht: str  # Weight (in kg) as string, e.g., "32,5"
+    kosten: str  # Cost (in euros) as string, e.g., "12,50"
 
 
 class _TableColumn(TypedDict):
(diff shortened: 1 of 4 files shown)


src/open_inwoner/mijn_afval/views.py
```diff
@@ -383,17 +389,30 @@ def _format_afval_profiel(profiel: AfvalProfiel) -> list[_ContainerLocationData]
...
                         "gewicht": _format_number(lediging.gewicht),
+                        "kosten": number_format(
+                            lediging.kosten, decimal_pos=2, force_grouping=False
+                        ),
                     }
                     ledigingen_data.append(lediging_data)
 
             # Format container data
             totaal_gewicht = _format_number(container.totaal_gewicht)
+            totaal_kosten = number_format(
```
swrichards: why the `_format_number` / `number_format` discrepancy?

swrichards (approved):

src/open_inwoner/mijn_afval/views.py
pi-sigma: We only show a single decimal point for weights, but for cost we need two decimal places. I can add this to  `_format_number()`, but it would mean adding a parameter and doing an `if/else` in the function. Thought it's simpler to just use `number_format` directly.

swrichards: I'd say we either do a singe function with a `decimal_points` kwarg (with a default), or make two semantically distinct functions: `format_weight` and `format_cost`.

pi-sigma: On second thought, I decided to amend `_format_number` after all. If you stumbled over this, other readers will be puzzled as well.

swrichards (approved):
LGTM, but holding off on merging until we've deployed the open afval changes to the open afval test server

PR #2653, opened 2026-06-24
Show all digital addresses on profile view page, make preferred address editable

### LINKED ISSUE #2652: Show digital addresses on profile view page

### LINKED ISSUE #2656: Make preferred address editable

swrichards:
Closes: #2652

### DIFF
--- a/src/open_inwoner/templates/pages/profile/me.html
+++ b/src/open_inwoner/templates/pages/profile/me.html
@@ -61,15 +61,19 @@
         </div>
         <div class="tabled__row tabled__row--blank">
             <div class="tabled__item tabled__key">{% trans "E-mail adres" %}</div>
-            <div class="tabled__item tabled__value">{{ user.email }}</div>
+            <div class="tabled__item tabled__value tabled__break">
+                {% for da in user.digital_addresses.all %}{% if da.type == "email" %}
+                    <div>{{ da.value }}{% if da.is_standard_for_type %} {% trans "(voorkeur)" %}{% endif %}</div>
+                {% endif %}{% endfor %}
+            </div>
         </div>
         <div class="tabled__row tabled__row--blank">
             <div class="tabled__item tabled__key">{% trans "Telefoonnummer" %}</div>
-            <div class="tabled__item tabled__value">{{ user.phonenumber }}</div>
-        </div>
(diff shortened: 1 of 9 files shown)


src/open_inwoner/templates/pages/profile/me.html
```diff
@@ -61,15 +61,19 @@ <h2 class="utrecht-heading-2" id="business-overview">{% trans "Bedrijfsgegevens"
         </div>
         <div class="tabled__row tabled__row--blank">
             <div class="tabled__item tabled__key">{% trans "E-mail adres" %}</div>
-            <div class="tabled__item tabled__value">{{ user.email }}</div>
+            <div class="tabled__item tabled__value tabled__break">
+                {% for da in user.digital_addresses.all %}{% if da.type == "email" %}
```
pi-sigma: I think it would be better if we fetch the addresses in the view and pass them to the template:

in `profile.py`:
```
context["email_addresses"] = user.digital_addresses.filter(type=DigitalAddressType.email)
context["phone_addresses"] = user.digital_addresses.filter(type=DigitalAddressType.phone)
```

Then in the template you can do `{% for da in email_addresses %}` with no type check needed.

The addresses are not pre-fetched, and `user` comes from the request. To avoid the additional query, we can use:
```
from django.db.models import prefetch_related_objects

prefetch_related_objects([user], "digital_addresses")
context["email_addresses"] = [da for da in user.digital_addresses.all() if da.type == DigitalAddressType.email]
context["phone_addresses"] = [da for da in user.digital_addresses.all() if da.type == DigitalAddressType.phone]
```
Not sure if it's worth it.

pi-sigma (changes requested):
I note that BSN and eHerkenning users now have exactly the same DA rendering (previously we had `user.email` vs `user.get_contact_email`). Perhaps worth extracting into a separate template and then `{% include %}` to prevent drift if we need to make changes

swrichards:
> I note that BSN and eHerkenning users now have exactly the same DA rendering (previously we had `user.email` vs `user.get_contact_email`). Perhaps worth extracting into a separate template and then `{% include %}` to prevent drift if we need to make changes

Did the extractionb: `get_contact_email` should otherwise resolved to the same email in all the other places it's needed (only the notifications really, and the contact list).

PR #2654, opened 2026-06-24
[#2655] Add e2e tests for profile edits

### LINKED ISSUE #2655: Add an end-to-end test for the profile pages

pi-sigma:


swrichards (changes requested):
Very happy with this! Would just like to add:

- Adding a duplicate within the user's set of DAs (should not change state and show a field validation error)
- Adding a malformed email/phone number (should show validation error)

swrichards (approved):

PR #2672, opened 2026-06-30
[#2665] Include role and confidentiality config in zaken cache keys

### LINKED ISSUE #2665: Missing cache key in fetch_zaken methods
The cache keys for `fetch_zaken_by_bsn` and `fetch_zaken_for_company` in `clients.py` do not include `limit_user_visible_cases_to_role`. When `OpenZaakConfig.limit_user_visible_cases_to_role` is enabled (e.g. set to initiator), the `rol__omschrijvingGeneriek` parameter is correctly added to the ZRC API request. However, because the config value is absent from the cache key, any cached result from before the setting was changed continues to be served until the entry expires. During that window, users may see zaken they should not have access to.

pi-sigma:
`fetch_zaken_by_bsn()` and `fetch_zaken_for_company()` filter requests by `zaak_max_confidentiality` and `limit_user_visible_cases_to_role`, but neither field was part of the cache key. Changing either config value would serve stale results until the cache expired.

Fixes #2665

### DIFF
--- a/src/open_inwoner/openzaak/clients.py
+++ b/src/open_inwoner/openzaak/clients.py
@@ -135,7 +135,7 @@ class ZakenClient(ZgwAPIClient):
                 )
 
     @cache_result(
-        "{self.base_url}:zaken:{user_bsn}:{max_requests}:{identificatie}",
+        "{self.base_url}:zaken:{user_bsn}:{max_requests}:{identificatie}:{self.zaak_max_confidentiality}:{self.limit_user_visible_cases_to_role}",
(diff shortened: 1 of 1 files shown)


swrichards (approved):
LGTM -- could you add an issue to our backlog to deprecate the cache decorator in favor of a caching key based directly on the URL + params?

swrichards:
> LGTM -- could you add an issue to our backlog to deprecate the cache decorator in favor of a caching key based directly on the URL + params?

Referencing this issue as an example

PR #2675, opened 2026-07-01
[#2663] Link question onderwerpobject to zaak by UUID

### LINKED ISSUE #2663: PP-446 Portaal: Vraag in portaal irt openklant gerelateerd aan een zaak wordt niet goed geregistreerd
### Beschrijving
Indien er een vraag in portaal gesteld wordt nav een zaak wordt deze “anoniem” aangemaakt.
Dit is te zien in ITA.

Image
Image
Image
Image
Image

pi-sigma:
## Summary

Stored the zaak UUID (extracted from its URL) as `onderwerpobjectidentificator.objectId` instead of the human-readable `identificatie`, and set `codeObjecttype` / `codeSoortObjectId` to match (`zgw-Zaak` / `uuid`).

 Added `ZGWService.get_zaak_by_uuid()` to resolve a zaak across all configured API groups by UUID. Use it in retrieve_question instead of `search_zaken()`, which filtered by `identificatie` and would no longer match. Update `list_questions_for_zaak()` to filter the OpenKlant API by UUID as well.

Restored VCR cassettes for OpenKlant API integration tests which were removed during a recent refactor.

Closes #2663

### DIFF
--- a/src/open_inwoner/openklant/tests/helpers.py
+++ b/src/open_inwoner/openklant/tests/helpers.py
@@ -1,0 +1,220 @@
+import json
+import subprocess
+import time
+from pathlib import Path
+
+from django.test import TestCase as DjangoTestCase
+
+import requests
+import structlog
+from openklant_client import OpenKlantClient
+from typing_extensions import TypedDict
+from vcr.record_mode import RecordMode
+from vcr.unittest import VCRMixin
+
+from open_inwoner.conf.utils import config
+
+logger = structlog.stdlib.get_logger(__name__)
+
+_REPO_ROOT = Path(__file__).resolve().parents[4]
+
+
+class OpenKlantServiceManager:
+    _in_server_context: bool = False
+    _django_service_name: str = "openklant-web"
+    _api_root: str = "http://localhost:8338"
+    _api_path: str = "/klantinteracties/api/v1"
+    # Must match the `secret` used by OpenKlant2ConfigFactory's ServiceFactory,
+    # since that's the token the client under test actually sends.
+    _api_token: str = "test-api-token"
+    _docker_compose_project_name: str = "openklant2-api-test"
+    _docker_compose_path: Path = _REPO_ROOT / "docker" / "docker-compose.openklant.yml"
+
+    def _docker_compose(self, *args: str, check: bool = True, input: str | None = None):
+        input_data = {"text": True, "input": input} if input else {}
+        try:
+            return subprocess.run(
+                args=[
+                    "docker",
+                    "compose",
+                    "-f",
+                    str(self._docker_compose_path),
+                    "-p",
+                    self._docker_compose_project_name,
+                    *args,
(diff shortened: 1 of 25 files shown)


src/open_inwoner/openzaak/services.py
```diff
@@ -823,6 +823,17 @@ def check_zaak_access(
 
         return zaak, api_group
 
+    def get_zaak_by_uuid(self, uuid: str) -> ZaakWithApiGroup | None:
+        for api_group in ZGWApiGroupConfig.objects.select_related("zrc_service"):
+            try:
+                zaak = self._zaken_client_factory(api_group).fetch_single_zaak(uuid)
+            except ZgwAPIClientError:
+                continue
```
swrichards: Let's log this, distinguishing between notfound (fine) and any other error (log as warning).

src/open_inwoner/openzaak/services.py
```diff
@@ -823,6 +823,17 @@ def check_zaak_access(
 
         return zaak, api_group
 
+    def get_zaak_by_uuid(self, uuid: str) -> ZaakWithApiGroup | None:
+        for api_group in ZGWApiGroupConfig.objects.select_related("zrc_service"):
```
swrichards: A bit more work, but these can sneakily become quite slow: I think it's worth doing these in parallel.

If any of the futures timeout/error: log a warning, emit an otel metric.
If both futures complete and both return a zaak: log this as an unexpected state, and return the first.

src/open_inwoner/openklant/tests/helpers.py
swrichards, line 1: Did you run these against the default version? If so, I think it's worth bumping the default version and committing that default in the recording script to `2.15.0` which is the most recent version PodiumD has pinned.

swrichards (changes requested):
Good to have the fixtures back in, otherwise some small things.

src/open_inwoner/openzaak/metrics.py
```diff
@@ -46,3 +46,12 @@
...
 )
+
+zaak_by_uuid_fetch_errors = meter.create_counter(
+    "oip.openzaak.zaak_by_uuid_fetch_errors",
+    unit="{error}",
+    description=(
+        "Errors encountered when fetching a zaak by UUID across API groups in parallel. "
+        "Attributes: reason ('timeout', 'api_error', 'unknown')"
+    ),
+)
```
swrichards: This is a bit too specific: this should be a general `zgw_api_error` where the reason and endpoint become attributes. However, that's probably best done in the underlying ZaKenClient. I would remove the mtrics from this PR entirely for now, and make a backlog ticket specifically for redesinging the ZGW API OTEL metrics.

swrichards (approved):
LGTM, just a comment to remove the OTEL metrics for reasons explained.

swrichards (approved):

PR #2687, opened 2026-07-06
[#2676] Show public container id in Mijn Afval container heading

### LINKED ISSUE #2676: Use `publicContainerId` in Mijn Afval container heading

pi-sigma:
The container table caption used the internal API id, which is an opaque identifier not meant for end users. Add publicContainerId to the AfvalContainer model and use it in the heading instead.

Closes #2676

### DIFF
--- a/src/open_inwoner/mijn_afval/api_models.py
+++ b/src/open_inwoner/mijn_afval/api_models.py
@@ -24,6 +24,7 @@ class Periode(PydanticModel):
 
 class AfvalContainer(PydanticModel):
     id: str
+    public_container_id: str
     afval_type: str
     is_verzamelcontainer: bool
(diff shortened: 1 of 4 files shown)


pi-sigma:
@swrichards `_AfvalContainerData["identifier"]` (used for the script-id in the template) still uses `container.id` for uniqueness. I left it as-is because this is for internal bookkeeping. Should we change it as well?

swrichards:
> @swrichards `_AfvalContainerData["identifier"]` (used for the script-id in the template) still uses `container.id` for uniqueness. I left it as-is because this is for internal bookkeeping. Should we change it as well?

Yes that's fine.

swrichards (approved):

PR #2716, opened 2026-07-15
2679 getallen in mijn afval tabel rechts uitlijnen

### LINKED ISSUE #2679: Getallen in Mijn Afval tabel rechts uitlijnen
Getallen moeten volgens de NEN-EN-ISO 80000-1 norm rechts uitgelijnd zijn. Momenteel staan ze links uitgelijnd. 
Hieronder zie je een screenshot van hoe de nieuwe tabel eruit komt te zien.

Image

Image

stefrado:
## Summary

- Getallen in mijn afval tabel rechts uitlijnen 
- Implement unit type to table cells (to be approved by UIUX)

## Issue References

- Github Issue: #2679

### DIFF
--- a/src/open_inwoner/react/components/Table/Table.tsx
+++ b/src/open_inwoner/react/components/Table/Table.tsx
@@ -70,7 +70,11 @@ const Table: AC<ITableProps> = ({
           <TableRow>
             {/* LOOP 1: Iterate over each column to render table headers */}
             {data.columns?.map(({ header, scope = 'col' }, index) => (
-              <TableHeaderCell key={index} scope={scope}>
+              <TableHeaderCell
+                key={index}
+                scope={scope}
+                className={index === 0 ? undefined : 'right'}
+              >
                 {header}
               </TableHeaderCell>
             ))}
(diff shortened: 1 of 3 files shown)


src/open_inwoner/react/components/Table/Table.tsx
```diff
@@ -61,6 +61,13 @@ const Table: AC<ITableProps> = ({
 
   const isEmpty = !data.rows || data.rows.length === 0;
 
+  const getFormattedValue = (label: string, unit: string) => {
+    if (!label) return label;
```
pi-sigma: What if the label is missing but the cell contains data? We'll discard the latter and show the empty label. It's not an issue today, but is this desired?

src/open_inwoner/react/components/Table/Table.tsx
```diff
@@ -61,6 +61,13 @@ const Table: AC<ITableProps> = ({
...
+    // Haalt de tekst tussen de haakjes uit de titel, bv. '€' uit 'Bedrag (€)'
+    const value = label.match(/\(([^)]*)\)/)?.[1] ?? '';
+    return value === '€' ? `${value} ${unit}` : `${unit} ${value}`;
```
pi-sigma: It's not ideal. Any header that later gets a parenthetical aside for some reason will leak that text into every row value in that column. Not blocking as it works for now, but please create an issue for the backlog with a reference to this issue; we might want to move this to the backend instead to simplify this and make it less fragile.

pi-sigma (approved):

src/open_inwoner/react/components/Table/Table.tsx
stefrado: Good comment, i'll add it to the backlog

src/open_inwoner/react/components/Table/Table.tsx
stefrado: No this was incorrect. I fixed the logic and now this will return the value when there is no columnLabel

src/open_inwoner/react/components/Table/Table.tsx
stefrado: https://github.com/maykinmedia/open-inwoner/issues/2723

pi-sigma (approved):

Issue #2719, opened 2026-07-17
GFT container is niet groen in de chart


stefrado:
### Describe the bug

<img width="672" height="362" alt="Image" src="https://github.com/user-attachments/assets/c506f2ab-bd0d-4de6-9986-b66e2029270d" />

PR #2720, opened 2026-07-17
fix: [#2719] chart gft bar color match is based on lower text

stefrado:
## Summary

chart gft bar color match is based on lower text

## Issue References

- Github Issue: #2719

### DIFF
--- a/src/open_inwoner/react/components/Chart/types.ts
+++ b/src/open_inwoner/react/components/Chart/types.ts
@@ -7,7 +7,7 @@ export interface AfvalLediging {
 
 export interface AfvalContainer {
   identifier: string;
-  type: 'Restafval' | 'GFT';
+  type: string;
   totaal_gewicht: string;
   ledigingen: AfvalLediging[];
 }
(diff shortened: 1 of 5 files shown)


src/open_inwoner/react/components/Chart/types.ts
pi-sigma, line 1: `export interface AfvalContainer` still has `type: 'Restafval' | 'GFT';` Should also be widened to `type: string;`

src/open_inwoner/react/components/Chart/useChart.ts
```diff
@@ -221,9 +221,11 @@ export class BarChartBuilder {
     return base;
   }
 
-  private getColor(type: 'GFT' | 'Restafval', containerIndex: number): string {
+  private getColor(type: string, containerIndex: number): string {
     const baseColor =
-      type === 'GFT' ? CHART_STYLES.colors.gft : CHART_STYLES.colors.restafval;
+      type.toLowerCase() === 'gft'
+        ? CHART_STYLES.colors.gft
+        : CHART_STYLES.colors.restafval;
```
pi-sigma: This will lump unknown afval types under restavfal. Needs an explicit lookup keyed on the lowercased type (`restafval`, `gft`), and give anything unmatched a neutral fallback color plus dev mode warning for debugging.

pi-sigma (changes requested):

pi-sigma (approved):

PR #2730, opened 2026-07-28
[#2728] Sanitize hrefs in prosemirror-rendered content

### LINKED ISSUE #2728: Render-side content sanitization

pi-sigma:
Prosemirror rich text fields (e.g. FAQ answers, category descriptions, product content) allow the LINK mark with no restriction on href scheme. Nothing in the editor, form validation, or render pipeline rejected schemes like javascript: or data:, so a stored answer could execute arbitrary JS in a citizen's session when they clicked the link on the public FAQ page (stored XSS).

Add an href allow-list (http(s), mailto, tel, relative/fragment) and apply it in get_rendered_content()/get_product_rendered_content(), plus the PDC API serializers that returned raw .html untouched.

This is a render-side backstop in open_inwoner; the root cause in `maykin-django-prosemirror` (unvalidated href on the link mark) is tracked separately for a fix upstream (https://github.com/maykinmedia/django-prosemirror/pull/56).

Closes #2728

### DIFF
--- a/src/open_inwoner/api/pdc/serializers.py
+++ b/src/open_inwoner/api/pdc/serializers.py
@@ -9,6 +9,7 @@ from open_inwoner.pdc.models.product import (
     ProductFile,
     ProductLocation,
 )
+from open_inwoner.utils.html import sanitize_html
 
 
 class FilerImageSerializer(serializers.ModelSerializer):
@@ -68,7 +69,7 @@ class Questionserializer(serializers.ModelSerializer):
     def get_answer(self, obj):
         if obj.answer:
             try:
-                return obj.answer.html
+                return sanitize_html(obj.answer.html)
             except Exception:
                 return None
         return None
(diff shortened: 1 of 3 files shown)


swrichards (commented):
@pi-sigma This is a very good find, but I think the best place to fix this would be in the library itself. If I am reading this correctly, we would still be storing the offending href tags in the prosemirror document and only removing them on render, which is treating the symptom. Let's discuss later if we have time to do this, or whether we first go with this quick fix and schedule the prosemirror fix for later.

pi-sigma:
@swrichards PR in the library exists (https://github.com/maykinmedia/django-prosemirror/pull/56), I updated the comment above. As stated in the description, this is just a stopgap to prevent malicious content which has already been saved from being rendered. Perhaps a clean up via a data migration is better?

swrichards (approved):

PR #2738, opened 2026-07-28
[#2735] Improve zaken list reliability

### LINKED ISSUE #2735: PP-461 Zaken list incomplete on first load
[\[#DRT-642\] Mijn Zaken toont niet alle zaken bij eerste laadactie.pdf](https://github.com/user-attachments/files/30457468/DRT-642.Mijn.Zaken.toont.niet.alle.zaken.bij.eerste.laadactie.pdf)

**Incident omschrijving**
Na release 4.8 zie ik dat wanneer ik Mijn Zaken voor de eerste keer open, niet alle beschikbare zaken worden geladen en getoond. Pas nadat ik naar een andere pagina binnen de paginering navigeer (bijvoorbeeld pagina 3), worden de ontbrekende zaken alsnog geladen en zichtbaar. Dit kan leiden tot verwarring over het ontbreken van zaken en zorgt ervoor dat gebruikers een onvolledig overzicht te zien krijgen van hun zaken.

**Beschrijving issue:**
Na release 4.8 zie ik dat wanneer ik Mijn Zaken voor de eerste keer open, niet alle beschikbare zaken worden geladen en getoond. Pas nadat ik naar een andere pagina binnen de paginering navigeer (bijvoorbeeld pagina 3), worden de ontbrekende zaken alsnog geladen en zichtbaar. Dit kan leiden tot verwarring over het ontbreken van zaken en zorgt ervoor dat gebruikers een onvolledig overzicht te zien krijgen van hun zaken.

Image

pi-sigma:
Warn and retry when the zaken list is incomplete.

Closes #2735

### DIFF
--- a/src/open_inwoner/openzaak/services.py
+++ b/src/open_inwoner/openzaak/services.py
@@ -6,7 +6,7 @@ import threading
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import date
-from typing import Generic, TypedDict, TypeVar, cast
+from typing import Generic, Self, TypedDict, TypeVar, cast
 
 from django.utils.functional import cached_property
 
@@ -117,6 +117,13 @@ class SkipReason(enum.Enum):
     FULL_RESOLUTION_FAILED = "full_resolution_failed"
     TIMEOUT = "timeout"
 
+    @classmethod
+    def transient_reasons(cls) -> frozenset[Self]:
+        """Zaken skipped for these reasons may show up on retry"""
+        return frozenset(
+            {cls.TIMEOUT, cls.FULL_RESOLUTION_FAILED, cls.ZAAKTYPE_RESOLUTION_FAILED}
+        )
+
 
 @dataclass
 class SkippedZaak:
@@ -132,20 +139,20 @@ _ZaakT = TypeVar("_ZaakT", bound=ZaakWithApiGroup)
 class ZakenResult(Generic[_ZaakT]):
     zaken: list[_ZaakT]
     skipped: list[SkippedZaak]
-    # Set when a whole-stage timeout dropped zaken whose URLs are unknown (the raw
-    # fetch stage), so `skipped` cannot enumerate them.
-    raw_fetch_timed_out: bool = False
+    # Set when a whole-stage timeout or fetch error dropped zaken whose URLs are
+    # unknown (the raw fetch stage), so `skipped` cannot enumerate them.
+    raw_fetch_incomplete: bool = False
 
     @property
-    def has_timeouts(self) -> bool:
-        """Whether results are incomplete because of a timeout.
+    def is_incomplete(self) -> bool:
+        """Whether results are incomplete because of a timeout or fetch/resolve error.
 
-        Only timeouts count: zaken excluded for confidentiality, an internal
-        zaaktype or a failed resolution are legitimately absent and retrying
-        will not bring them back.
+        Zaken excluded for confidentiality, an internal zaaktype or a
+        before-visible-from date are legitimately absent and retrying will not
+        bring them back.
         """
-        return self.raw_fetch_timed_out or any(
-            skipped.reason == SkipReason.TIMEOUT for skipped in self.skipped
+        return self.raw_fetch_incomplete or any(
+            skipped.reason in SkipReason.transient_reasons() for skipped in self.skipped
         )
 
 
@@ -424,7 +431,7 @@ class ZGWService:
         )
 
         fetched_zaken: list[ZaakWithApiGroup] = []
-        raw_fetch_timed_out = False
+        raw_fetch_incomplete = False
         with parallel(max_workers=self._max_workers) as executor:
             futures = [
                 executor.submit(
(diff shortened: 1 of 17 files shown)


src/open_inwoner/openzaak/tasks.py
```diff
@@ -139,7 +139,7 @@ def warm_cache_for_user(
     service = ZGWService()
 
     # get_raw_zaken handles multi-group concurrency and timeouts; populates fetch_zaken_by_bsn cache
-    raw_zaken = service.get_raw_zaken(identification)
+    raw_zaken = service.get_raw_zaken(identification).zaken
     if not raw_zaken:
         logger.info("Finished ZGW cache warm-up: no zaken found")
```
swrichards: This log line should also report if we had any timeouts, as the issue shows, we're not sure if it was "no zaken" or "timed out while fetching zaken".

src/open_inwoner/openzaak/services.py
swrichards, line 455: I think the `raw_fetched_time_out` case is too narrow: a server or network error would also be swallowed and lead to the same behaviour. For the issue at hand, it doesn't matter whether the raw cases fetch failed because of a timeout or some other transient server error. I think we have to set a flag in this exception branch as well, and rename the flag to something simpler like `results_incomplete`. Technically we would differentiate between transient cases (timeout, network error, server error) and more troubling errors (e.g. a 400, we did something wrong), but for the user does this not matter and the result is the same, they are not viewing the whole list.

The only visual we might want to make: if the number of results is 0, _and_ we encountered errors, the message should perhaps be more "We c ould not fetch your cases" rather than "we could not fetch _all_ your cases"

src/open_inwoner/openzaak/services.py
```diff
@@ -125,13 +125,28 @@ class SkippedZaak:
...
 
 
-_ZaakT = TypeVar("_ZaakT", bound=ZaakWithApiGroupZaakTypeResolved)
+_ZaakT = TypeVar("_ZaakT", bound=ZaakWithApiGroup)
 
 
 @dataclass
 class ZakenResult(Generic[_ZaakT]):
     zaken: list[_ZaakT]
     skipped: list[SkippedZaak]
```
swrichards, line 141: As discussed below, I think it would be simpler to distinguish between:

- `skipped` (normal case, no server errors) and;
- `has_errors` (including network or server errors, not just timeouts)

The reason being that, for timeouts, the cause is really irrelevant. Cruciallyt, this should mean that `SkipReason.TIMEOUT` is not actually a reason for skipping a case, but should also lead to `has_errors = True`.

swrichards (changes requested):
Basic approach is good, but I think we need to broaden this beyond just timeouts, see my comments on `services.py`.

src/open_inwoner/openzaak/services.py
pi-sigma: Makes sense. To be clear: broadening to `is_complete=True/False` means we would include non-transient errors in the re-try flow. The trade-off is between (a) "retry even if it's sometimes useless" vs. (b) more complex logic and the danger of misclassifying a transient error as permanent and silently dropping cases.

src/open_inwoner/openzaak/services.py
```diff
@@ -125,13 +125,41 @@ class SkippedZaak:
...
+# legitimate, permanent exclusion. Zaken skipped for these reasons may show up
+# on retry; the other reasons (confidentiality, internal zaaktype, etc.) are
+# final and retrying will not bring them back.
+_INCOMPLETE_SKIP_REASONS = frozenset(
+    {
+        SkipReason.TIMEOUT,
+        SkipReason.ZAAKTYPE_RESOLUTION_FAILED,
+        SkipReason.FULL_RESOLUTION_FAILED,
+    }
+)
```
swrichards: I think this would be nicer as a classmethod on `SkipReason`:

```python
class SkipReason(enum.Enum):
    TIMEOUT = "timeout"
    FULL_RESOLUTION_FAILED = "full_resolution_failed"
    # ...

    @classmethod
    def transient_reasons(cls) -> set[Self]:
        return frozenset({cls.TIMEOUT, cls.FULL_RESOLUTION_FAILED})
```

src/open_inwoner/openzaak/services.py
```diff
@@ -125,13 +125,41 @@ class SkippedZaak:
...
 @dataclass
 class ZakenResult(Generic[_ZaakT]):
     zaken: list[_ZaakT]
     skipped: list[SkippedZaak]
+    # Set when a whole-stage timeout or fetch error dropped zaken whose URLs are
+    # unknown (the raw fetch stage), so `skipped` cannot enumerate them.
+    raw_fetch_incomplete: bool = False
+
+    @property
+    def is_incomplete(self) -> bool:
```
swrichards, line 147: Sorry for the back and forth, but I wonder if `retryable` isn't the label we want, as that better captures the intent here?

```suggestion
    raw_fetch_has_retryable_errors: bool = False

    @property
    def is_retryable(self) -> bool:
```

Up to you though, it's a bit of a nitpick/bikeshed.

swrichards (approved):
LGTM, just two optional readability improvements.

src/open_inwoner/openzaak/services.py
pi-sigma: I think it's the other way around: `is_incomplete` is also used for logging in the cache warm-up and the search, where `is_retryable` would not be appropriate because there is no re-try logic. The broader meaning is what we need.

PR #2739, opened 2026-07-29
[#2736] Improve performance of Klantinteracties component

### LINKED ISSUE #2736: PP-456 Klantinteracties performance
**Incident omschrijving:** 
Voor het acceptatietest voor de release ben ik in Portaal aan het testen. De omgeving is extreem traag. Ik krijg geen foutmelding, maar door de traagheid wordt het testen negatief beïnvloed.
Kan iemand hiernaar kijken. Dan pas kunnen we met zekerheid zeggen of het aan de traagheid ligt of dat dit komt doordat er in Portaal iets niet goed zit.

**Beschrijving issue**
Het bewonersportaal (Open Inwoner) is traag voor gebruikers die met DigiD zijn ingelogd: de pagina's "Mijn zaken" en vooral "Mijn vragen" (contactmomenten) doen er 20–25 seconden over en geven soms een foutmelding. Oorzaak: de applicatie haalt bij elke paginaweergave tientallen gegevens één voor één op bij het achterliggende e-Suite-systeem in plaats van in één keer. Dit is een fout in de applicatie zelf, geen infrastructuurprobleem; de traagheid bestond al langer, maar de zwaarste pagina is nieuw sinds de release van 16 juli. De oplossing moet van de leverancier van Open Inwoner komen (efficiënter ophalen of cachen); inspanning ligt daarmee buiten ons team. Analyse staat hieronder. 

Voor het acceptatietest voor de release ben ik in Portaal aan het testen. De omgeving is extreem traag. Ik krijg geen foutmelding, maar door de traagheid wordt het testen negatief beïnvloed. 

Kan iemand hiernaar kijken. Dan pas kunnen we met zekerheid zeggen of het aan de traagheid ligt of dat dit komt doordat er in Portaal iets niet goed zit.

[\[#DRT-624\] Portaal extreem traag.pdf](https://github.com/user-attachments/files/30456812/DRT-624.Portaal.extreem.traag.pdf)

pi-sigma:
### Summary

Addresses reported 20-25s load times and intermittent errors on "Mijn vragen" (PP-456/#2736), caused by an N+1 fetch to e-Suite plus inefficient DB access pattern. Fixes the root causes on both the eSuite and OpenKlant2 backends, adds caching and partial-results handling so failures degrade gracefully instead of losing data, and cleans up the  resulting duplication.

Closes #2736 

### Overview of changes

**feat: fetch contactmomenten in parallel [eSuite]**
  Replaces the sequential per-contactmoment fetch with a worker pool, and introduces `KlantContactMomentenResult` so
  partial failures are reported instead of dropped silently.

**feat: filter klantcontacten server-side [OpenKlant]**
  Applies the `hadBetrokkene__wasPartij__uuid` filter now supported by OpenKlant, cutting an unbounded per-partij scan
  down to a scoped query while keeping the client-side check as a privacy cross-check.

  **feat: resolve per-page lookups once in the questions listing**
  Hoists the answer-mapping and subject-mapping lookups out of the per-row loop, and moves the shared
  `get_or_create` into a transactional manager method used by both backends.

  **feat: short-circuit the eSuite detail page and guard per-klant fetch failures**
  The detail page no longer resolves every contactmoment just to show one, and a single klant's listing failure now
  surfaces as `list_fetch_failed` instead of aborting the whole request.

  **fix: guard against StopIteration in OpenKlant retrieve_question**
  An unmatched `question_uuid` raised `StopIteration` instead of returning "not found", which surfaced as a 500 rather
  than a 404.

  **feat: cache resolved contactmomenten**
  Caches `retrieve_contactmoment` with a configurable, admin-tunable timeout, mirroring the existing zaken cache
  pattern.

  **feat: show a partial-results banner with manual retry for Mijn vragen**
  Replaces the generic "something went wrong" message with a warning banner and a retry link; auto-retry (htmx) is left
  for once `fix/2735-zaken-list-incomplete` merges.

  **fix: guard KlantContactMomentRedirectView against unmatched uuids**
  Both backend branches of the redirect view 500'd on an unmatched uuid instead of 404ing; neither had test coverage.

  **refactor: fold openklant/wrap.py into services.py**
  `wrap.py` was down to one function and a type alias already duplicated in `services.py`; folds it in and deletes the
  module.

**refactor: drop redundant partij retrieval in OpenKlant lookup**
  The partij listing already returns the full representation, so retrieving it again by uuid spent a second request on a response we had; no caller reads the `_expand` it added.

  **fix: only let the primary backend raise the partial-results banner**
  Both backends are still queried, since an instance migrated to OpenKlant2 keeps its history in the eSuite, but a leftover secondary backend that stopped answering put a permanent banner in front of a complete list.

  **feat: cache the partij uuid for a user [OpenKlant]**
  Every klantinteracties page resolves the user's partij before doing anything else; only the uuid is cached (the profile pages patch the partij) under a hashed key, and a missing partij is not cached.

**feat: expose cache_key and invalidate on the cache decorator**
  The cache decorator only supported get-or-set; callers that mutate the underlying resource had no way to evict a stale entry without duplicating the key-formatting logic by hand.

**feat: cache klantcontactmomenten listing per klant [eSuite]**
"Mijn vragen" re-lists a klant's klantcontactmomenten on every page load; only the per-contactmoment resolution was cached so far. Cached separately, keyed on klant_url, and invalidated right after a new question is created so it still appears immediately.

**feat: warm klantcontactmomenten listing cache on login [eSuite]**
Populates the same listing cache during the existing off-request-cycle login task, so "Mijn vragen" gets a cache hit on first view instead of paying for the listing itself. The warm-up and page-view calls now share one max_requests bound, since caching them under different bounds would have populated an entry the page view never reads.


src/open_inwoner/openklant/clients.py
```diff
@@ -187,9 +198,14 @@ def retrieve_objectcontactmomenten_for_contactmoment(
     #
     # klantcontactmomenten
     #
-    def retrieve_klantcontactmomenten_for_klant(
+    def list_klantcontactmomenten_for_klant(
         self, klant: Klant
```
swrichards: If we want caching here, we could pass `klant_url: str` instead, and possibly add a `max_requests: None | int = None` as with ZGW, so we can build a cache decorator for it with both params in the cache key,

swrichards: (We could then also make this part of the cache warming).

src/open_inwoner/openklant/services.py
```diff
@@ -585,31 +666,180 @@ def retrieve_objectcontactmomenten_for_contactmoment(
...
-            kcm.contactmoment = self.retrieve_contactmoment(kcm.contactmoment)
+        failed_urls: set[str] = set()
+        with parallel(max_workers=max_workers) as executor:
+            futures = [
+                executor.submit(
+                    self._resolve_contactmomenten_chunk, chunk, client, deadline
+                )
+                for chunk, client in zip(chunks, clients)
+            ]
+            for future in concurrent.futures.as_completed(futures):
```
swrichards: As discussed: we won't do the custom deadline logic and the chunking here. We'll, for now, adopt the zaken service pattern: 1 resolution = 1 future (no chunking), and no home grown timeout checking, just `as_completed` within the executor contextmanager.

We'll then apply the new-style `TimedBatch` logic to both zaken and klantcontacten services.

swrichards (commented):
As discussed, revert to the ZGW pattern for now (see comments), then we apply the new-style time batched logic separately.

src/open_inwoner/openklant/services.py
```diff
@@ -143,6 +168,27 @@ class Question(TypedDict):
...
+    klantcontact__uuid: str
+
+
+@dataclass
+class QuestionsResult:
+    questions: list[Question]
+    # Set when the backing data could not be retrieved in full, so callers can tell
+    # the user they are looking at less than they have.
+    is_incomplete: bool = False
+
```
swrichards, line 189: I saw there was a release of the openklant-client, would that help?

pi-sigma: No the release added support for the `DigitaalAdres.referentie` attribute; unrelated

src/open_inwoner/openklant/services.py
```diff
@@ -1057,6 +1412,41 @@ def get_partij_for_user(self, user: User) -> Partij | None:
 
         return None
 
+    def _partij_uuid_cache_key(self, user: User) -> str:
+        # The bsn/kvk is hashed rather than used directly: cache keys turn up in
+        # logs and monitoring, and these identify a citizen.
+        identity = hashlib.sha256(
+            f"{user.bsn}|{user.kvk}|{user.vestiging}".encode()
+        ).hexdigest()
+        return f"{self.config.service.api_root}:partij_uuid:{identity}"
```
swrichards, line 1423: This does not actually obscure much, because somebody can just iterate over the whole space of numbers and infer the original values. Needs to use a secret/salt, e.g. a hmac digest using `settings.SECRET_KEY`

src/open_inwoner/openklant/services.py
```diff
@@ -587,25 +636,115 @@ def retrieve_objectcontactmomenten_for_contactmoment(
...
-        data = self.client.parse_json(response)
-        all_data = list(pagination_helper(self.client, data))
-        klanten_contact_moments = self.client.factory(KlantContactMoment, all_data)
 
-        # resolve linked resources
-        for kcm in klanten_contact_moments:
-            if not kcm.klant == klant.url:
-                raise ValueError("klantcontactmoment not linked to klant")
+        future_to_kcm: dict[concurrent.futures.Future, KlantContactMoment] = {}
+        remaining: dict[int, int] = {}
```
swrichards, line 750: I think this might read easier as `unprocessed_processed = {id(kcm)_ for kcm in kcms}` and then later you can do `unprocessed_processed.pop(id(kcm))` in the future handling, and checking for timeouts is then `id(kcm) in unprocessed_process`, which might be a bit more readable.

src/open_inwoner/openklant/services.py
```diff
@@ -2171,7 +2184,60 @@ def _has_initiator(row) -> bool:
...
+
+            if foreign:
+                logger.error(
+                    "Server-side partij filter was not applied; klantcontacten for "
+                    "other partijen were returned and discarded locally",
+                    partij_uuid=partij_uuid,
+                    discarded=foreign,
+                )
+
+        return _filtered()
```
swrichards, line 2401: I am a bit weary of this: we'll keep it here as I see the rational, but it's giving off eSuite/Openzaak vibes in terms of branching. I think the correct approach here (which we'll have to do when there's less time pressure) is to explicitly store the connected openklant version and branch on that, with different clients/types in the client library.

src/open_inwoner/openklant/services.py
```diff
@@ -1358,6 +1361,41 @@ def get_partij_for_user(self, user: User) -> Partij | None:
...
+        # The bsn/kvk is hashed rather than used directly: cache keys turn up in
+        # logs and monitoring, and these identify a citizen.
+        identity = hashlib.sha256(
+            f"{user.bsn}|{user.kvk}|{user.vestiging}".encode()
+        ).hexdigest()
+        return f"{self.config.service.api_root}:partij_uuid:{identity}"
+
+    def resolve_partij_uuid(
+        self, user: User, lookup: Callable[[], Partij | None]
+    ) -> str | None:
```
swrichards, line 1427: Passing the callable feels a bit awkward to me, I feel it would be cleaner to set the cached valued whenever we do an actual partij lookup, and have this method fall-back to fetching it if uncached so we always have a value (but that would require additional tests that the value is actually cached on all the partij retrieval paths).

swrichards (changes requested):
Just some small things: not a lot of space to do a bigger rethink of how this works given the constraints. A lot of good stuff in here, but it would benefit from splitting up into smaller PRs, but for reasons discussed, we'll move ahead now.

src/open_inwoner/openklant/services.py
pi-sigma: Structure and naming are exactly parallel to the code in openzaak:
```
        future_to_zaak: dict[concurrent.futures.Future, ZaakWithApiGroup] = {}
        visible_ids: set[int] = set()
        processed_ids: set[int] = set()
        skipped: list[SkippedZaak] = []
        with parallel(max_workers=self._max_workers) as executor:
```
I'm leaving this as-is; we can change it as part of the follow-up refactor of the parallelism implementation.

src/open_inwoner/openklant/services.py
pi-sigma: It's just defensive, no branching here. I added a `TODO` however that this can be removed once we're sure every deploy is on a sufficiently new version of OpenKlant

src/open_inwoner/openklant/services.py
pi-sigma: Two reasons against:
1. `list_questions()` and `retrieve_question()` use different lookups and therefore pass different callables: a `get-or-create` in the former, a simple `get` lookup in the later case. A hard-coded fallback can only be one of these, and there's then no simple way to express the difference. If we pick the `get-or-create` variant, then `retrieve_question()` starts creating where it shouldn't; if we pick the simple `get` variant,  `list_questions()` will fail to create partijen where it should
2. cache writes are centralized now but would become scattered (`get_or_create_partij_for_user`, `find_persoon_for_bsn`, `find_organisatie_for_kvk_and_vestiging`), making it easier to introduce bugs

src/open_inwoner/accounts/views/contactmoments.py
```diff
@@ -178,11 +177,10 @@ def get_context_data(self, **kwargs):
...
 
-        if fetch_error:
-            messages.error(
-                self.request,
-                _("Something went wrong. You might not see all of your questions."),
-            )
+        # Manual retry for now: a full auto-retry (mirroring the htmx polling on the
+        # cases list) is blocked on fix/2735-zaken-list-incomplete landing first, so
+        # both lists can share the same pattern instead of diverging.
+        ctx["partial_results"] = fetch_error
```
swrichards: Comment out of date

src/open_inwoner/openklant/services.py
swrichards: I guess this is an issue for a lot of other sensitive cache key calculations as well. Might be something to look at later.

swrichards: See #2777

src/open_inwoner/accounts/signals.py
```diff
@@ -295,6 +296,25 @@ def warm_zgw_cache_on_login(
         logger.warning("Failed to dispatch ZGW cache warm-up task", exc_info=True)
 
 
+@receiver(user_logged_in, dispatch_uid="user_logged_in.warm_klantcontactmomenten_cache")
+def warm_klantcontactmomenten_cache_on_login(
+    sender: type[User], request: HttpRequest | None, user: User, **kwargs
+) -> None:
+    config = KlantenSysteemConfig.get_solo()
+    if config.primary_backend != KlantenServiceType.ESUITE.value:
+        return
```
swrichards: I'd prefer to move this into the `try/except` -- better safe than sorry, signals shouldn't crash the login flow.

src/open_inwoner/openklant/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+logger = structlog.stdlib.get_logger(__name__)
+
+
+@app.task(
+    base=QueueOnce,
+    once={"keys": ["user_id"], "graceful": True},
+)
+def warm_klantcontactmomenten_cache_for_user(user_id: int) -> None:
+    """
+    Warm the eSuite klantcontactmomenten listing cache for a user on login.
```
swrichards, line 21: I wouldn't mind having logging on all the branches here

PR #2748, opened 2026-08-03
[#2747] Align Ga Naar arrow via flexbox instead of absolute positioning

### LINKED ISSUE #2747: PP-387 Icons misalign
Backports:

- `2.4.x`: #2824

pi-sigma:
The "Ga Naar" plugin's arrow (`ExternalLink.scss`) was positioned with hard-coded `position: absolute; bottom: 28px; right: 24px`, and the text content had a matching hard-coded `width: calc(100% - var(--spacing-mega)`) to leave room for it. This contradicted the component's own design token, which already specifies to use relative positioning with flexbox.

Fixes #2747


src/open_inwoner/scss/components/ExternalLink/ExternalLink.scss
```diff
@@ -117,10 +117,10 @@
     }
 
     &__arrow {
```
stefrado, line 119: The alignment looks correct and the arrow is always centered. Only problem is that there is no spacing between the text and the arrow. If there is available spacing it looks ok, but when this is not the case it looks like this:

Image

This is solvable with the design token: `--oip-external-link-arrow-icon-margin-inline-start` (should be at least `4px`)

stefrado (approved):
Looks good, one small comment

src/open_inwoner/scss/components/ExternalLink/ExternalLink.scss
swrichards: @stefrado This one? https://github.com/maykinmedia/open-inwoner-design-tokens/blob/101afa9e72da59a353fe1559926cf3100b2aa55f/src/components/external-link-plugin.tokens.json#L100-L103

stefrado: Yes that one! `auto` should be changed to `4px`

stefrado (approved):

src/open_inwoner/scss/components/ExternalLink/ExternalLink.scss
```diff
@@ -38,6 +38,7 @@
     display: var(--oip-external-link-display);
     flex-direction: var(--oip-external-link-flex-direction);
     align-items: center;
+    gap: var(--oip-external-link-row-gap) var(--oip-external-link-column-gap);
```
stefrado: I do not think that this gap is the solution:

Image

Image

src/open_inwoner/scss/components/ExternalLink/ExternalLink.scss
stefrado, line 125: To help you out, i created the PR in oip-design-tokens which pushes the arrow `4px` -> https://github.com/maykinmedia/open-inwoner-design-tokens/pull/48


Once that PR is merged and a new version is pushed to NPM, the `oip-design-tokens` version should be updated in `package.json`

stefrado (changes requested):

stefrado (approved):

